PR #236, opened 2020-02-11
add an auto setter for gcloud.project.projectNumber

tangiel:
This allows kpt to automatically set a project number if the relevant
setter appears in the config, saving users from having to go look it
up themselves. A future change should automatically set this whenever
gcloud.core.project is set.

### DIFF
--- a/internal/util/setters/setters.go
+++ b/internal/util/setters/setters.go
@@ -20,6 +20,7 @@ import (
 	"os/exec"
 	"strings"
 
+	"github.com/pkg/errors"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/setters"
 )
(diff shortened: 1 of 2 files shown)


internal/util/setters/setters.go
```diff
@@ -62,13 +63,35 @@ func PerformSetters(path string) error {
...
+
+func GetProjectNumberFromProjectID(projectID string) (string, error) {
+	gcloudCmd := exec.Command("gcloud",
+		"projects", "describe", projectID, "--format", "value(projectNumber)")
+	b, err := gcloudCmd.Output()
+	if err != nil {
+		fmt.Printf("err: %v\n", err)
+		return "", err
+	}
+	fmt.Println(strings.TrimSpace(string(b)))
```
pwittrock: why is this printed?

internal/util/setters/setters.go
```diff
@@ -62,13 +63,35 @@ func PerformSetters(path string) error {
...
 	return kio.Pipeline{Inputs: []kio.Reader{rw}, Filters: fltrs, Outputs: []kio.Writer{rw}}.
 		Execute()
 }
+
+func GetProjectNumberFromProjectID(projectID string) (string, error) {
+	gcloudCmd := exec.Command("gcloud",
+		"projects", "describe", projectID, "--format", "value(projectNumber)")
+	b, err := gcloudCmd.Output()
+	if err != nil {
+		fmt.Printf("err: %v\n", err)
```
pwittrock: this error should get printed further up, use `errors.Wrap` to wrap the error and attach a stack-trace

pwittrock:
looks good, couple small things

internal/util/setters/setters.go
tangiel: Woops, just debugging. Removed.

PR #241, opened 2020-02-16
Add documentation for kpt live commands

mortent:
First draft. Still WIP

### DIFF
--- a/docs/live/README.md
+++ b/docs/live/README.md
@@ -1,0 +1,104 @@
+## pkg
+
+Fetch, update, and sync configuration files using git
+
+<link rel="stylesheet" type="text/css" href="/kpt/gifs/asciinema-player.css" />
+<asciinema-player src="coming..." speed="1" theme="solarized-dark" cols="60" rows="26" font-size="medium" idle-time-limit="1"></asciinema-player>
+<script src="/kpt/gifs/asciinema-player.js"></script>
+
+    # run the tutorial from the cli
+    kpt tutorial live
+
+[tutorial-script]
+
+### Synopsis
+
+Tools to safely apply and remove packages from clusters.
+
+| Command   | Description                                       |
+|-----------|---------------------------------------------------|
+| [apply]   | apply a package to the cluster                    |
+| [preview] | preview the operations that apply will perform    |
+| [destroy] | remove the package from the cluster               |
+
+**Data Flow**: local configuration or stdin -> kpt live -> apiserver (Kubernetes cluster)
+
+| Configuration Read From | Configuration Written To |
+|-------------------------|--------------------------|
+| local files or stdin    | apiserver                |
+| apiserver               | stdout                   |
+
+#### Pruning
+kpt live apply will automatically delete resources which have been
+previously applied, but which are no longer included. This clean-up
+functionality is called pruning. For example, consider a package
+which has been applied with the following three resources:
+
+```
+service-1 (Service)
+deployment-1 (Deployment)
+config-map-1 (ConfigMap)
+```
+
+Then imagine the package is updated to contain the following resources,
+including a new ConfigMap named `config-map-2` (Notice that `config-map-1`
+is not part of the updated package):
+
+```
+service-1 (Service)
+deployment-1 (Deployment)
+config-map-2 (ConfigMap)
+```
+
+When the updated package is applied, `config-map-1` is automatically
+deleted (pruned) since it is omitted.
+
+
+In order to take advantage of this automatic clean-up, a package must contain
+a **grouping object template**, which is a ConfigMap with a special label. An example is:
+
+```
+apiVersion: v1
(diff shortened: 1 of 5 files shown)


docs/live/README.md
```diff
@@ -0,0 +1,81 @@
...
+and that contains a list of identifiers for the applied resources. In order for 
+apply to know the name and namespace where the generated grouping object should
+be created, it is required that the user supplies a ConfigMap manifest as part
+of the set passed to apply which must contain the annotation 
+`cli-utils.sigs.k8s.io/inventory-id: <name>`. This is just the base used by
+apply to create the grouping objects in the cluster, so you will see ConfigMaps
+created with the name of the grouping object being the name provided in the 
+base ConfigMap and a generated suffix. 
+
+An example of a ConfigMap with the appropriate annotation is:
```
seans3: annotation -> label

docs/live/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Data Flow**: local configuration or stdin -> kpt live -> apiserver (Kubernetes cluster)
+
+| Configuration Read From | Configuration Written To |
+|-------------------------|--------------------------|
+| local files or stdin    | apiserver                |
+| apiserver               | stdout                   |
+
+#### Pruning
+kpt live apply handles pruning of resources. That means that
```
seans3: Consider the following wording for paragraph 1:

kpt live apply will automatically delete resources which have been previously applied, but which are currently omitted. This clean-up functionality is called pruning. For example, consider a package which has been applied with the following three resources:


```
service-1 (Service)
deployment-1 (Deployment)
config-map-1 (ConfigMap)
```

Then imagine the package is updated to contain the following resources, including a new ConfigMap named `config-map-2` (Notice that `config-map-1` is not part of the updated package):

```
service-1 (Service)
deployment-1 (Deployment)
config-map-2 (ConfigMap)
```

When the updated package is applied, config-map-1 is automatically deleted (pruned) since it is omitted.

docs/live/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+#### Pruning
+kpt live apply handles pruning of resources. That means that 
+if a package has already been applied to a cluster, but a new version
+becomes available with some resources no longer present, kpt live apply
+will make sure to delete those resources. This fixes the problem where
+deleted resources are not known to kubectl apply, and therefore end up
+orphaned in the cluster.
+
+kpt live apply relies on a grouping object to be able to keep track
```
seans3: Consider the following wording for paragraph 2:

In order to take advantage of this automatic clean-up, a package must contain a **grouping** object, which is a ConfigMap with a special label. An example of this grouping object is:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-grouping-object
  labels:
    cli-utils.sigs.k8s.io/inventory-id: test-group
```

And the special label is:

`cli-utils.sigs.k8s.io/inventory-id: *group-name*`

`kpt live apply` recognizes this package grouping object from the special label, and it stores the metadata of all applied objects in the ConfigMap's data field. Subsequent `kpt live apply` commands can then query the grouping object, and calculate the omitted objects, cleaning up accordingly. When the grouping object is created, a hash suffix is added to the name. Example:

`test-grouping-object-17b4dba8`

docs/live/README.md
mortent: Fixed.

docs/live/README.md
mortent: Yeah, I think using an example like this is a good idea. I have incorporated this with just a tiny adjustment.

docs/live/README.md
mortent: So I think we should try to find terminology to distinguish the ConfigMap which is the template for the grouping objects and the actual grouping objects that are created in the cluster. I struggled with this when I wrote this and I think this suggestion also mixes the two in a way that can make it confusing.

mortent: How about something like this?

In order to take advantage of this automatic clean-up, a package must contain a grouping object template, which is a ConfigMap with a special label. An example is:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-grouping-object
  labels:
    cli-utils.sigs.k8s.io/inventory-id: test-group
```

And the special label is:

```
cli-utils.sigs.k8s.io/inventory-id: *group-name*
```

kpt live apply recognizes this template from the special label, and based on this kpt will create new grouping object with the metadata of all applied objects in the ConfigMap's data field. Subsequent kpt live apply commands can then query the grouping object, and calculate the omitted objects, cleaning up accordingly. When a grouping object is created in the cluster, a hash suffix is added to the name. Example:

```
test-grouping-object-17b4dba8
```

seans3: This looks good.

mortent: Updated the PR

PR #242, opened 2020-02-19
feat: adding homebrew formula

tfhartmann:
Added a homebrew formula and updated the README to include steps on how to add the tap and install with `brew`


pwittrock:
This is so cool.

PR #260, opened 2020-02-20
wrap cfg set to add implicit project number set

tangiel:
This is a bit of a hacky change, but it's simpler than a fullblown
callback-based system. For cfg set, it runs the original set, but if
it detects that gcloud.core.project is being set, it will also update
gcloud.project.projectNumber.


tangiel:
@pwittrock a bit hackier than what we discussed, let me know if it's not acceptable.

pwittrock:
awesome

pwittrock:
Would you update the set that gets called from sync as well?

PR #272, opened 2020-02-24
fix: updating brew shasum

tfhartmann:
updating shasum so that brew will install the new vesion correctly


pwittrock:
@tfhartmann Thanks.  We should probably add a script for performing a release that updates the binary url and shasum.

PR #441, opened 2020-03-23
Fix indentation in generated Long message

mortent:
This changes the formatting used in the Hugo markdown from defining blocks with indentation to using "```". This approach works better with mdtogo since the content of the generated Long message gets the same indentation as the rest of the Cobra output. We should also go through all the commands and use a consistent way to describe usage, args, flags and env vars. I will do that separately from this.

Will do a follow-up PR to regenerate the Hugo content.

@pwittrock


pwittrock:
@mortent run `make docs` to regenerate the site

mortent:
@pwittrock Done

PR #472, opened 2020-04-02
Updates to use newest kyaml, cmd/config, and cli-utils

seans3:
* Updates to use newest kyaml, cmd/config, and cli-utils for upcoming v0.16.0 release


pwittrock:
I think this is obsolete.

PR #485, opened 2020-04-03
Doc updates to pkg sub-commands

drekle:
Updating the docs to reflect my own investigation of this tool.

Still investigating this tool and it is my understanding that all references to `kpt get` are `kpt pkg get`.

Updated a single reference to `kpt sync` to `kpt pkg sync`.


drekle:
@googlebot I signed it!

pwittrock:
Thanks!

PR #511, opened 2020-04-12
Add kubeconfig flags to relevant live cmds.

### LINKED ISSUE #455: The --context flag is not implemented
When running `kpt live --help`, it mentions:

```
      --cluster string                 The name of the kubeconfig cluster to use
      --context string                 The name of the kubeconfig context to use
```

However:

```
$ kpt live apply --context=foo dir
Error: unknown flag: --context
$ kpt live --context=foo apply dir
Error: unknown flag: --context
```

I'd triage this as a bug as opposed to a feature request :)

bzub:
Fixes #455

### DIFF
--- a/commands/livecmd.go
+++ b/commands/livecmd.go
@@ -47,9 +47,8 @@ func GetLiveCommand(name string) *cobra.Command {
 
 	// Create the factory and IOStreams for the "live" commands. The factory
 	// is created using the config flags.
-	flags := liveCmd.Flags()
 	kubeConfigFlags := genericclioptions.NewConfigFlags(true).WithDeprecatedPasswordFlag()
-	kubeConfigFlags.AddFlags(flags)
+	kubeConfigFlags.AddFlags(liveCmd.Flags())
 	matchVersionKubeConfigFlags := util.NewMatchVersionFlags(kubeConfigFlags)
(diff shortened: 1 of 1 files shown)


bzub:
@googlebot I signed it!

seans3:
Do we want to do this in `cli-utils` or do we want to do it here?

mortent:
I think we would like to make this change in https://github.com/kubernetes-sigs/cli-utils rather than here. All the other flags are defined there, so it seems to make the most sense.
@bzub would you be willing to move this to a PR against cli-utils?

bzub:
@mortent Yup will do.

pwittrock:
@bzub thanks!

seans3:
Actually, now that I look at it, I think your original approach in `kpt live` makes more sense than `cli-utils`--reopening.

/reopen

seans3:
For more context, the kubeconfig flags are created in `kpt live` -- not `cli-utils`. And the fix is to use `persistent flags` to persist to the flags to the subcommands (they exist on `kpt live`, but not `kpt live <subcommand>`. So we should perform the fix in `kpt`. Although it looks I don't have permissions to reopen this PR.

/cc @mortent

PR #520, opened 2020-04-15
fix typo

bkcsfi:


bkcsfi:
@googlebot I signed it!

pwittrock:
Thanks!

pwittrock:
I think the existing doc is correct -- *get* is referring to `kpt pkg get` (as opposed to `git`).  LMK if you think there is a way to make this more clear.  (maybe change `get` to `kpt pkg get`)

PR #535, opened 2020-04-22
Ignore whitespace diffs in demos

ian-howell:
The whitespace changes that occasionally occur during the demos can make the actual `git diff` a bit confusing to read. For example, the `git diff` at the end of `02-get-create-setter/demo.sh` currently looks like the following:

image

With this PR, the `git diff` now looks as so:
image


ian-howell:
@googlebot I signed it!

PR #555, opened 2020-04-28
Set useragent for kpt requests

mortent:
This updates the `util.Factory` to return a `rest.Config` with the useragent set to something like `kpt/v0.23.0`. This works as expected for the normal clients.
The issue is that the `RESTClientGetter` interface also provides functions for getting a DiscoveryClient and a RESTMapper, so these will not use the updated `rest.Config` and therefore not pick up the right useragent. I haven't found a way to address this that doesn't require a change in the upstream code.

Another option we could use is to set the variables in `k8s.io/client-go/pkg/version/base.go` during build and then just rely on the regular way the useragent is set for kubernetes. But it does feel a bit strange and it would be different from the way we currently set the version in `run.go` during build.

@pwittrock @seans3

### DIFF
--- a/run/run.go
+++ b/run/run.go
@@ -16,6 +16,7 @@ package run
 
 import (
 	"bytes"
+	"flag"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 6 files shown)


run/run.go
```diff
@@ -175,7 +179,28 @@ func newHelp(e []string, c *cobra.Command) func(command *cobra.Command, strings
...
+}
+
+func getUserAgent() string {
+	if version == "" {
+		return "kpt/<unknown>"
+	}
+	return fmt.Sprintf("kpt/%s", version)
+}
+
+var version string = "v3.2.1"
```
pwittrock, line 203: why are we setting this version?

pwittrock: wdyt of just making the default version "uknown", and removing the ifs

run/run.go
```diff
@@ -111,7 +115,7 @@ func GetMain() *cobra.Command {
 
 	// help and documentation
 	cmd.InitDefaultHelpCmd()
-	cmd.AddCommand(kptcommands.GetKptCommands("kpt")...)
+	cmd.AddCommand(kptcommands.GetKptCommands("kpt", newFactory(cmd))...)
```
pwittrock, line 118: Should we use the newFactory to refresh OpenAPI and discovery in the background immediately?

run/run.go
mortent: The version set here is just a leftover from some testing that I forgot to remove. But I like the suggestion of just setting it to unknown and remove the ifs.

mortent: Fixed

mortent:
@pwittrock I have updated the PR based on the comments and added a test.

PR #556, opened 2020-04-28
WIP: docs: flag is enable-star, not enable-starlark

justinsb:
Update a typo in the docs that had the wrong flag name.


pwittrock:
Fixed

PR #598, opened 2020-05-01
update references of `kpt run` to `kpt fn run`

dustinspecker:
I ran `make docs` locally as well and I got a whole bunch more changes. Please let me know if some of those are needed with part of this commit.


dustinspecker:
@googlebot I signed it!

PR #613, opened 2020-05-06
Fix broken link for `example repository`

zzxwill:


zzxwill:
@googlebot I signed it!

PR #614, opened 2020-05-06
Update functions docs for writing go functions

pwittrock:


site/content/en/guides/producer/functions/golang/_index.md
```diff
@@ -0,0 +1,314 @@
...
+...
+err := node.PipeE(yaml.Lookup("spec", "replicas"), yaml.FieldSetter{StringValue: "3"})
+```
+
+```go
+// Set the spec.replicas field to 3, creating it if it doesn't exist
+var node *yaml.RNode
+...
+// pass in the type of the node to create if it doesn't exist (e.g. Sequence, Map, Scalar)
+err := node.PipeE(yaml.LookupCreate(yaml.ScalarNode, "spec", "replicas"), yaml.FieldSetter{StringValue: "3"})
```
frankfarzan, line 251: Related to our offline discussion. Reviewing the API here, I think it's still exposing low-level YAML AST constructs.

I'm not opposed to having this as a complementary API if people don't mind thinking in terms of YAML AST. But ideally the user should be thinking in terms of objects in the native language (Javascript objects or Go structs), and not in the terms of the low-level file format. I think we can get close to that reality by the orchestrator doing the heavy lifting in terms of preserving file format features like comments in YAML.

pwittrock: Can you give an example of what this means?  `Unstructured` and `map[string]interface{}` do not seem like better alternatives to `yaml.Node`.  Providing both a high-level constructr as well as access to low-level constructs for break glass isn't a bad thing.

site/content/en/guides/producer/functions/golang/_index.md
```diff
@@ -0,0 +1,314 @@
+---
```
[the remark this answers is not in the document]
pwittrock: Done

site/content/en/guides/producer/functions/golang/_index.md
frankfarzan: Yea, unstructured in Go is also cumbersome to use. This is why we started with TS as the first SDK language.

Going back to the other comment, this is what I suggest: Present the different choice of languages and runtime (TS, Go/kyaml, star/python) with the same prominence (Same level in the hierarchy in hugo). They're at heterogeneous in terms of API and abstractions, but it's fine for now to get validation. This is in contrast to other tools (e.g. Pulumi and other SDKs) that attempt to provide the same abstractions in different languages.

site/content/en/guides/producer/functions/golang/_index.md
```diff
@@ -0,0 +1,314 @@
...
+for functions implementing abstractions (e.g. client-side CRD equivalents).
+
+```go
+...
+	type NestedValue struct {
+		Value string `yaml:"value,omitempty"`
+	}
+	type Spec struct {
+		NestedValue string `yaml:"nestedValue,omitempty"`
+        MapValues map[string]string  `yaml:"mapValues,omitempty"`
```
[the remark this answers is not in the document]
pwittrock: Done

site/content/en/guides/producer/functions/golang/_index.md
pwittrock: I am not sure i understand the issues you are concerned about.

With kyaml in go:

```go
v, err := value.Pipe(yaml.Lookup("path", "to", "value"))
```

With  lodash in js:

```js
v = _.get(value, "path", "to", "value")
```

Is the concern the package name: **yaml**.Lookup vs **resource**.Lookup?

pwittrock:
@frankfarzan the TypeScript SDK page should probably have a quick start guide the same as the Go SDK does.  Redirecting to another site isn't great UX.

site/content/en/guides/producer/functions/golang/_index.md
pwittrock: re: consistency across SDKs, I'd argue SDKs should stay as close to the IO spec as possible, with libraries offering value-adds on top.  e.g. using the nomenclature `functionConfig`, `ResourceList`, `results`, etc.  This will ensure consistency between the language SDKs.

PR #622, opened 2020-05-08
a fix bc the foo label description was repeated

linde:


linde:
if this is generated and the source of truth for this content is elsewhere, just let me know

PR #649, opened 2020-05-20
Update to version 0.11.0 of cli-utils

mortent:
@seans3


seans3:
looks like a minor lint issue with go import order in `run.go`

PR #763, opened 2020-06-25
Document for kpt delete setters

jijiew:
Add document for kpt delete setters


jijiew:
will add the feature first

PR #804, opened 2020-07-09
Add --workflow flag to kpt fn export

ZhuGongpu:
Use the `--workflow` flag to specify the workflow orchestrator. Now `DIR` is the first argument in consistent with other subcommands.

### DIFF
--- a/internal/cmdexport/cmdexport.go
+++ b/internal/cmdexport/cmdexport.go
@@ -18,6 +18,8 @@ package cmdexport
 import (
 	"fmt"
 	"os"
+	"sort"
+	"strings"
 
 	"github.com/GoogleContainerTools/kpt/internal/cmdexport/orchestrators"
 	"github.com/GoogleContainerTools/kpt/internal/cmdexport/types"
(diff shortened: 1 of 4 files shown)


ZhuGongpu:
@Liujingfang1 @mortent @monopole @frankfarzan

internal/cmdexport/cmdexport.go
```diff
@@ -83,14 +61,37 @@ func GetExportRunner() *ExportRunner {
...
+	case "cloud-build":
+		{
+			r.Pipeline = new(orchestrators.CloudBuild)
+		}
+	case "gitlab-ci":
+		{
+			r.Pipeline = new(orchestrators.GitLabCI)
+		}
+	default:
+		return fmt.Errorf("unsupported orchestrator %v", r.WorkflowOrchestrator)
```
Liujingfang1, line 92:
```
unsupported orchestrator %v. It must be one of github-actions, cloud-build and gitlab-ci.
```

Liujingfang1: Maybe put "github-action, cloud-build and gitlab-ci" into a string variable, since it's used several times in the file.

internal/cmdexport/cmdexport.go
```diff
@@ -83,14 +68,38 @@ func GetExportRunner() *ExportRunner {
...
-	OutputFilePath string
-	Command        *cobra.Command
+	WorkflowOrchestrator string
+	OutputFilePath       string
+	Command              *cobra.Command
 	*types.PipelineConfig
 	Pipeline orchestrators.Pipeline
 }
 
+var supportedOrchestrators = map[string]orchestrators.Pipeline{
```
Liujingfang1, line 78: Let's avoid using a global variable. This map can be put inside a function like
```
func supportedOrchestrators() map[string]orchestrators.Pipeline{
   return map[string]...
}
```

internal/cmdexport/cmdexport.go
frankfarzan: Cobra doesn't have a built-in utility for taking in an enum? interesting.

PR #805, opened 2020-07-09
Support exporting Jenkins pipelines

ZhuGongpu:


ZhuGongpu:
@Liujingfang1 @monopole @frankfarzan

internal/cmdexport/orchestrators/jenkins.go
```diff
@@ -0,0 +1,160 @@
...
+}
+`
+
+	result, err := renderTemplate("pipeline", templateString, p)
+	out = []byte(result)
+
+	return
+}
+
+type JenkinsStage struct {
```
Liujingfang1, line 64: Why do we need both `Jenkins` and `JenkinsStage`? It seems that `Name` and `Scripts` can be put into `Jenkins` directly. Is it with the consideration of potential multi stages? Then you can change 
```
type Jenkins struct {
	Stage JenkinsStage
}
```

to 
```
type Jenkins struct {
	Stages []JenkinsStage
}
```

internal/cmdexport/orchestrators/jenkins.go
```diff
@@ -0,0 +1,160 @@
...
+func (p *Jenkins) Init(config *types.PipelineConfig) Pipeline {
+	step := JenkinsStageStep{
+		MountedWorkspace: "/app",
+		Dir:              config.Dir,
+		FnPaths:          config.FnPaths,
+	}
+	p.Stage = JenkinsStage{
+		Name: "Run kpt functions",
+		Scripts: []string{
+			step.Generate(),
```
Liujingfang1, line 40: Is that possible to assign the step and stages in `Init` function? Then in `Generate()`, we can call all underlying `Generate()` from step or stage?

internal/cmdexport/orchestrators/jenkins.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return
+}
+
+type JenkinsStage struct {
+	Name    string
+	Scripts []string
+}
+
+func (stage *JenkinsStage) Generate() (result string, err error) {
```
Liujingfang1, line 69: The stage `Generate` function can trigger the JenkinsStageStep `Generate` function.

internal/cmdexport/orchestrators/jenkins.go
```diff
@@ -0,0 +1,167 @@
...
+	"bytes"
+	"fmt"
+	"path"
+	"strings"
+	"text/template"
+
+	"github.com/GoogleContainerTools/kpt/internal/cmdexport/types"
+)
+
+type Jenkins struct {
```
Liujingfang1, line 27: Please add some comment for this type.

internal/cmdexport/orchestrators/jenkins.go
```diff
@@ -0,0 +1,167 @@
...
+}
+`
+
+	result, err := renderTemplate("pipeline", templateString, p)
+	out = []byte(result)
+
+	return
+}
+
+type JenkinsStage struct {
```
Liujingfang1, line 54: ditto

internal/cmdexport/orchestrators/jenkins.go
```diff
@@ -0,0 +1,167 @@
...
+        '''{{end}}
+    }
+}`
+
+	result, err = renderTemplate("stage", templateString, stage)
+
+	return
+}
+
+type JenkinsStageStep struct {
```
Liujingfang1, line 85: ditto

internal/cmdexport/orchestrators/jenkins.go
```diff
@@ -0,0 +1,167 @@
...
+		KptImage,
+		"fn run",
+		path.Join(step.MountedWorkspace, step.Dir),
+	}}
+	multilineScript.lines = append(multilineScript.lines, fnPaths...)
+
+	return multilineScript.Generate()
+}
+
+type JenkinsMultilineScript struct {
```
Liujingfang1, line 123: ditto

internal/cmdexport/orchestrators/jenkins.go
```diff
@@ -0,0 +1,167 @@
...
+		fnPaths = append(fnPaths, fnPath)
+	}
+
+	multilineScript := &JenkinsMultilineScript{lines: []string{
+		"docker run",
+		fmt.Sprintf("-v $PWD:%s", step.MountedWorkspace),
+		"-v /var/run/docker.sock:/var/run/docker.sock",
+		KptImage,
+		"fn run",
+		path.Join(step.MountedWorkspace, step.Dir),
```
Liujingfang1, line 116: Let's put `"fn run"` and `path.Join(step.MountedWorkspace, step.Dir)` into the same line. So in the generated pipeline will be similar to
```   fn run  /app/resources  \```

PR #819, opened 2020-07-14
Render inline typescript functions with syntax highlighting

prachirp:
- Render inline typescript functions with syntax highlighting
- Remove license and usage leaving only function implementations
- Fix whitespace issues when inlining
- Fix various links


site/content/en/guides/producer/functions/ts/quickstart/_index.md
```diff
@@ -55,9 +55,19 @@ Currently supported platforms: amd64 Linux/Mac
    given `label_name` and `label_value` to add the appropriate label to `Namespace` objects using
    the SDK's `addLabel` function.
 
-   {{< code type="ts" >}}
-   {{< readfile file="static/ts/label_namespace.ts" >}}
-   {{< /code >}}
+  ```typescript
```
frankfarzan, line 58: This will eventually become out of date. Is there no way to include files directly? (Not link a copy, but actual source files).

prachirp: We could theoretically have a short code which fetches and shows the raw source file but then we'd have to live with license boilerplate. This way is better.

Issue #226, opened 2020-02-05
Support auto-generating function pipelines for different workflow orchestrators


frankfarzan:
e.g.:

`kpt fn export [cloud-build,jenkins-pipeline,github-actions]`

This would generate e.g. Cloud Build configuration for running a pipeline of kpt functions.

There are various design options we should explore:

1. Each workflow step runs the function wrapped with a utility that wires functions input/output in a shared volume
2. Each workflow step runs the function directly using unique input/output files in a shared volume. As an example see:
https://github.com/GoogleCloudPlatform/csp-config-management/blob/1.0.0/ci-pipeline-unstructured/cloudbuild.yaml
3. A single workflow step that runs the entire kpt functions pipeline. This would lose some UX benefit of separate reporting/notifications in the workflow UI, but maybe acceptable if the workflow orchestrator is an implementation detail. For example, if everything is bubbled up into the GitHub UI instead of user going to Jenkins or Cloud Build. A technical challenge: Running docker-in-docker is not supported by all orchestrators.

A general issue is that workflow orchestrators do not typically support stdin/stdout, and the configuration function spec doesn't enforce a consistent flag name for input/output.

frankfarzan:
4. Mandate flags and/or env vars to [functions spec](https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/api-conventions/functions-spec.md):

-i, --input
-o, --output

Downside is that existing tools (e.g. kubectl, helm, kustomize) need to be wrapped.

prachirp:
@Liujingfang1 is this done?

Liujingfang1:
@prachirp Yes, this is done. The document is available https://googlecontainertools.github.io/kpt/guides/consumer/function/export/

PR #840, opened 2020-07-21
Bundle docker into the kpt image

ZhuGongpu:
Ref #226, bundle docker into the kpt image to enable `kpt fn run` in containers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ RUN go build -v -o /usr/local/bin/kpt ./
 
 FROM alpine:3.11
 RUN apk update && apk upgrade && \
-        apk add --no-cache git less man diffutils bash openssh && \
+    apk add --no-cache git less man diffutils bash openssh docker-cli && \
     rm -rf /var/lib/apt/lists/* && \
     rm /var/cache/apk/*
 COPY --from=0 /usr/local/bin/kpt /usr/local/bin/kpt


ZhuGongpu:
@mortent @Liujingfang1 @monopole @frankfarzan

Dockerfile
```diff
@@ -25,8 +25,10 @@ RUN go build -v -o /usr/local/bin/kpt ./
 
 FROM alpine:3.11
 RUN apk update && apk upgrade && \
-        apk add --no-cache git less man diffutils bash openssh && \
+    apk add --no-cache git less man diffutils bash openssh && \
     rm -rf /var/lib/apt/lists/* && \
     rm /var/cache/apk/*
+RUN apk --update add docker openrc && \
```
Liujingfang1, line 31: What is the impact on image size when including docker?

ZhuGongpu: Its size increases from `83.9MB` (`35.24MB` compressed) to `402MB` (`125.74MB` compressed).

wstrange: Do you need all of docker, or just the client? Not sure if that is a seperate package or not..

mortent: Maybe we should consider publishing a separate image that includes docker if we can't find a way to reduce the size?

ZhuGongpu: @wstrange I found something [here](https://stackoverflow.com/a/43594065). It turns out the image size can be reduced into `150MB` (`52.81MB` compressed). @mortent 

#226 requires the docker client to be installed in the kpt container, and the docker client talks to the docker daemon through the mounted `docker.sock`. But I am not sure if there are any other cases that also need the docker daemon installed. @mortent Do you have any suggestions?

khrisrichardson: @ZhuGongpu there's a separate docker-cli package, so you wouldn't necessarily need to download and unarchive the tgz

```bash
apk info docker-cli
docker-cli-19.03.11-r0 description:
Docker CLI

docker-cli-19.03.11-r0 webpage:
https://www.docker.io/

docker-cli-19.03.11-r0 installed size:
72921088
```

PR #861, opened 2020-07-23
Add fn export example blueprint

ZhuGongpu:


ZhuGongpu:
@Liujingfang1 @monopole @frankfarzan

package-examples/function-export/resources/resources.yaml
```diff
@@ -0,0 +1,29 @@
...
+      labels:
+        app: nginx
+    spec:
+      containers:
+        - name: nginx
+          image: nginx:1.14.2
+          ports:
+            - containerPort: 80
+---
+apiVersion: v1
```
Liujingfang1, line 26: Let's move the namespace to be the first object. It won't affect anything though. It is consistent with the order that they should be applied.

package-examples/function-export/Kptfile
```diff
@@ -0,0 +1,6 @@
+apiVersion: kpt.dev/v1alpha1
+kind: Kptfile
+metadata:
+  name: function-export
```
Liujingfang1, line 4: name: function-export-example or function-export-blueprint

package-examples/function-export/README.md
```diff
@@ -0,0 +1,9 @@
+# Function Export Example
+
+The blueprint is part of Kpt Function Export Guide:
+
+- `resources/resources.yaml`: declares a `Deployment` and a `Namespace`.
+- `resources/constraints/`: declares constraints used by the `gatekeeper-validate` function.
+- `functions.yaml`: runs two functions from [Kpt Functions Catalog](../../catalog) declaratively:
+  - `gatekeeper-validate` enforces constraints over all resources.
+  - `label-namespace` adds a label to all Namespaces.
```
Liujingfang1, line 9: Let's put the order of functions as

 - `label-namespace` adds a label to all Namespaces.
 - `gatekeeper-validate` enforces constraints over all resources.

Since `label-namespace` is a transformer and `gatekeeper-validate` is a validator. The transformer should be before validator.

PR #864, opened 2020-07-24
Update kpt image in fn export

ZhuGongpu:
Update the kpt image used in `kpt fn export` to `gcr.io/kpt-dev/kpt:latest`.


ZhuGongpu:
@Liujingfang1 @mortent @monopole @frankfarzan

PR #898, opened 2020-08-02
update kpg live init docs

jijiew:
This PR updates the outdated kpt init docs.

see kubernetes-sigs/cli-utils#205 for the related change.


jijiew:
> Can we also include a description of how we decide on the name and namespace for the inventory if the user does not specify it with flags?
> Also, can we verify that the guides are also updated?

Done, also updated the example live init output in guides

mortent:
Looks good to me, but needs a rebase.

jijiew:
> Looks good to me, but needs a rebase.

Done.

PR #924, opened 2020-08-07
Add instructions for downloading kpt using gsutil 

janetkuo:


site/content/en/installation/binaries/_index.md
```diff
@@ -16,7 +16,15 @@ Download pre-compiled binaries.
 | [Windows (x64)][windows]
 
 ```sh
-# For linux/mac
+# Download binaries using gsutil:
+# Linux (x64)
+gsutil cp gs://kpt-dev/latest/linux_amd64/kpt .
```
prachirp, line 21: If we download kpt to the current directory then can we also update the `kpt version` command to be `./kpt version` or add a note about moving the kpt binary into the user's path?

mortent:
So I'm not sure if we should make this change. While we still publish the binaries directly to GS, we prefer if users download the binaries from github instead. That makes it easier for us to keep track of the popularity of kpt.

If it is inconvenient for users to download the `tar.gz` version currently published in the github release, is there an option where we can publish the binaries directly on github?

janetkuo:
I was following the instruction to download `kpt` from GCS bucket and thought using `gsutil` rather than clicking on the download link would be easier given that I want to use `kpt` in a remote machine. If we prefer users to download from GitHub, should we remove the link in the doc? 

I'd assume there's a way to track downloads from GCS bucket.

prachirp:
I'm more in favor of adding these instructions and track the logs on the gcs bucket (https://cloud.google.com/storage/docs/access-logs#should-you-use). Users can already get kpt through gcloud or homebrew so it makes sense to give them the option to download through gsutil.

site/content/en/installation/binaries/_index.md
janetkuo: Done

prachirp (approved):

mikebz:
@frankfarzan shall we merge this?

site/content/en/installation/binaries/_index.md
frankfarzan: Why gsutil? These are public artifacts, so we can use curl.

frankfarzan (changes requested):

site/content/en/installation/binaries/_index.md
janetkuo: Windows doesn't have `curl` by default.

janetkuo: Updated - use `curl` and removed instructions for Windows

janetkuo:
PTAL

mikebz:
@frankfarzan still relevant?  Should we merge?

mortent:
I would prefer if we made the binaries available through github instead of GCS. I have created an issue for this: https://github.com/GoogleContainerTools/kpt/issues/1349

mikebz:
@mortent question about your comment.  Is your concern that if we explain how to do this with GCS that we will increase the number of clients we would need to migrate?  If so we should close this issue.  If we will have to continue supporting GCS and this is just helpful for people who are want to know how to do this right now - maybe we should add the help?

frankfarzan:
@janetkuo Is this stale? I'm the process of deleting branches in the kpt repo. In general, PR should be based on your personal fork. Let me know if it's safe to delete this branch.

janetkuo:
Please feel free to delete the branch.

PR #943, opened 2020-08-14
Add ordering docs

mortent:
Our current documentation doesn't explain how `kpt live` sorts resources for apply and prune.

### DIFF
--- a/site/content/en/reference/live/apply/_index.md
+++ b/site/content/en/reference/live/apply/_index.md
@@ -91,6 +91,48 @@ is added to the name. Example:
 test-grouping-object-17b4dba8
 ```
 
+### Ordering
+
+`kpt live apply` will sort the resources before applying them. This makes sure
+namespaces are applied before resources going into the namespace, configmaps
+are applied before Deployments and StatefulSets, and other known dependencies
+between the builtin kubernetes resource types. Kpt does not analyze the actual
+dependencies between the resources, but sorts the resources based on the Kind
+of resources. Custom ordering of resources is not supported.
+
+During pruning, the same rules are used, but resources will be deleted in
+reverse order. Note that this does not wait for a resource to be deleted
+before continuing to delete the remaining resources.
+
+The following resources will be applied first in this order:
+
+* Namespace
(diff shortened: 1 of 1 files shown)


site/content/en/reference/live/apply/_index.md
```diff
@@ -91,6 +91,47 @@ is added to the name. Example:
...
+
+`kpt live apply` will sort the resources before applying them. This makes sure
+namespaces are applied before resources going into the namespace, configmaps
+are applied before Deployments and StatefulSets that might reference them, and
+other known dependencies between the builtin kubernetes resource types.
+Custom ordering of resources is not supported.
+
+During pruning, the same rules are used, but resources will be deleted in
+reverse order. Note that this does not wait for a resource to be deleted
+before continuing to delete the remaining resources.
```
frankfarzan, line 104: what's a concrete scenario where it's useful to have non-blocking deletion order?

site/content/en/reference/live/apply/_index.md
```diff
@@ -91,6 +91,47 @@ is added to the name. Example:
 test-grouping-object-17b4dba8
 ```
 
+### Ordering
+
+`kpt live apply` will sort the resources before applying them. This makes sure
+namespaces are applied before resources going into the namespace, configmaps
+are applied before Deployments and StatefulSets that might reference them, and
```
frankfarzan, line 98: "Deployments and StatefulSets that might reference them"

On first read, this sounds like you're inspecting Deployments for refs to ConfigMaps. But later on, it seems the ordering is purely based on kind and not their contents. May want to clarity this.

mortent: I have updated the text to make this clearer.

site/content/en/reference/live/apply/_index.md
mortent: Yeah, this helps in some situations, but it doesn't help when there are finalizers to delay deletion of resources. We do want to support that too, but it requires bigger changes.

frankfarzan (approved):

PR #965, opened 2020-08-20
Use latest kyaml and cmd/config

mortent:
Updates to the latest versions on master of kyaml and cmd/config.


monopole:
/lgtm
/approve

not sure why all the red; we'll see.

PR #973, opened 2020-08-24
Generate site with Cloud Build

mortent:
Updates the Cloud Build job that runs after every commit to generate the web site for kpt. It does this in the following way:
 * On every commit, the job (which already runs the tests) will also generate the site.
 * The cloud build job will then commit the changes back into the kpt repo. But instead of committing back to the master branch, it will commit the changes to the gh-pages branch.
 * The site will be served from the gh-pages branch instead of master, so we will be able to delete the docs folder from the master branch.

This means users will no longer need to include the generated content with the PRs. Since the job commits changes to a separate branch, we can run the job for a while to verify that it works before switching the site to be served from the gh-pages branch.

### DIFF
--- a/release/branch/cloudbuild.yaml
+++ b/release/branch/cloudbuild.yaml
@@ -13,73 +13,214 @@
 # limitations under the License.
 
 steps:
+  # Fetch the private ssh key for the kpt-robot account from secret manager.
+  - name: 'gcr.io/cloud-builders/gcloud'
+    entrypoint: 'bash'
+    args: [ '-c', 'mkdir -p /root/.ssh && gcloud secrets versions access latest --secret=kpt-robot-ssh > /root/.ssh/id_github' ]
+    volumes:
+      - name: home
+        path: /root
+
+  # Set up git with key and domain
+  - name: 'gcr.io/cloud-builders/git'
+    entrypoint: 'bash'
+    args:
+      - '-c'
+      - |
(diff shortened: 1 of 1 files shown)


frankfarzan:
Addressed https://github.com/GoogleContainerTools/kpt/issues/940

release/branch/cloudbuild.yaml
```diff
@@ -13,73 +13,214 @@
 # limitations under the License.
```
prachirp, line 14: Can you update the PR template and CONTRIBUTING.md?

release/branch/cloudbuild.yaml
```diff
@@ -13,73 +13,214 @@
...
+          git push origin gh-pages -f
+        else
+          echo "no changes, skipping push"
+        fi
+    dir: 'kpt/public'
+    volumes:
+      - name: home
+        path: /root
 
 timeout: '60m'
```
prachirp, line 226: How long does the job usually take?

release/branch/cloudbuild.yaml
mortent: We can do that when we switch to serve the site from the new branch. Until then, we should use the same process as before.

release/branch/cloudbuild.yaml
mortent: Around 25 minutes. Eventually we might consider updating the site in a separate job from building the binaries.

PR #997, opened 2020-09-01
Use Runner in apply command like other commands

seans3:
* Uses `GetApplyRunner` to retrieve cobra command, like other commands. Will allow updates to the `ApplyRunner` before retrieving the command in the future.


seans3:
I will implement this small PR once we update the API to `ApplyRunner` to take a `Provider` instead of a `Factory`.

PR #1004, opened 2020-09-03
Add github action for markdown linting

mortent:
This sets up a github action that runs the markdownlinter. It also updates the Makefile target for linting docs to run the markdownlinter-cli from inside node_modules under the `site` folder, instead of cloning the git repo.

In order to make the linting pass, this disables some of the linting rules by introducing a `markdownlint.json` file in the `site` folder the deactivates some of the rules where we currently have violations. Eventually we should clean this up and then active the rules again.


.github/workflows/markdown.yml
```diff
@@ -0,0 +1,26 @@
...
+  pull_request:
+    branches:
+      - master
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: xt0rted/markdownlint-problem-matcher@v1
+      - run: (cd site && npm run lint-check)
```
frankfarzan, line 26: nit: newline at end of files

frankfarzan (approved):
Are you planning to eventually use GH actions instead of Cloud Build for everything? I recommend converging on a single CI orchestrator, and GH Actions is used in other kpt repos.

PR #1023, opened 2020-09-11
Udpate fn run doc

Shell32-Natsu:
Add a section for using imperative and declarative functions at the same time.

### DIFF
--- a/site/content/en/reference/fn/run/_index.md
+++ b/site/content/en/reference/fn/run/_index.md
@@ -177,6 +177,58 @@ metadata:
       deferFailure: true
 ```
 
+## Scoping Rules
+
+Functions which are nested under some sub directory are scoped only to
+Resources under that same sub directory. This allows fine grain control over
+how functions are executed.
+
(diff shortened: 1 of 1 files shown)


site/content/en/reference/fn/run/_index.md
```diff
@@ -266,6 +266,12 @@ details in the following:
 
 * [Issue 757]
 
+#### Use Imperative and Declarative at the same time
```
frankfarzan, line 269: Title is misleading because this is not allowed

mortent:
Use the `docs:` prefix on the commit message(s) so we can filter out this from the release notes.

prachirp (changes requested):

prachirp (approved):
LGTM. Can you move Scoping Rules section above Imperative Run Specifics?

Shell32-Natsu:
@mortent @frankfarzan Please review.

frankfarzan (approved):

PR #1030, opened 2020-09-14
Move internal/kptfile to pkg/kptfile to make it vendorable

barney-s:
Moving kptfile schema and utils under pkg to make it vendorable.

Planning to vendor KPTFile schema in another tool that parses kptfile and extracts metadata from it.


barney-s:
Verified this on my tool. It works.

mortent:
We are doing a release today. Let's hold off merging this until we have cut the release.

seans3:
I think this looks good. Do we want to modify the current kptfile API, or is the current one sufficient? Once we start using the kptfile API outside of kpt, it will be much more difficult to change.

barney-s:
Iam not blocked on this. We can merge later.
I forked and am iterating on the fork.

Regarding the API sufficiency, it is too early to comment.
I will review the API and see if we can move non-critical functions back into internal.
That we we can reduce the surface and only expose well defined ones.

Alternatively we could have a pkg/kptfile/beta file for non well defined ones.

mortent (approved):
This looks good to me. As Sean mentioned, we are still pre-1.0 so there almost certainly will be changes to the API.

PR #1057, opened 2020-09-23
Adds ResourceGroupPathManifestReader

seans3:
* Adds `ManifestReader` which generates a `ResourceGroup` custom resource during path `Read()`.
* Updates `Kptfile` to add `Inventory` parameters.
* Currently not hooked up into `kpt`.

### DIFF
--- a/internal/live/rgpath.go
+++ b/internal/live/rgpath.go
@@ -1,0 +1,88 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package live
+
+import (
+	"fmt"
+	"strings"
+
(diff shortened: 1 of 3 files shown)


internal/live/rgpath.go
```diff
@@ -0,0 +1,92 @@
...
+			},
+			"spec": map[string]interface{}{
+				"resources": []interface{}{},
+			},
+		},
+	}
+	var invInfo = &resource.Info{
+		Namespace: namespace,
+		Name:      name,
+		Mapping: &meta.RESTMapping{
```
mortent, line 86: Do we need to set this? I think we handle this when we amend the infos with the client.

seans3: Removed.

mortent (approved):

PR #1068, opened 2020-09-24
ResourceGroup implementation of Provider interface

seans3:
* `ResourceGroup` implementation of the `Provider` interface
* `ResourceGroup` version of `InventoryClient`
* `ResourceGroup` version of `ManifestReader`


internal/live/rgprovider.go
```diff
@@ -0,0 +1,72 @@
...
+func (f *ResourceGroupProvider) ToRESTMapper() (meta.RESTMapper, error) {
+	return f.factory.ToRESTMapper()
+}
+
+// ManifestReader returns the ResourceGroup inventory object version of
+// the ManifestReader.
+func (f *ResourceGroupProvider) ManifestReader(reader io.Reader, args []string) manifestreader.ManifestReader {
+	namespace, enforceNamespace, err := f.factory.ToRawKubeConfigLoader().Namespace()
+	if err != nil {
+		namespace = metav1.NamespaceDefault
```
mortent, line 50: Why continue here? Shouldn't we just return an error?

seans3: Fixed.

mortent (approved):

PR #1069, opened 2020-09-24
Update cmd/config, kyaml and cli-utils dependencies

mortent:
Update the dependencies

### DIFF
--- a/commands/livecmd.go
+++ b/commands/livecmd.go
@@ -64,7 +64,7 @@ func GetLiveCommand(name string, f util.Factory) *cobra.Command {
 	applyCmd.Long = livedocs.ApplyShort + "\n" + livedocs.ApplyLong
 	applyCmd.Example = livedocs.ApplyExamples
 
-	previewCmd := preview.GetPreviewRunner(f, ioStreams).Command
+	previewCmd := preview.PreviewCommand(f, ioStreams)
 	previewCmd.Short = livedocs.PreviewShort
 	previewCmd.Long = livedocs.PreviewShort + "\n" + livedocs.PreviewLong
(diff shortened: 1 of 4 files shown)


commands/livecmd.go
```diff
@@ -64,7 +64,7 @@ func GetLiveCommand(name string, f util.Factory) *cobra.Command {
 	applyCmd.Long = livedocs.ApplyShort + "\n" + livedocs.ApplyLong
 	applyCmd.Example = livedocs.ApplyExamples
 
-	previewCmd := preview.GetPreviewRunner(f, ioStreams).Command
+	previewCmd := preview.PreviewCommand(f, ioStreams)
```
seans3, line 67: Let's keep the `Runner`. We will need it to inject the `ResourceGroup` versions of structures with the `ResourceGroupProvider`.

commands/livecmd.go
```diff
@@ -74,7 +74,7 @@ func GetLiveCommand(name string, f util.Factory) *cobra.Command {
 	diffCmd.Long = livedocs.DiffShort + "\n" + livedocs.DiffLong
 	diffCmd.Example = livedocs.DiffExamples
 
-	destroyCmd := destroy.GetDestroyRunner(f, ioStreams).Command
+	destroyCmd := destroy.DestroyCommand(f, ioStreams)
```
seans3, line 77: Same

seans3 (changes requested):

commands/livecmd.go
mortent: But we can just update that when we have the provider for ResourceGroup ready. This already is using the default provider.

seans3 (approved):
/lgtm

PR #1071, opened 2020-09-24
kpt init command updates Kptfile inventory parameters

seans3:
* New `kpt live init` command set the `Inventory` parameters in the `Kptfile`.
* Not connected into `kpt` command yet.
* Adds unit tests for `KptInitOptions`.

### DIFF
--- a/commands/initcmd.go
+++ b/commands/initcmd.go
@@ -1,0 +1,183 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package commands
+
+import (
+	"crypto/sha1"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/GoogleContainerTools/kpt/pkg/kptfile"
+	"github.com/GoogleContainerTools/kpt/pkg/kptfile/kptfileutil"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util/i18n"
+	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/config"
+)
+
+const defaultInventoryName = "inventory"
+
+var invExistsError = `ResourceGroup configuration has already been created. Changing
+them after a package has been applied to the cluster can lead to
+undesired results. Use the --force flag to suppress this error.
+`
+
+// KptInitOptions encapsulates fields for kpt init command. This init command
+// fills in inventory values in the Kptfile.
+type KptInitOptions struct {
+	factory     cmdutil.Factory
+	ioStreams   genericclioptions.IOStreams
+	dir         string // Directory with Kptfile
+	force       bool   // Set inventory values even if already set in Kptfile
+	name        string // Inventory object name
+	namespace   string // Inventory object namespace
+	inventoryID string // Inventory object unique identifier label
+}
+
+// NewKptInitOptions returns a pointer to an initial KptInitOptions structure.
+func NewKptInitOptions(f cmdutil.Factory, ioStreams genericclioptions.IOStreams) *KptInitOptions {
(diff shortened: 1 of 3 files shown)


commands/initcmd.go
```diff
@@ -0,0 +1,136 @@
...
+// NewKptInitOptions returns a pointer to an initial KptInitOptions structure.
+func NewKptInitOptions(f cmdutil.Factory, ioStreams genericclioptions.IOStreams) *KptInitOptions {
+	io := config.NewInitOptions(f, ioStreams)
+	return &KptInitOptions{
+		defaultInitOptions: io,
+	}
+}
+
+// Complete fills in fields for KptInitOptions based on the passed "args".
+func (io *KptInitOptions) Complete(args []string) error {
```
mortent, line 43: I'm not sure why this is split out into `Complete` and `Run` functions. Seems like we can do all in one.

commands/initcmd.go
```diff
@@ -0,0 +1,136 @@
...
+	force              bool   // Set inventory values even if already set in Kptfile
+	name               string // Inventory object name
+	namespace          string // Inventory object namespace
+	inventoryID        string // Inventory object unique identifier label
+	defaultInitOptions *config.InitOptions
+}
+
+// NewKptInitOptions returns a pointer to an initial KptInitOptions structure.
+func NewKptInitOptions(f cmdutil.Factory, ioStreams genericclioptions.IOStreams) *KptInitOptions {
+	io := config.NewInitOptions(f, ioStreams)
```
mortent, line 36: The naming here is a little confusing. Can we chose a different name on this variable than we use in the function definitions. It is the same name, but different types.

commands/initcmd.go
```diff
@@ -0,0 +1,136 @@
...
+
+// KptInitOptions encapsulates fields for kpt init command. This init command
+// fills in inventory values in the Kptfile.
+type KptInitOptions struct {
+	dir                string // Directory with Kptfile
+	force              bool   // Set inventory values even if already set in Kptfile
+	name               string // Inventory object name
+	namespace          string // Inventory object namespace
+	inventoryID        string // Inventory object unique identifier label
+	defaultInitOptions *config.InitOptions
```
mortent, line 31: I don't understand why we are nesting the old inventory template version inside the new Kptfile version. If we need to share code, we should just split it out into a separate library that can be shared.

commands/initcmd.go
```diff
@@ -0,0 +1,136 @@
...
+		return err
+	}
+	io.dir = io.defaultInitOptions.Dir
+	// Add random suffix to default inventory name to avoid collisions
+	if io.name == defaultInventoryName {
+		randomSuffix := common.RandomStr(time.Now().UTC().UnixNano())
+		io.name = fmt.Sprintf("%s-%s", defaultInventoryName, randomSuffix)
+	}
+	io.namespace = strings.TrimSpace(io.defaultInitOptions.Namespace)
+	io.inventoryID = strings.TrimSpace(io.defaultInitOptions.InventoryID)
```
mortent, line 54: So the id as we want to use for the Kptfile version should be based on the actual values, and not just a UID like we do with the inventory template solution.

seans3: Please explain this. I do not understand your comment.

mortent: Based on the discussions and the design, we want the generated id to be a combination of the hash of the name+namespace and the timestamp when this command is run. That will allow us to use the id as a checksum and detect situations when either the name or the namespace has been changed.

commands/initcmd.go
seans3: Fixed.

commands/initcmd.go
seans3: Fixed.

commands/initcmd.go
seans3: Fixed.

commands/initcmd.go
seans3: Initial implementation of `generateID()` function to create the `namespace/name`+`unixtimestamp` identifier (plus unit tests).

mortent (approved):

PR #1072, opened 2020-09-25
ManifestReader based on kyaml that handles nested packages and ignorefiles

mortent:
This leverages the LocalManifestReader from kyaml to read manifests. Similar to how we read files for the cfg commands, it will first identify all the nested packages and then read each of them. This will also automatically handle ignorefiles once #1024 is merged.


mikebz:
is this still relevant?  Given that we are on the `next` branch?

PR #1073, opened 2020-09-25
Docs: Add top-level index page for KRM functions.

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1066

image

image


site/content/en/guides/consumer/function/catalog/_index.md
```diff
@@ -0,0 +1,7 @@
+---
+title: "Function Catalog"
+linkTitle: "Function Catalog"
+type: docs
```
prachirp, line 4: Add weight field so this page appears above exporting workflows page

prachirp (changes requested):
Can you update catalog pages with sops function and new examples too? I would consider sops as a transformer fn.

site/content/en/guides/consumer/function/catalog/_index.md
etefera: Done.

prachirp (approved):
LGTM, will be a separate change to update catalog fns and examples

frankfarzan:
I don't see changes to this page:

https://googlecontainertools.github.io/kpt/guides/consumer/function/

Which is the point of entry for catalog at this point.

frankfarzan (approved):

PR #1080, opened 2020-09-29
Docs: Remove kustomize-build from catalog.

etefera:
Addresses https://github.com/GoogleContainerTools/kpt/issues/1078


frankfarzan:
Also need to delete the actual function code:

https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/functions/ts/src

frankfarzan (approved):

PR #1101, opened 2020-10-01
docs: Hide outdated guides

mortent:
It is unclear if the blueprint guide is still aligned with the direction of kpt, so this PR hides it. It also hides the variants guide since it doesn't have any content.


prachirp (approved):
LGTM, I learned about the draft field.

frankfarzan:
I suggest just deleting it. Git history is sufficient to revive it if necessary.

mortent:
@frankfarzan updated

frankfarzan (approved):

PR #1121, opened 2020-10-12
StreamManifestReader for ResourceGroup inventory

seans3:
* Creates `ResourceGroupStreamManifestReader`. This is a stream manifest reader which transforms the `Kptfile` into a `ResourceGroup` custom resource as the `Inventory` object.
* Adds unit tests.
* Not yet connected to kpt binary, so it will not yet affect kpt.
* Necessary for the `ResourceGroupProvider` to implement the `ResourceGroup` as the `Inventory` object.

### DIFF
--- a/internal/live/rgstream.go
+++ b/internal/live/rgstream.go
@@ -1,0 +1,92 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package live
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+
+	"github.com/GoogleContainerTools/kpt/pkg/kptfile"
+	"k8s.io/cli-runtime/pkg/resource"
+	"k8s.io/klog"
+	"sigs.k8s.io/cli-utils/pkg/manifestreader"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// ResourceGroupStreamManifestReader encapsulates the default stream
+// manifest reader.
+type ResourceGroupStreamManifestReader struct {
+	streamReader *manifestreader.StreamManifestReader
+}
+
+var ResourceSeparator = []byte("\n---\n")
+
+// Read reads the manifests and returns them as Info objects.
+// Transforms the Kptfile into the ResourceGroup inventory object,
+// and appends it to the rest of the standard StreamManifestReader
+// generated objects. Returns an error if one occurs. If the
+// ResourceGroup inventory object does not exist, it is NOT an error.
+func (p *ResourceGroupStreamManifestReader) Read() ([]*resource.Info, error) {
+	var resourceBytes bytes.Buffer
+	_, err := io.Copy(&resourceBytes, p.streamReader.Reader)
+	if err != nil {
+		return []*resource.Info{}, err
+	}
+	// Split the bytes into resource configs, and if the resource
+	// config is a Kptfile, transform it into a ResourceGroup object.
+	var rgInfo *resource.Info
+	var filteredBytes bytes.Buffer
+	resources := bytes.Split(resourceBytes.Bytes(), ResourceSeparator)
+	for _, r := range resources {
+		if !isKptfile(r) {
+			r = append(r, ResourceSeparator...)
+			_, err := filteredBytes.Write(r)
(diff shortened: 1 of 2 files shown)


internal/live/rgstream.go
```diff
@@ -0,0 +1,88 @@
...
+	if rgInfo != nil {
+		infos = append(infos, rgInfo)
+	}
+	return infos, err
+}
+
+var kptFileTemplate = kptfile.KptFile{ResourceMeta: kptfile.TypeMeta}
+
+// isKptfile returns true if the passed resource config is a Kptfile; false otherwise
+func isKptfile(resource []byte) bool {
```
mortent, line 68: It is not clear to me exactly how we are making this check. Is it checking based on the presence of the correct GVK in the resource, or are we just checking that the resource can be decoded into the Kpt struct?

seans3: We're checking that it can be decoded into the Kpt struct. This idiom is used elsewhere in the kptfileutils.

internal/live/rgstream.go
```diff
@@ -0,0 +1,88 @@
...
+			rgInfo, err = transformKptfile(r)
+			if err != nil {
+				return []*resource.Info{}, err
+			}
+		}
+	}
+	// Reset the stream reader, and generate the infos. Append the
+	// ResourceGroup inventory info if it exists.
+	p.streamReader.Reader = bytes.NewReader(filteredBytes.Bytes())
+	infos, err := p.streamReader.Read()
```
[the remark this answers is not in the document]
seans3: Kptfile has the wrong file extension, so the `Builder` just ignores it and does not create an `Info` for it. The `Builder` will only build infos for `.yaml` and `.json`.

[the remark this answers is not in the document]
seans3: Actually, maybe the stream reader is different than the path reader. Let me look further into this.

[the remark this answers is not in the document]
seans3: The error is:
```
can't find scope for resource Kptfile.kpt.dev test1
```
This is an error in the `SetNamespaces()` function for the common `ManifestReader`. This is the comment directly above the error block:
```
       // If we get here, it means the resource does not have the namespace                                                                                                                                  
        // set and we didn't find the resource type in the RESTMapper. As                                                                                                                                     
        // a last try, we look at all the CRDS that are part of the set and                                                                                                                                   
        // see if we get a match on the resource type. If so, we can determine                                                                                                                                
        // from the CRD whether the resource type is cluster-scoped or                                                                                                                                        
        // namespace-scoped. If it is the latter, we set the namespace                                                                                                                                        
        // to the provided default.                               
```
So I think this is a problem with the GVK not resolving with the RESTMapper.

internal/live/rgstream.go
mortent: That seems to be a different use-case. In that example we have already decided that the file is a Kptfile and we just want to parse it into a struct. I worry that any KRM resource that contains only typemeta and objectmeta will be identfied as a Kptfile with this test.

internal/live/rgstream.go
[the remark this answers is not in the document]
mortent: Yeah, that is what I expected. I think the reuse of the existing StreamManifestReader is a bit strange here,

internal/live/rgstream.go
seans3: I think the linked code is the correct code to validate the bytes in the `io.Reader` can be marshalled into the golang `Kptfile` struct correctly. I've added unit tests to prove this works.

mortent: So the example I had in mind here is whether a resource such as:
```
apiVersion: foo/v1
kind: FooBar
metadata:
  name: test1
```
will be identified as a Kptfile. I'm pretty sure the answer is yes with the current `isKptfile` check, but that would be incorrect.

seans3: You're right. This does incorrectly pass. I've added another check, and added a unit tests to show it works.

We are not doing this validation in `kptfileutils.ReadFile()`, and we should add it wherever we parse the kptfile. Eventually, these helper functions should move to `kptfileutils`, but I believe that should be another PR.

PR #1142, opened 2020-10-20
move ResourceGroupProvider to pkg/live

Liujingfang1:
This PR moves the `live` package from `internal/live` to `pkg/live` so that I can start testing it in ConfigSync.

/assign @mortent @seans3


mortent:
Looks good to me. I'll let @seans3 do the final approval.

seans3 (approved):
/lgtm

PR #1150, opened 2020-10-26
include labels and annotations when generating ResourceGroup

Liujingfang1:
This change is to inject the labels and annotations from Kptfile into ResourceGroup


seans3:
Looks good. Thanks for the tests. I think this needs to be rebased because of your previously merged PR.

seans3 (approved):
/lgtm

PR #1166, opened 2020-10-29
removing travis build instructions

### LINKED ISSUE #1153: Migrate off deprecated travis-ci.org
Both kpt and kpt-functions-sdk use travis-ci.org, which is deprecated. We should migrate off, taking into account guidance from the Github Admin team.

mikebz:
This is a clean up pr that removes the Travis CI build instructions

Resolves #1153 

WARNING: we still have builds running https://travis-ci.com/github/GoogleContainerTools/kpt


mortent:
I assume we are covering the tasks in the travis build with github actions? Can we add a link to another job that make the same checks?

mikebz:
Let me double check if the same checks are covered.  From what @prachirp told me it seems like it was old/prohibited/retired but good thing to double check.

mikebz:
@mortent it looks like the Go action does 
```
make all
```

which is more than what the travis file did with a default target of just build.  So the Github action does more:
```
all: license fix vet fmt lint test build buildall tidy
```

It looks like until we turn off TravisCI as a check we will have PRs that are "hung".  Travis without the travis.yml file doesn't know what to do and you can see it in this PR.

mortent (approved):

mortent:
@mikebz I'll merge this without the input from Travis CI and then remove Travis from the project.

mikebz:
🙇  thanks for the look @mortent

PR #1177, opened 2020-11-03
Dual-delegating provider allows ConfigMap and ResourceGroup inventory

seans3:
* Adds `DualDelegatingProvider`; a `Provider` which handles both `ResourceGroup` and `ConfigMap` at the same time.
* `InventoryClient` is created from knowledge of what type of `Inventory` has read with `ManifestReader`.
* Removes errors when `Kptfile` can not become `ResourceGroup` inventory object, since we delegate to the `ResourceGroupProvider` to generate the objects. And it is valid to only have a `ConfigMap` as an inventory object.
* Is NOT currently connected, so it can not affect the current functionality.
* Adds unit tests.

### DIFF
--- a/pkg/live/dual-delegating-provider.go
+++ b/pkg/live/dual-delegating-provider.go
@@ -1,0 +1,130 @@
+// Copyright 2020 Google LLC.
+// SPDX-License-Identifier: Apache-2.0
+
+package live
+
+import (
+	"fmt"
+	"io"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog"
+	"k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
+	"sigs.k8s.io/cli-utils/pkg/manifestreader"
+	"sigs.k8s.io/cli-utils/pkg/provider"
+)
+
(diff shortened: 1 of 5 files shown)


pkg/live/dual-delegating-provider_test.go
```diff
@@ -0,0 +1,155 @@
+// Copyright 2020 The Kubernetes Authors.
```
mikebz, line 1: should the attribution here be to the Kubernetes Authors?  This is in kpt repo.

seans3: Good catch. It should be: `// Copyright 2020 Google LLC`

seans3: Fixed for all files in this PR. A small subsequent PR will fix others.

pkg/live/dual-delegating-provider.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil, err
+	}
+	objs, err := r.Read()
+	if err != nil {
+		return nil, err
+	}
+	klog.V(4).Infof("ManifestReader read %d objects", len(objs))
+	rgInv := findResourceGroupInv(objs)
+	if rgInv != nil {
+		cp.wrapInv = WrapInventoryObj
```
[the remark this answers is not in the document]
seans3: Done. Let me know what you think.

mortent (approved):

Issue #1176, opened 2020-11-02
Support enabling network access when using `kpt fn run --fn-path`


yuwenma:
kpt version: 0.36.1

declarative kpt function file
validations/fn.yaml
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: Kubeval
  annotations:
    config.kubernetes.io/function: |-
      container:
        image: gcr.io/kpt-functions/kubeval
```

hydrated/service.yaml
```yaml
apiVersion: v1
kind: Service
metadata:
  name: frontend
  labels:
    app: guestbook
    tier: frontend
spec:
  type: LoadBalancer
  ports:
  - port: 80
  selector:
    app: guestbook
    tier: frontend
``` 

Error message:
```bash
kpt fn run hydrated/ --fn-path validations/ --network
[ERROR] Failed to parse raw kubeval output:
Unexpected token E in JSON at position 0

ERR  - stdin: Failed initializing schema https://kubernetesjsonschema.dev/master-standalone/service-v1.json: Get "https://kubernetesjsonschema.dev/master-standalone/service-v1.json": dial tcp: lookup kubernetesjsonschema.dev on 192.168.65.1:53: dial udp 192.168.65.1:53: connect: network is unreachable
 in object 'v1/Service//frontend' in file hydrated.yaml
error: exit status 1
```

I also tried 
1. run `kpt fn run hydrated/ --fn-path validations/ --network=true`
2. add "network: true" in the validations/fn.yaml
```yaml
...

  annotations:
    config.kubernetes.io/function: |-
      container:
        image: gcr.io/kpt-functions/kubeval
        network: true
```
3. add "network.required=true" in the validations/fn.yaml like @dflemstr did in issues #334 
```yaml
...
      container:
        image: gcr.io/kpt-functions/kubeval
        network:
          required: true
```

None of the above works. I used to have the same command works, not sure what happened.

@phanimarupaka @Shell32-Natsu @mortent

fsommar:
I can confirm that I'm seeing the same behavior in 0.36.1 where the previously working combination of

```yaml
  annotations:
    config.kubernetes.io/function: |-
      container:
        image: gcr.io/kpt-functions/kubeval
        network: 
          required: true
```

and `--network` flag are no longer working.

Shell32-Natsu:
What you need to do is using `--network` flag with the function config like this

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: Kubeval
[...]

PR #1185, opened 2020-11-06
docs: update network option in function config

Shell32-Natsu:
Update the `network` option in the function config. Now it's a single boolean.

#1176


prachirp (approved):
Has the "--network" flag changed?

Shell32-Natsu:
The `--network` flag is not changed.

frankfarzan (approved):

PR #1217, opened 2020-11-16
WIP: Experimental support for kustomize in kpt

justinsb:
Consider this a placeholder for evolving the experience for kustomize
users and packages in kpt.  I started with the absolute minimum, which
is to make `kpt render` behave similarly to `kustomize build`.

Feature-flagged behind exporting an env var: KPT_EXPERIMENTS=render


justinsb:
Let's figure out what a more complete integration looks like here :-)

cc @frankfarzan @mortent @mikebz @phanimarupaka @monopole @seans3

frankfarzan:
This is not aligned with the planned roadmap in kpt. Spirit of this work should be tracked as design proposal instead.

justinsb:
Can we discuss in the community meeting?

PR #1243, opened 2020-12-01
Move test setup manager

mortent:
The `TestSetupManager` is useful for general testing of kpt pkg functionality. This moves it out of the `update_test.go` file into a separate package.


internal/util/update/update_test.go
```diff
@@ -24,13 +24,11 @@ import (
 	"github.com/GoogleContainerTools/kpt/internal/gitutil"
 	"github.com/GoogleContainerTools/kpt/internal/testutil"
 	"github.com/GoogleContainerTools/kpt/internal/testutil/pkgbuilder"
-	"github.com/GoogleContainerTools/kpt/internal/util/get"
+	testmgr "github.com/GoogleContainerTools/kpt/internal/util/testutil"
```
droot, line 27: I see testutil pkg already being there, any reason we can't move this there.

Also,

testutil.TestSetupManager{} reads better than testmgr.TestSetupManager :) if we are able to move it in testutil.

mortent: Updated the PR by moving the `TestSetupManager` into `github.com/GoogleContainerTools/kpt/internal/testutil` and eliminating the package at `github.com/GoogleContainerTools/kpt/internal/util/testutil`.

phanimarupaka (approved):

PR #1244, opened 2020-12-01
Adds kpt live install-resource-group command

seans3:
* Adds `kpt live install-resource-group` command to add the ResourceGroup CRD to the cluster.
* Adds end-to-end test to test `kpt live install-resource-group` command
* Refactors `kpt live migrate` command to place the common CRD installation code into `inventoryrg.go`
* Refactors cluster deletion/creation for clean test suites in `end-to-end-test.sh`

Example output from relevant part of end-to-end test:
```
Setting Up Test Suite...
Deleting kind cluster...
Deleting kind cluster...COMPLETED
Creating kind cluster...
Creating kind cluster...COMPLETED
Setting Up Test Suite...COMPLETED

Testing kpt live install-resource-group
kpt live install-resource-group
.....SUCCESS
```

### DIFF
--- a/commands/installrg.go
+++ b/commands/installrg.go
@@ -1,0 +1,58 @@
+// Copyright 2020 Google LLC.
+// SPDX-License-Identifier: Apache-2.0
+
+package commands
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/GoogleContainerTools/kpt/pkg/live"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util/i18n"
+)
+
+// InstallRGRunner encapsulates fields for the kpt live install-resource-group command.
+type InstallRGRunner struct {
+	Command   *cobra.Command
+	ioStreams genericclioptions.IOStreams
+	factory   cmdutil.Factory
+}
+
+// GetInstallRGRunner returns a pointer to an initial InstallRGRunner structure.
+func GetInstallRGRunner(factory cmdutil.Factory, ioStreams genericclioptions.IOStreams) *InstallRGRunner {
+	r := &InstallRGRunner{
+		factory:   factory,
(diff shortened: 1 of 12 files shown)


commands/installrg.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// GetInstallRGRunner returns a pointer to an initial InstallRGRunner structure.
+func GetInstallRGRunner(factory cmdutil.Factory, ioStreams genericclioptions.IOStreams) *InstallRGRunner {
+	r := &InstallRGRunner{
+		factory:   factory,
+		ioStreams: ioStreams,
+	}
+	cmd := &cobra.Command{
+		Use:                   "install-resource-group",
```
Liujingfang1, line 31: In the `init` command, should kpt install `ResourceGroup` when the CRD is not installed? If it fails, it prints the instruction to users for running `install-resource-group` by an admin.

If `init` doesn't install `ResourceGroup` by default, which `kpt` command gives instruction to run `kpt live install-resource-group`? I think both `init` and `apply` need to do this. The package can be applied to a different cluster after `init` or the installed CRD can be removed, which will break the `apply` command. By giving instructions for the `install-resource-group` command, it is clear for users to take actions when the cluster doesn't `ResourceGroup` installed.

frankfarzan: 'init' command shouldn't require talking to a cluster. With GitOps, you may have such a flow:

1. kpt live init
2. Commit to Git
3. Async controller syncs package

'apply' command is ultimately responsible for checking its preconditions. As a best effort, it should attempt to install the CRD if needed. If it fails, it should output a clear error message fo why it failed (It could be authz issues, or admission controller blocking it, etc.). e.g.:

_Failed to install XXX CRD: <underlying error>. You can manually install the CRD by invoking 'kpt live install...'._

Liujingfang1: That sounds good. For GitOps users with ConfigSync, the CRD should have been installed by the apply is triggered. For GitOps users not using ConfigSync, we can add this precondition to the kpt doc.

mortent (approved):

PR #1254, opened 2020-12-03
change the inventory field to pointer

Liujingfang1:
This change is to use `inventory==nil` to determine if the inventory information is empty.
This check is also needed in Config Sync.


seans3 (approved):
/lgtm

PR #1268, opened 2020-12-09
Setup golang build action for 'next' branch

mortent:
Makes sure we run the build on PRs against the `next` branch.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -15,23 +15,23 @@
 name: Go
 on:
   # Trigger the workflow on push or pull request,
-  # but only for the master branch
+  # but only for the 'next' branch
   push:
     branches:
-      - master
+      - next
   pull_request:
     branches:
-      - master
+      - next
 jobs:
 
   build:
     name: Build
     runs-on: ubuntu-latest
     steps:
-    - name: Set up Go 1.13
+    - name: Set up Go 1.15
       uses: actions/setup-go@v1
       with:
-        go-version: 1.13
+        go-version: 1.15
       id: go
     - name: Check out code into the Go module directory
       uses: actions/checkout@v1


mikebz:
Can we consider using the `main` branch and kill two birds with one stone?

mortent:
I think we should consider moving to a `main` branch, but I don't think we want to have both a `master` and a `main` branch at the same time. The plan is to move to use `main` when we reconcile the `next` branch with `master`.

.github/workflows/go.yml
```diff
@@ -15,23 +15,23 @@
 name: Go
 on:
   # Trigger the workflow on push or pull request,
-  # but only for the master branch
+  # but only for the 'next' branch
   push:
```
droot, line 19: We don't do any push from github actions.. right ?

droot (approved):
/lgtm

runewake2:
Re moving to `main`. GitHub has said they plan to [introduce tooling for migrating `master` -> `main` for existing projects](https://github.com/github/renaming#later-this-year-seamless-move-for-existing-repositories-). They mention planning to deliver this tooling by EoY.

.github/workflows/go.yml
```diff
@@ -15,23 +15,23 @@
...
     branches:
-      - master
+      - next
```
runewake2, line 21: To prevent CI drift between the branches it may make sense to include all CI covered branches so the file can stay consistent between them.

Ex:

```diff
  branches:
    - master
+   - next
```

This shouldn't impact [execution of GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#:~:text=The%20following%20steps%20occur%20to%20trigger%20a%20workflow,%22Using%20environment%20variables.%22).

This way if new steps are introduced in `master` or `next` they can be introduced to the other more easily.

PR #1282, opened 2020-12-10
kpt run CRD installation before live apply

seans3:
* Adds `ApplyRunnerWrapper` in `kpt` to run a pre-step before invoking the cli-utils `ApplyRunner`.
* if the `ResourceGroup` inventory environment variable exists, `kpt live apply` installs the `ResourceGroup` custom resource definition before running the rest of cli-utils apply.
* Tested by running the entire e2e test suite...passed.


mortent (approved):
Do we want this in `master`? If so, we also need to cherry-pick it into `next`.

PR #1294, opened 2020-12-14
Update error handling for e2e test; adds tests

seans3:
* Fixes `kpt live` e2e test error handling. Example:
```
...
Testing basic ConfigMap init with flags
kpt live init e2e/live/testdata/test-case-1a
.E.............ERROR
error: expected missing text "foo bar"
...
```
* Adds tests for the flags of `kpt live init` command
* Adds tests for `kpt live preview --destroy` command
* Adds TODO list to keep track of future improvements
* Fixes wrong flag in `Examples:`


mortent (approved):
We can merge this, but I'm a little concerned about how we are going to maintain this long-term. It is quite a bit of bash script and example files by now.

PR #1295, opened 2020-12-14
Updates migrate no ConfigMap error to be success

seans3:
* Intercept `NoInventoryObjError`, since this means the migration has already previously run and successfully completed.
* Adds a test case in the end-to-end test to prove this works.

### DIFF
--- a/commands/migratecmd.go
+++ b/commands/migratecmd.go
@@ -15,6 +15,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/klog"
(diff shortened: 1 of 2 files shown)


commands/migratecmd.go
```diff
@@ -123,6 +124,12 @@ func (mr *MigrateRunner) Run(reader io.Reader, args []string) error {
 	// Retrieve the current ConfigMap inventory objects.
 	cmInvObj, err := mr.retrieveConfigMapInv(bytes.NewReader(stdinBytes), args)
 	if err != nil {
+		if _, ok := err.(inventory.NoInventoryObjError); ok {
+			// No ConfigMap inventory means the migration has already run before.
+			klog.V(4).Infoln("swallowing no ConfigMap inv error")
```
Liujingfang1, line 129: inv -> inventory

mortent (approved):

PR #1302, opened 2020-12-14
Breaking: Remove functions from kptfile

frankfarzan:
This is an undocumented behavior that will be superseded by hydration redesign.


mortent:
Do we have a clear understanding of how this feature has been used and how those use-cases can be addressed with the hydration redesign?
Code looks good.

frankfarzan:
No, such undocumented behavior is a bug IMO, so I would flip the burden of proof. I haven't seen this UX discussed anywhere in external or internal docs. The spirit of this codepath (A package has functions, and interaction of setters and functions) needs to be addressed by v1 hydration story regardless.

In addition, given this is on `next` branch, we shouldn't hesitate to do such cleanups.

pkg/kptfile/pkgfile.go
```diff
@@ -270,10 +252,9 @@ func shouldRemoveValue(updatedDef, localDef, originalDef *yaml.MapNode, key stri
 type Dependency struct {
 	Name            string `yaml:"name,omitempty"`
 	Upstream        `yaml:",inline,omitempty"`
-	EnsureNotExists bool       `yaml:"ensureNotExists,omitempty"`
-	Strategy        string     `yaml:"updateStrategy,omitempty"`
-	Functions       []Function `yaml:"functions,omitempty"`
-	AutoSet         bool       `yaml:"autoSet,omitempty"`
+	EnsureNotExists bool   `yaml:"ensureNotExists,omitempty"`
+	Strategy        string `yaml:"updateStrategy,omitempty"`
```
droot, line 256: I see it now, so one can specify in the dependencies what functions to invoke (set is one trigger point).

droot (commented):
Took a quick look, looks good to me.

I think we should add special label or prefix the title of the PR (:breaking) to track these changes. It will help us in the end to screen all breaking changes that are superseded by new behavior.

mortent (approved):

commands/configcmd.go
```diff
@@ -149,22 +148,13 @@ func SetCommand(parent string) *cobra.Command {
 	kustomizeCmd.Example = cfgdocs.SetExamples
 	kustomizeCmd.SilenceUsage = true
 	kustomizeCmd.SilenceErrors = true
-	var autoRun bool
-	setCmd.Flags().BoolVar(&autoRun, "auto-run", true,
-		`Automatically run functions after setting (if enabled for the package)`)
```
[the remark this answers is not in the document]
frankfarzan: Instead of commenting on this parameter, let me uplevel and discuss the spirit of this featur. If resources in a package are parameterized, and the user sets those parameters, then the package needs to be re-hydrated i.e. "reconciled". The hydration step is mandatory after _any_ change to resources, not just setters e.g. if you vim edit a file, you need to re-run hydration. Depending on setter and hydration interaction model, we may need to do something similar in `next branch`, where:

kpt cfg set foo bar

is porcelain for:

1. Sett parameter foo to value bar
2. Rerun the hydration pipeline

pkg/kptfile/pkgfile.go
frankfarzan: Per comment above, we need to uplevel this instead of having a separate mechanism for specifying functions in the Kptfile.

PR #1307, opened 2020-12-16
trigger apply in migrate command to fix the owning-inventory annotation

### LINKED ISSUE #1306: Regression: pruning fails for ResourceGroup inventory object
* After a recent PR: https://github.com/kubernetes-sigs/cli-utils/pull/291, pruning fails for `ResourceGroup` inventory object after migration.

Steps to reproduce:
1. set environment variable `export RESOURCE_GROUP_INVENTORY=1`
2. start/restart cluster with kind: `kind delete cluster; kind create cluster`
3. setup the "app": `cp -f e2e/live/testdata/template-rg-namespace.yaml e2e/live/testdata/migrate-case-1a/inventory-template.yaml`
5. apply the "app" with ConfigMap inventory: `kpt live apply e2e/live/testdata/migrate-case-1a`
6. migrate the "app" to ResouceGroup: `kpt live migrate e2e/live/testdata/migrate-case-1a`
6. copy Kptfile to next version: `cp -f e2e/live/testdata/migrate-case-1a/Kptfile e2e/live/testdata/migrate-case-1b`
7. apply/prune the next version of the "app":  `kpt live apply e2e/live/testdata/migrate-case-1b`

Output:
```
namespace/test-rg-namespace unchanged
pod/pod-b unchanged
pod/pod-c unchanged
pod/pod-d created
4 resource(s) applied. 1 created, 3 unchanged, 0 configured, 0 failed
pod/pod-c prune skipped
pod/pod-b prune skipped
pod/pod-a prune skipped
namespace/test-rg-namespace prune skipped
0 resource(s) pruned, 4 skipped, 0 failed
```

All the resources are erroneously `prune skipped`

Expected Output:
```
namespace/test-rg-namespace unchanged
pod/pod-b unchanged
pod/pod-c unchanged
pod/pod-d created
4 resource(s) applied. 1 created, 3 unchanged, 0 configured, 0 failed
pod/pod-a pruned
1 resource(s) pruned, 0 skipped, 0 failed
```

Reason: It looks like a new `CanPrune` function is erroneously identifying all resources as having a different inventory object
```
I1215 16:24:38.369507 1266578 prune.go:230] skip pruning object that doesn't belong to current inventory: 5b339645-7b28-4de2-a1f9-5f5ce0763140
pod/pod-c prune skipped
I1215 16:24:38.371694 1266578 prune.go:230] skip pruning object that doesn't belong to current inventory: 1c16ea30-5583-499a-a0ea-1283e4292f13
pod/pod-b prune skipped
I1215 16:24:38.373388 1266578 prune.go:230] skip pruning object that doesn't belong to current inventory: 506cc2ed-6aba-4d7c-a55c-966a21d4b8cf
pod/pod-a prune skipped
I1215 16:24:38.374488 1266578 prune.go:230] skip pruning object that doesn't belong to current inventory: 02c9183e-7903-4b35-9af1-986b5eb0154a

```

Liujingfang1:
After we add the `InventoryPolicy` in the cli-utils library, we need to update the migrate command so that the `owning-inventory` annotations are pointed to the ResourceGroup object.

Fix #1306


commands/migratecmd.go
```diff
@@ -250,18 +272,26 @@ func (mr *MigrateRunner) migrateObjs(cmObjs []object.ObjMetadata, reader io.Read
...
-	if err != nil {
+
+	applier := apply.NewApplier(mr.rgProvider)
+
+	if err := applier.Initialize(); err != nil {
 		return err
 	}
-	inv := live.WrapInventoryInfoObj(rgInv)
-	_, err = rgInvClient.Merge(inv, cmObjs)
+	ch := applier.Run(context.TODO(), inv, objs, apply.Options{
```
mortent, line 285: If there are local changes when we do this, is this still safe?

Liujingfang1: Good question. Since the `objs` are loaded by `rgLoader` from the local not from the cluster, so the `applier.Run` applies whatever in the local.
If the change is for updating or adding resources, this is safe. If the change includes removing a resource, this will lead to the object being orphaned after the apply.

So we need to document there shouldn't be any local changes when calling the `migrate` command.

mortent: So requiring that we update the actual resources as well as the inventory as part of the migration is unfortunate. A nice property of the existing migration path is that it only updates the inventory and does not touch any of the workloads. 

Would it be a better solution to just force users to do this in two steps? So instead of trying to update the `owning-inventory` annotation, we can just verify either that the resource containing the inventory information has the `managed-by` annotation or that all resources listed in the inventory list have the `owning-inventory` annotation? Fixing this for users should just be to run `apply`.

seans3, line 285: This worries me because it is a major abstraction violation. `kpt` should not know about the lower-level guts of how apply is implemented. I think this points to a bigger problem with the `InventoryPolicy` which Morten has pointed out. The `InventoryPolicy` annotations significantly increase the scope of the inventory mechanism by now including every object that is applied. Before, the algorithm only dealt with one (inventory) object during apply. Now it deals with every object that is applied.

seans3 (changes requested):
I am worried about the scope creep that this PR illustrates in the inventory mechanism.

commands/migratecmd.go
Liujingfang1: I see what you mean. We should be able to do a better approach here. How about that the `migrate` command only updates the inventory object. What it does is to get the current inventory list from the cluster and then assign that list to ResourceGroup. So it doesn't bother with local changes since they are not applied to the cluster.

Then the `owning-inventory` annotation should be added by the next `apply`. We need to handle the pruning and applying correctly in the library. So that the first time apply after the migration doesn't break.

Liujingfang1: How about the `migrate` command only updates the inventory object. It gets the inventory list from the ConfigMap inventory object and then assign it to the ResourceGroup CR. Then the `owning-inventory` is handled by the next `apply` after the migration.

mortent: So while I agree that kpt knowing about the underlying inventory is not ideal, I don't think it is a big issue for a migration code. The way the InventoryPolicy is designed, kpt will not have to know about it once we have migrated. 

The apply mechanism in general already deals with every resource. This adds an extra annotation to the resources, but otherwise doesn't really affect the core mechanism of the inventory algorithm. The most important way the apply logic is affected is around checking for this annotation, but that seems pretty well in line with how we handle lifecycle directives.

Liujingfang1: The above approach covers the case where the previous apply was by an old version of `kpt`, where the `owning-inventory` was not added. 

If the previous apply was by the same version of `kpt` with ConfigMap inventory object, the above approach doesn't work. Since the resources already have the `owning-inventory` annotations that point to the ConfigMap. We need to fix these annotation. This can't be done by `apply`. 

So I think we need to update the `owning-inventory` in the resources from empty or the ConfigMap to the ResourceGroup CR. I'll update this PR for that.

Liujingfang1: I tried to updating the `owning-inventory` annotation of objects in the cluster, but it is not a simple operation and we may not do it correctly. Here is the approach I used.
- read the list of object meta from the previous ConfigMap inventory object
- for each object meta, get the object from the cluster. If the object can be found, add the `owning-inventory` annotation to the object and update its `last-applied-configuration` annotation.

However the objects read from the cluster can't be used to run a re-apply. Or even the current update using a dynamic client passed, the next apply will fail.

Liujingfang1: We need to update the `owning-inventory` at the same time when migrating the inventory object since the next `apply` can't update the `owning-inventory` object correctly. See https://github.com/GoogleContainerTools/kpt/pull/1307#discussion_r546922602.

Liujingfang1:
Closing this one since we use a different approach to update the `owning-inventory` annotation in https://github.com/GoogleContainerTools/kpt/pull/1318

PR #1310, opened 2020-12-16
Add pipeline io

### LINKED ISSUE #1304: Read/Write API for the pipeline struct
Part of #1278

Shell32-Natsu:
close #1304


internal/pipeline/builder.go
```diff
@@ -0,0 +1,108 @@
...
+// NewBuilder returns a builder with default Pipeline set
+func NewBuilder() *Builder {
+	return &Builder{p: DefaultPipeline()}
+}
+
+// NewBuilderWithPipeline eturns a builder with a pre-defined
+// Pipeline
+func NewBuilderWithPipeline(p *Pipeline) *Builder {
+	return &Builder{p: p}
+}
```
droot, line 108: I thought a bit more about our use-cases. We will mostly be constructing a pipeline either by reading from a file on disk. If it doesn't exist, we want to create one with the defaults. 
and second set of use-cases are manipulating an existing pipeline where helper methods on an existing pipeline are going to be handy and then a validate method in the end to ensure pipeline is a valid one.
Sorry, I know I sent you down this path, but I am realizing now that builder is not going to be super useful for us. We can define these helper methods on the pipeline object itself. WDYT ? (or we can still continue with this...and take that call later once I have some pieces of pipeline executor in some shape).

droot (commented):
Did a quick pass, have some comments.

internal/pipeline/builder.go
Shell32-Natsu: I was thinking about putting these methods in to Pipeline when I wrote these stuffs. Will do it in this way.

droot: Ack. Thanks for accommodating.

internal/pipeline/io.go
```diff
@@ -0,0 +1,92 @@
...
+const (
+	kptAPIVersion string = "kpt.dev/v1alpha1"
+	pipelineKind  string = "Pipeline"
+	defaultName   string = "pipeline"
+)
+
+var defaultSources []string = []string{"./*"}
+
+// newEmptyPipeline returns a new empty Pipeline
+func newEmptyPipeline() *Pipeline {
```
frankfarzan, line 36: What's this function getting you? Just inline this.

internal/pipeline/io.go
```diff
@@ -0,0 +1,92 @@
...
+
+// NewPipeline returns a pointer to a new default Pipeline.
+// The default Pipeline should be:
+// apiVersion: kpt.dev/v1alpha1
+// kind: Pipeline
+// metadata:
+//   name: pipeline
+// sources:
+//   - '.*'
+func NewPipeline() *Pipeline {
```
frankfarzan, line 48: This can just be 'New', the package is already qualifying it:

Pipeline.New()

internal/pipeline/io.go
```diff
@@ -0,0 +1,92 @@
...
+	p.Name = defaultName
+	p.Sources = defaultSources
+	return p
+}
+
+// FromBytes returns a Pipeline parsed from bytes
+func FromBytes(b []byte) (*Pipeline, error) {
+	p := newEmptyPipeline()
+	err := yaml.Unmarshal(b, p)
+	if err != nil {
```
frankfarzan, line 61: General note: I see a lot of:

	if err != nil {
		return nil, err
	}

This may be valid in many places, but for each instance consider whether you can add more context on the failure instead of passing it through.

internal/pipeline/types_test.go
```diff
@@ -22,170 +22,190 @@ import (
...
-				},
-			},
-		},
-		{
-			Input: `
+		Expected: *NewPipeline().
+			SetAPIVersion("kpt.dev/v1alpha1").
+			SetKind("Pipeline").
+			SetName("pipeline").
+			SetSources(nil),
```
frankfarzan, line 120: why is this nil instead of default?

internal/pipeline/types.go
```diff
@@ -60,6 +60,85 @@ type Pipeline struct {
...
+func (p *Pipeline) String() (string, error) {
+	b, err := yaml.Marshal(p)
+	if err != nil {
+		return "", err
+	}
+	return string(b), nil
+}
+
+// SetName sets Pipeline name
+func (p *Pipeline) SetName(n string) *Pipeline {
```
frankfarzan, line 73: why do we need setters and getters? It seems like very Java-esque.

internal/pipeline/types.go
```diff
@@ -60,6 +60,85 @@ type Pipeline struct {
...
+	return p
+}
+
+// SetValidators replaces the validators in Pipeline by v
+func (p *Pipeline) SetValidators(v []Function) *Pipeline {
+	p.Validators = v
+	return p
+}
+
+func addFunctions(orig *[]Function, new []Function) {
```
frankfarzan, line 138: This just wraps append, I would inline it.

internal/pipeline/io.go
frankfarzan, line 48: It's kind of unorthodox to have the different constructors be in a different file, any reason not do have these in types.go? I think `io.go` is bit of a misnomer anyways.

frankfarzan (changes requested):

internal/pipeline/types_test.go
Shell32-Natsu: Because if there is no `sources` in the yaml input, the `Sources` slice will be `nil`. `nil` is different from empty slice.

internal/pipeline/io.go
Shell32-Natsu: Will merge these 2 files.

internal/pipeline/types_test.go
frankfarzan: I'd expected this would default to './*'.

internal/pipeline/pipeline.go
```diff
@@ -60,6 +73,69 @@ type Pipeline struct {
 	Validators []Function `yaml:"validators,omitempty"`
 }
 
+// String returns the string representation of Pipeline
+func (p *Pipeline) String() (string, error) {
+	b, err := yaml.Marshal(p)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to convert pipeline to string")
```
droot, line 80: Since Go 1.13 onwards, Go supports wrapping/unwrapping errors and also provides some helpers methods in errors package for ex. errors.Is(..), errors.As(...) that makes it easy to deal with errors. Pl. use that for wrapping errors here. Here is a blog post that talks about it https://blog.golang.org/go1.13-errors. Highly recommended, it also improves the errors handling code readability as well.

internal/pipeline/pipeline.go
```diff
@@ -60,6 +73,69 @@ type Pipeline struct {
 	Validators []Function `yaml:"validators,omitempty"`
 }
 
+// String returns the string representation of Pipeline
+func (p *Pipeline) String() (string, error) {
```
droot, line 77: hmm...It isn't expected from a stringer method to return an error. Generally people use it as part of fmt.Printf(...), but now caller has to worry about handling this error and they don't even know what to do if this fails :)

May be we can default to just printing the name of pipeline ? instead of returning an error.

internal/pipeline/pipeline.go
```diff
@@ -60,6 +73,69 @@ type Pipeline struct {
...
+//   name: pipeline
+// sources:
+//   - '.*'
+func New() *Pipeline {
+	p := &Pipeline{}
+	p.APIVersion = kptAPIVersion
+	p.Kind = pipelineKind
+	p.Name = defaultName
+	p.Sources = defaultSources
+	return p
```
droot, line 99: nit: just to improve readability
```
return &Pipeline{
APIVersion:  kptAPIVersion,
...
}
```

internal/pipeline/pipeline.go
```diff
@@ -60,6 +73,69 @@ type Pipeline struct {
...
+}
+
+// FromBytes returns a Pipeline parsed from bytes
+func FromBytes(b []byte) (*Pipeline, error) {
+	p := New()
+	err := yaml.Unmarshal(b, p)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to construct pipeline from bytes")
```
droot, line 107: use the new errors wrapping utilities and in other places in this change.

internal/pipeline/pipeline.go
```diff
@@ -60,6 +73,69 @@ type Pipeline struct {
...
+	}
+	return p, nil
+}
+
+// FromString returns a Pipeline parsed from string
+func FromString(s string) (*Pipeline, error) {
+	return FromBytes([]byte(s))
+}
```
droot, line 115: I would suggest keeping `FromString` and `FromBytes` as internal methods because our use-cases mostly involves constructing pipeline either from a file or max from a reader. And if at anyplace we have bytes/string, it is trivial to construct io.Reader from string/bytes. This also reduces extra tests. 

We should start with minimum API surface for a package and gradually add things as we learn more.

droot (commented):
Looking good. I have some more comments, then we are good to go. Happy to chat offline if you want to resolve this faster.

droot:
Can you pl. squash all the commits.

Shell32-Natsu:
> Can you pl. squash all the commits. 

I believe Frank has enabled squash and merge so we can squash when we merge the pr.

internal/pipeline/pipeline.go
```diff
@@ -60,6 +73,69 @@ type Pipeline struct {
...
+	p := New()
+	err := yaml.Unmarshal(b, p)
+	if err != nil {
+		return nil, fmt.Errorf("failed to construct pipeline from bytes: %w", err)
+	}
+	return p, nil
+}
+
+// FromReader returns a Pipeline parsed from the content in reader
+func FromReader(r io.Reader) (*Pipeline, error) {
```
frankfarzan, line 118: Do these ctor's perform any validations on the input?

Please document this here.

For example, at which point are you going to validate the syntax. For example, `sources` can only be relative directories? At which point are you going to assert required fields? etc.

frankfarzan (changes requested):

internal/pipeline/pipeline.go
Shell32-Natsu: I will leave a todo here since validating pipeline schema is the next task.

droot (approved):
/lgtm

internal/pipeline/pipeline.go
```diff
@@ -60,6 +73,74 @@ type Pipeline struct {
 	Validators []Function `yaml:"validators,omitempty"`
 }
 
+// String returns the string representation of Pipeline struct
+// The string returnes is
```
frankfarzan, line 77: ?

frankfarzan (approved):
minor typo

PR #1311, opened 2020-12-16
Fix e2e test that depends on env

mortent:
This fixes some e2e tests that relied on specific error messages, but that was dependent on the local setup. This instead stubs out the `RunE` function on the Cobra command. This also required moving the check for required setters from `RunE` to `PreRunE`, but that is the right place to put this check.
It also updates the `go.sum` file as a result of running `make tidy`.


e2e/e2e_test.go
```diff
@@ -601,3 +612,21 @@ openAPI:
 		})
 	}
 }
+
+func findSubCmd(cmd *cobra.Command, names ...string) *cobra.Command {
```
droot, line 616: you can probably can use https://godoc.org/github.com/spf13/cobra#Command.Find

e2e/e2e_test.go
```diff
@@ -559,7 +561,7 @@ openAPI:
           required: true
           isSet: true
  `,
-			errMsg: `Package uninitialized`,
+			errMsg: `completed`,
```
droot, line 564: reading the name field in the above two test cases suggests the test should fail with some error message ?

e2e/e2e_test.go
mortent: I looked for something like this, but for some reason I didn't find it. I have updated the PR to use the built-in cobra function.

e2e/e2e_test.go
mortent: Yeah, the name and the expected error is a bit misleading here. I have updated the PR to just expect no error in the last two test cases.

droot (approved):

PR #1314, opened 2020-12-21
Docs: Replace incorrect kpt pkg init command.

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1313 by replacing the nonexistent `kpt init` command with `kpt pkg init`.
image

Also makes half of the text portion of the instructions easier to edit.


frankfarzan (approved):
FTR, can you include the sources to these image as a comment somewhere in the docs?

PR #1315, opened 2020-12-21
add e2e test for kpt live apply

Liujingfang1:
The added test covers the following flow:
- `kpt live apply` with a previous version of kpt
- `kpt live preview` with built version of kpt for the same package
- `kpt live apply` with built version of kpt for the same package.

/assign @mortent @seans3

### DIFF
--- a/e2e/live/end-to-end-test.sh
+++ b/e2e/live/end-to-end-test.sh
@@ -113,6 +113,16 @@ shift $((OPTIND-1))
 #  Helper functions
 ###########################################################################
 
+function downloadPreviousKpt {
+  set -e
+  echo "Downloading latest kpt binary..."
+  curl https://storage.googleapis.com/kpt-dev/latest/linux_amd64/kpt -o latestkpt 2>&1
+  chmod +x latestkpt
+  mv latestkpt $BIN_DIR/.
+  echo "Downloading latest kpt binary...SUCCESS"
+  set +e
+}
+
 # buildKpt builds the kpt binary, storing it in the temporary directory.
 # To check the stdout output of the build check $OUTPUT_DIR/kptbuild.
 # stderr output will be output to the terminal.
@@ -375,6 +385,8 @@ echo "Temp Dir: $TMP_DIR"
 echo
 
 buildKpt
+
+downloadPreviousKpt
 
 echo
 set +e                          # Do not stop the test for errors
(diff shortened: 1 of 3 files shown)


Liujingfang1:
This one should be merged after #1307

e2e/live/end-to-end-test.sh
```diff
@@ -113,6 +113,16 @@ shift $((OPTIND-1))
 #  Helper functions
 ###########################################################################
 
+function downloadPreviousKpt {
```
mortent, line 116: So once we release the next version of kpt that includes support for the inventory policy, would this test still work?

mortent (commented):
Would this test be better in cli-utils? It doesn't seem like this is something that is actually specific to kpt and it would be easier to set up the correct scenarios using the ginkgo tests in cli-utils.

e2e/live/end-to-end-test.sh
Liujingfang1: We can pin the kpt version used in this test. The last released version is v0.37.1. WDYT?

mortent: That would address the first part. But is there a good reason to have this test here rather than in cli-utils?

Liujingfang1: The cli-utils tests the library functionality. It doesn't test a package applied by a previous version of  the cli `kpt` can be applied by the new version of the cli `kpt`. Later the cli-utils may only contain the library part. So the `kpt` repo is a good place for this test.

mortent: ok, that makes sense. But I do think we should make sure we have tests in cli-utils as well to cover the situation where a set of resources was added to the cluster without the `owning-inventory` annotation and then later is updated in way where the annotation are added. We might already cover this, but we should verify.

mortent (approved):

e2e/live/end-to-end-test.sh
Liujingfang1: Just verified that there is some e2e test covered the case without the `owning-inventory` annotation. https://github.com/kubernetes-sigs/cli-utils/blob/master/test/e2e/inventory_policy_test.go#L88

Liujingfang1: Changed the version to v0.37.1. PTAL

PR #1318, opened 2020-12-23
Update the owning-inventory for existing objects during migration

### LINKED ISSUE #1306: Regression: pruning fails for ResourceGroup inventory object

Liujingfang1:
After we add the InventoryPolicy in the cli-utils library, we need to update the migrate command so that the owning-inventory annotations are pointed to the ResourceGroup object. The change is to loop over the resources that are included in the ConfigMap inventory object and change their `owning-inventory` annotation to the new ResourceGroup CR.

When updating the `owning-inventory` annotation, we also update the last-applied-configuration annotation.

Fix #1306 This is a different approach compared with #1307.

### DIFF
--- a/commands/migratecmd.go
+++ b/commands/migratecmd.go
@@ -5,14 +5,17 @@ package commands
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 
+	"github.com/GoogleContainerTools/kpt/pkg/client"
 	"github.com/GoogleContainerTools/kpt/pkg/live"
 	"github.com/spf13/cobra"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/klog"
@@ -139,7 +142,7 @@ func (mr *MigrateRunner) Run(reader io.Reader, args []string) error {
 	}
 	if len(cmObjs) > 0 {
 		// Migrate the ConfigMap inventory objects to a ResourceGroup custom resource.
-		if err = mr.migrateObjs(cmObjs, bytes.NewReader(stdinBytes), args); err != nil {
+		if err = mr.migrateObjs(cmObjs, cmInvObj.ID(), bytes.NewReader(stdinBytes), args); err != nil {
 			return err
 		}
 		// Delete the old ConfigMap inventory object.
@@ -232,7 +235,7 @@ func (mr *MigrateRunner) retrieveInvObjs(invObj inventory.InventoryInfo) ([]obje
 // migrateObjs stores the passed objects in the ResourceGroup inventory
 // object and applies the inventory object to the cluster. Returns
 // an error if one occurred.
-func (mr *MigrateRunner) migrateObjs(cmObjs []object.ObjMetadata, reader io.Reader, args []string) error {
+func (mr *MigrateRunner) migrateObjs(cmObjs []object.ObjMetadata, oldID string, reader io.Reader, args []string) error {
 	fmt.Fprint(mr.ioStreams.Out, "  migrate inventory to ResourceGroup...")
 	if len(cmObjs) == 0 {
 		fmt.Fprintln(mr.ioStreams.Out, "no inventory objects found")
@@ -260,8 +263,13 @@ func (mr *MigrateRunner) migrateObjs(cmObjs []object.ObjMetadata, reader io.Read
 		return err
 	}
 	inv := live.WrapInventoryInfoObj(rgInv)
+	err = updateOwningInventoryAnnotation(mr.rgProvider.Factory(), cmObjs, oldID, inv.ID())
+	if err != nil {
+		return err
+	}
 	_, err = rgInvClient.Merge(inv, cmObjs)
 	if err != nil {
+		fmt.Fprintln(mr.ioStreams.Out, "failed", err.Error())
 		return err
 	}
 	fmt.Fprintln(mr.ioStreams.Out, "success")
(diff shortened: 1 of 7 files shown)


commands/migratecmd.go
```diff
@@ -264,6 +267,11 @@ func (mr *MigrateRunner) migrateObjs(cmObjs []object.ObjMetadata, reader io.Read
 	if err != nil {
 		return err
 	}
+	err = updateOwningInventoryAnnotation(mr.rgProvider.Factory(), cmObjs, oldID, inv.ID())
```
mortent, line 270: So in this case we add the `owning-inventory` annotation regardless of whether the resource had one already right?
If something should fail during migration and kpt crashes just before this line, would re-running the migration still result in the correct end state? I think so, but I just want to make sure we have thought that through.

pkg/client/client.go
```diff
@@ -0,0 +1,103 @@
...
+// Update updates an object using dynamic client
+func (uc *client) Update(ctx context.Context, meta object.ObjMetadata, obj *unstructured.Unstructured, options *metav1.UpdateOptions) error {
+	r, err := uc.resourceInterface(meta)
+	if err != nil {
+		return err
+	}
+	if options == nil {
+		options = &metav1.UpdateOptions{}
+	}
+	_, err = r.Update(ctx, obj, *options)
```
mortent, line 42: Does this result in an `UPDATE` or a `PATCH` call to the apiserver? Just want to make sure we do this the same way as a regular apply operation.

mortent (commented):
Some small questions, but I think this looks good. I think this approach is simpler and safer than in #1307 
I'll let @seans3 take a look too

commands/migratecmd.go
Liujingfang1: This function checks if the resource has the `owning-inventory` annotation. If it doesn't have or if the `owning-inventory` points to the previous ConfigMap inventory object, it updates the `owning-inventory` to the new ResourceGroup object. The reason of doing this is that the `owning-inventory` might be other inventory object and we can't always change it to the ResourceGroup object during migration. If the `owning-inventory` is already the ResourceGroup object, we don't need to update it; if the `owning-inventory` is another object, the migration can't adopt it to the current ResourceGroup object. 

Since the current logic is to create the ResourceGroup CR first, then update the `owning-inventory` annotation. In the case you mentioned, if the ResourceGroup object is created, re-running the migration would skip the part for adding `owning-inventory` annotation.  If the ResourceGroup object wasn't created, re-running the migration would result in the correct end state.

If we change it to first update the `owning-inventory` annotations; then create the ResourceGroup CR. If there is any failure during the migration, re-running the migration would fix it.

pkg/client/client.go
Liujingfang1: It's an `UPDATE` call. I didn't use the `PATCH` call since we should only change the `metadata.annotation`.  Only two annotations are updated: the `owning-inventory` and the last-applied-config annotation. If we compute them correctly, we don't need to use the `PATCH` call.

commands/migratecmd.go
Liujingfang1: Changed the order of the two operations. PTAL.

pkg/client/client.go
seans3: For more context: the `Update` call translates to a REST `Put` with the entire object in the body.

seans3 (approved):
/lgtm

seans3:
Should we squash the commits before merging?

Liujingfang1:
@seans3 I squashed the commits. Let's wait for the tests to pass.

PR #1320, opened 2020-12-23
Create arm64 binaries for kpt releases

mortent:
Adds arm64 arch to goreleaser.

Ref: https://github.com/GoogleContainerTools/kpt/issues/1319

### DIFF
--- a/release/tag/goreleaser.yaml
+++ b/release/tag/goreleaser.yaml
@@ -22,6 +22,7 @@ builds:
       - windows
     goarch:
       - amd64
+      - arm64
     ldflags: -s -w -X github.com/GoogleContainerTools/kpt/run.version={{.Version}}
 archives:
   - files:


mengqiy:
Should we also update https://github.com/GoogleContainerTools/kpt/blob/master/release/tag/cloudbuild.yaml#L80-L139 and https://github.com/GoogleContainerTools/kpt/blob/master/release/branch/cloudbuild.yaml#L93-L122?

runewake2:
Also may want to consider supporting ARM in our e2e tests.

mortent:
@mengqiy We should consider how we should handle the binaries that we are currently pushing to GCS. Ideally I think we want to just push everything to github, so it will be easier for us to get data on the number of downloads. Also, the approach we currently use in the cloud build scripts doesn't really scale..

@runewake2 Yeah, we probably should try to do that, but looks like we need to set up custom runners for github actions to be able to do it. More importantly we should make sure we run the build on macos and windows in addition to linux.

mortent:
I have created separate issues for the other questions brought up here: https://github.com/GoogleContainerTools/kpt/issues/1349 and https://github.com/GoogleContainerTools/kpt/issues/1350. I suggest we merge this as is.

mikebz:
I agree with merging this now and then doing other incremental improvements.

droot (approved):
/lgtm

PR #1321, opened 2020-12-23
pipeline run command

droot:
This is a skeleton implemention for `kpt pipeline run` command. There are lot of places you will see `TODO` and notes and placeholder methods which may not be in the scope of the PR.

NOTE: I have test leaf level hydration manually on a pkg using `kpt pipeline run` command.

### DIFF
--- a/commands/pipelinecmd.go
+++ b/commands/pipelinecmd.go
@@ -1,0 +1,45 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import (
+	"github.com/GoogleContainerTools/kpt/internal/pipeline"
+	"github.com/spf13/cobra"
+)
+
+// GetPipelineCommand returns a command that implements "kpt pipeline <run|...>" subcommands.
+func GetPipelineCommand(name string) *cobra.Command {
+	cmd := &cobra.Command{
+		Use: "pipeline",
+		// TODO(droot): wire docs with docs machinery
+		Short:   "Pipeline (coming soon)",
+		Long:    "Pipeline (coming soon)",
+		Example: "pipeline examples (coming soon)",
+		Aliases: []string{"p"}, // TODO(droot): kpt p run ?
+		RunE: func(cmd *cobra.Command, args []string) error {
+			h, err := cmd.Flags().GetBool("help")
+			if err != nil {
+				return err
+			}
+			if h {
+				return cmd.Help()
+			}
+			return cmd.Usage()
+		},
+	}
+
+	cmd.AddCommand(pipeline.NewCommand(name))
+	return cmd
+}
--- a/internal/pipeline/cmd.go
+++ b/internal/pipeline/cmd.go
@@ -1,0 +1,66 @@
+// Package cmdget contains the get command
+package pipeline
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/GoogleContainerTools/kpt/internal/util/cmdutil"
+	"github.com/spf13/cobra"
+	"k8s.io/klog"
+)
+
+// NewRunner returns a command runner
+func NewRunner(parent string) *Runner {
+	r := &Runner{}
+	c := &cobra.Command{
+		Use:     "run [DIR]",
+		Short:   "run",
+		Long:    "run",
+		Example: "run",
+		RunE:    r.runE,
+		PreRunE: r.preRunE,
+	}
+	cmdutil.FixDocs("kpt", parent, c)
+	r.Command = c
+	return r
+}
+
+func NewCommand(parent string) *cobra.Command {
+	return NewRunner(parent).Command
+}
+
+// Runner contains the run function pipeline run command
+type Runner struct {
+	pkgPath string
+	Command *cobra.Command
+}
+
+func (r *Runner) preRunE(c *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		// no pkg path specified, default to current working dir
+		wd, err := os.Getwd()
+		if err != nil {
+			fmt.Fprintf(c.OutOrStderr(), "error getting current dir: %v \n", err)
+			return err
+		}
+		r.pkgPath = wd
+	} else {
+		// resolve and validate the provided path
+		r.pkgPath = args[0]
+	}
+	return nil
+}
+
+func (r *Runner) runE(c *cobra.Command, args []string) error {
+	klog.Infof("running pipeline command")
+	executor := Executor{
+		PkgPath: r.pkgPath,
+	}
+	err := executor.Execute()
+	if err != nil {
+		fmt.Fprintf(c.OutOrStderr(), "failed to run pipeline %v \n", err)
+		return err
+	}
+	return nil
+}
--- a/internal/pipeline/executor.go
+++ b/internal/pipeline/executor.go
@@ -1,0 +1,291 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pipeline
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"k8s.io/klog"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// Executor hydrates a given pkg.
+type Executor struct {
+	PkgPath string
+}
+
+// Execute runs a pipeline.
+func (e *Executor) Execute() error {
+	klog.Infof("Running pipeline for pkg %s", e.PkgPath)
+
+	p, err := newPkg(e.PkgPath)
+	if err != nil {
+		return fmt.Errorf("failed to initialize pkg %w", err)
+	}
+
+	// initialize hydration context
+	hctx := &hydrationContext{
+		root:            p,
+		hydrationStates: map[string]hydrationState{},
+		pkgs:            map[string]*pkg{},
+	}
+
+	resources, err := hydrate(p, hctx)
+	if err != nil {
+		return fmt.Errorf("failed to hydrate the pkg %s %w", e.PkgPath, err)
+	}
+	// TODO(droot): support different sink modes
+	// for now do in-place
+	klog.Infof("hydrated reources: %d", len(resources))
+
+	return nil
+}
+
+//
+// hydrationContext contains bits to track state of a package hydration.
+// This is sort of global state that is available to hydration step at
+// each pkg along the hydration walk.
+type hydrationContext struct {
+	// root points to the root of hydration graph where we bagan our hydration journey
+	root *pkg
+
+	// TODO (droot): wire sink-mode here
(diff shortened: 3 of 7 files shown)


commands/pipelinecmd.go
```diff
@@ -0,0 +1,118 @@
...
+	"os"
+
+	"github.com/GoogleContainerTools/kpt/internal/pipeline"
+
+	"github.com/spf13/cobra"
+	"go.uber.org/zap"
+)
+
+// GetPipelineCommand returns a command that implements "kpt pipeline <run|...>" subcommands.
+func GetPipelineCommand(name string) *cobra.Command {
```
mortent, line 29: I think we should try to stick with the `runner`-pattern for the cobra commands: https://github.com/GoogleContainerTools/kpt/blob/655b215629ac693b736154a453dbf5a20d08e9a4/internal/cmdget/cmdget.go#L32

commands/pipelinecmd.go
```diff
@@ -0,0 +1,118 @@
...
+			return nil
+		},
+	}
+	return c
+}
+
+// makeLogger returns a logger.
+// Primary reason for having this is to help in debugging.
+// TODO(droot): reconcile this with kpt logger (if one exists).
+func makeLogger() *zap.Logger {
```
mortent, line 95: What are the benefits of using the zap logging framework over klog?

internal/pipeline/executor.go
```diff
@@ -0,0 +1,271 @@
...
+
+	"go.uber.org/zap"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// Executor hydrates a given pkg.
+type Executor struct {
+	PkgPath string
+	Logger  *zap.Logger
```
mortent, line 29: I'm thinking the logger should be passed through as a function parameter instead of being set on the structs. Maybe as a value on a context?

internal/pipeline/executor.go
```diff
@@ -0,0 +1,271 @@
...
+				return
+			}
+			input = append(input, depResources...)
+		}
+	}
+
+	if includeLocalResources {
+		var localResources []*yaml.RNode
+		// TODO(droot): figure out how to avoid reading sub packages here ?
+		pkgReader := &kio.LocalPackageReader{PackagePath: p.path, MatchFilesGlob: kio.MatchAll}
```
mortent, line 184: There is a property called `IncludeSubpackages` that I think should work.

internal/pipeline/executor.go
```diff
@@ -0,0 +1,271 @@
...
+type hydrationState int
+
+// constants for all the hydration states
+const (
+	Dry hydrationState = iota
+	Hydrating
+	Wet
+)
+
+func (s hydrationState) String() string {
```
mortent, line 116: Consider using the golang stringer here: https://godoc.org/golang.org/x/tools/cmd/stringer

internal/pipeline/executor.go
```diff
@@ -0,0 +1,271 @@
...
+// expands takes a list of sources (./*, ./, ...) and current pkg path
+// and returns list of package paths that this pkg depends on.
+func expand(sources []string, pkgPath string) []string {
+	// TODO(droot): implement this with tests
+	// This is one of the critical pieces of code
+	var pkgPaths []string
+	for _, s := range sources {
+		switch s {
+		case "./*":
+			// TODO(droot): figure out how to enumerate sub packages for current pkgPath.
```
mortent, line 239: I think this should work here: https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/pathutil/pathutil.go#L14

commands/pipelinecmd.go
droot: Ack. I started without Runner with a plan to add it when I need it, but didn't encounter any need. 
I think I will probably add it to wrap this command to include some auxiliary stuff such as logger, commandline options. Curious where else did you find runner useful ?

commands/pipelinecmd.go
droot: We used zap in kubebuilder related projects and were sort of happy with it (it's very flexible and doesn't come in the way), so I used it here. Are we using any logging in kpt yet ? Haven't used klog. If klog solves all the problems of glog, then may be that is a better alternative.

internal/pipeline/executor.go
droot: The downside of passing in context will be that every function will have to take a context, even the helper functions.
I will check klog and see if it can be a good alternative without requiring passing logger.

internal/pipeline/executor.go
droot: Ack.

internal/pipeline/executor.go
droot: hmmm...interesting. It does introduces dependency on `stringer` tool, wondering if it is worth the benefits (saving just one fn.)

internal/pipeline/executor.go
droot: Thanks.

internal/pipeline/executor.go
```diff
@@ -0,0 +1,271 @@
...
+			depResources, err = hydrate(depPkg, hctx)
+			if err != nil {
+				err = fmt.Errorf("failed to hydrated sub pkg %s %w", s, err)
+				return
+			}
+			input = append(input, depResources...)
+		}
+	}
+
+	if includeLocalResources {
```
frankfarzan, line 181: We should extract out such common operations into functions or methods.

If going with OOP approach:

p.readLocalResources()

internal/pipeline/executor.go
```diff
@@ -0,0 +1,271 @@
...
+		// TODO(droot): this will be the place where we filter kpt resources
+		localResources = filter(localResources)
+		input = append(input, localResources...)
+	}
+
+	output := &kio.PackageBuffer{}
+	// TODO(droot): parameterize the sink-mode (hctx.sinkMode) to determine whether to
+	// write it in-place or not. Currently in-place.
+	pkgWriter := &kio.LocalPackageWriter{PackagePath: p.path}
+	pipe := kio.Pipeline{
```
frankfarzan, line 199: there's an overload of 'pipeline' here. Should probably trying to disambiguate through variable naming and comments.

internal/pipeline/executor.go
```diff
@@ -0,0 +1,271 @@
...
+	// and read an actual pipeline
+	p.pipeline = New()
+
+	// adjust the pipeline.sources to point to
+	return nil
+}
+
+// expands takes a list of sources (./*, ./, ...) and current pkg path
+// and returns list of package paths that this pkg depends on.
+func expand(sources []string, pkgPath string) []string {
```
frankfarzan, line 232: p.normalizedSources()?

Regardless of where validation logic runs, this needs to guarantee that source are normalized/clean/expanded. This logic will need to be well unit-tested and it will grow in complexity over time

internal/pipeline/executor.go
frankfarzan: That's looking for specific filename though.

You basically want `ls -d`. I foresee this getting more complicated in the future. For example, we may want to ignore 'hidden' directories (those start with `.`).

internal/pipeline/fn.go
```diff
@@ -0,0 +1,182 @@
...
+
+// annotator is built kpt function for annotating KRM resources.
+type annotator struct {
+	key    string
+	value  string
+	logger *zap.SugaredLogger
+}
+
+func (a *annotator) Run(r io.Reader, w io.Writer) error {
+	rw := &kio.ByteReadWriter{
```
frankfarzan, line 55: Based on just this example, why isn't `rw' passed as an argument to this method?

internal/pipeline/fn.go
```diff
@@ -0,0 +1,182 @@
...
+const (
+	// annotation used to track resources for preserving comments during the function execution
+	idAnnotation = "config.k8s.io/id"
+)
+
+// KRMFn defines capabilities of a KRM function. It primarily knows
+// how to run itself. It accepts input from an io.Reader that is
+// a stream of KRM objects in the form of KRM FunctionSpec. It also accepts
+// a io.Rriter where it writes the result of running a KRM function.
+type KRMFn interface {
```
frankfarzan, line 37: General question: Are you extracting a common library for use by 'kpt fn' and 'kpt pipeline' in the kpt repo?

commands/pipelinecmd.go
droot: Moved the `run` subcommand to runner pattern as you suggested.

droot (commented):
@frankfarzan Thanks for the review. Addressed majority of the comments.

internal/pipeline/executor.go
```diff
@@ -0,0 +1,290 @@
...
+// to a given pipeline.
+func fnFilters(_ *Pipeline) []kio.Filter {
+	// TODO(droot): hardcoding built-in set-annotation function for testing
+	// this will be replaced from the pipeline functions eventually.
+	fn := &annotator{
+		key:   "builtin/setter-1",
+		value: "test",
+	}
+
+	return []kio.Filter{
```
Shell32-Natsu, line 285: Maybe another TODO here for adding other functions from real pipeline.

internal/pipeline/executor.go
```diff
@@ -0,0 +1,271 @@
...
+	"path/filepath"
+
+	"go.uber.org/zap"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// Executor hydrates a given pkg.
+type Executor struct {
+	PkgPath string
```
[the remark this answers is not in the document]
frankfarzan: https://github.com/GoogleContainerTools/kpt/pull/1340

internal/pipeline/executor.go
frankfarzan: sgtm

internal/pipeline/fn.go
frankfarzan: If the function implementation always creates a rw from r and w and operates on that, that's a signal that we should just be passing rw as an argument?

internal/pipeline/executor.go
```diff
@@ -0,0 +1,271 @@
...
+	// here we explore the p.sources to determine sub packages etc.
+	for _, s := range expand(p.pipeline.Sources, p.path) {
+		// TODO(droot): sync with pipeline's defaults once Donny's PR is merged
+		if s == "." {
+			// current pkg
+			// will be hydrated as we drop of this loop
+			includeLocalResources = true
+		} else {
+			// TODO(droot): recursive hydration is broken for now, needs more work
+			var depResources []*yaml.RNode
```
frankfarzan, line 228: nit: "dep" and "dependency" are overloaded. Maybe 'transitiveResources'?

internal/pipeline/executor.go
droot: Done.

internal/pipeline/executor.go
droot: Done.

droot:
I think I have addressed most of the feedback. Treat this PR to be skeleton PR, so all the TODOs left in the PR will be done in the follow up PRs including the test coverage.

I think this lays out the shape of the solution very well and we can already see work done in other PRs fitting well in this skeleton. I would like to get this merged now so that @natasha41575 @Shell32-Natsu can start integrating their PRs and address TODOs in this PR.

^^ @frankfarzan @mortent

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.6.1
+	go.uber.org/zap v1.10.0
```
mortent, line 18: I don't think this is still needed.

internal/pipeline/cmd.go
```diff
@@ -0,0 +1,66 @@
...
+		// no pkg path specified, default to current working dir
+		wd, err := os.Getwd()
+		if err != nil {
+			fmt.Fprintf(c.OutOrStderr(), "error getting current dir: %v \n", err)
+			return err
+		}
+		r.pkgPath = wd
+	} else {
+		// resolve and validate the provided path
+		r.pkgPath = args[0]
```
mortent, line 50: It seems like `args[0]` here will give us a relative path, while `os.Getwd()` will return an absolute path. I think we should always use absolute paths.

mortent (approved):
Some minor comments. We should consider adding tests early to make sure we have a design that lends itself to unit tests.

go.mod
droot: Yep, you are right. Just ran `go mod tidy` and it's gone.

internal/pipeline/cmd.go
droot: Down the line the pipeline.Executor is converting the path provided here to absolute path.

droot:
fixes #1321

PR #1335, opened 2021-01-06
add .spec.subgroups field in ResourceGroup CRD

Liujingfang1:
Add the field `.spec.subgroups` for child ResourceGroup CRs, which represent sub groups.

/assign @mortent @seans3

### DIFF
--- a/pkg/live/example-resource-group-crd-for-v1.yaml
+++ b/pkg/live/example-resource-group-crd-for-v1.yaml
@@ -45,26 +45,25 @@ spec:
           metadata:
             type: object
           spec:
-            description: ResourceGroupSpec defines the desired state of ResourceGroup
+            description: spec defines the desired state of ResourceGroup
             properties:
               descriptor:
-                description: Descriptor regroups the information and metadata about
+                description: descriptor regroups the information and metadata about
                   a resource group
                 properties:
                   description:
-                    description: Description is a brief description of a group of
-                      resources
+                    description: description is a brief description of a group of resources
                     type: string
                   links:
-                    description: Links are a list of descriptive URLs intended to
-                      be used to surface additional information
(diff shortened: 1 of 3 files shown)


seans3:
I think we should add more comments in the files to make clear the difference between ResourceGroup subgroup objects and regular inventory objects, but I'm OK with this in general. I'll let Morten comment, and if he's OK with it, we'll merge it.

pkg/live/example-resource-group-crd-for-v1.yaml
```diff
@@ -80,8 +80,31 @@ spec:
...
-                  resource group
+                description: Resources contains a list of resources that form the resource
+                  group
+                items:
+                  description: ObjMetadata organizes and stores the identifying information
+                    for an object. This struct (as a string) is stored in a grouping
+                    object to keep track of sets of applied objects.
+                  properties:
+                    group:
+                      type: string
```
mortent, line 91: Do we need `GroupKind` for this? Is there any situation where the subgroup wouldn't be a ResourceGroup?

Liujingfang1: The subgroup should always be ResourceGroup. The reason of adding this is for simplicity in the controller so that it handles the subgroups and resources in the same way.

Liujingfang1: I'll remove the unnecessary `GroupKind` in the controller code. Will update this CRD afterwards.

Liujingfang1: Updated this. PTAL

mortent (approved):

seans3 (approved):
/lgtm

PR #1336, opened 2021-01-07
Allow test framework to better handle the main branch

mortent:
The existing test framework doesn't allow setting up the `main` branch properly, as it is automatically created from `master`. This leverages the existing `Content` struct to provide better support for setting up different branches.

### DIFF
--- a/e2e/e2eutils.go
+++ b/e2e/e2eutils.go
@@ -34,7 +34,10 @@ func SetupGitRepo(t *testing.T) (*testutil.TestGitRepo, string, func()) {
 		t.FailNow()
 	}
 
-	g, _, c := testutil.SetupDefaultRepoAndWorkspace(t, testutil.Dataset1, map[string]string{})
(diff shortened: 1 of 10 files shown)


e2e/e2eutils.go
```diff
@@ -34,7 +34,10 @@ func SetupGitRepo(t *testing.T) (*testutil.TestGitRepo, string, func()) {
 		t.FailNow()
 	}
 
-	g, _, c := testutil.SetupDefaultRepoAndWorkspace(t, testutil.Dataset1, map[string]string{})
+	g, _, c := testutil.SetupDefaultRepoAndWorkspace(t, testutil.Content{
```
frankfarzan, line 37: Let's followup on the general scope of this test framework, there are plans for having e2e testing of functions, we need to rationalize the different appraoches.

frankfarzan (approved):

PR #1339, opened 2021-01-07
Add pipeline validation

### LINKED ISSUE #1331: Validate pipeline schema
Part of #1278

Shell32-Natsu:
close #1331

### DIFF
--- a/internal/pipeline/pipeline.go
+++ b/internal/pipeline/pipeline.go
@@ -94,20 +94,19 @@ func (p *Pipeline) Validate() error {
 		return fmt.Errorf("kind %s is not valid, should be %s",
 			p.Kind, pipelineKind)
 	}
-	for _, s := range p.Sources {
-		if s == "." || s == "./*" {
+	for i, s := range p.Sources {
+		if s == "./*" {
 			continue
 		}
 		if err := ValidatePath(s); err != nil {
 			return fmt.Errorf("source path invalid: %w", err)
 		}
+		p.Sources[i] = path.Clean(s)
 	}
 	fns := []Function{}
-	for _, fs := range [][]Function{p.Generators, p.Transformers, p.Validators} {
-		for _, fn := range fs {
-			fns = append(fns, fn)
-		}
-	}
+	fns = append(fns, p.Generators...)
+	fns = append(fns, p.Transformers...)
+	fns = append(fns, p.Validators...)
 	for _, f := range fns {
 		err := f.Validate()
 		if err != nil {
@@ -270,9 +269,6 @@ func isNodeZero(n *yaml.Node) bool {
 func ValidatePath(p string) error {
 	if path.IsAbs(p) {
 		return fmt.Errorf("path is not relative: %s", p)
-	}
-	if path.Clean(p) != p {
-		return fmt.Errorf("path is not clean: %s", p)
 	}
 	if strings.Contains(p, "\\") {
 		return fmt.Errorf("path cannot have backslash: %s", p)
--- a/internal/pipeline/pipeline.go
+++ b/internal/pipeline/pipeline.go
@@ -21,8 +21,9 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"path/filepath"
+	"path"
 	"regexp"
+	"strings"
 
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
@@ -81,15 +82,7 @@ func (p *Pipeline) String() string {
 	return fmt.Sprintf("%+v", *p)
 }
 
-// Validate will validate the Pipeline `p`. This function will validate
-// fields 'apiVersion', 'kind', 'sources', 'generators', 'transformers'
-// and 'validators'.
-//
-// 'sources' are valid if and only if every source is:
-// 1. a relative path to local package OR
-// 2. '.' OR
-// 3. './*'
-//
+// Validate will validate all fields in the Pipeline `p`
 // 'generators', 'transformers' and 'validators' share same schema and
 // they are valid if all functions in them are ALL valid.
 func (p *Pipeline) Validate() error {
@@ -105,26 +98,20 @@ func (p *Pipeline) Validate() error {
 		if s == "." || s == "./*" {
 			continue
 		}
-		if filepath.IsAbs(s) {
-			return fmt.Errorf("source path must be relative: %s", s)
+		if err := ValidatePath(s); err != nil {
+			return fmt.Errorf("source path invalid: %w", err)
 		}
 	}
-	for _, f := range p.Generators {
-		err := f.Validate()
-		if err != nil {
-			return fmt.Errorf("generator %s is not valid: %w", f.Image, err)
+	fns := []Function{}
+	for _, fs := range [][]Function{p.Generators, p.Transformers, p.Validators} {
+		for _, fn := range fs {
+			fns = append(fns, fn)
 		}
 	}
-	for _, f := range p.Transformers {
+	for _, f := range fns {
 		err := f.Validate()
 		if err != nil {
-			return fmt.Errorf("transformer %s is not valid: %w", f.Image, err)
-		}
-	}
-	for _, f := range p.Validators {
-		err := f.Validate()
-		if err != nil {
-			return fmt.Errorf("validator %s is not valid: %w", f.Image, err)
+			return fmt.Errorf("function %s is not valid: %w", f.Image, err)
 		}
 	}
 	return nil
@@ -211,30 +198,29 @@ type Function struct {
 	ConfigMap map[string]string `yaml:"configMap,omitempty"`
 }
 
-// Validate will validate the function `f`. A function is valid if:
-// 1. 'image' is a valid function name OR a fully qualified name of a function AND
-// 2. have zero OR one of 'config', 'configPath' and 'configMap' AND
-// 3. if 'configPath' is used, the value MUST be a relative path
+// Validate will validate the function `f`.
 func (f *Function) Validate() error {
 	err := ValidateFunctionName(f.Image)
 	if err != nil {
 		return fmt.Errorf("function name is not valid: %w", err)
 	}
-	var cnt = 0
+	// configCnt is used to count the number of configs provided. More than one
+	// config is not allowed.
+	var configCnt = 0
 	if f.ConfigPath != "" {
-		cnt++
+		configCnt++
+		if err := ValidatePath(f.ConfigPath); err != nil {
+			return fmt.Errorf("configPath is invalid: %w", err)
+		}
 	}
 	if len(f.ConfigMap) != 0 {
-		cnt++
+		configCnt++
 	}
 	if !isNodeZero(&f.Config) {
-		cnt++
+		configCnt++
 	}
-	if cnt > 1 {
-		return fmt.Errorf("only zero or one config is allowed, %d provided", cnt)
-	}
-	if f.ConfigPath != "" && filepath.IsAbs(f.ConfigPath) {
-		return fmt.Errorf("configPath must be relative: %s", f.ConfigPath)
+	if configCnt > 1 {
+		return fmt.Errorf("only zero or one config is allowed, %d provided", configCnt)
 	}
 
 	return nil
@@ -278,4 +264,18 @@ func isNodeZero(n *yaml.Node) bool {
 		n.Anchor == "" && n.Alias == nil && n.Content == nil &&
 		n.HeadComment == "" && n.LineComment == "" && n.FootComment == "" &&
 		n.Line == 0 && n.Column == 0
+}
+
+// ValidatePath validates `path` and return an error if it's invalid
+func ValidatePath(p string) error {
+	if path.IsAbs(p) {
+		return fmt.Errorf("path is not relative: %s", p)
+	}
+	if path.Clean(p) != p {
+		return fmt.Errorf("path is not clean: %s", p)
+	}
+	if strings.Contains(p, "\\") {
+		return fmt.Errorf("path cannot have backslash: %s", p)
+	}
+	return nil
 }
--- a/internal/pipeline/pipeline.go
+++ b/internal/pipeline/pipeline.go
@@ -21,6 +21,8 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
+	"regexp"
 
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
@@ -79,6 +81,55 @@ func (p *Pipeline) String() string {
 	return fmt.Sprintf("%+v", *p)
 }
 
+// Validate will validate the Pipeline `p`. This function will validate
+// fields 'apiVersion', 'kind', 'sources', 'generators', 'transformers'
+// and 'validators'.
+//
+// 'sources' are valid if and only if every source is:
+// 1. a relative path to local package OR
+// 2. '.' OR
+// 3. './*'
+//
+// 'generators', 'transformers' and 'validators' share same schema and
+// they are valid if all functions in them are ALL valid.
+func (p *Pipeline) Validate() error {
+	if p.APIVersion != kptAPIVersion {
+		return fmt.Errorf("apiVersion %s is not valid, should be %s",
+			p.APIVersion, kptAPIVersion)
+	}
+	if p.Kind != pipelineKind {
+		return fmt.Errorf("kind %s is not valid, should be %s",
+			p.Kind, pipelineKind)
+	}
+	for _, s := range p.Sources {
+		if s == "." || s == "./*" {
+			continue
+		}
+		if filepath.IsAbs(s) {
+			return fmt.Errorf("source path must be relative: %s", s)
+		}
+	}
+	for _, f := range p.Generators {
+		err := f.Validate()
+		if err != nil {
+			return fmt.Errorf("generator %s is not valid: %w", f.Image, err)
+		}
+	}
+	for _, f := range p.Transformers {
+		err := f.Validate()
+		if err != nil {
+			return fmt.Errorf("transformer %s is not valid: %w", f.Image, err)
+		}
+	}
+	for _, f := range p.Validators {
+		err := f.Validate()
+		if err != nil {
+			return fmt.Errorf("validator %s is not valid: %w", f.Image, err)
+		}
+	}
+	return nil
+}
+
 // New returns a pointer to a new default Pipeline.
 // The default Pipeline should be:
 // apiVersion: kpt.dev/v1alpha1
@@ -86,7 +137,7 @@ func (p *Pipeline) String() string {
 // metadata:
 //   name: pipeline
 // sources:
-//   - '.*'
+//   - './*'
 func New() *Pipeline {
 	return &Pipeline{
 		ResourceMeta: yaml.ResourceMeta{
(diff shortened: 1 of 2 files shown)


internal/pipeline/pipeline.go
```diff
@@ -79,14 +81,63 @@ func (p *Pipeline) String() string {
...
+	}
+	if p.Kind != pipelineKind {
+		return fmt.Errorf("kind %s is not valid, should be %s",
+			p.Kind, pipelineKind)
+	}
+	for _, s := range p.Sources {
+		if s == "." || s == "./*" {
+			continue
+		}
+		if filepath.IsAbs(s) {
```
frankfarzan, line 108: Need more validation for source. Should extract this in its own function.

These paths are slash-separated OS-agnostic paths so 'filepath' package (OS-specific) is not appropriate. Look at the 'path' package instead, keep in mind we define special characters like './*'.

Need to check it's a) Slash delimited b) Relative c) Clean 

Examples of bad sources:

a\b\c
c:\a\b\c\
/home/frankf
../a/../b

There are probably other corner cases that needs to be considered. Shouldn't assume kpt is run on a Unix system.

internal/pipeline/pipeline.go
```diff
@@ -164,3 +210,72 @@ type Function struct {
...
+// 1. 'image' is a valid function name OR a fully qualified name of a function AND
+// 2. have zero OR one of 'config', 'configPath' and 'configMap' AND
+// 3. if 'configPath' is used, the value MUST be a relative path
+func (f *Function) Validate() error {
+	err := ValidateFunctionName(f.Image)
+	if err != nil {
+		return fmt.Errorf("function name is not valid: %w", err)
+	}
+	var cnt = 0
+	if f.ConfigPath != "" {
```
frankfarzan, line 224: Just combine these if into one

internal/pipeline/pipeline.go
```diff
@@ -164,3 +210,72 @@ type Function struct {
...
+	if len(f.ConfigMap) != 0 {
+		cnt++
+	}
+	if !isNodeZero(&f.Config) {
+		cnt++
+	}
+	if cnt > 1 {
+		return fmt.Errorf("only zero or one config is allowed, %d provided", cnt)
+	}
+	if f.ConfigPath != "" && filepath.IsAbs(f.ConfigPath) {
```
frankfarzan, line 236: Ditto if this path

internal/pipeline/pipeline.go
```diff
@@ -79,14 +81,63 @@ func (p *Pipeline) String() string {
...
+	}
+	for _, s := range p.Sources {
+		if s == "." || s == "./*" {
+			continue
+		}
+		if filepath.IsAbs(s) {
+			return fmt.Errorf("source path must be relative: %s", s)
+		}
+	}
+	for _, f := range p.Generators {
```
frankfarzan, line 112: Just combine into into one loop and remove duplication

internal/pipeline/pipeline.go
```diff
@@ -79,14 +81,63 @@ func (p *Pipeline) String() string {
 	return fmt.Sprintf("%+v", *p)
 }
 
+// Validate will validate the Pipeline `p`. This function will validate
+// fields 'apiVersion', 'kind', 'sources', 'generators', 'transformers'
+// and 'validators'.
```
frankfarzan, line 86: This should just validate all fields, no need to enumerate this in comment

internal/pipeline/pipeline.go
```diff
@@ -79,14 +81,63 @@ func (p *Pipeline) String() string {
...
+//
+// 'sources' are valid if and only if every source is:
```
frankfarzan, line 88: Avoid duplication of comments. Source of truth is the Pipeline struct.

frankfarzan (changes requested):

internal/pipeline/pipeline.go
Shell32-Natsu: How does kpt handle paths on non-Unix system?

internal/pipeline/pipeline.go
Shell32-Natsu: These 3 ifs are used to count the number of configs provided so they cannot be merged. I will add comments and improve variable name.

internal/pipeline/pipeline.go
Shell32-Natsu: How do you define `Clean`? if it's `path.Clean()` then `./base` is not a clean path.

internal/pipeline/pipeline.go
```diff
@@ -79,14 +82,49 @@ func (p *Pipeline) String() string {
...
+		if err := ValidatePath(s); err != nil {
+			return fmt.Errorf("source path invalid: %w", err)
+		}
+	}
+	fns := []Function{}
+	for _, fs := range [][]Function{p.Generators, p.Transformers, p.Validators} {
+		for _, fn := range fs {
+			fns = append(fns, fn)
+		}
+	}
```
droot, line 110: This can be simplified by:
```Go
fns = append(fns, p.Generators...)
fns = append(fns, p.Transformers...)
fns = append(fns, p.Validators...)

```

internal/pipeline/pipeline.go
```diff
@@ -164,3 +197,85 @@ type Function struct {
...
+		n.HeadComment == "" && n.LineComment == "" && n.FootComment == "" &&
+		n.Line == 0 && n.Column == 0
+}
+
+// ValidatePath validates `path` and return an error if it's invalid
+func ValidatePath(p string) error {
+	if path.IsAbs(p) {
+		return fmt.Errorf("path is not relative: %s", p)
+	}
+	if path.Clean(p) != p {
```
droot, line 274: hmm...this will prevent `./` and `./base` to be valid paths which are completely valid. 
This is getting trickier :)

Another way to validate this would be to construct a path such as `path.Join(curr-pkg-path, rel(source-in-the-pipeline)))` and checking if that exist on the filesystem or not. [Now whether or not creates cycle can be caught during the hydration]

internal/pipeline/pipeline.go
```diff
@@ -164,3 +197,85 @@ type Function struct {
...
+// >= 0.10.5
+func isNodeZero(n *yaml.Node) bool {
+	return n != nil && n.Kind == 0 && n.Style == 0 && n.Tag == "" && n.Value == "" &&
+		n.Anchor == "" && n.Alias == nil && n.Content == nil &&
+		n.HeadComment == "" && n.LineComment == "" && n.FootComment == "" &&
+		n.Line == 0 && n.Column == 0
+}
+
+// ValidatePath validates `path` and return an error if it's invalid
+func ValidatePath(p string) error {
```
droot, line 270: Do we need to export these (both Validate) functions ?

internal/pipeline/pipeline.go
Shell32-Natsu: Yeah, I noticed this. https://github.com/GoogleContainerTools/kpt/pull/1339#discussion_r555357290

internal/pipeline/pipeline.go
Shell32-Natsu: Exposing these functions makes it can be tested in `pipeline_test` package. I am ok to use `pipeline` package in `pipeline_test.go`. Which one do you think is better?

internal/pipeline/pipeline.go
frankfarzan: Re How does kpt handle paths on non-Unix system?

Depends on which level you're talking about. Arguments to CLI or OS-specific. In this context, we're talking about an OS-agnostic, portable format for Pipeline.sources. 

How do you define Clean?

We are defining our own format in Pipeline.sources which may have nothing to do with Golang or its std libraries. We can decide how strict we want the validation to be.
For example, which of the following should we accept:

`base`, `./base`, `./base/`, `base/`, `base1/.../base2`?

There's a trade-off here:
1) Strict validation ensures ensures uniformity and consistency.
2) Strict validation may be unnecessary pedantic or "dumb". For example rejecting, `base/` even though it's clear
what the user intent was.
 
The most pragmatic (easiest to implement) is:
1. Treat `./*` as a special case.
2. Accept any other relative path (not strict) and run Clean on it. All cases above are acceptable.

Also note:
- Reason we chose `./base` was because it makes it obvious this is a relative to the current directory. With `base`, there may be an ambiguity whether package `base` lives in the filesystem.

Thoughts?

Shell32-Natsu: +1 for the pragmatic way.

internal/pipeline/pipeline.go
```diff
@@ -79,14 +82,48 @@ func (p *Pipeline) String() string {
 	return fmt.Sprintf("%+v", *p)
 }
 
+// Validate will validate all fields in the Pipeline `p`
```
frankfarzan, line 85: nit: You don't need to refer to the parameter `p`. This is a receiver method, so it's obvious it acts on Package. Same for other method receivers.

internal/pipeline/pipeline.go
droot: I think it is fine in this case because we don't expect lot of refactoring in this part of the code. 

Testing internal function can introduce maintenance headache if the code has a potential for lot of refactoring because now the maintainer has to not only change test code whenever they refactor internal func (and it's usage). So need to be done with careful judgement.

internal/pipeline/pipeline_test.go
```diff
@@ -315,3 +317,266 @@ func isPipelineEqual(t *testing.T, p1, p2 Pipeline) bool {
...
+
+func TestValidatePath(t *testing.T) {
+	type input struct {
+		Path  string
+		Valid bool
+	}
+
+	cases := []input{
+		{
+			Path:  "a/b/c",
```
frankfarzan, line 530: Currently this allows having subdirectory named `*`.  Is that intentional?

Can you think of other corner cases?

internal/pipeline/pipeline_test.go
```diff
@@ -315,3 +317,266 @@ func isPipelineEqual(t *testing.T, p1, p2 Pipeline) bool {
...
+			Valid: true,
+		},
+		{
+			Path:  "a/b/",
```
frankfarzan, line 534: to reduce verbosity of table driven tests, you can omit field names Path and Valid

internal/pipeline/pipeline.go
```diff
@@ -164,3 +196,82 @@ type Function struct {
...
+		}
+	}
+	if len(f.ConfigMap) != 0 {
+		configCnt++
+	}
+	if !isNodeZero(&f.Config) {
+		configCnt++
+	}
+	if configCnt > 1 {
+		return fmt.Errorf("only zero or one config is allowed, %d provided", configCnt)
```
frankfarzan, line 222: General rule: Refer to path of specific fields you're validating in error messages. e.g. `spec.....ConfigMap`. Otherwise, "zero or one config is allowed" is ambigious.

Same for all other error messages.

frankfarzan (changes requested):

internal/pipeline/pipeline_test.go
Shell32-Natsu: `*` is a valid file name in Unix/Linux. It's possible that there is a subdirectory `./a/*` so I think it's valid.

frankfarzan: A more reasonable scenario is the intended `./*` but didn't get it quite right (e.g. `*`, or `./*`), as opposed to having a directory with an asterisk which is extremely unlikely.

internal/pipeline/pipeline.go
```diff
@@ -79,14 +82,48 @@ func (p *Pipeline) String() string {
 	return fmt.Sprintf("%+v", *p)
 }
 
+// Validate will validate all fields in the Pipeline
+// 'generators', 'transformers' and 'validators' share same schema and
+// they are valid if all functions in them are ALL valid.
+func (p *Pipeline) Validate() error {
+	if p.APIVersion != kptAPIVersion {
+		return fmt.Errorf("'apiVersion' %s is not valid, should be %s",
```
frankfarzan, line 90: General rule: Any user specified value should be quoted using %q

internal/pipeline/pipeline.go
```diff
@@ -79,14 +82,48 @@ func (p *Pipeline) String() string {
...
+			p.APIVersion, kptAPIVersion)
+	}
+	if p.Kind != pipelineKind {
+		return fmt.Errorf("'kind' %s is not valid, should be %s",
```
frankfarzan, line 94: not valid -> invalid?

Standardize all error messages in this file.

internal/pipeline/pipeline.go
```diff
@@ -79,14 +82,48 @@ func (p *Pipeline) String() string {
...
+			p.Kind, pipelineKind)
+	}
+	for i, s := range p.Sources {
+		if s == "./*" {
+			continue
+		}
+		if err := ValidatePath(s); err != nil {
+			return fmt.Errorf("'sources[%d]' invalid: %w", i, err)
```
frankfarzan, line 102: is invalid

internal/pipeline/pipeline.go
```diff
@@ -164,3 +196,82 @@ type Function struct {
 	// `ConfigMap` is a convenient way to specify a function config of kind ConfigMap.
 	ConfigMap map[string]string `yaml:"configMap,omitempty"`
 }
+
+// Validate will validate all fields in function.
+func (f *Function) Validate() error {
+	err := ValidateFunctionName(f.Image)
+	if err != nil {
+		return fmt.Errorf("'image' is not valid: %w", err)
```
frankfarzan, line 204: Error message should have the user-provided value (f.Image). Use %q per the other comment.

internal/pipeline/pipeline.go
```diff
@@ -164,3 +196,82 @@ type Function struct {
...
+		configFields = append(configFields, "configPath")
+	}
+	if len(f.ConfigMap) != 0 {
+		configFields = append(configFields, "configMap")
+	}
+	if !isNodeZero(&f.Config) {
+		configFields = append(configFields, "config")
+	}
+	if len(configFields) > 1 {
+		return fmt.Errorf("only zero or one config is allowed but got [%s]",
```
frankfarzan, line 221: Other comment about this is not resolved

internal/pipeline/pipeline.go
```diff
@@ -164,3 +196,82 @@ type Function struct {
...
+	domainComponentRegexp := `(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])`
+	domainRegexp := fmt.Sprintf(`%s(?:\.%s)*(?:\:[0-9]+)?`, domainComponentRegexp, domainComponentRegexp)
+	nameRegexp := fmt.Sprintf(`^(?:%s\/)?%s(?:\/%s)*$`, domainRegexp,
+		pathComponentRegexp, pathComponentRegexp)
+	matched, err := regexp.MatchString(nameRegexp, name)
+	if err != nil {
+		return err
+	}
+	if !matched {
+		return fmt.Errorf("function name %s is invalid", name)
```
frankfarzan, line 250: Use %q consistently everywhere.

Also do an audit of error messages.

internal/pipeline/pipeline_test.go
```diff
@@ -315,3 +317,294 @@ func isPipelineEqual(t *testing.T, p1, p2 Pipeline) bool {
...
+			"././././",
+			true,
+		},
+		{
+			"./!&^%$/#(@)/_-=+|<;>?:'\"/'`",
+			true,
+		},
+		{
+			"",
+			true,
```
frankfarzan, line 595: Why is this valid?

internal/pipeline/pipeline_test.go
```diff
@@ -315,3 +317,294 @@ func isPipelineEqual(t *testing.T, p1, p2 Pipeline) bool {
...
+		{
+			"a//b",
+			true,
+		},
+		{
+			"a/b/.",
+			true,
+		},
+		{
+			"a/*/b",
```
frankfarzan, line 574: I don't think this should be valid

internal/pipeline/pipeline_test.go
```diff
@@ -315,3 +317,294 @@ func isPipelineEqual(t *testing.T, p1, p2 Pipeline) bool {
...
+		{
+			"./!&^%$/#(@)/_-=+|<;>?:'\"/'`",
+			true,
+		},
+		{
+			"",
+			true,
+		},
+		{
+			"\t \n",
```
frankfarzan, line 598: why is this valid?

internal/pipeline/pipeline_test.go
```diff
@@ -315,3 +317,294 @@ func isPipelineEqual(t *testing.T, p1, p2 Pipeline) bool {
...
+		{
+			"./a/.../b",
+			true,
+		},
+		{
+			".",
+			true,
+		},
+		{
+			"a\\b",
```
frankfarzan, line 554: a\b is also invalid.

frankfarzan (changes requested):

internal/pipeline/pipeline_test.go
Shell32-Natsu: I think this is about how restrict this validation function should be. Is a path to a non-exist directory valid? Is a path to a invalid kpt package valid? Is the accidental typo in path but that path is a kpt package valid? I think these scenarios are out of this functions's duty. If there are errors, they will be reported when kpt try to read and process the sources and we shouldn't check every possible situation. If we find that some scenarios happen frequently we can add more restrictions in the future.

frankfarzan: IMO this layer should be responsible for lexical validation. Concerns like does package actually exists is a separate concern. The comments here are only concerned with lexical correctness.

internal/pipeline/pipeline.go
```diff
@@ -164,3 +196,90 @@ type Function struct {
...
+	if path.IsAbs(p) {
+		return fmt.Errorf("path %q is not relative", p)
+	}
+	if strings.TrimSpace(p) == "" {
+		return fmt.Errorf("path %q cannot have only white spaces", p)
+	}
+	if p != "./*" && strings.Contains(p, "*") {
+		return fmt.Errorf("path %q contains asterisk, asterisk is only allowed in './*'", p)
+	}
+	for _, c := range "\\\a\b\f\v" {
```
frankfarzan, line 279: maybe add a comment why these characters are special?

internal/pipeline/pipeline.go
```diff
@@ -164,3 +196,90 @@ type Function struct {
...
+	return n != nil && n.Kind == 0 && n.Style == 0 && n.Tag == "" && n.Value == "" &&
+		n.Anchor == "" && n.Alias == nil && n.Content == nil &&
+		n.HeadComment == "" && n.LineComment == "" && n.FootComment == "" &&
+		n.Line == 0 && n.Column == 0
+}
+
+// ValidatePath validates input path and return an error if it's invalid
+func ValidatePath(p string) error {
+	if path.IsAbs(p) {
+		return fmt.Errorf("path %q is not relative", p)
```
frankfarzan, line 271: In general, you can remove some duplication, the caller can wrap errors returned by this function and add additional context.

return fmt.Errorf("path is not relative")

And then caller of ValidatePath can wrap the error:

sources[0] is invalid: '/a/b/c'. path is relative

frankfarzan (approved):

droot (approved):
/lgtm

PR #1340, opened 2021-01-08
Skeleton data structure representing a pkg metadata (WIP)

frankfarzan:


internal/pipeline/pipeline.go
```diff
@@ -43,8 +43,8 @@ var defaultSources []string = []string{"./*"}
 type Pipeline struct {
 	yaml.ResourceMeta `yaml:",inline"`
 	//  1. Sources to resolve as input to the pipeline. Possible values:
-	//  a) A relative path to a local package e.g. './base', '../foo'
-	//    The source package is resolved recursively.
+	//  a) A slash-seprated, OS-agnostic relative package path which may include '.' and '..' e.g. './base', '../foo'
```
droot, line 46: s/seprated/separated

internal/pkg/pkg.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/GoogleContainerTools/kpt/internal/pipeline"
+	"github.com/GoogleContainerTools/kpt/pkg/kptfile"
+)
+
+// pkg represents the static metadata for a kpt package.
+type pkg struct {
```
droot, line 27: s/pkg/Pkg

internal/pkg/pkg.go
```diff
@@ -0,0 +1,91 @@
...
+	// A nil value represents an implicit package.
+	Kptfile *kptfile.KptFile
+
+	// A package can contain zero or one Pipeline meta resource.
+	Pipeline *pipeline.Pipeline
+}
+
+// New returns a pkg given an absolute or relative OS-defined path.
+// Use ReadKptfile or ReadPipeline on the return value to read meta resources from filesystem.
+func New(path string) (pkg, error) {
```
droot, line 46: We should be returning `*pkg` here or you are thinking something else here ?

internal/pkg/pkg.go
```diff
@@ -0,0 +1,91 @@
...
+	d = filepath.ToSlash(d)
+	return pkg{UniquePath: u, DisplayPath: d}, nil
+}
+
+// ReadKptfile reads the Kptfile meta resource from filesystem.
+func (p *pkg) ReadKptfile() *pkg {
+	// TODO
+	// p.Kptfile = ...
+	return p
+}
```
droot, line 79: Thinking from usage perspective, having `p.Kptfile()` method that returns the p.Kptfile. It reads the kptfile if not read already.

internal/pkg/pkg.go
```diff
@@ -0,0 +1,91 @@
...
+
+// ReadPipeline reads the Pipeline meta resource from filesystem.
+func (p *pkg) ReadPipeline() *pkg {
+	// TODO
+	// p.Pipeline = ...
+	return p
+}
```
droot, line 86: Similar to above commment, I can see `p.Pipeline()` method very useful here.

In light of this, we can then make fields `pipeline, kptfile` as internal on the pkg struct. WDYT ?

internal/pkg/pkg.go
frankfarzan: Done. Are you concerned about passing large values? Currently, this struct contains couple of string values and everything else is a pointer, so it's not a large struct. But it may expand in the future.

internal/pkg/pkg.go
frankfarzan: I like the lazy loading suggestion.

droot (approved):

PR #1341, opened 2021-01-08
Set `metadata.name` in Kptfile to name of directory on Get

mortent:
When a package is fetched with `kpt pkg get`, the name of the Kptfile should be set to the name of the directory.

### DIFF
--- a/internal/util/get/get.go
+++ b/internal/util/get/get.go
@@ -96,7 +96,7 @@ func (c Command) Run() error {
 	}
 
 	// create or update the KptFile with the values from git
-	if err = (&c).upsertKptfile(r); err != nil {
+	if err = upsertKptfile(c.Destination, c.Name, r); err != nil {
 		return errors.Wrap(err)
 	}
 	return nil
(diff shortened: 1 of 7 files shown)


internal/util/get/get.go
```diff
@@ -258,30 +258,30 @@ func (c *Command) DefaultValues() error {
 
 // upsertKptfile populates the KptFile values, merging any cloned KptFile and the
 // cloneFrom values.
-func (c *Command) upsertKptfile(spec *git.RepoSpec) error {
+func upsertKptfile(path, name string, spec *git.RepoSpec) error {
```
frankfarzan, line 261: Currently, we pass around a path of type `string` everywhere in the codebase. It's ambiguous whether it's absolute, relative, cleaned, .... Does it work for windows (I'm pretty sure it wouldn't work in general)? Does it work when there are symlinks? etc.

We should move towards making this an explicit type, so the compiler helps us. Relevant:

https://github.com/GoogleContainerTools/kpt/pull/1340/files#diff-f440b5f78da4e52ed45c778be5d7b49373a9427073d651b6166eb5bbd93b5d6a

frankfarzan (approved):

internal/util/get/get.go
mortent: I have been working on making sure all paths passed around is always absolute, and any translation from relative to absolute happens in the cobra commands. Agree that we can consider a custom type for an absolute path.

frankfarzan: Depending on the usage, we can pass around:
1. Pkg: https://github.com/GoogleContainerTools/kpt/blob/next/internal/pkg/pkg.go#L35
2. UniquePath: https://github.com/GoogleContainerTools/kpt/blob/next/internal/pkg/pkg.go#L27

PR #1351, opened 2021-01-11
Add generic function for updating the annotations only

Liujingfang1:
Add a function to update the annotation of an object.  Here is how it should be used.
```
obj := client.Get()

var newAnnotations map[string]string
// get the new annotations that obj should have
// not including the last-applied-config annotation

UpdateAnnotation(obj, newAnnotations)

client.Update(obj)
```

Also add unit tests for the new function.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -61,10 +61,10 @@ func (uc *client) resourceInterface(meta object.ObjMetadata) (dynamic.ResourceIn
 	return namespacedClient, nil
 }
 
-// UpdateAnnotation updates the object owning inventory annotation
+// ReplaceOwningInventoryID updates the object owning inventory annotation
 // to the new ID when the owning inventory annotation is either empty or the old ID.
-// It returns if the annotation is updated.
-func UpdateAnnotation(obj *unstructured.Unstructured, oldID, newID string) (bool, error) {
+// It returns true if the annotation is updated.
+func ReplaceOwningInventoryID(obj *unstructured.Unstructured, oldID, newID string) (bool, error) {
 	key := "config.k8s.io/owning-inventory"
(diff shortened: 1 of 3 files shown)


pkg/client/client.go
```diff
@@ -61,10 +61,10 @@ func (uc *client) resourceInterface(meta object.ObjMetadata) (dynamic.ResourceIn
 	return namespacedClient, nil
 }
 
-// UpdateAnnotation updates the object owning inventory annotation
+// ReplaceOwningInventoryID updates the object owning inventory annotation
 // to the new ID when the owning inventory annotation is either empty or the old ID.
 // It returns if the annotation is updated.
```
seans3, line 66: `// It returns true if the annotation is updated`

pkg/client/client.go
```diff
@@ -73,19 +73,23 @@ func UpdateAnnotation(obj *unstructured.Unstructured, oldID, newID string) (bool
 	val, found := annotations[key]
 	if !found || val == oldID {
 		annotations[key] = newID
+		return true, UpdateAnnotation(obj, annotations)
+	}
+	return false, nil
+}
+
+func UpdateAnnotation(obj *unstructured.Unstructured, annotations map[string]string) error {
```
seans3, line 81: Since this is exported, we should probably have comments to describe its use.

seans3 (commented):
Small nitpicks.

pkg/client/client.go
Liujingfang1: Good catch. Fixed it

pkg/client/client.go
Liujingfang1: done

seans3 (approved):
/lgtm

PR #1355, opened 2021-01-12
add container runtime for pipeline

### LINKED ISSUE #1334: Implement support for docker function execution
part of #1279 

The docker runtime implementation will be put in kpt repo instead of kyaml

Shell32-Natsu:
close #1334


internal/pipeline/runtime/container.go
```diff
@@ -0,0 +1,111 @@
...
+	AsCurrentUser bool
+}
+
+// ContainerRunner implements a KRMFn which run a containerized
+// KRM function
+type ContainerRunner struct {
+	// Image is the container image to run
+	Image string
+
+	Exec ExecRunner
```
droot, line 42: Not sure why we have this here ?

internal/pipeline/runtime/container.go
```diff
@@ -0,0 +1,111 @@
...
+
+// ContainerNetworkName is a type for network name used in container
+type containerNetworkName string
+
+const (
+	networkNameNone containerNetworkName = "none"
+	networkNameHost containerNetworkName = "host"
+)
+
+type containerFunctionPermission struct {
```
droot, line 31: May be shorten it `containerFnPermission` ?

internal/pipeline/runtime/container.go
```diff
@@ -0,0 +1,111 @@
...
+	AllowNetwork  bool
+	AsCurrentUser bool
+}
+
+// ContainerRunner implements a KRMFn which run a containerized
+// KRM function
+type ContainerRunner struct {
```
droot, line 38: Thinking of naming here: would calling it `ContainerFn` be more appropriate ?

internal/pipeline/runtime/container.go
Shell32-Natsu: 1. This piece of code is modified from kyaml function runtime and the container runtime in kyaml is based on exec runtime.
2. Since we have plan to support exec function so I think it's better to have a separate runner for exec so I don't change this structure.

internal/pipeline/runtime/container.go
```diff
@@ -0,0 +1,111 @@
...
+	// Image is the container image to run
+	Image string
+
+	Exec ExecRunner
+
+	containerFunctionPermission
```
droot, line 44: Any reasons for preferring embedding vs explicit field `perms` ? I do want to refresh my notes on this topic.

internal/pipeline/runtime/exec.go
```diff
@@ -0,0 +1,39 @@
...
+	Args []string `yaml:"args,omitempty"`
+}
+
+// Run runs the command
+func (c *ExecRunner) Run(reader io.Reader, writer io.Writer) error {
+	cmd := exec.Command(c.Path, c.Args...)
+	cmd.Stdin = reader
+	cmd.Stdout = writer
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
```
droot, line 38: We should use `ctx` that allows to set timeout etc. For now pick a large timeout but we should do it now avoid it plumbing it later.

internal/pipeline/runtime/container.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return fmt.Errorf("error when get command and args: %w", err)
+	}
+	f.Exec.Path = path
+	f.Exec.Args = args
+	return nil
+}
+
+// getCommand returns the command + args to run to spawn the container
+func (f *ContainerRunner) getCommand() (string, []string, error) {
```
droot, line 73: may be have this function return `exec.Cmd`

droot (changes requested):
Took a quick look. Some early comments.

internal/pipeline/runtime/container.go
Shell32-Natsu: I think return an ExecFn may be better? Or just merge this function with `setupExec` and directly setup the `Exec` member. `exec.Cmd` should be encapsulated by `ExecFn`.

internal/pipeline/runtime/container.go
droot: My main motivation for that suggestion was decoupling the execFn and containerFn so that changes to one doesn't affect the another. Also note that the execFn is a very lightweight wrapper over golang's exec primitives, so this extra layering doesn't buy us much instead increases overhead for anyone understanding `containerFn`.

I know current code is inspired by the code in kyaml but we can diverge from that depending on our requirements.

> Since we have plan to support exec function so I think it's better to have a separate runner for exec so I don't change this structure.

Totally agree with having separate runner for execFn. I am suggesting even greater degree of separation between the two :)

internal/pipeline/runtime/container.go
droot: I shared my view in an earlier comment. If we decide to go that route, this can be revisited.

internal/pipeline/runtime/exec.go
droot: I realize, we will have to change `KRMFn` to accept `ctx` and relay that here.

internal/pipeline/runtime/exec_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/GoogleContainerTools/kpt/internal/pipeline/runtime"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestExecRunner(t *testing.T) {
```
droot, line 25: We can add test for `containerFn` that requires docker on the machine and we can use Golang's build tags to selectively invoke that. You can read more about https://mickey.dev/posts/go-build-tags-testing/

droot (changes requested):

internal/pipeline/runtime/container.go
```diff
@@ -0,0 +1,111 @@
...
+	Image string
+	// Container function will be killed after this timeour.
+	// The default value is 5 minutes.
+	Timeout time.Duration
+	Perm    ContainerFnPermission
+}
+
+// Run implements KRMFn. It will run the container function with
+// stdin from r and write the output to w
+func (f *ContainerFn) Run(r io.Reader, w io.Writer) error {
```
frankfarzan, line 60: Are you implementing this to work on Windows?

internal/pipeline/runtime/container_test.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"bytes"
+	"testing"
+
+	"github.com/GoogleContainerTools/kpt/internal/pipeline/runtime"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestContainerFn(t *testing.T) {
+	// use sed to test instead of calling real docker
```
frankfarzan, line 28: how do you use `sed`?

frankfarzan (changes requested):

internal/pipeline/runtime/container_test.go
Shell32-Natsu: Updated the test but forgot this comment which was copied from kyaml. Will remove it.

internal/pipeline/runtime/container.go
Shell32-Natsu: @droot Do we want to support Windows?

droot: That's a good question. Didn't think about this. We would like to support windows, but reality is testing it regularly requires access to windows machines etc and I have found that to be challenging. Actual work required to support it won't be much IMO. From my previous project experience, what has helped is to create an issue for windows support and our friends from Windows eco-system help to support that.

TL;DR: windows support will be best effort basis and not a priority at this point.

frankfarzan: I think we should avoid regression if kpt supports windows today, but it can be punted for now and tracked as a support issue.

Shell32-Natsu: https://github.com/GoogleContainerTools/kpt/issues/1367

internal/pipeline/runtime/container_test.go
```diff
@@ -0,0 +1,65 @@
+// +build docker
```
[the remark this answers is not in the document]
Shell32-Natsu: done

Shell32-Natsu:
@droot @frankfarzan PTAL

droot (approved):

frankfarzan (approved):

PR #1360, opened 2021-01-14
modify client function to update both labels and annotations

Liujingfang1:
The current function only updates annotations, we need to update both labels and annotations.


seans3 (approved):
/lgtm

Issue #1266, opened 2020-12-08
add e2e test for continuing on error


Liujingfang1:
An example could be Apply two Deployments a, b. Deployment a has some invalid field so it should fail the apply. Deployment b is correctly configured. As a result, we should see the Deployment b is applied and Deployment a is not applied.

/assign @seans3

mikebz:
assigning to @runewake2 since I think this is a good ticket to get familiar with e2e tests.

PR #1370, opened 2021-01-21
Continue on error test

runewake2:
**WIP - please don't merge yet.**

quick test case covering #1266

This introduces one test case over `kpt live apply` where a resource is unable to be successfully applied.

It is expected that

| pod | result |
| :-- | --: |
| `pod-a` | succeeds |
| `pod-B` | fails |

Currently both `pod-a` and `pod-B` are added to the inventory object. This test expects only `pod-a` to be added to the inventory object. This means the test currently fails. Leave this PR until that behavior has been merged, some changes may need to occur after that feature has completed.

`pod-B` has an invalid pod name and should fail to create - because of this failure it should not be added to the inventory object.

### DIFF
--- a/e2e/live/end-to-end-test.sh
+++ b/e2e/live/end-to-end-test.sh
@@ -872,6 +872,15 @@ assertPodNotExists "pod-c" "test-namespace"
 assertPodNotExists "pod-d" "test-namespace"
 printResult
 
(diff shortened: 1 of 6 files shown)


seans3 (approved):
/lgtm

PR #1372, opened 2021-01-21
implemented executor resolveSources function with tests

### LINKED ISSUE #1347: Implement function to resolve/normalize pipeline.Sources
The task here is to implement the following function from [the PR](https://github.com/GoogleContainerTools/kpt/pull/1321/files) with exhaustive unit-tests.

```
// expands takes a list of sources (./*, ./, ...) and current pkg path
// and returns list of package paths that this pkg depends on.
func (p *pkg) resolveSources() []string {
```

natasha41575:
Close #1347

This is an implementation of the resolveSources function in the pipeline run command skeleton code

### DIFF
--- a/internal/pipeline/executor.go
+++ b/internal/pipeline/executor.go
@@ -16,11 +16,14 @@ package pipeline
 
 import (
 	"fmt"
+	"io/ioutil"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"k8s.io/klog"
 	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/pathutil"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
@@ -125,9 +128,8 @@ func (p *pkg) Pipeline() *Pipeline {
 	return p.pipeline
 }
 
-// expands takes a list of sources (./*, ./, ...) and current pkg path
+// resolveSources takes a list of sources (./*, ./, ...) and current pkg path
 // and returns list of package paths that this pkg depends on.
-// TODO(droot): implement this with tests
 // This is one of the critical pieces of code
 func (p *pkg) resolveSources() []string {
 	pipeline := p.Pipeline()
@@ -139,9 +141,19 @@ func (p *pkg) resolveSources() []string {
 			// include only this pkg sources
 			pkgPaths = append(pkgPaths, p.Path())
 		case sourceAllSubPkgs:
-			// TODO(droot): figure out how to enumerate sub packages for current pkgPath.
-			// For now including current pkg
+			// For now including current pkg and subpkgs in current directory
 			pkgPaths = append(pkgPaths, p.Path())
+			files, _ := ioutil.ReadDir(p.Path())
+			for _, f := range files {
+				if f.IsDir() {
+					// A directory is a package if it has a Kptfile.
+					// This may change as the concept of a package is expanded such that
+					// every directory is its own kpt package
+					subPaths, _ := pathutil.DirsWithFile(p.Path()+"/"+f.Name(), "Kptfile", false)
+					sort.Strings(subPaths)
+					pkgPaths = append(pkgPaths, subPaths...)
+				}
+			}
 		default:
 			// s points to a specific sub pkg
 			pkgPaths = append(pkgPaths, s)
--- a/internal/pipeline/executor.go
+++ b/internal/pipeline/executor.go
@@ -17,10 +17,12 @@ package pipeline
 import (
 	"fmt"
 	"io/ioutil"
+	"path"
 	"path/filepath"
 	"sort"
 	"strings"
 
+	"github.com/GoogleContainerTools/kpt/pkg/kptfile"
 	"k8s.io/klog"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/pathutil"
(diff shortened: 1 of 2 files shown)


internal/pipeline/executor.go
```diff
@@ -138,9 +140,19 @@ func (p *pkg) resolveSources() []string {
 			// include only this pkg sources
 			pkgPaths = append(pkgPaths, p.Path())
 		case sourceAllSubPkgs:
-			// TODO(droot): figure out how to enumerate sub packages for current pkgPath.
-			// For now including current pkg
+			// For now including current pkg and subpkgs in current directory
 			pkgPaths = append(pkgPaths, p.Path())
+			files, _ := ioutil.ReadDir(p.Path())
```
droot, line 145: Change the signature of the `resolveResources` to return this error. Don't ignore this error :)

internal/pipeline/executor.go
```diff
@@ -138,9 +140,19 @@ func (p *pkg) resolveSources() []string {
...
+			for _, f := range files {
+				if f.IsDir() {
+					// A directory is a package if it has a Kptfile.
+					// This may change as the concept of a package is expanded such that
+					// every directory is its own kpt package
+					subPaths, _ := pathutil.DirsWithFile(p.Path()+"/"+f.Name(), "Kptfile", false)
```
droot, line 151: Use the `path` pkg helper functions to construct these paths (I think there is `Join`) method, it will help us later to support different types of OS etc. better.

droot: Also, don't ignore error :)

internal/pipeline/executor_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	defer os.RemoveAll(subDir2)
+
+	err = ioutil.WriteFile(subDir2+"/Kptfile", []byte(`
+apiVersion: kpt.dev/v1alpha1
+kind: Kptfile
+`), 0600)
+	if !assert.NoError(t, err) {
+		t.FailNow()
+	}
```
droot, line 62: Can you pl. check pkgbuilder related test setup @mortent added for `pkg get/update` tests. We need utilities to setup these tests because we need these in other places as well.

internal/pipeline/executor_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// sorting for testing purposes
+	subDirs := []string{subDir1, subDir2}
+	sort.Strings(subDirs)
+
+	tests := map[string]struct {
+		p			*pkg
+		expected 	[]string
+	}{
+		"package without pipeline": {
```
droot, line 72: +1 on mapping tests like these to make it easy to distinguish.

droot (commented):
Looking good. Have some comments.

internal/pipeline/executor_test.go
natasha41575: Acknowledged. I will add a file to testutil that is similar to the current pkgbuilder (since that one has a different pkg struct with different fields)

I think it would make the codebase less confusing if we had different names for these two structs; right now "pkg" is overloaded. WDYT?

natasha41575: Acknowledged. I will add a file to testutil that is similar to the current pkgbuilder (since that one has a different pkg struct with different fields)

internal/pipeline/executor.go
natasha41575: done

internal/pipeline/executor.go
natasha41575: done, thanks!

internal/pipeline/executor_test.go
natasha41575: updated the test structure, thanks

internal/pipeline/executor.go
```diff
@@ -139,15 +142,32 @@ func (p *pkg) resolveSources() []string {
 			// include only this pkg sources
 			pkgPaths = append(pkgPaths, p.Path())
 		case sourceAllSubPkgs:
-			// TODO(droot): figure out how to enumerate sub packages for current pkgPath.
-			// For now including current pkg
+			// For now including current pkg and subpkgs in current directory
```
droot, line 145: nit: remove `For now` because `sourceAllSubPkgs` requires us to include current pkg and subpkgs.

internal/pipeline/pkgbuilder.go
```diff
@@ -0,0 +1,32 @@
...
+
+func (p pkg) withPipeline() *pkg {
+	p.pipeline = &Pipeline{}
+	return &p
+}
+
+func (p pkg) withSources(sources []string) *pkg {
+	p.pipeline.Sources = sources
+	return &p
+}
```
droot, line 32: `Pipeline` and `sources` should be read from the pkg itself, not sure why do we need these methods. If it is required to create arbitrary pipeline/source for a pkg, then we should avoid that. It will make maintaining tests harder.

internal/pipeline/pkgbuilder.go
```diff
@@ -0,0 +1,32 @@
+package pipeline
```
droot, line 1: license it missing

internal/pipeline/pkgbuilder.go
droot: I realized, we can make the `resourceSources` to take just path and not be defined at the `pkg` level. That will make it easier to test.

droot (commented):
Not defining resolveResource at the pkg level can make this thing easier. We can briefly talk about it.

internal/pipeline/executor.go
```diff
@@ -139,15 +142,32 @@ func (p *pkg) resolveSources() []string {
...
+			if err != nil {
+				return nil, err
+			}
+			for _, f := range files {
+				if f.IsDir() {
+					// A directory is a package if it has a Kptfile.
+					// This may change as the concept of a package is expanded such that
+					// every directory is its own kpt package
+					absolutePath := path.Join(p.Path(), f.Name())
+					subPaths, err := pathutil.DirsWithFile(absolutePath, "Kptfile", false)
```
mortent, line 157: We should use the constant for the name of the Kptfile. I think it is `kptfile.KptFileName`

internal/pipeline/executor.go
```diff
@@ -139,15 +142,32 @@ func (p *pkg) resolveSources() []string {
...
+					if err != nil {
+						return nil, err
+					}
+					sort.Strings(subPaths)
```
mortent, line 161: We do we sort the paths here? In particular, why do we need them in alphabetical order?

internal/pipeline/executor_test.go
```diff
@@ -128,3 +130,72 @@ spec:
...
+			expected: []string{p.Path()},
+		},
+
+		"directory with subpackages, all sources": {
+			p: p.withPipeline().withSources([]string{"./*"}),
+			expected: append([]string{p.Path()}, subDirs...),
+		},
+	}
+
+	for testName, test := range tests {
```
mortent, line 189: We should use subtests here and avoid including the name of the test in the error messages.

internal/pipeline/executor.go
natasha41575: done

internal/pipeline/pkgbuilder.go
natasha41575: updated the PR to add a helper that takes a path to make tests easier. We can talk about this if you have further comments or suggestions

internal/pipeline/executor.go
natasha41575: done, thanks

internal/pipeline/executor_test.go
natasha41575: done, thank you

internal/pipeline/executor.go
natasha41575: This was a discussion I had with @droot. There's no strong reason to sort the subpaths apart from having deterministic output to make testing easier

droot (approved):

natasha41575:
thank you @droot!

PR #1380, opened 2021-01-21
added basic e2e test for pipeline run command

### LINKED ISSUE #1369: Basic end to end test for pipeline run command
We merged a basic implementation of `kpt pipeline run` in last sprint. We need a basic test that tests end to end workflow for `pipeline run` command.

At the very basic, it starts with a simple kpt pkg, invokes the pipeline run command and compares the output to expected output.

natasha41575:
Close #1369 

Adds the most basic e2e test for pipeline run

Note: Currently pipeline run leaves behind a `builtin/setter-1: 'test'` annotation in all the yaml files, I'm not sure if this is expected or desired

Edited to include what I wrote in a comment later in the thread:
Based on our discussion today, I've updated the PR to include the test harness from kpt-functions-catalog that Donny wrote with a small tweak to allow it to work for pipeline. I've also included a basic pipeline run command test here. The kpt-functions-catalog repo can import and reuse this test harness when it is ready.


e2e/e2e_test.go
```diff
@@ -451,3 +451,138 @@ spec:
...
+
+func TestPipeline(t *testing.T) {
+	var tests = []struct {
+		name     string
+		command  string
+		args     []string
+		input    map[string]string
+		expected map[string]string
+		errMsg   string
+	}{
```
droot, line 463: I am still thinking how we would like to structure these tests. We will have to do some work upfront to create a test harness that does the following:
 1. Takes input pkg (either some structured format specified in the `Go` test files or from a directory under `testData` on filesystem)
 2. A way to specify corresponding expected output (either some form in `Go` file or under a directory say `golden` under `testData`)
 3. Runs `pipeline run` how we are doing today where setting up the input dirs etc and then comparing the output to the golden files. I remember seeing utility method about comparing the two pkgs to be Equal or not in one of @mortent PR. 
 
Think, if we were to test pkgs with hierarchy that has sub-packages etc., how we can setup multiple such tests and invoke it with different pipeline setup.

droot:
> Note: Currently pipeline run leaves behind a builtin/setter-1: 'test' annotation in all the yaml files, I'm not sure if this is expected or desired

Yes, currently it's hardcoded, but Donny is working on a task that will enable running functions specified in the `pipeline.yaml` so that's another reason flexible test setup is important because expected output will depend on the content of the `pipeline.yaml`.

e2e/e2e_test.go
natasha41575: That makes sense, thanks. I will give more thought to this and would love to sync up if/when you have ideas for how these tests should be structured.

natasha41575: I've updated the test structure to use directories testData and golden under testutils

e2e/e2e_test.go
```diff
@@ -451,3 +452,50 @@ spec:
 		})
 	}
 }
+
+func getPipelineDataPath(subdir string) string {
+	testdatapath, _ := testutil.GetTestDataPath()
```
droot, line 457: Can the testData directory live in `pipeline` pkg itself ?

droot (commented):
This is looking good. I have one minor question about the location of the testData.

e2e/e2e_test.go
frankfarzan: @mengqiy is setting up such e2e test harness with test packages that contain expected output/results.

e2e/e2e_test.go
```diff
@@ -451,3 +452,50 @@ spec:
...
+		command      string
+		args         []string
+	}{
+		"basic pipeline run command": {
+			dataset: testutil.BasicPipeline,
+			command: "run",
+		},
+	}
+
+	for _, test := range tests {
```
mortent, line 478: Use subtests here.

e2e/e2e_test.go
```diff
@@ -451,3 +452,50 @@ spec:
...
+		dir, err := ioutil.TempDir("", "")
+		if !assert.NoError(t, err) {
+			t.FailNow()
+		}
+
+		testdata := filepath.Join(getPipelineDataPath("testdata"), test.dataset)
+		golden := filepath.Join(getPipelineDataPath("golden"), test.dataset)
```
mortent, line 485: Is using a golden directory structure for the output scalable here? Our experience with packaging is that there are many cases to test and that providing a large number of golden results is hard to maintain.

mengqiy:
@natasha41575 In case you are curious about what we did in kpt-functions-catalog.
Our package is self-contained and it has everything needed for testing.
Let's use [kubeval](https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/examples/kubeval) as an example:
```
.
├── .expected
│   ├── config.yaml
│   └── results.yaml
├── .krmignore
├── README.md
├── configs
│   └── example-config.yaml
└── fn-config.yaml
```
And https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/f4b9e3af8ee2ed405ba264691181f5fc90677367/tests/e2e-test/tests/e2e_test.go#L14-L31 describes what these files are.

natasha41575:
@mengqiy Thank you so much! I will take a look and use that as a reference

frankfarzan:
> @mengqiy Thank you so much! I will take a look and use that as a reference

I expect the functions catalog and kpt repo to share the same test harness for testing hydration pipeline. Is that the plan?

Shell32-Natsu:
> > @mengqiy Thank you so much! I will take a look and use that as a reference
> 
> I expect the functions catalog and kpt repo to share the same test harness for testing hydration pipeline. Is that the plan?

Technically there is no blocker to use the same harness.

natasha41575:
@frankfarzan We can make that the plan. I will talk to @droot to get his input

mengqiy:
> > > @mengqiy Thank you so much! I will take a look and use that as a reference
> >
> > I expect the functions catalog and kpt repo to share the same test harness for testing hydration pipeline. Is that the plan?
> 
> Technically there is no blocker to use the same harness.

These 2 repos eventually should use the same test harness. I expect the primary use case of kpt function in the future will be through the hydration pipeline. Using the same test harness will make more sense. 
But it should not be a blocker that stop either project from moving forward at the moment.

@natasha41575 If you need my input or help with the test harness design, I'm happy to help.

natasha41575:
@droot @Shell32-Natsu @mengqiy @frankfarzan 
Based on our discussion today, I've updated the PR to include the test harness from kpt-functions-catalog with a small tweak to allow it to work for pipeline. I've also included a basic pipeline run command test here. The kpt-functions-catalog repo can import and reuse this test harness when it is ready.

e2e/e2e_test.go
```diff
@@ -24,12 +24,42 @@ import (
...
 	"sigs.k8s.io/kustomize/kyaml/openapi"
 )
 
+// for these tests to pass, a version of kpt supporting the pipeline run
+// command will be installed, because they call the kpt binary
+func TestPipeline(t *testing.T) {
+	runPipelineTests(t, "../internal/pipeline/examples/")
+}
+
+// runTests will scan test cases in 'path' and run all the
```
Shell32-Natsu, line 40: This comment needs to be updated

pkg/runner/config.go
```diff
@@ -0,0 +1,99 @@
+package runner
```
Shell32-Natsu, line 1: It's a little bit confusing to put these files under `pkg/runner`. Maybe `pkg/test/runner` or something.

pkg/runner/runner.go
```diff
@@ -0,0 +1,222 @@
...
+func (r *Runner) Run(cmd string, args []string, t *testing.T) error {
+	fmt.Printf("Running test against package %s\n", r.pkgName)
+	tmpDir, err := ioutil.TempDir("", "kpt-fn-catalog-e2e-*")
+	if err != nil {
+		return fmt.Errorf("failed to create temporary dir: %w", err)
+	}
+	defer os.RemoveAll(tmpDir)
+	tmpPkgPath := filepath.Join(tmpDir, r.pkgName)
+	// create dir to store untouched pkg to compare against
+	orgPkgPath := filepath.Join(tmpDir, "results")
```
Shell32-Natsu, line 54: The `results` dir is not used to store original package content. It's used to hold the kpt output when it has error. Is this changed in pipeline?

Shell32-Natsu (changes requested):

pkg/runner/runner.go
```diff
@@ -0,0 +1,222 @@
...
+	return &Runner{
+		pkgName:  filepath.Base(testCase.Path),
+		testCase: testCase,
+	}, nil
+}
+
+// Run runs the test.
+func (r *Runner) Run(cmd string, args []string, t *testing.T) error {
+	fmt.Printf("Running test against package %s\n", r.pkgName)
+	tmpDir, err := ioutil.TempDir("", "kpt-fn-catalog-e2e-*")
```
mengqiy, line 47: `"kpt-fn-catalog-e2e-*"` this should be changed.

internal/pipeline/examples/basicpipeline/.expected/diff.patch
```diff
@@ -0,0 +1,30 @@
+diff --git a/.expected/config.yaml b/.expected/config.yaml
+index 04229ba..2df37b4 100644
+--- a/.expected/config.yaml
++++ b/.expected/config.yaml
+@@ -1,2 +1,5 @@
+ exitCode: 0
+ network: false
++metadata:
```
mengqiy, line 8: I don't see a pipeline definition that can cause these diff to happen.
Did I miss something?

Shell32-Natsu: There is a *builtin* function for testing. I think that's the reason.

natasha41575: Currently fnFilters is hardcoded to add these annotations. After #1388 they should disappear. Probably better to wait until that is merged before merging this one

pkg/runner/runner.go
natasha41575: I thought results dir was meant to originally meant for the functions to write their result to that dir. 
https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/127a1afe906e8b2a7d1d37fc14106b03d9631ca8/tests/e2etest/runner/runner.go#L70

In the original harness that you wrote, my understanding is that it runs kpt fn run on tmpPkgPath but stores the results in results dir and then compares them. 

kpt pipeline does not have a --results-dir option. The idea I had was to alter the harness so that it saves the original package in orgPkgPath, runs kpt <cmd> run on tmpPkgPath, and then compares them. 

If this makes sense I can change the comment and the name of the dir to make it more clear, but please correct me if I am misunderstanding something

e2e/e2e_test.go
natasha41575: updated

pkg/runner/config.go
natasha41575: moved to pkg/test/runner

pkg/runner/runner.go
natasha41575: done

pkg/runner/runner.go
natasha41575: in pipeline there is no --results-dir argument so I changed it to transform tmpPkgPath and compare it against orgPkgPath. but it is different and that doesn't work for fn so I updated the PR to handle fn differently

droot (approved):
Awesome!
/lgtm

pkg/test/runner/config.go
```diff
@@ -0,0 +1,99 @@
...
+
+func newTestCaseConfig(path string) (TestCaseConfig, error) {
+	configPath := filepath.Join(path, expectedDir, expectedConfigFile)
+	b, err := ioutil.ReadFile(configPath)
+	if os.IsNotExist(err) {
+		// return default config
+		return TestCaseConfig{
+			ExitCode: 0,
+			Network:  false,
+			RunTimes: 2,
```
mengqiy, line 27: We default RunTimes to 2 for functions catalog to ensure idempotency.
Does 2 still make sense for pipeline use case?

pkg/test/runner/config.go
```diff
@@ -0,0 +1,99 @@
...
+		return TestCaseConfig{}, fmt.Errorf("filed to read test config file: %w", err)
+	}
+
+	var config TestCaseConfig
+	err = yaml.Unmarshal(b, &config)
+	if err != nil {
+		return config, fmt.Errorf("failed to unmarshal config file: %s\n: %w", string(b), err)
+	}
+	if config.RunTimes == 0 {
+		config.RunTimes = 2
```
mengqiy, line 40: Same question here

mengqiy (commented):
1 minor comment, otherwise LGTM

frankfarzan:
> @droot @Shell32-Natsu @mengqiy @frankfarzan
> Based on our discussion today, I've updated the PR to include the test harness from kpt-functions-catalog that Donny wrote with a small tweak to allow it to work for pipeline. I've also included a basic pipeline run command test here. The kpt-functions-catalog repo can import and reuse this test harness when it is ready.

pkg/test/runner/config.go
natasha41575: Talked to Sunil and this is not a requirement for pipeline. As a result I've changed the default to 1, thanks for pointing this out. But that means on your functions-catalog tests you will have to specify your tests to run twice when you need to check idempotency

e2e/e2e_test.go
```diff
@@ -24,12 +24,42 @@ import (
...
+	}
+	for _, c := range *cases {
+		c := c // capture range variable
+		t.Run(c.Path, func(t *testing.T) {
+			t.Parallel()
+			r, err := runner.NewRunner(c)
+			if err != nil {
+				t.Fatalf("failed to create test runner: %s", err)
+			}
+			err = r.Run("pipeline", t)
```
Shell32-Natsu, line 55: The runner type probably should be passed when we create the runner (`runner.NewRunner(c)`). And it will be good to have a const value and type defined instead of strings.

pkg/test/runner/runner.go
```diff
@@ -0,0 +1,272 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to prepare package: %w", err)
+	}
+
+	// run function
+	command := run.GetMain()
+	kptArgs := []string{"pipeline", "run", tmpPkgPath}
+	for i := 0; i < r.testCase.Config.RunTimes; i++ {
+		command.SetArgs(kptArgs)
+		e2e.Exec(t, command)
```
Shell32-Natsu, line 145: This will make the test fail directly if there is an error. What if an error is expected?

e2e/e2e_test.go
natasha41575: NewRunner now takes a second argument for the command. And I defined consts for the command names

pkg/test/runner/runner.go
natasha41575: changed it so that the error gets returned and test for pipeline can be written to expect errors

natasha41575:
@Shell32-Natsu please take another look when you get the chance, thank you

Shell32-Natsu (approved):
lgtm

PR #1382, opened 2021-01-21
use preprocessing to handle the first time apply with inventorypolicy

Liujingfang1:
With `inventory-policy` flags added, users have two options: `strict` and `adopt`. `strict` means an object can be updated/pruned only when the `owning-inventory` annotation matches with the current inventory object. `adopt` means that an object can be updated/pruned when the `owning-inventory` annotation matches or this annotation is not present. The default option is `strict`. 

When package was applied by previous version of `kpt`, which doesn't have the `inventory-policy` flag. We need to handle the first time apply when user switches to the new version of `kpt`. In the first time apply, the inventory-policy should be `adopt` so that the resources will be added the `owning-inventory` annotation. 

In this change, the first time apply is detected by checking the label "apps.kubernetes.io./managed-by". If this label is missing, we treat it as the first time apply and add this label to the inventory object.

This PR depends on https://github.com/kubernetes-sigs/cli-utils/pull/316

### DIFF
--- a/commands/applycmd.go
+++ b/commands/applycmd.go
@@ -8,14 +8,21 @@ import (
 
 	"github.com/GoogleContainerTools/kpt/internal/util/setters"
 	"github.com/GoogleContainerTools/kpt/pkg/live"
+	"github.com/GoogleContainerTools/kpt/pkg/live/preprocess"
 	"github.com/spf13/cobra"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/klog"
-	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"sigs.k8s.io/cli-utils/cmd/apply"
+	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
 	"sigs.k8s.io/cli-utils/pkg/manifestreader"
 	"sigs.k8s.io/cli-utils/pkg/provider"
+)
+
+const (
+	inventoryPolicyFlag = "inventory-policy"
+	strictPolicy        = "strict"
 )
 
 // Get ApplyRunner returns a wrapper around the cli-utils apply command ApplyRunner. Sets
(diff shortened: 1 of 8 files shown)


commands/applycmd.go
```diff
@@ -8,23 +8,30 @@ import (
...
 	"sigs.k8s.io/cli-utils/cmd/apply"
+	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
 	"sigs.k8s.io/cli-utils/pkg/manifestreader"
 	"sigs.k8s.io/cli-utils/pkg/provider"
 )
 
+const (
+	inventoryPolicyFlag = "inventory-policy"
+	strictPolicy        = "strict"
```
seans3, line 25: Does these consts belong here? Or in a more central location (e.g. `pkg/inventory` or `pkg/common`)

seans3 (commented):
Small question about consts

commands/applycmd.go
Liujingfang1: Agreed that they should be from some common locations. Moved them as constants in cli-utils.

Liujingfang1:
@seans3 Addressed the comments. It's ready for another review.

mortent:
I think this looks good. Would the use of cobra command wrappers be simplified if we just implement the commands in kpt instead of cli-utils?
And is there a way we can add tests for some of this?

Liujingfang1:
> I think this looks good. Would the use of cobra command wrappers be simplified if we just implement the commands in kpt instead of cli-utils?

Yes, it would simplify the implementation a lot. We don't need to inject the preprocessing from the wrapper to the actual cli.

> And is there a way we can add tests for some of this?
I'm also thinking about this. The existing e2e test, specifically the one for testing `kpt live` of different kpt versions, covers the first-time apply case. Besides that, we can add a unit test for the `PreProcess` function.

Liujingfang1:
@mortent I added unit tests for the pre processing function. PTAL

seans3:
looks good to me. I'll let Morten also sign-off.

mortent (approved):

PR #1385, opened 2021-01-23
Binary only release artifacts

### LINKED ISSUE #1349: Improve availability of kpt binaries through github
Our release process currently use goreleaser to build and publish kpt artifacts to github. But we also do a step after this during the release to make the kpt binary available through GCS directly. This has two benefits for users:
 * Makes it easier to fetch just the binary without getting the `tar.gz` bundle.
 * It is possible to use single URL to always fetch the latest release.

We should move away from publishing binaries through GCS and instead always publish through github. We have some options here:
 * Stop releasing just the binaries and just ask users to fetch the artifacts already available through github
 * Publish the binaries as well as the `tar.gz` artifacts through github. If we name the binaries without including the version (so something like `kpt_<platform>_<arch>`), users will be able to download the binaries using the url: `https://github.com/GoogleContainerTools/kpt/releases/latest/download/kpt_<platform>_<arch>`

I think we should aim for the last option. I'm not sure if this is something that can be handled by goreleaser or if we need to add additional cloud build steps to do this.

runewake2:
- GitHub tagged releases now include direct binary downloads. More info: https://goreleaser.com/customization/archive/
  - New binary only releases should be available at `kpt_darwin_amd64/kpt` (exact name varies by platform).
- kpt release/README includes documentation for dry-running goreleaser. This can be used for locally testing changes to goreleaser configs.

Fixes #1349

### DIFF
--- a/release/README.md
+++ b/release/README.md
@@ -34,3 +34,17 @@ also updated with the `latest` tag for tagged releases.
 - `kpt-dev` release buckets
   - `gs://kpt-dev/latest`
   - `gs://kpt-dev/releases`
+
+# Dry-Run Goreleaser
+
+To test local changes to the [`goreleaser.yaml`](./tag/goreleaser.yaml) config. You may
+[install goreleaser](https://goreleaser.com/install/) locally and provide the
+`--skip-verify --skip-publish` flags.
+
+From the kpt directory you would run:
+
+```sh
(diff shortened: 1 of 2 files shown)


mortent (approved):
With this change, I don't think we need to upload binaries to GCS. We should look at just removing that from the cloud build job.

PR #1388, opened 2021-01-25
Add function to get function filter chain

### LINKED ISSUE #1346: Implement function that returns function chain for given pipeline
Main ask here is to implement a function (refer to this [PR](https://github.com/GoogleContainerTools/kpt/pull/1321/files)) that return a function chain as slice of kio.Filter that can be provided to a `kio.Pipeline` to execute.

Copying the placeholder function from the [PR](https://github.com/GoogleContainerTools/kpt/pull/1321/files). Function will read the function specs specified in the `validators, generators, transformer` section and convert that to corresponding kio.Filter (builtin, docker-runtime etc.)

```Go
// fnFilters returns chain of functions that are applicable
// to a given pipeline.
func fnFilters(_ *Pipeline) []kio.Filter
```

Shell32-Natsu:
close #1346


internal/pipeline/pipeline.go
```diff
@@ -228,6 +247,64 @@ func (f *Function) Validate() error {
...
+	case f.ConfigPath != "":
+		file, err := os.Open(f.ConfigPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to open config file path %s: %w", f.ConfigPath, err)
+		}
+		b, err := ioutil.ReadAll(file)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read content from config file: %w", err)
+		}
+		dataNode, err = yaml.Parse(string(b))
```
droot, line 277: I am thinking of a CUJ where someone implements a generator function that takes a KRM file as input (that is CR in this case), wrapping it in a configmap seems odd and will require them to implement extra unwrapping logic in their code.
I think relaying the config content as-is without wrapping is desirable here.

internal/pipeline/pipeline_test.go
```diff
@@ -620,3 +619,87 @@ func TestValidatePath(t *testing.T) {
...
+			expected: `apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: function-input
+data: {foo: bar}
+`,
+		},
+	}
+
+	for _, c := range cases {
```
mortent, line 691: Use golang subtests here instead of adding the test name to the error messages.

Shell32-Natsu: done

Shell32-Natsu:
Added the reverted docker runtime in this PR. @droot

internal/pipeline/executor.go
```diff
@@ -283,18 +288,22 @@ func filterMetaData(resources []*yaml.RNode) []*yaml.RNode {
...
-	return []kio.Filter{
-		&fnRunner{
-			fn: fn,
-		},
+	var ans []kio.Filter
+	for _, f := range filters {
+		ans = append(ans, f)
 	}
+
+	ans = append(ans, &fnRunner{
```
droot, line 302: I am thinking may be we can support some builtin functions with special image path `set-annotation`, that can help us test the pipeline (without hardcoding) without running docker.

internal/pipeline/executor.go
```diff
@@ -283,18 +288,22 @@ func filterMetaData(resources []*yaml.RNode) []*yaml.RNode {
...
 
-	// TODO(droot): Implement this with the logic to create
-	// function chain from a pipeline
-	// hardcoding built-in set-annotation function for testing
-	return []kio.Filter{
-		&fnRunner{
-			fn: fn,
-		},
+	var ans []kio.Filter
+	for _, f := range filters {
```
droot, line 298: It's not clear to me why we need this extra loop given that fnRunner implements `kio.Filter`.

internal/pipeline/cmd.go
```diff
@@ -1,3 +1,17 @@
+// Copyright 2019 Google LLC
```
droot, line 1: nit: 2019

There should be some tooling to catch these :)

droot (commented):
This is looking in good shape. I need to look at the tests once again, but some minor comments.

internal/pipeline/cmd.go
Shell32-Natsu: I believe this line is added by some tool :(. It's added after running `make all`.

Shell32-Natsu: Added by this line

https://github.com/GoogleContainerTools/kpt/blob/739cb5e352b599a96f6f3e33e7593489c6481fd9/scripts/update-license.sh#L18

internal/pipeline/executor.go
Shell32-Natsu: I am not familiar wit this part in Go. `filters` is in type `[]*fnRunner` and we need to return type `[]kio.Filter`. Type assertion doesn't work with this error message: `invalid type assertion: filters.([]kio.Filter) (non-interface type []*fnRunner on left)`. If I use `fns = append(fns, filters...)` the error message will be `cannot use filters (type []*fnRunner) as type []kio.Filter in append`.

I don't know why a loop can work :(

internal/pipeline/executor.go
Shell32-Natsu: How about `kpt.dev/builtin/set-annotation`?

internal/pipeline/executor.go
droot: I think I know now why this won't work unless in all the functions we use kio.Filter instead of fnRunner.

Not a big deal. Lets get on a call to try that.

droot (commented):
This is looking good. Let's get on a call to resolve that type issue between kio.Filter and fnRunner.

internal/pipeline/executor.go
```diff
@@ -283,18 +288,22 @@ func filterMetaData(resources []*yaml.RNode) []*yaml.RNode {
...
-		},
+	var fns []kio.Filter
+	for _, f := range filters {
+		fns = append(fns, f)
 	}
+
+	fns = append(fns, &fnRunner{
+		fn: &annotator{
+			key:   "builtin/setter-1",
+			value: "test",
```
natasha41575, line 305: are we keeping this function in for testing? is the plan to remove it later?

Shell32-Natsu: I think we plan to remove it later.

internal/pipeline/pipeline.go
```diff
@@ -228,6 +248,68 @@ func (f *Function) Validate() error {
...
 }
 
+// runner returns a fnRunner from the image and configs of
+// this function.
+func (f *Function) runner() (kio.Filter, error) {
+	config, err := f.config()
+	if err != nil {
+		return nil, err
+	}
+	return &fnRunner{
```
droot, line 258: I think it will be better to look at f.Image to see if it uses `set-annotator`, then return the buildin annotator function. That way we can avoid hardcoding. WDYT ?

droot (commented):
Awesome. small suggestion about removing the hard coding for the built-in fn.

Makefile
```diff
@@ -69,6 +69,9 @@ license-check:
 test:
 	go test -cover ./...
 
+test-docker:
```
frankfarzan, line 72: can you add a description for top-level make targets? Why would someone run this?

internal/pipeline/executor.go
frankfarzan: I suggest not conflating product code with examples. It makes it hard to follow the purpose of this PR. Instead leave TODO(s) with work that needs to be done in a future PR.

internal/pipeline/pipeline.go
```diff
@@ -119,6 +121,24 @@ func (p *Pipeline) Validate() error {
 	return nil
 }
 
+// fnChain returns a slice of function runners from the
+// functions and configs defined in pipeline.
+func (p *Pipeline) fnChain() ([]kio.Filter, error) {
+	fns := []Function{}
+	fns = append(fns, p.Generators...)
+	fns = append(fns, p.Transformers...)
+	fns = append(fns, p.Validators...)
```
frankfarzan, line 130: Do you currently treat validators differently from executors? Validators are not allowed to modify ResourceList.items

Leave TODO(s) for any functionality that's punted to future PR(s). Otherwise, it makes it hard to distinguish if something is a potential bug.

internal/pipeline/pipeline.go
```diff
@@ -228,6 +248,68 @@ func (f *Function) Validate() error {
 	return nil
 }
 
+// runner returns a fnRunner from the image and configs of
+// this function.
+func (f *Function) runner() (kio.Filter, error) {
```
frankfarzan, line 253: Split Function and its methods into a separate file. This file can then only include Pipeline

internal/pipeline/pipeline.go
```diff
@@ -228,6 +248,68 @@ func (f *Function) Validate() error {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse config %s: %w", string(b), err)
+		}
+		// directly use the config from file
+		return node, nil
+	case !isNodeZero(&f.Config):
+		// directly use the inline config
+		return yaml.NewRNode(&f.Config), nil
+	case len(f.ConfigMap) != 0:
+		node = yaml.NewMapRNode(&f.ConfigMap)
```
frankfarzan, line 288: Logic after the switch should only be done for this case:

return wrapConfigMap(node)

internal/pipeline/pipeline.go
```diff
@@ -228,6 +248,68 @@ func (f *Function) Validate() error {
...
+	default:
+		node = nil
```
frankfarzan, line 290: you don't need to pass a ConfigMap if there's no function config.

internal/pipeline/pipeline_test.go
```diff
@@ -620,3 +636,88 @@ func TestValidatePath(t *testing.T) {
...
+		configFileContent string
+		expected          string
+	}
+
+	cases := []input{
+		{
+			name: "no config",
+			fn:   Function{},
+			expected: `apiVersion: v1
+kind: ConfigMap
```
frankfarzan, line 653: you don't need to populate ResourceList.functionConfig at all.

internal/pipeline/pipeline.go
```diff
@@ -228,6 +248,68 @@ func (f *Function) Validate() error {
...
+			Image: f.Image,
+		},
+		fnConfig: config,
+	}, nil
+}
+
+func (f *Function) config() (*yaml.RNode, error) {
+	var node *yaml.RNode
+	switch {
+	case f.ConfigPath != "":
```
frankfarzan, line 269: Is there a TODO to special handle these resources?

    # 'configPath' specifies a relative path to a file in the current directory containing a K8S resource used
    # as the function config.
    # This resource is excluded when resolving 'sources', and as a result
    # cannot be operated on by the pipeline.

frankfarzan (changes requested):

internal/pipeline/pipeline.go
Shell32-Natsu: If we want to use a special image name then I think it's better to make it more distinct, like `builtin-set-annotator`.

internal/pipeline/pipeline.go
Shell32-Natsu: We haven't implemented the function to resolve the `sources`. I think here is not the good place to add such todo.

internal/pipeline/executor.go
```diff
@@ -283,18 +288,22 @@ func filterMetaData(resources []*yaml.RNode) []*yaml.RNode {
...
-			fn: fn,
-		},
+	var ans []kio.Filter
+	for _, f := range filters {
+		ans = append(ans, f)
 	}
+
+	ans = append(ans, &fnRunner{
+		fn: &annotator{
+			key:   "builtin/setter-1",
```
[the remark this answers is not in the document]
Shell32-Natsu: This has been removed

internal/pipeline/pipeline.go
Shell32-Natsu: TODO added.

internal/pipeline/pipeline.go
Shell32-Natsu: done

internal/pipeline/pipeline.go
Shell32-Natsu: done

natasha41575:
lgtm, will leave for Frank or Sunil to approve

Shell32-Natsu:
@frankfarzan @droot PTAL

internal/pipeline/pipeline_function.go
```diff
@@ -0,0 +1,220 @@
...
+
+	return nil
+}
+
+// runner returns a fnRunner from the image and configs of
+// this function.
+func (f *Function) runner() (kio.Filter, error) {
+	// TODO: remove this builtin placeholder function
+	if f.Image == builtinSetAnnotatorImage {
+		for k, v := range f.ConfigMap {
```
droot, line 93: As discussed, update the annotator function implementation to take configMap as fnConfig so that it behaves like any other function.

Shell32-Natsu: done

droot (approved):

droot:
After this PR is merged, I would like to refactor the code to bring in line with the new design before we start any new work on it. /cc @natasha41575

internal/pipeline/fn.go
```diff
@@ -45,10 +45,7 @@ type KRMFn interface {
 // Built-in functions that are compiled into kpt
 
 // annotator is built kpt function for annotating KRM resources.
-type annotator struct {
-	key   string
-	value string
-}
+type annotator struct{}
```
frankfarzan, line 48: I made this comment somewhere else: Without having TODO(s), it's not clearly why there's an example function in a file called fn.go, and it's hard for reviewers to actually give feedback. If we need a placeholder, then it needs to be communicated properly.

internal/pipeline/pipeline_function.go
```diff
@@ -0,0 +1,216 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/GoogleContainerTools/kpt/internal/pipeline/runtime"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+const (
+	builtinSetAnnotatorImage string = "builtin/set-annotator"
```
frankfarzan, line 33: I wouldn't start by having concept of built-ins. Instead just support container-based functions for now. Per the other comment, need to have a clean between the orchestration logic and example functions.

internal/pipeline/executor.go
```diff
@@ -246,12 +246,17 @@ func hydrate(p *pkg, hctx *hydrationContext) (resources []*yaml.RNode, err error
 	// TODO(droot): parameterize the sink-mode (hctx.sinkMode) to determine whether to
 	// write it in-place or not. Currently in-place.
 	pkgWriter := &kio.LocalPackageWriter{PackagePath: p.Path()}
+	// we will gather filters from the pipeline
```
frankfarzan, line 249: Such comments is not making things more clear (it's just rephrasing the function call in words). Either expand or remove it.

internal/pipeline/pipeline.go
frankfarzan: Can you add a TODO where it's most appropriate? This is a signal to reviewers that this is not yet done instead of a an oversight. It also makes it easy to grep for TODOs to figure out what's left to do.

frankfarzan (changes requested):

internal/pipeline/pipeline_function.go
Shell32-Natsu: Using a builtin function allows us to run pipeline tests without docker.

Shell32-Natsu:
According to the offline discussion, we decided to remove builtin `set-annotator` and only support docker functions. @droot @frankfarzan 

The e2e test is also updated. @natasha41575

frankfarzan (approved):

internal/pipeline/pipeline_function.go
```diff
@@ -0,0 +1,205 @@
...
+		},
+		fnConfig: config,
+	}, nil
+}
+
+func (f *Function) config() (*yaml.RNode, error) {
+	var node *yaml.RNode
+	switch {
+	case f.ConfigPath != "":
+		file, err := os.Open(f.ConfigPath)
```
frankfarzan, line 103: ConfigPath is a slash-delimited relative path so you can't just os.Open() on it.

PR #1391, opened 2021-01-25
update functions for wrapping an inventory ResourceGroup

Liujingfang1:
Replace the existing function
```
func WrapInventoryInfoObj(obj *unstructured.Unstructured) inventory.InventoryInfo
```
by
```
func WrapInventoryResourceGroup(obj *unstructured.Unstructured) *InventoryResourceGroup 
```

The latter one can cover the use case of the former one. The latter one also helps simplify the integration with ConfigSync.


seans3 (approved):
/lgtm

PR #1393, opened 2021-01-26
Add Github Action to check for broken site links

etefera:
and fix broken links in documentation.

Fixes https://github.com/GoogleContainerTools/kpt/issues/1190


.github/workflows/sitecheck.yml
```diff
@@ -0,0 +1,42 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: SiteCheck
+on:
+  pull_request:
+    branches:
+      - master
```
frankfarzan, line 19: Are you planning to enable on 'next' branch as well?

frankfarzan (changes requested):

.github/workflows/sitecheck.yml
etefera: For now, just this branch, since we expect the `next` branch to have different build steps.

frankfarzan: Most of the doc changes going forward will be in `next`, so that's where checking for broken links is very useful.

frankfarzan (approved):

PR #1397, opened 2021-01-27
Update kpt live init documentation.

### LINKED ISSUE #1362: kpt live init --help Displays Two Different Sets of Flags
When running `kpt live init --help` the help content includes two sets of documented `Flags`. The flag options do not match.

```
➜ kpt live init --help
Initialize a package with a object to track previously applied resources

  kpt live init DIRECTORY [flags]

Args:

  DIR:
    Path to a package directory.  The directory must contain exactly
    one ConfigMap with the grouping object annotation.

Flags:

  --inventory-id:
    Identifier for group of applied resources. Must be composed of valid label characters.
  --namespace:
    namespace for the inventory object. If not provided, kpt will check if all the resources
    in the package belong in the same namespace. If they are, that namespace will be used. If
    they are not, the namespace in the user's context will be chosen.

Usage:
  kpt live init DIRECTORY

Examples:

  # initialize a package
  kpt live init my-dir/

  # initialize a package with a specific name for the group of resources
  kpt live init --namespace=test my-dir/


Flags:
  -h, --help                  help for init
  -i, --inventory-id string   Identifier for group of applied resources. Must be composed of valid label characters.
```

Specifically:

```
Flags:

  --inventory-id:
    Identifier for group of applied resources. Must be composed of valid label characters.
  --namespace:
    namespace for the inventory object. If not provided, kpt will check if all the resources
    in the package belong in the same namespace. If they are, that namespace will be used. If
    they are not, the namespace in the user's context will be chosen.
```

and

```
Flags:
  -h, --help                  help for init
  -i, --inventory-id string   Identifier for group of applied resources. Must be composed of valid label characters.
```

Should be displayed together to make this help content more clear.

runewake2:
This updates the `kpt live init` cli help. This changes the text that is selected from online documentation via `mdtogo`. Flags are documented via the Cobra command config:

```txt
Initialize a package with a object to track previously applied resources

The init command initializes a package with a template resource which will
be used to track previously applied resources so that they can be pruned
when they are deleted.

The template resource is required by other live commands
such as apply, preview and destroy.

Usage:
  kpt live init DIRECTORY

Examples:

  # initialize a package
  kpt live init my-dir/

  # initialize a package with a specific name for the group of resources
  kpt live init --namespace=test my-dir/


Flags:
  -h, --help                  help for init
  -i, --inventory-id string   Identifier for group of applied resources. Must be composed of valid label characters.
```

This removes the duplicated Flag values and provides additional context for the command.

> Note: other commands which have similar behavior to the original `kpt live init --help` have not been updated (for example `kpt pkg init --help`).

Fixes #1362

### DIFF
--- a/internal/docs/generated/livedocs/docs.go
+++ b/internal/docs/generated/livedocs/docs.go
@@ -137,22 +137,12 @@ var FetchK8sSchemaExamples = `
 
 var InitShort = `Initialize a package with a object to track previously applied resources`
 var InitLong = `
-  kpt live init DIRECTORY [flags]
+The init command initializes a package with a template resource which will
+be used to track previously applied resources so that they can be pruned
+when they are deleted.
 
-Args:
-
-  DIR:
-    Path to a package directory.  The directory must contain exactly
-    one ConfigMap with the grouping object annotation.
-
-Flags:
-
-  --inventory-id:
-    Identifier for group of applied resources. Must be composed of valid label characters.
-  --namespace:
-    namespace for the inventory object. If not provided, kpt will check if all the resources
-    in the package belong in the same namespace. If they are, that namespace will be used. If
-    they are not, the namespace in the user's context will be chosen.
+The template resource is required by other live commands
+such as apply, preview and destroy.
 `
 var InitExamples = `
   # initialize a package
--- a/site/content/en/reference/live/init/_index.md
+++ b/site/content/en/reference/live/init/_index.md
@@ -11,12 +11,14 @@ description: >
 
 {{< asciinema key="live-init" rows="10" preload="1" >}}
 
+<!--mdtogo:Long-->
 The init command initializes a package with a template resource which will
 be used to track previously applied resources so that they can be pruned
 when they are deleted.
 
 The template resource is required by other live commands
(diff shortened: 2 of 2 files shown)


runewake2:
The docs in this case are not necessarily out of date as much as they are confusing. Our website (and original mdtogo generated content) uses `inventory-id` and `namespace` as flags. Cobra uses the `inventory-id` flag and has a global flag for `namespace`. Because we have two Flags headers that are different reading it isn't clear which is the source of truth.

I'm not sure that the website docs are incorrect here. Calling out that you can specify the namespace of an inventory seems like relevant and useful information. We just render that information differently on the kpt cli.

There are a few other issues with mdtogo that may need to be fixed first before we try to redo the rest of the docs.

mortent:
Yeah, the `--namespace` flag is a bit tricky here. Adding the additional explanation here does make sense, but it isn't quite aligned with the other commands. But I think this is an improvement, so let's merge this and we can revisit the bigger question on how we should structure the output from commands later.

mortent (approved):

frankfarzan (approved):

PR #1398, opened 2021-01-27
Fix parsing of repos with .git

### LINKED ISSUE #1323: Handling of hidden directories in kpt get
When using `kpt pkg get` it seems that hidden directories are ignored.

As an example, I tried to include some Github Actions into a package, which are by convention located in `.github/workflows`. Locally it works without problems, as Github workflow files are just YAML files, which can be used and edited with `kpt`

So I have two issues:

- As I am new to kpt, I thought it might be good idea to include some Github workflows into the package. Is is OK to do that or an abuse according to `kpt`?
- If it is not an abuse, is there any possibility to fetch hidden directories such as `.github`?

This is probably related to https://github.com/GoogleContainerTools/kpt/issues/517, just in this case I want to "un-ignore" hidden directories.

I have created a repo, which can be used to reproduce this if needed:
https://github.com/msonnleitner/kpttest

natasha41575:
Close #1323 

Parsing should check that a repository name ends in .git, rather than blindly checking if the package name contains .git anywhere.

E.g. `kpt pkg get https://github.com/GoogleContainerTools/kpt/.github/workflows ./`

### DIFF
--- a/internal/cmdget/cmdget_test.go
+++ b/internal/cmdget/cmdget_test.go
@@ -165,6 +165,15 @@ func TestCmd_fail(t *testing.T) {
 	assert.Contains(t, err.Error(), "failed to lookup master(or main) branch")
 }
 
+// TestCmd_git tests that if a package path contains .git, it is still parsed correctly
+func TestCmd_git(t *testing.T) {
+	r := cmdget.NewRunner("kpt")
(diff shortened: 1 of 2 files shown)


mikebz:
thanks for a quick fix!  Is there a way to add some tests here so we don't regress in case this gets refactored?

phanimarupaka (approved):

PR #1416, opened 2021-02-04
added function to filter out function configs

### LINKED ISSUE #1410: `kpt pipeline run` should filter out function config resources
function config should not be mutated during the pipeline execution and good way to ensure that will be to not provide them as input to the functions.

natasha41575:
Filter out function config by not including resources of type configMap containing the annotation config.kubernetes.io/function

### DIFF
--- a/internal/pipeline/executor.go
+++ b/internal/pipeline/executor.go
@@ -320,20 +320,8 @@ func filterMetaData(resources []*yaml.RNode) []*yaml.RNode {
 func filterFnConfig(resources []*yaml.RNode) []*yaml.RNode {
 	var filtered []*yaml.RNode
 	for _, r := range resources {
-		var mn *yaml.MapNode
-		metaData := r.Field(yaml.MetadataField)
-		if metaData != nil {
-			mn = metaData.Value.Field(yaml.AnnotationsField)
-		}
-		annotations := map[string]string{}
-		if !mn.IsNilOrEmpty() {
-			_ = mn.Value.VisitFields(func(node *yaml.MapNode) error {
-				annotations[yaml.GetValue(node.Key)] = yaml.GetValue(node.Value)
-				return nil
-			})
-		}
 		m, _ := r.GetMeta()
-		if _, ok := annotations["config.kubernetes.io/function"]; !ok || m.Kind != "ConfigMap" {
+		if _, ok := m.Annotations["config.kubernetes.io/function"]; !ok || m.Kind != "ConfigMap" {
 			filtered = append(filtered, r)
 		}
 	}
--- a/internal/pipeline/executor.go
+++ b/internal/pipeline/executor.go
@@ -190,8 +190,8 @@ func (p *pkg) localResources(includeMetadata bool) (resources []*yaml.RNode, err
 		err = fmt.Errorf("failed to read resources for pkg %s %w", p.Path(), err)
 		return resources, err
 	}
+	resources = filterFnConfig(resources)
 	if !includeMetadata {
-		// TODO(droot): this will be the place where we filter kpt resources
 		resources = filterMetaData(resources)
 	}
 	return resources, err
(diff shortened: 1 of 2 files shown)


internal/pipeline/executor.go
```diff
@@ -308,8 +308,32 @@ func hydrate(p *pkg, hctx *hydrationContext) (resources []*yaml.RNode, err error
...
+	}
+	return filtered
+}
+
+// filterFnConfig filters out function configs
+func filterFnConfig(resources []*yaml.RNode) []*yaml.RNode {
+	var filtered []*yaml.RNode
+	for _, r := range resources {
+		var mn *yaml.MapNode
+		metaData := r.Field(yaml.MetadataField)
```
Shell32-Natsu, line 324: Please use `RNode.GetMeta` and `RNode.GetAnnotations`

internal/pipeline/executor.go
```diff
@@ -190,8 +190,8 @@ func (p *pkg) localResources(includeMetadata bool) (resources []*yaml.RNode, err
 		err = fmt.Errorf("failed to read resources for pkg %s %w", p.Path(), err)
 		return resources, err
 	}
+	resources = filterFnConfig(resources)
 	if !includeMetadata {
```
Shell32-Natsu, line 194: Not sure about here. Maybe @droot can clarify. Should we put `filterFnConfig` in this if?

internal/pipeline/executor.go
```diff
@@ -308,8 +308,20 @@ func hydrate(p *pkg, hctx *hydrationContext) (resources []*yaml.RNode, err error
 func filterMetaData(resources []*yaml.RNode) []*yaml.RNode {
```
droot, line 308: let's rename this function to `filterMetaResources` to reflect what it does.

internal/pipeline/executor.go
```diff
@@ -190,8 +190,8 @@ func (p *pkg) localResources(includeMetadata bool) (resources []*yaml.RNode, err
 		err = fmt.Errorf("failed to read resources for pkg %s %w", p.Path(), err)
 		return resources, err
 	}
+	resources = filterFnConfig(resources)
```
droot, line 193: I am still thinking the correct place to do it, but the intent is if there are resources that are provided as input to a function in the pipeline, we want to exclude such resources. So this will require reading the function config file and fetch the GVKNN (could be multiple) and filter those out. Basically, treat those function config as input to the pipeline, so pipeline shouldn't be applied to them.

Filtering Configmap on the basis of annotation `config.kubernetes.io/function` will filter our old style function definitions. I missed that, we need that filtering as well.

internal/pipeline/executor.go
```diff
@@ -308,8 +308,20 @@ func hydrate(p *pkg, hctx *hydrationContext) (resources []*yaml.RNode, err error
...
+		m, _ := r.GetMeta()
+		if !strings.Contains(m.APIVersion, "kpt.dev") {
+			filtered = append(filtered, r)
+		}
+	}
+	return filtered
+}
+
+// filterFnConfig filters out function configs
+func filterFnConfig(resources []*yaml.RNode) []*yaml.RNode {
```
droot, line 320: This is definitely needed. pl. see my comment above about additional filtering. Feel free to do that in separate PR if that helps.

internal/pipeline/executor.go
```diff
@@ -308,8 +308,20 @@ func hydrate(p *pkg, hctx *hydrationContext) (resources []*yaml.RNode, err error
...
+	var filtered []*yaml.RNode
+	for _, r := range resources {
+		m, _ := r.GetMeta()
+		if _, ok := m.Annotations["config.kubernetes.io/function"]; !ok || m.Kind != "ConfigMap" {
```
droot, line 324: Somehow my brain took little bit more time to parse the above conditional logic :) consider writing it in selection form (seems simpler to read):

```Go
_, hasFnAnnotation := m.Annotations["config..."]
if m.Kind == "ConfigMap" && hasFnAnnotation {
 continue
}
filtered = append(...)
```

internal/pipeline/executor.go
natasha41575: resolving this conversation for now because such filtering will be included on a different PR if we decide to do it

internal/pipeline/executor.go
natasha41575: done

internal/pipeline/executor.go
natasha41575: moved the isfnconfig check to inside the filtermetaresources function. @droot wdyt?

internal/pipeline/executor.go
natasha41575: done

internal/pipeline/executor.go
```diff
@@ -303,19 +302,25 @@ func hydrate(p *pkg, hctx *hydrationContext) (resources []*yaml.RNode, err error
...
 			filtered = append(filtered, r)
 		}
 	}
 	return filtered
 }
 
+// isFnConfig filters out function configs
+func isFnConfig(m *yaml.ResourceMeta) bool {
+	_, hasFnAnnotation := m.Annotations["config.kubernetes.io/function"]
+	return m.Kind == "ConfigMap" && hasFnAnnotation
```
frankfarzan, line 321: Are you assuming that function configs MUST of type ConfigMap and MUST have such an annotation?

In general, a function config can be of any kind. And we shouldn't need this annotation anymore given we have an explicit Pipeline declaration. Function configs can be found by following the `config` field in the Pipeline.

@droot 

We still need a mechanism for such files to not be applied when calling `kpt live apply`. Current mechanism is to use `config.kubernetes.io/local-config: "true"` annotation.

internal/pipeline/executor_test.go
```diff
@@ -79,6 +79,18 @@ kind: Custom
 spec:
   image: nginx:1.2.3`,
 				`
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-func-config
+  annotations:
+    config.kubernetes.io/function: |
```
frankfarzan, line 87: This annotation is superseded by the Pipeline declaration.

frankfarzan (changes requested):

internal/pipeline/executor.go
frankfarzan: Before getting too deep into implementation, we need to have examples packaging using the v1alpha2 checked into `package-examples`. Otherwise, there's chance of misunderstanding the design spec.

frankfarzan: #1421

natasha41575: understood, thank you for the input. will discuss with Sunil

natasha41575:
closing for now

Issue #1377, opened 2021-01-21
Bug: Inscrutable remote diff output


karlkfi:
```
$ kpt pkg diff --diff-type remote
Only in /tmp/kpt-602117789: kpt-gen.sh
Only in /tmp/kpt-602117789: tenant.yaml
```

What does this output mean? 
Does it mean "these files are only remote or only local"?

If so, can we name the temp file directory local-<suffix> and remote-<suffix> so it's possible to tell which is which?

runewake2:
Hi Karl, I've put together a change that hopefully adds some extra context around what the `diff` command is doing. There's two parts to the change:

The first prints out where the information in these directories is being sourced from before the diff results: 
`Staging https://github.com/kubernetes/examples/staging/cockroachdb:master at /tmp/kpt-upstream-examples-cockroachdb-master-161806574`

The second changes the directory names we use to be more descriptive: `kpt-12348765` -> `/kpt-upstream-examples-cockroachdb-master-12348765` for example.

In your case your output might look like:

```txt
$ kpt pkg diff --diff-type remote
Only in /tmp/kpt-upstream-githubProject-parentDirectory-master-602117789: kpt-gen.sh
Only in /tmp/kpt-upstream-githubProject-parentDirectory-master-602117789: tenant.yaml
```

Do these changes help clarify the results of a `kpt pkg diff`? Is the output this shows clear now or are there other changes that would help improve the experience here (for example: is the new directory name too long or the "Staging..." text helpful)?

mikebz:
@runewake2 is there a way to inercept `/tmp/kpt-upstream-githubProject-parentDirectory-master-602117789` and just make it `upstream`?

runewake2:
Yes, I think we could do that but it wouldn't catch all issues. `pkg diff` uses the local diffing tool and that varies across systems. That behavior can also be overwritten if you'd prefer a graphical diff like [`meld`](https://meldmerge.org/), replacing stdout output would have no effect on these graphical tools.

If we are transforming diffs output I think we'd want to include that in addition to the changes mentioned above.

The change you describe might look like:

```txt
Only in local: deployment.yaml
Only in upstream@master: tenant.yaml
Only in upstream@ProjectVersion: pod.yaml
```

> we can not just use `upstream` here because `remote` and `3way` diffs will calculate the diffs between two different versions of the upstream repository.

Or something like

```txt
diff local/deployment.yaml upstream@ProjectVersion/deployment.yaml
```

[...]

PR #1419, opened 2021-02-05
Update Diffing Display

runewake2:
Attempts to improve the UX of kpt package diffs like experienced in #1377

#### update upstream directory names

`kpt-12348765` might now appear as `kpt-upstream-examples-cockroachdb-ProjectVersion-12348765` where:
  - `upstream` references that the package was fetched from the source package. The local copy will appear as `local` instead.
  - `examples` references the Git project name. In this case taken from: https://github.com/kubernetes/examples
  - `cockroachdb` references the directory containing the upstream package. In this case taken from: staging/cockroachdb
  - `ProjectVersion` is a string constant that represents a version stored in the projects Kptfile. During remote or 3way diffs this may also appear as the provided reference (`main`, `master` etc).

#### add output of staged directories to stdout, this information looks like this:

```txt
Staging cockroachdb at /tmp/kpt-local-943465024
Staging https://github.com/kubernetes/examples/staging/cockroachdb:PackageVersion at /tmp/kpt-upstream-examples-cockroachdb-ProjectVersion-535434399
Staging https://github.com/kubernetes/examples/staging/cockroachdb:master at /tmp/kpt-upstream-examples-cockroachdb-master-555934580
```

  - `PackageVersion` represents a package which matches the package version stored in the projects Kptfile
  - `master` will represent the remote reference used during a 3way diff.

> The `/tmp` directory will appear differently depending on the system (MacOS, Linux or Windows OS).

#### There are some known issues with this approach:

1. When a diffing UI is used the "Staging..." text may not be obvious (hidden by the UI window) and the directories do not capture the complete information yet.
2. Performing diffs against a series of temporary directories is still confusing. Note here: Kpt's diff is not atomic and will change the directory it is run against - this is the motivation for tmp directories - this prevents expected differences from being diffed like in the Kptfile but also means that a diff should not be run directly against the local package.

### DIFF
--- a/internal/util/diff/diff.go
+++ b/internal/util/diff/diff.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -123,12 +124,11 @@ func (c *Command) Run() error {
 	// get the upstreamPkg at current version
 	upstreamPkg, err := c.PkgGetter.GetPkg(kptFile.Upstream.Git.Repo,
 		kptFile.Upstream.Git.Directory,
-		kptFile.Upstream.Git.Commit,
-		"PackageVersion")
-	fmt.Printf("Staging %s/%s:%s at %s\n",
+		kptFile.Upstream.Git.Commit)
+	fmt.Printf("Staging %s/%s@%s at %s\n",
 		kptFile.Upstream.Git.Repo,
 		kptFile.Upstream.Git.Directory,
-		"PackageVersion",
+		kptFile.Upstream.Git.Commit,
 		upstreamPkg)
 	if err != nil {
 		return err
@@ -154,12 +154,11 @@ func (c *Command) Run() error {
 		// get the upstream pkg at the target version
 		upstreamTargetPkg, err = c.PkgGetter.GetPkg(kptFile.Upstream.Git.Repo,
 			kptFile.Upstream.Git.Directory,
-			c.Ref,
 			c.Ref)
 		if err != nil {
 			return err
 		}
-		fmt.Printf("Staging %s/%s:%s at %s\n",
+		fmt.Printf("Staging %s/%s@%s at %s\n",
 			kptFile.Upstream.Git.Repo,
 			kptFile.Upstream.Git.Directory,
 			c.Ref,
@@ -296,7 +295,7 @@ func (d *defaultPkgDiffer) prepareForDiff(dir string) error {
 
 // PkgGetter knows how to fetch a package given a git repo, path and ref.
 type PkgGetter interface {
-	GetPkg(repo, path, ref, refDesc string) (dir string, err error)
+	GetPkg(repo, path, ref string) (dir string, err error)
 }
 
 // defaultPkgGetter uses get.Command abstraction to implement PkgGetter.
@@ -308,13 +307,9 @@ type defaultPkgGetter struct{}
 // path is the sub directory of the git repository that the package was cloned from
 // ref is the git ref the package was cloned from
 // refDesc is a human readable name of the reference
-func (pg defaultPkgGetter) GetPkg(repo, path, ref, refDesc string) (string, error) {
-	repoSrc := strings.Split(repo, "/") // For github repo's this will be the project name
-	pkgSrc := strings.Split(path, "/")  // This will be the directory the package is contained in
-	tmpPath := fmt.Sprintf("kpt-upstream-%s-%s-%s-",
-		repoSrc[len(repoSrc)-1],
-		pkgSrc[len(pkgSrc)-1],
-		refDesc)
+func (pg defaultPkgGetter) GetPkg(repo, path, ref string) (string, error) {
+	tmpPath := fmt.Sprintf("kpt-upstream-%s-",
+		ref[0:int(math.Min(float64(len(ref)), 7))])
 	dir, err := ioutil.TempDir("", tmpPath)
 	if err != nil {
 		return dir, err
--- a/internal/util/diff/diff.go
+++ b/internal/util/diff/diff.go
@@ -104,7 +104,7 @@ func (c *Command) Run() error {
 
 	// Stage current package
 	// This prevents prepareForDiff from modifying the local package
-	currPkg, err := ioutil.TempDir("", "kpt-local-")
+	currPkg, err := ioutil.TempDir("", "kpt-")
 	if err != nil {
 		return errors.Errorf("failed to create stage dir for current package: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


runewake2:
One note: there is a potential change here that would include part or all of the remote reference in the directory name. This would solve the 2nd listed issue above but makes the directory names potentially very long.

During a remote or three way diff the output would change like this:

```diff
- Staging https://github.com/kubernetes/examples/staging/cockroachdb:PackageVersion at /var/folders/5_/kg1tj70j275ctm9tqs59vz6h00k8mw/T/kpt-upstream-examples-cockroachdb-213532293
+ Staging https://github.com/kubernetes/examples/staging/cockroachdb:PackageVersion at /var/folders/5_/kg1tj70j275ctm9tqs59vz6h00k8mw/T/kpt-upstream-examples-cockroachdb-75f116a8b55b6a31ab64544d60d1f6ce60053507-213532293
- Staging https://github.com/kubernetes/examples/staging/cockroachdb:master at /var/folders/5_/kg1tj70j275ctm9tqs59vz6h00k8mw/T/kpt-upstream-examples-cockroachdb-525830738
+ Staging https://github.com/kubernetes/examples/staging/cockroachdb:master at /var/folders/5_/kg1tj70j275ctm9tqs59vz6h00k8mw/T/kpt-upstream-examples-cockroachdb-master-525830738
```

Interested in thoughts on this change.

mikebz:
is there a test we need to develop or update to make sure we don't regress this functionality in the future?

runewake2:
We do not currently have tests over the specific parts of code I've touched here. However there are a few tests covering the complete diff functionality.

I'm working on developing tests in parallel but wanted to create this first to get feedback on the experience it creates.

internal/util/diff/diff.go
```diff
@@ -102,34 +103,44 @@ func (c *Command) Run() error {
...
+	currPkg, err := stageDirectory(stagingDirectory, "local-package")
+	if err != nil {
+		return errors.Errorf("failed to create stage dir for current package: %v", err)
+	}
+
 	err = copyutil.CopyDir(c.Path, currPkg)
 	if err != nil {
 		return errors.Errorf("failed to stage current package: %v", err)
 	}
+	fmt.Printf("Staging %s at %s\n", c.Path, currPkg)
```
mikebz, line 129: is this the correct ordering of arguments.  Seems like: Staging a <package> at <directory>?

internal/util/diff/diff.go
```diff
@@ -102,34 +103,44 @@ func (c *Command) Run() error {
...
-	upstreamPkg, err := c.PkgGetter.GetPkg(kptFile.Upstream.Git.Repo,
+	upstreamPkg, err := c.PkgGetter.GetPkg(stagingDirectory,
+		kptFile.Upstream.Git.Repo,
 		kptFile.Upstream.Git.Directory,
 		kptFile.Upstream.Git.Commit)
+	fmt.Printf("Staging %s/%s@%s at %s\n",
+		kptFile.Upstream.Git.Repo,
+		kptFile.Upstream.Git.Directory,
+		kptFile.Upstream.Git.Commit,
+		upstreamPkg)
```
mikebz, line 140: same question as above.  Do people expect to see the directory where things are staged after the "at..."

internal/util/diff/diff.go
```diff
@@ -282,14 +294,21 @@ func (d *defaultPkgDiffer) prepareForDiff(dir string) error {
...
-	GetPkg(repo, path, ref string) (dir string, err error)
+	GetPkg(stagingDir, repo, path, ref string) (dir string, err error)
 }
 
 // defaultPkgGetter uses get.Command abstraction to implement PkgGetter.
 type defaultPkgGetter struct{}
 
-func (pg defaultPkgGetter) GetPkg(repo, path, ref string) (string, error) {
-	dir, err := ioutil.TempDir("", "kpt-")
+// GetPkg checks out a repository into a temporary directory for diffing
```
mikebz, line 303: <3 the extra comments.  Thanks for relaying the context!

internal/util/diff/diff.go
```diff
@@ -301,3 +320,17 @@ func (pg defaultPkgGetter) GetPkg(repo, path, ref string) (string, error) {
...
+
+// shortSha returns a shortened version of a commit SHA
+func shortSha(sha string) string {
+	return sha[0:int(math.Min(float64(len(sha)), 7))]
+}
+
+// stageDirectory creates a subdirectory of the provided path for temporary operations
+// path is the parent staged directory and should already exist
+// subpath is the subdirectory that should be created inside path
+func stageDirectory(path, subpath string) (string, error) {
```
mikebz, line 332: seems like there is a handy function but earlier there is an ioutils.TempDir call still.  Is this supposed to be the one place that standardizes our calls?  Why not use it?

internal/util/diff/diff.go
runewake2: This will print: `Staging dir at /tmp/local-package`. The format I have used is "Staging `some-source` at `some-staging-location`"

internal/util/diff/diff.go
runewake2: The `pkg` variables store the path to the directory containing the staged package. This may be refactored in the future. The path is returned from the `stageDirectory` and `GetPkg` functions.

internal/util/diff/diff.go
runewake2: I create a single parent temp directory (the `ioutils.TempDir` call you are referring to) and then inside that generated directory create additional staged directories. This function was written so the returned types match the signature of the `TempDir` function so that they may be swapped interchangeably as needed. The `TempDir` call still serves a purpose and does additional work to ensure conflicts are avoided when creating the directory. Once we have an empty working directory we can interact with it with less safeguards so random strings and additional safety checks were removed from this wrapper.

This change means that `/tmp/kpt-foo-12348765` will now appear as `/tmp/kpt-12348765/foo` and `/tmp/kpt-12348765/bar` for example.

This change also means that duplicate logic to cleanup our staging directories was removed and located in a single spot that cleans up the top level staging directory.

runewake2: To expand a little: if we remove the creation of a temp directory entirely then if the directory cleanup fails OR kpt exits prior to cleaning up the directories then diffs may become broken until that is fixed by the user/OS. Maintaining this behavior helps avoid that situation.

internal/util/diff/diff.go
mikebz: got it.  The variable naming is a bit misleading but it seems like it's printing the right stuff.  Thanks for clarifying!

runewake2:
After latest changes directories will be staged similar to:

```
/tmp/kpt-165532019/local-master-75f116a
/tmp/kpt-165532019/remote-master-75f116a
/tmp/kpt-165532019/target-master
```

The `target` directory is only used when `3way`, `combined` or `remote` diffs are run.

`target-master` relies upon the passed in ref to diff against. This defaults to `master` or `main` but can be a tag, branch or sha depending on the values provided.

mortent (approved):

PR #1425, opened 2021-02-07
Adds RBAC error during apply end-to-end test case

seans3:
* Adds test case in end-to-end test when an RBAC error happens during apply.
* Validates the case is reported as `failed` and the object does not end up in the inventory.

### DIFF
--- a/e2e/live/end-to-end-test.sh
+++ b/e2e/live/end-to-end-test.sh
@@ -882,6 +882,39 @@ assertPodNotExists "pod-c" "test-namespace"
 assertPodNotExists "pod-d" "test-namespace"
 printResult
 
+# Test 21: RBAC error applying a resource
+echo "Testing RBAC error during apply"
+echo "kpt live apply e2e/live/testdata/rbac-error-step-1"
+echo "kpt live apply e2e/live/testdata/rbac-error-step-2"
+# Setup: create a service account and bind a Role to it so it has administrative
+# privileges on the "test" namespace, but no permissions on the default
+# namespace.
+kubectl apply -f e2e/live/testdata/rbac-error-step-1 > $OUTPUT_DIR/status
+assertContains "namespace/rbac-error created"
+assertContains "rolebinding.rbac.authorization.k8s.io/admin created"
+assertContains "serviceaccount/user created"
+wait 2
+
+# Setup: use the service account just created. It does not have permissions
+# on the default namespace, so it will give a permissions error on apply
+# for anything attempted to apply to the default namespace.
+kubectl config set-credentials user --token="$(kubectl get secrets -ojsonpath='{.data.token}' \
+  "$(kubectl get sa user -ojsonpath='{.secrets[0].name}')" \
+  | base64 -d)" > $OUTPUT_DIR/status
+kubectl config set-context kind-kind:user --cluster=kind-kind --user=user > $OUTPUT_DIR/status
+kubectl config use-context kind-kind:user > $OUTPUT_DIR/status
+wait 2
+
+# Attempt to apply two ConfigMaps: one in the default namespace (fails), and one
+# in the test namespace (succeeds).
+${BIN_DIR}/kpt live apply e2e/live/testdata/rbac-error-step-2 > $OUTPUT_DIR/status
+assertCMInventory "rbac-error" "1"
+assertContains "configmap/error-config-map failed"
+assertContains "configmap/valid-config-map created"
+assertContains "2 resource(s) applied. 1 created, 0 unchanged, 0 configured, 1 failed"
+assertContains "0 resource(s) pruned, 0 skipped, 0 failed"
+printResult
+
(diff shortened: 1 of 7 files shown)


e2e/live/end-to-end-test.sh
```diff
@@ -878,6 +878,38 @@ ${BIN_DIR}/kpt live apply e2e/live/testdata/continue-on-error > $OUTPUT_DIR/stat
...
+echo "kpt live apply e2e/live/testdata/rbac-error-step-1"
+echo "kpt live apply e2e/live/testdata/rbac-error-step-2"
+# Setup: create a service account and bind a Role to it so it has administrative
+# privileges on the "test" namespace, but no permissions on the default
+# namespace.
+kubectl apply -f e2e/live/testdata/rbac-error-step-1 > $OUTPUT_DIR/status
+assertContains "namespace/test created"
+assertContains "rolebinding.rbac.authorization.k8s.io/admin created"
+assertContains "serviceaccount/user created"
+wait 2
```
runewake2, line 893: I'm not sure I understand what this `wait` is doing. Can you add a comment here to make it easier to understand?

e2e/live/end-to-end-test.sh
```diff
@@ -878,6 +878,38 @@ ${BIN_DIR}/kpt live apply e2e/live/testdata/continue-on-error > $OUTPUT_DIR/stat
...
+
+# Setup: use the service account just created. It does not have permissions
+# on the default namespace, so it will give a permissions error on apply
+# for anything attempted to apply to the default namespace.
+kubectl config set-credentials user --token="$(kubectl get secrets -ojsonpath='{.data.token}' \
+  "$(kubectl get sa user -ojsonpath='{.secrets[0].name}')" \
+  | base64 -d)" > $OUTPUT_DIR/status
+kubectl config set-context kind-kind:user --cluster=kind-kind --user=user > $OUTPUT_DIR/status
```
runewake2, line 901: I think you can drop `> $OUTPUT_DIR/status` from these lines. Each of these commands will overwrite the contents of that output anyways so if we're not reading the contents the output shouldn't need to be saved.

e2e/live/end-to-end-test.sh
```diff
@@ -878,6 +878,38 @@ ${BIN_DIR}/kpt live apply e2e/live/testdata/continue-on-error > $OUTPUT_DIR/stat
...
+kubectl config use-context kind-kind:user > $OUTPUT_DIR/status
+wait 2
+
+# Attempt to apply two ConfigMaps: one in the default namespace (fails), and one
+# in the test namespace (succeeds).
+${BIN_DIR}/kpt live apply e2e/live/testdata/rbac-error-step-2 > $OUTPUT_DIR/status
+assertCMInventory "test" "1"
```
runewake2, line 908: Nit: It may make sense to use a more descriptive namespace here to avoid this test interfering with others like what was discussed in #1426.

e2e/live/end-to-end-test.sh
seans3: The main reason for the `> $OUTPUT_DIR/status`: it keeps the output from being polluted. We'd rather just have `.` or `E` than the entire command output, which is what would happen without the redirection. This could probably be changed with a `verbose` flag, but I haven't created it yet.

e2e/live/end-to-end-test.sh
seans3: Good point.

e2e/live/end-to-end-test.sh
seans3: We may not need the `wait`, but it helps insure there is sufficient time for the resources to actually be spun up on the cluster (e.g `namespace/test`) after they've been created in the API Server. The best way to do this is to do a `kubectl wait for=<CONDITION>`, but this is a shortcut. Checkout the bash code for `assertPodExists` and `assertPodNotExists` which uses `kubectl wait ...`.

seans3:
Output for this test case:
```
Testing RBAC error during apply
kpt live apply e2e/live/testdata/rbac-error-step-1
kpt live apply e2e/live/testdata/rbac-error-step-2
.........SUCCESS
```

e2e/live/end-to-end-test.sh
runewake2: I was going to suggest `kubectl wait` depending on the use case, but I think you may be intending to [`sleep 2`](https://linux.die.net/man/3/sleep) instead of [`wait 2`](https://linux.die.net/man/3/wait) here. `wait` is going to delay until process id 2 exits. I'm not sure what `pid 2` refers to in this case or if that's what was intended? Maybe I'm misunderstanding?

seans3: The `wait` is a bash function defined earlier in the script which calls `sleep`. The number is the number of seconds to sleep.

runewake2 (approved):

PR #1426, opened 2021-02-08
Updates continue-on-error e2e test case

seans3:
* Fixes flaky "continue-on-error" end-to-end test case. Flakiness caused by re-using namespace from previous test which is deleted before being re-created, causing a race condition. Moves test to position before deletion.
* Updates number of inventory items from 1 to 2, since the namespace is also stored in the inventory if it applied with other items.
* Adds a couple more assert statements.
* Adds missing licenses to resource yamls.

Current output:
```
Testing continue-on-error
kpt live apply e2e/live/testdata/continue-on-error
......SUCCESS
```

### DIFF
--- a/e2e/live/end-to-end-test.sh
+++ b/e2e/live/end-to-end-test.sh
@@ -853,7 +853,18 @@ assertPodExists "pod-c" "test-namespace"
 assertPodExists "pod-d" "test-namespace"
 printResult
 
-# Test 19: Basic kpt live destroy
+# Test 19: kpt live apply continue-on-error
+echo "Testing continue-on-error"
+echo "kpt live apply e2e/live/testdata/continue-on-error"
+${BIN_DIR}/kpt live apply e2e/live/testdata/continue-on-error > $OUTPUT_DIR/status
+assertCMInventory "test-namespace" "2"
+assertContains "pod/pod-a created"
+assertContains "pod/pod-B failed"
+assertPodExists "pod-a" "test-namespace"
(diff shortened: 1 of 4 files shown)


runewake2:
If reusing the namespace creates a race condition maybe it'd make sense to move this test into a unique namespace instead? Otherwise this may reoccur after more tests are added or things are refactored.

runewake2 (approved):

mortent:
I am concerned about the maintainability of this test suite. It is currently almost 1000 lines of bash script and several folders with test data. 
Should we look at turning this into Ginkgo tests, similar to how we do this in [cli-utils](https://github.com/kubernetes-sigs/cli-utils/tree/master/test/e2e) and in OSS Kubernetes? We can just use Golang `os/exec` to run the kpt cli commands.

runewake2:
I had looked into adopting a go based `exec` pattern when I implimented this originally. The hope was reduced complexity and ability to get test results in testing tools/CI. 

It seemed like adopting that would significantly increase the time it took to deliver these tests in the short term and I wasn't sure it made sense to tackle right now. I agree we want to move away from this pattern but without a wrapper around `exec` tests were a lot less easy to read then they are here.

Happy to discuss more.

mortent:
I don't think we should hold up merging this PR, but it seems like it is time to consider how we can do e2e tests at scale.

Liujingfang1:
/lgtm

I think we can first enable these e2e tests in some Github jobs. When we enable it, these tests should be run against different K8s versions.
Then we consider making them more efficient and maintainable.

PR #1448, opened 2021-02-10
check if docker command is available before running pipeline run

### LINKED ISSUE #1442: `kpt pipeline run` should check docker command is available on the system
kpt pipeline uses docker runtime for executing kpt functions. So we should ensure `kpt` is able to run `docker` command (may be run `docker ps`) before invoking `kpt pipeline run` command. If `docker` isn't available, we should present the user with a message saying `docker` is required to run this command and provide with links to docker installation page.

natasha41575:
Close #1442

### DIFF
--- a/e2e/e2e_docker_test.go
+++ b/e2e/e2e_docker_test.go
@@ -1,3 +1,5 @@
+// +build docker
+
 // Copyright 2021 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
--- a/e2e/e2e_docker_test.go
+++ b/e2e/e2e_docker_test.go
@@ -1,5 +1,3 @@
-// +build docker
-
 // Copyright 2021 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
--- a/internal/pipeline/cmd.go
+++ b/internal/pipeline/cmd.go
@@ -67,11 +67,15 @@ func (r *Runner) preRunE(c *cobra.Command, args []string) error {
 }
 
 func (r *Runner) runE(c *cobra.Command, args []string) error {
+	err := cmdutil.DockerCmdAvailable()
+	if err != nil {
+		return err
+	}
 	klog.Infof("running pipeline command")
 	executor := Executor{
 		PkgPath: r.pkgPath,
 	}
-	err := executor.Execute()
+	err = executor.Execute()
 	if err != nil {
 		fmt.Fprintf(c.OutOrStderr(), "failed to run pipeline %v \n", err)
 		return err
--- a/internal/util/cmdutil/cmdutil.go
+++ b/internal/util/cmdutil/cmdutil.go
@@ -93,13 +93,9 @@ func DockerCmdAvailable() error {
 	cmd.Stderr = &stderr
 	err := cmd.Run()
 	if err != nil {
-		if strings.Contains(stderr.String(), "Cannot connect to the Docker daemon") {
-			return fmt.Errorf(stderr.String())
-		}
-		return fmt.Errorf("docker is required to run this command\n" +
-			"To install docker, please follow the instructions on the docker installation page.\n" +
-			"Docker installation instructions can be found here:\n" +
-			"https://docs.docker.com/get-docker/")
+		return fmt.Errorf(stderr.String() +
+			`Docker is required to run this command. 
+To install docker, follow the instructions at https://docs.docker.com/get-docker/`)
 	}
 	return nil
 }
--- a/internal/util/cmdutil/cmdutil.go
+++ b/internal/util/cmdutil/cmdutil.go
@@ -15,6 +15,7 @@
 package cmdutil
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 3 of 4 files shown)


e2e/e2e_docker_test.go
```diff
@@ -1,5 +1,3 @@
-// +build docker
```
Shell32-Natsu, line 1: Better to keep this tag

internal/util/cmdutil/cmdutil.go
```diff
@@ -82,3 +83,16 @@ func ResolveAbsAndRelPaths(path string) (string, string, error) {
 
 	return relPath, absPath, nil
 }
+
+// DockerCmdAvailable runs `docker ps` to check that the docker command is
+// available, and returns an error with installation instructions if it is not
+func DockerCmdAvailable() error {
+	_, err := exec.Command("docker", "ps").Output()
+	if err != nil {
+		return fmt.Errorf("docker is required to run this command\n" +
```
Shell32-Natsu, line 92: `docker ps` runs with an error doesn't necessarily show that there is no docker. It's also possible that the docker daemon is not running. It will be good to wrap the stderr from the command.

e2e/e2e_docker_test.go
natasha41575: I will keep it in, but out of curiosity, what is it for?

Shell32-Natsu: This tag means the test in this file will not be run when we use `go test ./...` but will be included in `go test --tags=docker ./...`. It's added to allow us to run the tests that require docker, like this, only in the environment that has docker.

We have a target `test-docker` in the top level Makefile for it.

internal/util/cmdutil/cmdutil.go
natasha41575: done, wrapped stderr for such a case

e2e/e2e_docker_test.go
natasha41575: got it, added it back. Thanks!

internal/util/cmdutil/cmdutil.go
Shell32-Natsu: Daemon us not running is not the only possible reason. Some users may have docker run in root privilege. I am thinking add the error message before the installation directions no matter what the error is.

internal/util/cmdutil/cmdutil.go
```diff
@@ -82,3 +84,22 @@ func ResolveAbsAndRelPaths(path string) (string, string, error) {
...
+	cmd.Stderr = &stderr
+	err := cmd.Run()
+	if err != nil {
+		if strings.Contains(stderr.String(), "Cannot connect to the Docker daemon") {
+			return fmt.Errorf(stderr.String())
+		}
+		return fmt.Errorf("docker is required to run this command\n" +
+			"To install docker, please follow the instructions on the docker installation page.\n" +
+			"Docker installation instructions can be found here:\n" +
+			"https://docs.docker.com/get-docker/")
```
droot, line 102: tip: you can use Go's string literal with backtick to make it easier craft the error message.

Also, we can shorten the error message saying:
`docker is required to run this command.
To install docker, follow the instructions at https://docs.docker.com/get-docker/
`

internal/pipeline/cmd.go
```diff
@@ -53,11 +53,15 @@ func (r *Runner) preRunE(c *cobra.Command, args []string) error {
 }
 
 func (r *Runner) runE(c *cobra.Command, args []string) error {
+	err := cmdutil.DockerCmdAvailable()
+	if err != nil {
+		return err
+	}
```
droot, line 59: I am wondering what exactly does a user see when they encounter this error ? Can you share the exact output.

internal/util/cmdutil/cmdutil.go
natasha41575: done

internal/pipeline/cmd.go
natasha41575: ran `kpt pipeline run` from my terminal while the docker daemon was not running and received the following:
```
error: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
docker is required to run this command. 
To install docker, follow the instructions at https://docs.docker.com/get-docker/
```

internal/util/cmdutil/cmdutil.go
natasha41575: updated

natasha41575:
@droot @Shell32-Natsu please take a look

Shell32-Natsu (approved):
lgtm

internal/util/cmdutil/cmdutil.go
```diff
@@ -82,3 +84,18 @@ func ResolveAbsAndRelPaths(path string) (string, string, error) {
 
 	return relPath, absPath, nil
 }
+
+// DockerCmdAvailable runs `docker ps` to check that the docker command is
+// available, and returns an error with installation instructions if it is not
+func DockerCmdAvailable() error {
+	cmd := exec.Command("docker", "ps")
```
frankfarzan, line 91: Is there a particular reason you use `docker ps` which can potentially be expensive, instead of `docker version`?

natasha41575: there is no reason, the choice was arbitrary. I've changed it to `docker version`, thank you for the suggestion

Issue #1413, opened 2021-02-04
Implement pipeline definition changes


droot:
We recently revised the `pipeline` definition with two major changes:

- dropped `sources` field (no major code change except this will be now a private field and not read from the config file)
- function grouping changes. Now `pipeline` supports only two groups: `mutators` and `validators`.

Another change is that `pipeline` itself is now part of `Kptfile`, but I think we can do that separately because I expect to be a bigger change.

Shell32-Natsu:
Will change the pipeline and pipeline functions from OOP style to function style after #1417 being merged.

Shell32-Natsu:
https://github.com/GoogleContainerTools/kpt/issues/1431 should be done at the same time.

PR #1449, opened 2021-02-10
Refactor pipeline and pipeline function

Shell32-Natsu:
#1413

Waiting for #1417.


droot:
Since #1417 is merged, I will wait for you to rebase this PR on top of it and then review.

Shell32-Natsu:
@droot @frankfarzan This is ready for review

internal/pipeline/executor.go
```diff
@@ -309,8 +308,8 @@ func filterMetaData(resources []*yaml.RNode) []*yaml.RNode {
 
 // fnFilters returns chain of functions that are applicable
 // to a given pipeline.
-func fnFilters(p *Pipeline) ([]kio.Filter, error) {
-	filters, err := p.fnChain()
+func fnFilters(p *v1alpha2.Pipeline, pkgPath string) ([]kio.Filter, error) {
```
droot, line 311: As a convention I am thinking, if we want to use short name for `Pipeline`, we should use `pl` instead of `p`. We can reserve `p` for pkg.

internal/pipeline/executor.go
```diff
@@ -120,16 +121,14 @@ func (p *pkg) Pipeline() *Pipeline {
...
-
 	var pkgPaths []string
-	for _, s := range pipeline.Sources {
-		paths, err := resolveSource(s, p.Path())
-		if err != nil {
-			return nil, err
-		}
-		pkgPaths = append(pkgPaths, paths...)
+	// TODO: since we are considering only support './*', we can simplify
+	// this part
```
droot, line 126: Let's not try to simplify further because we will learn more about `sources` as we go along, so this is good for now.

internal/pipeline/pipeline.go
```diff
@@ -18,99 +18,27 @@ package pipeline
...
-		}
-		if err := validatePath(s); err != nil {
-			return fmt.Errorf("'sources[%d]': path %q is invalid: %w", i, s, err)
-		}
-		p.Sources[i] = path.Clean(s)
-	}
-	fns := []Function{}
-	fns = append(fns, p.Generators...)
-	fns = append(fns, p.Transformers...)
+func ValidatePipeline(p *v1alpha2.Pipeline) error {
```
droot, line 35: nit: comment above needs to be changed. We might be using `generators`, `transformers` in other places as well, so do a scan.

internal/pipeline/pipeline.go
```diff
@@ -18,99 +18,27 @@ package pipeline
...
+	fns := []v1alpha2.Function{}
+	fns = append(fns, p.Mutators...)
 	fns = append(fns, p.Validators...)
-	for _, f := range fns {
-		err := f.Validate()
+	for i := range fns {
+		f := fns[i]
+		err := validateFunction(&f)
```
droot, line 41: Not a big deal: but you can take struct as input here because validateFunction doesn't mutate the function. While typing this, I realized this can be made receiver and moved in types.go because this is self-contained.

internal/pipeline/pipeline.go
```diff
@@ -137,59 +65,16 @@ func (p *Pipeline) fnChain() ([]kio.Filter, error) {
...
-}
-
-// FromFile returns a Pipeline read from file
-func FromFile(path string) (*Pipeline, error) {
-	f, err := os.Open(path)
-	if err != nil {
-		return nil, fmt.Errorf("failed to open path %s: %w", path, err)
-	}
-	return FromReader(f)
+func fromString(s string) (*v1alpha2.Pipeline, error) {
```
droot, line 78: I don't see any use of these two functions, since pipeline will be read as part of reading the Kptfile, so should we remove them ?

internal/pipeline/pipeline_function.go
```diff
@@ -25,36 +25,13 @@ import (
...
-	ConfigPath string `yaml:"configPath,omitempty"`
-
-	// `ConfigMap` is a convenient way to specify a function config of kind ConfigMap.
-	ConfigMap map[string]string `yaml:"configMap,omitempty"`
-}
-
-// Validate will validate all fields in function.
-func (f *Function) Validate() error {
+// validateFunction will validate all fields in function.
+func validateFunction(f *v1alpha2.Function) error {
```
droot, line 34: This seems to be self-contained and can be moved to types.go as a method on the Function type?

pkg/api/kptfile/v1alpha2/types.go
```diff
@@ -207,7 +215,7 @@ type Function struct {
 
 	// `Config` specifies an inline KRM resource used as the function config.
 	// Config, ConfigPath, and ConfigMap fields are mutually exclusive.
-	Config *yaml.Node `yaml:"config,omitempty"`
+	Config yaml.Node `yaml:"config,omitempty"`
```
droot, line 218: Adding a note why this can not be a pointer will be helpful given that it's OneOf(config, configPath, configMap) scenario. I remember discussing this on some PR about how yaml.Zero() can be used to test empty YAML input.

internal/pipeline/pipeline_test.go
```diff
@@ -66,138 +31,107 @@ func checkOutput(t *testing.T, tc testcase, actual *Pipeline, err error) {
...
-		tc := tc
-		name := name
-		t.Run(name, func(t *testing.T) {
-			r := bytes.NewBufferString(tc.Input)
-			actual, err := FromReader(r)
-			checkOutput(t, tc, actual, err)
-		})
-	}
-}
+func TestPipelineValidate(t *testing.T) {
```
droot, line 40: I feel like these tests can be moved to where Pipeline type exists. Also, reading the pipeline as part of Kptfile will be better because that will be our regular usage.

droot (commented):
I have a few comments, happy to discuss on call.

internal/pipeline/pipeline.go
Shell32-Natsu: These 2 functions are used for testing. I will move them to test files.

internal/pipeline/pipeline.go
Shell32-Natsu: I don't think this is real self-contained because it requires some helpers like `validateFunctionName` and `isNodeZero`.

Shell32-Natsu: moved the validation to v1alpha2 package.

irvifa:
This one https://github.com/GoogleContainerTools/kpt/pull/1417 is merge already

internal/pipeline/pipeline_function.go
```diff
@@ -21,70 +21,17 @@ import (
...
-
-	return nil
-}
-
-// runner returns a fnRunner from the image and configs of
+// newFnRunner returns a fnRunner from the image and configs of
 // this function.
-func (f *Function) runner() (kio.Filter, error) {
-	config, err := f.config()
+func newFnRunner(f *v1alpha2.Function, pkgPath string) (kio.Filter, error) {
```
irvifa, line 33: Do we need to do a validation for the config name for example?

Shell32-Natsu: Do you mean validating is the config file name a valid file name?

droot (approved):
/lgtm

PR #1453, opened 2021-02-11
added simple package e2e test for pipeline

### LINKED ISSUE #1440: Add e2e test case for `kpt pipeline run` for simple package UX
The test case should test simple package UX where we have an `app` package that contains deployment.yaml and Kptfile with pipeline that uses set-labels, set-namespace functions from the function catalog as mutators and ensure `pipeline run` results in expected output.

natasha41575:
close #1440


internal/pipeline/examples/simplepackage/configs/Kptfile
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: kptfile
+pipeline:
+  mutators:
+  - image: gcr.io/kpt-fn/set-namespace
+    configMap:
+      namespace: staging
+  - image: gcr.io/kpt-fn/set-label
+    configMap:
+      tier: db
```
natasha41575, line 12: @Shell32-Natsu @droot can you verify that this is the correct way to invoke the functions? I was unsure

e2e/e2e_docker_test.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/GoogleContainerTools/kpt/pkg/test/runner"
 )
 
+// These e2e tests currently fail. When pipeline is implemented they should pass.
 func TestPipeline(t *testing.T) {
 	runPipelineTests(t, "../internal/pipeline/examples/")
```
droot, line 27: Can we rename `examples` to `testdata` ? to make it more idiomatic.

internal/pipeline/examples/simplepackage/configs/Kptfile
```diff
@@ -0,0 +1,12 @@
+apiVersion: kpt.dev/v1alpha1
```
droot, line 1: we need to use newer version of Kptfile here v1alpha2

internal/pipeline/examples/simplepackage/configs/Kptfile
```diff
@@ -0,0 +1,12 @@
...
+kind: Kptfile
+metadata:
+  name: kptfile
```
droot, line 4: This should be name of the package. `simplepackage` in this case.

While typing this I realized why not call this package something useful such as `app` or something.

internal/pipeline/examples/simplepackage/configs/Kptfile
droot: syntax look ok to me. I would use `tier: backend` may be so that we can use `tier: db` for the db package.

natasha41575: done

e2e/e2e_docker_test.go
natasha41575: done

internal/pipeline/examples/simplepackage/configs/Kptfile
natasha41575: done

internal/pipeline/examples/simplepackage/configs/Kptfile
natasha41575: done

natasha41575:
@droot please take a look

e2e/e2e_docker_test.go
droot: I see the files moved to `configdata` below, but this line still refers to `examples` dir.

natasha41575: fixed, thanks for the catch

internal/pipeline/testdata/simplepackage/.expected/config.yaml
```diff
@@ -0,0 +1,2 @@
+exitCode: 0
+network: false
```
Shell32-Natsu, line 2: Do we need to check idempotence for pipeline? @frankfarzan

internal/pipeline/testdata/simplepackage/.expected/diff.patch
```diff
@@ -0,0 +1,22 @@
...
++  labels:
++    tier: backend
+ spec:
+   replicas: 3
+ ---
+@@ -9,4 +12,8 @@ apiVersion: custom.io/v1
+ kind: Custom
+ spec:
+   image: nginx:1.2.3
++metadata:
```
Shell32-Natsu, line 19: In the test framework for pipeline, do you run the formatter after running pipeline? This order of fields looks weird.

internal/pipeline/testdata/simplepackage/.expected/config.yaml
natasha41575: Spoke briefly with @droot about this and he said it is not a requirement for pipeline run because it is a property of the kpt functions

internal/pipeline/testdata/simplepackage/.expected/diff.patch
natasha41575: updated it to run the formatter and changed the order of fields in the diff

Shell32-Natsu (approved):
lgtm

pkg/test/runner/runner.go
```diff
@@ -140,15 +140,23 @@ func (r *Runner) runPipeline() error {
...
 	command := run.GetMain()
 	kptArgs := []string{r.cmd, "run", tmpPkgPath}
 	for i := 0; i < r.testCase.Config.RunTimes; i++ {
 		command.SetArgs(kptArgs)
-		err = command.Execute()
+		fnErr = command.Execute()
+	}
+
+	// run formatter
+	command.SetArgs([]string{"cfg", "fmt", tmpPkgPath})
```
frankfarzan, line 152: Is there a reason why `kpt pipeline run` and `kpt fn run` shouldn't automatically format the output. 

@droot 
@mortent

natasha41575: thanks for bringing it up, I will talk to @droot more about this when he comes back tomorrow

PR #1457, opened 2021-02-11
[WIP] Include gcloud in kpt image

mortent:
Using `kpt live` from the kpt docker image requires gcloud for auth. This changes the source image we use for the kpt to use the cloud-sdk:alpine image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,9 +25,9 @@ RUN go mod download
 COPY . .
 RUN go build -v -o /usr/local/bin/kpt -ldflags="-s -w -X github.com/GoogleContainerTools/kpt/run.version=$KPT_VERSION" ./
 
-FROM alpine:3.11
+FROM gcr.io/google.com/cloudsdktool/cloud-sdk:alpine
 RUN apk update && apk upgrade && \
-    apk add --no-cache git less man diffutils bash openssh docker-cli && \
+    apk add --no-cache git less diffutils bash openssh docker-cli && \
     rm -rf /var/lib/apt/lists/* && \
     rm /var/cache/apk/*
 COPY --from=0 /usr/local/bin/kpt /usr/local/bin/kpt


mikebz (approved):
👍

mortent:
There are two challenges with this approach:
 * This increases the size of the image from ~50Mb to almost 200Mb.
 * It depends on the `gcr.io/google.com/cloudsdktool/cloud-sdk:alpine` base image, which always pulls down the latest alpine-based image. This leads to builds that aren't fully reproducible. We can get around that by depending on a specific version (like `gcr.io/google.com/cloudsdktool/cloud-sdk:327.0.0-alpine`), but this require that we always update the version as part of each release.

I think the best option is to include gcloud in a separate image.

Dockerfile
```diff
@@ -25,9 +25,9 @@ RUN go mod download
 COPY . .
 RUN go build -v -o /usr/local/bin/kpt -ldflags="-s -w -X github.com/GoogleContainerTools/kpt/run.version=$KPT_VERSION" ./
 
-FROM alpine:3.11
+FROM gcr.io/google.com/cloudsdktool/cloud-sdk:alpine
```
frankfarzan, line 28: We have started standardizing on specific versions of alpine for functions, we should follow the same approach here. @mengqiy 

@mortent Can you also elaborate (probably add to PR description) why you're making this change given the binary bloat to 200MB? Have you looked at re

mengqiy: In kpt functions, we are using alpine:3.12 for go-based and node:15.3-alpine3.12 for ts-based.
For Googlers, to see the complete description: https://docs.google.com/document/d/1q22he3sQF-w2Jn78YprG6orxx0Avy3N105QIqe51Vso/edit#heading=h.v1kcq2swzui1

https://github.com/GoogleCloudPlatform/cloud-sdk-docker/blob/master/alpine/Dockerfile is how gcr.io/google.com/cloudsdktool/cloud-sdk:alpine is built. It seems it installs a lot of other stuffs besides gcloud. If we can avoid install unnecessary stuffs, we make the size smaller.

mortent: So using `kpt live` with the kpt image requires gcloud for auth. I expect that the stuff installed into the cloud-sdk image is there for a reason, so I think it is safer for us to use the validated image for this instead of rolling our own, even if the cloud-sdk image uses a different version of alpine.

I suggest we just create a separate image for kpt that installs gcloud. That will allow users who don't need the `kpt live` functionality to use the small image, but we provide an image for users how want to use `kpt live`.

mikebz:
@mortent got a ping from Martin about the ETA on this.  How should we manage his expectations?

mortent:
As I suggested in another comment, I think we should create a separate image that includes gcloud. So we just need to create an additional Dockerfile and update build to create both images and push them to gcr. I just haven't had time to do it yet, so maybe someone else can pick this up.

It looks like the cloud-sdk image already includes git (https://github.com/GoogleCloudPlatform/cloud-sdk-docker/blob/master/alpine/Dockerfile#L16).

mikebz:
Thanks for clarification.  I was mostly concerned with the goal rather than the PR changes.  I am going to create a ticket for that and we can close this PR or someone take take it over to produce a separate image.

mortent:
Closing this as we want to create a separate image that includes gcloud instead of adding it to the current image.

PR #1467, opened 2021-02-16
Update diff-tool-opts docs for correctness.

runewake2:
Fix for incorrect `kpt pkg diff` documentation that suggests using `--diff-opts`. This replaces that with the `--diff-tool-opts` flag instead.

Fixes: #1468


runewake2:
The SiteCheck is failing but I do not think that's an issue related to this change (errors are in `live` docs).

frankfarzan (approved):

PR #1473, opened 2021-02-18
Fix doc link 404's

runewake2:
Fix for live alpha links 404ing causing `sitecheck` CI job to fail.

Should resolve CI errors seen in #1467


seans3 (approved):
/lgtm

Thanks for the fix.

frankfarzan (approved):

Issue #1441, opened 2021-02-09
Add e2e test for `kpt pipeline run` for nested package UX


droot:
We need to add e2e test case for nested package UX for `kpt pipeline run` where:

- top level package `app` that contains deployment.yaml and kptfile
- subpackage `db` under `app` package that contains some KRM resource say statefulset.yaml and kptfile
- `app`'s pipeline uses set-labels, set-namespace functions from function catalog
- `db`'s pipeline uses set-labels, set-namespace functions from function catalog 

test should ensure `kpt pipeline run` produces expected results i.e. recursive hydration where `db` package is hydrated first and `app` package is hydrated later and KRM resources in both packages should be transformed accordingly.

natasha41575:
I believe this is covered by https://github.com/GoogleContainerTools/kpt/pull/1483

PR #1483, opened 2021-02-21
implement pipeline run

droot:
This PR implements new feature `config rendering`.  Hydration can be invoked using command: `kpt fn render`.


natasha41575:
noticed this has test case for nested package, does it cover https://github.com/GoogleContainerTools/kpt/issues/1441? Checking to see if I still need to work on that or if I should assign that issue to you

droot:
@natasha41575 yes, I have added a basic test for nested UX so it covers #1441 to get some coverage on the code in this PR. Also note that the expected output is wrong because I am still debugging `set-label` function. 
We have to add more test cases. I do want to think if our current approach for adding tests is scalable (mainly from verbosity standpoint).

internal/pipeline/executor.go
```diff
@@ -22,174 +22,174 @@ import (
...
+	// written back to the originating source dir.
+	SinkInplace SinkMode = "in-place"
+
+	// SinkOutplace means that result of hydration will be
+	// written to specified dir.
+	SinkOutplace = "out-place"
+
+	// SinkStdout means that the result of hydration will be
+	// written to stdout.
+	SinkStdout = "stdout"
```
natasha41575, line 49: any reason why SinkInplace is type SinkMode but the other two aren't?

internal/pipeline/testdata/app-with-db/.expected/config.yaml
```diff
@@ -0,0 +1,2 @@
+exitCode: 0
+network: false
```
Shell32-Natsu, line 2: `network` is default to false

internal/pipeline/executor.go
droot: Other two also need to have the Type. It was an oversight.
I took out the `sinkMode` functionality for now. `SinkMode` support will be added in a separate PR with `out-of-place` support.

internal/pipeline/testdata/app-with-db/.expected/config.yaml
droot: Got it. Took it out.

internal/pipeline/executor.go
```diff
@@ -36,159 +38,129 @@ type Executor struct {
...
 
-// TODO(droot): this will be replaced when pkg abstraction PR is merged
-func newPkg(path string) (*pkg, error) {
-	pkgPath, err := filepath.Abs(path)
+// newPkgNode returns a pkgNode instance given a path.
+func newPkgNode(path string) (*pkgNode, error) {
+	p, err := pkg.New(path)
 	if err != nil {
-		return nil, fmt.Errorf("failed to find abs path %s %w", path, err)
+		return nil, fmt.Errorf("failed to read pkg %w", err)
```
frankfarzan, line 102: general comment: choose the failure error messages shown to the user carefully. It's an important part of the UX. In this case, `pkg` is an internal variable name.

internal/pipeline/executor.go
```diff
@@ -36,159 +38,129 @@ type Executor struct {
...
-	// Read the pipeline for the current package else assume default
-	// TODO(droot): integrate with Donny's code once that is merged
-	// and read an actual pipeline
-	p.pipeline = &kptfilev1alpha2.Pipeline{}
-
-	return nil
-}
+// subPackages returns paths to sub packages including current package's path.
+// It includes a package only if it has a Kptfile.
+func (pn *pkgNode) subPackages() ([]string, error) {
```
frankfarzan, line 118: This functionality should be extracted out/abstracted and used across different features (I'm guessing this is re-implemented multiple times already):

pkg.New(path).SubPackages() -> []Pkg

I don't think there's any reason for hydration logic to do any low-level file operations.

internal/pipeline/executor.go
```diff
@@ -36,159 +38,129 @@ type Executor struct {
...
+	sort.Strings(pkgPaths)
+	return pkgPaths, nil
 }
 
-// localResources reads
-func (p *pkg) localResources(includeMetadata bool) (resources []*yaml.RNode, err error) {
-	// TODO(droot): figure out how to avoid reading sub packages here ?
-	pkgReader := &kio.LocalPackageReader{PackagePath: p.Path(), MatchFilesGlob: kio.MatchAll}
+// localResources returns resources that belong to this package excluding the subpackage resources.
+func (pn *pkgNode) localResources(includeMetaResources bool) (resources []*yaml.RNode, err error) {
```
frankfarzan, line 149: Same here:

pkg.New(path).LocalResources()

internal/pipeline/executor.go
```diff
@@ -207,78 +179,84 @@ func (s hydrationState) String() string {
...
+			return resources, err
 		case Wet:
-			// already hydrated, we sort of assume pkgs in hctx will be populated but we
-			// can be defensive here
-			resources = hctx.pkgs[p.Path()].resources
-		default:
-			err = fmt.Errorf("invalid pkg state %v detected", state)
+			resources = currPkg.resources
+			return resources, err
+		case Dry:
```
frankfarzan, line 194: why remove the default? It's more defensive in case you the enum changes in the future

internal/pipeline/executor.go
```diff
@@ -207,78 +179,84 @@ func (s hydrationState) String() string {
...
+			// Is this a dead path ?
```
frankfarzan, line 195: ?

when would the state ever be Dry?

internal/pipeline/executor.go
```diff
@@ -288,25 +266,42 @@ func hydrate(p *pkg, hctx *hydrationContext) (resources []*yaml.RNode, err error
...
+		newPath := path.Join(relPath, meta.Annotations[kioutil.PathAnnotation])
+		err = transRes.PipeE(yaml.SetAnnotation(kioutil.PathAnnotation, newPath))
+		if err != nil {
+			return resources, err
+		}
+	}
+	return resources, nil
+}
+
+// filterMetaResources filters kpt metadata files such as pipeline,
```
frankfarzan, line 302: pipeline is no longer a separate resource?
what's permissions?

internal/pipeline/executor.go
```diff
@@ -288,25 +266,42 @@ func hydrate(p *pkg, hctx *hydrationContext) (resources []*yaml.RNode, err error
...
 // Kptfile, permissions etc.
-func filterMetaData(resources []*yaml.RNode) []*yaml.RNode {
+func filterMetaResources(resources []*yaml.RNode) []*yaml.RNode {
```
frankfarzan, line 304: This should filter out function config specified in`pipeline.config` right? Please have a test for all the cases here.

internal/pipeline/executor.go
```diff
@@ -207,78 +179,84 @@ func (s hydrationState) String() string {
...
-			currPkgResources, err = p.localResources(false)
+			currPkgResources, err = currPkg.localResources(false)
 			if err != nil {
 				return resources, err
 			}
+			// Resources are read from local filesystem at a package level, so the
+			// path annotation in each resource points to path relative to that package.
+			// But the resources are written to the file system at the root package level, so
+			// the path annotation in each resources needs to be adjusted to be relative to the rootPkg.
+			relPath, err := filepath.Rel(hctx.root.Path(), s)
```
frankfarzan, line 225: This is the kind of low-level operation that needs to be done once and reused instead of re-implemented multiple times.

Also how do you handle new resources generated in the package after executing the pipeline? Please compare to to how 'kpt fn run' works today.

internal/pipeline/executor.go
```diff
@@ -207,78 +179,84 @@ func (s hydrationState) String() string {
...
 
-			if depPkg, err = newPkg(s); err != nil {
+			if subPkg, err = newPkgNode(s); err != nil {
 				return resources, err
 			}
 
-			transitiveResources, err = hydrate(depPkg, hctx)
+			transitiveResources, err = hydrate(subPkg, hctx)
 			if err != nil {
 				err = fmt.Errorf("failed to hydrated sub pkg %s %w", s, err)
```
frankfarzan, line 241: ditto on error messages being part of the UX: does "hydrated" make sense to the end-user?

internal/pipeline/executor.go
```diff
@@ -207,78 +179,84 @@ func (s hydrationState) String() string {
...
 		}
 	}
 
 	output := &kio.PackageBuffer{}
-	// TODO(droot): parameterize the sink-mode (hctx.sinkMode) to determine whether to
-	// write it in-place or not. Currently in-place.
-	pkgWriter := &kio.LocalPackageWriter{PackagePath: p.Path()}
 
-	filters, err := fnFilters(p.Pipeline(), p.Path())
+	pl, err := currPkg.pkg.Pipeline()
```
frankfarzan, line 255: assuming the following lines work correctly if pl is nil or empty

internal/pipeline/testdata/app-with-db/configs/Kptfile
```diff
@@ -0,0 +1,19 @@
...
+kind: Kptfile
+metadata:
+  name: app-with-db
+pipeline:
+  mutators:
+  - image: gcr.io/kpt-fn/set-namespace:unstable
+    configMap:
+      namespace: staging
+  - image: gcr.io/kpt-fn/set-label:unstable
+    config:
```
frankfarzan, line 11: this should be a configMap for brevity

internal/pipeline/testdata/app-with-db/configs/db/Kptfile
```diff
@@ -0,0 +1,19 @@
+apiVersion: kpt.dev/v1alpha2
+kind: Kptfile
+metadata:
+  name: db
+pipeline:
```
frankfarzan, line 5: you want to test:

- a generator function in subpackage `db`
- a generator function in package `app`
- pipeline that uses `configs` to exercise filtering meta resources.
- validation functions: shouldn't be allowed to mutate resources
- failure cases

internal/pkg/pkg.go
```diff
@@ -77,23 +80,57 @@ func New(path string) (*Pkg, error) {
 
 // Kptfile returns the Kptfile meta resource by lazy loading it from the filesytem.
 // A nil value represents an implicit package.
-func (p *Pkg) Kptfile() *kptfilev1alpha2.KptFile {
-	if !p.kptfileLoaded {
-		// TODO
-		// p.kptfile = ...
-		p.kptfileLoaded = true
+func (p *Pkg) Kptfile() (*kptfilev1alpha2.KptFile, error) {
```
frankfarzan, line 83: Is this used in the hydration code?

frankfarzan (changes requested):

droot:
@frankfarzan addressed most of the comments (will mark addressed one resolved). I have addressed the review comments in a separate commit to make it easier to review https://github.com/GoogleContainerTools/kpt/pull/1483/commits/80c01ce072eac55f15f73688b4fb59dc52883a6d

internal/pipeline/executor.go
droot: I agree. I have improved error messages to make it useful for the user for now, but I will scrub them again once we have the finalized the command UX and terminology.

I realize that errors are serving two users (implementor and the end-user), so striking the balance sometimes is getting difficult. I found this post from Rob Pike https://commandcenter.blogspot.com/2017/12/error-handling-in-upspin.html very useful. I want to explore if we can apply those lessons to kpt error handling.

internal/pipeline/executor.go
droot: Good point. I moved this to `pkg` abstraction.

internal/pipeline/executor.go
droot: Good pt. Done.

internal/pipeline/executor.go
droot: Added `default`.

internal/pipeline/executor.go
droot: When package are initialized, they are in `Dry` state, but the way our recursive hydration code is implemented, we never encounter the `Dry` state so if we come here, that represent `internal error`.

internal/pipeline/executor.go
droot: I am using `running pipeline` instead of `hydrated` for now and will scrub once again when we settle on the terminology.

internal/pipeline/testdata/app-with-db/configs/Kptfile
droot: @Shell32-Natsu is addressing this, will use the `configMap` version once the PR is in.

internal/pkg/pkg.go
droot: Actually wasn't being used earlier in hydration code, but added in this commit to ensure that `hydration` is invoked on valid kpt packages. Valid kpt package will have `Kptfile`. I am on the fence if I want to make this `invariant` part of `pkg.New()` constructor.

internal/pipeline/testdata/app-with-db/configs/db/Kptfile
droot: This is not resolved yet. I will be addressing this soon and capture that in a separate commit.

droot: Added testcases for filtering meta resources test case and generator in the subpackage `db` . Rest of the test cases I will be adding in the separate PR.

internal/pipeline/executor.go
droot: fixed the comment.

internal/pipeline/executor.go
droot: I tweaked this to handle generated resources and also added a basic testcase for generator in a nested package. Also changed the order, it is not being done as a post-hydration step at the pkg level. I can potentially move it to a helper method now.

internal/pipeline/testdata/app-with-db/configs/Kptfile
Shell32-Natsu: We have already updated to support ConfigMap. @mengqiy Have we released them?

pkg/test/runner/runner.go
```diff
@@ -114,7 +118,10 @@ func (r *Runner) runPipeline() error {
 	if err != nil {
 		return fmt.Errorf("failed to create temporary dir: %w", err)
 	}
-	defer os.RemoveAll(tmpDir)
+	if !r.testCase.Config.Debug {
+		// if debug is true, keep the test directory around for debugging
+		defer os.RemoveAll(tmpDir)
```
Shell32-Natsu, line 123: It's better to output the path of temporary directory

pkg/test/runner/runner.go
```diff
@@ -23,8 +23,8 @@ const (
 	expectedResultsFile string = "results.yaml"
 	expectedDiffFile    string = "diff.patch"
 	expectedConfigFile  string = "config.yaml"
-	CommandFn           string = "fn"
-	CommandPipeline     string = "pipeline"
+	CommandFnEval       string = "run"
+	CommandFnRender     string = "render"
```
Shell32-Natsu, line 27: Shouldn't this be `CommandPipelineRender`?

internal/pipeline/testdata/app-with-db/configs/Kptfile
mengqiy: Yes, we have released them. Tag `unstable` is always from HEAD.

internal/pkg/pkg.go
frankfarzan: Are you thinking about the commands such as `kpt pkg init` where the Kptfile does not exist?

internal/pkg/pkg.go
```diff
@@ -77,26 +92,177 @@ func New(path string) (*Pkg, error) {
...
+// readKptfile reads the KptFile in the given pkg.
+// TODO(droot): This method exists for current version of Kptfile.
+// Need to reconcile with the team how we want to handle multiple versions
+// of Kptfile in code. One option is to follow Kubernetes approach to
+// have an internal version of Kptfile that all the code uses. In that case,
+// we will have to implement pieces for IO/Conversion with right interfaces.
+func (p *Pkg) readKptfile() (*kptfilev1alpha2.KptFile, error) {
+	kf := &kptfilev1alpha2.KptFile{}
+
+	f, err := os.Open(path.Join(p.UniquePath.String(), kptfilev1alpha2.KptFileName))
```
frankfarzan, line 115: Do you actually need `UniquePath.String()`? I find think of String() to be mostly relevant for printing purposes not for computation. Why not just `string(p.UniquePath)`.

internal/pkg/pkg.go
```diff
@@ -77,26 +92,177 @@ func New(path string) (*Pkg, error) {
...
+		MatchFilesGlob:     kio.MatchAll,
+	}
+	resources, err = pkgReader.Read()
+	if err != nil {
+		err = fmt.Errorf("failed to read resources for pkg %s %w", p, err)
+		return resources, err
+	}
+	if !includeMetaResources {
+		resources = filterMetaResources(resources)
+		resources, err = filterFunctionConfigFiles(resources, functionConfigFilePaths(pl))
```
frankfarzan, line 211: Should this be folded into filterMetaResources()?

internal/pkg/pkg.go
```diff
@@ -77,26 +92,177 @@ func New(path string) (*Pkg, error) {
...
+		if err != nil {
+			return resources, fmt.Errorf("failed to filter function config files: %w", err)
+		}
+	}
+	return resources, err
+}
+
+// filterMetaResources filters kpt metadata files such as Kptfile, function configs.
```
frankfarzan, line 219: I think we have defined meta resources as including function configs, no?

frankfarzan (commented):
Looking good

internal/pipeline/testdata/app-with-db/configs/Kptfile
droot: Thanks guys! Using `configMap` now, just verified with the latest `set-label` and it works.

internal/pkg/pkg.go
droot: No, my comment was more about checking if `Kptfile` is missing in `pkg.New(...)` constructor.

pkg/test/runner/runner.go
droot: Good pt. Added.

pkg/test/runner/runner.go
droot: These are actually now `fn` subcommands. We no longer have `pipeline run` command.

internal/pipeline/executor.go
```diff
@@ -207,78 +135,107 @@ func (s hydrationState) String() string {
...
-			// can be defensive here
-			resources = hctx.pkgs[p.Path()].resources
+			resources = currPkg.resources
+			return resources, err
 		default:
-			err = fmt.Errorf("invalid pkg state %v detected", state)
+			return resources, fmt.Errorf("package %s detected in invalid state", currPkg.pkg)
 		}
-		return resources, err
+	} else {
```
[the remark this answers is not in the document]
droot: Done.

internal/pipeline/executor.go
```diff
@@ -207,78 +135,107 @@ func (s hydrationState) String() string {
...
-			// TODO(droot): recursive hydration is broken for now, needs more work
-			var transitiveResources []*yaml.RNode
-			var depPkg *pkg
+	// include current package's resources in the input resource list
+	input = append(input, currPkgResources...)
 
-			if depPkg, err = newPkg(s); err != nil {
-				return resources, err
-			}
+	resources, err = currPkg.runPipeline(input, hctx)
```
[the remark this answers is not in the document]
droot: Using `input` and `output`, reading much better. Thanks.

internal/pipeline/executor.go
```diff
@@ -36,161 +35,90 @@ type Executor struct {
...
+	pn = &pkgNode{
+		pkg:   p,
+		state: Dry, // package starts in dry state
 	}
-	return resources, err
+	return pn, nil
+}
+
+func (pn *pkgNode) Path() string {
+	return string(pn.pkg.UniquePath)
```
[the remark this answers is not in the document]
droot: I agree. Last time I tried using `UniquePath`, I was running into issues where I had to convert `UniquePath` to string in ton of places because hydration use filesystem package extensively, so I gave up. Now that I have code almost final, I want to try again. I am keeping this comment open.

internal/pipeline/executor.go
```diff
@@ -207,78 +135,107 @@ func (s hydrationState) String() string {
...
-				err = fmt.Errorf("failed to hydrated sub pkg %s %w", s, err)
-				return resources, err
-			}
-			input = append(input, transitiveResources...)
+	if hctx.root != currPkg {
+		// Resources are read from local filesystem or generated at a package level, so the
+		// path annotation in each resource points to path relative to that package.
+		// But the resources are written to the file system at the root package level, so
+		// the path annotation in each resources needs to be adjusted to be relative to the rootPkg.
+		relPath, err := filepath.Rel(hctx.root.Path(), currPkg.Path())
```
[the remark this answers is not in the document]
droot: Defined pkg level `RelativePathTo` method as you suggested in `1`. Pl. take a look and let me know.

internal/pipeline/executor.go
```diff
@@ -207,78 +135,107 @@ func (s hydrationState) String() string {
...
-	pkgWriter := &kio.LocalPackageWriter{PackagePath: p.Path()}
 
-	filters, err := fnFilters(p.Pipeline(), p.Path())
+	pl, err := pn.pkg.Pipeline()
+	if err != nil {
+		return nil, fmt.Errorf("failed to read pipeline for package %s %w", pn.pkg, err)
+	}
+
+	// empty pipeline
+	if len(pl.Mutators) == 0 && len(pl.Validators) == 0 {
```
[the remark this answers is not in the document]
droot: Added `pipeline.IsEmpty()` method to encapsulate the logic.

pkg/test/runner/runner.go
```diff
@@ -45,10 +45,14 @@ func NewRunner(testCase TestCase, c string) (*Runner, error) {
 
 // Run runs the test.
 func (r *Runner) Run() error {
-	if r.cmd == CommandFn {
+	switch r.cmd {
+	case CommandFnEval:
 		return r.runFn()
+	case CommandFnRender:
+		return r.runPipeline()
```
natasha41575, line 52: Maybe these method names should change. Before they were calling `kpt fn run` and `kpt pipeline run` and now the names here don't make sense. E.g. instead of `r.runFn` and `r.runPipeline` they can be `r.fnRun` and `r.fnRender`

internal/pipeline/executor.go
```diff
@@ -207,78 +135,107 @@ func (s hydrationState) String() string {
...
+		return input, nil
+	}
+
+	filters, err := fnFilters(pl, pn.Path())
```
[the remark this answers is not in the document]
droot: Yeah. Removed an unnecessary function and calling it `fnChain`.

internal/pipeline/executor.go
```diff
@@ -288,33 +245,35 @@ func hydrate(p *pkg, hctx *hydrationContext) (resources []*yaml.RNode, err error
...
-		meta, _ := r.GetMeta()
-		if !strings.Contains(meta.APIVersion, "kpt.dev") {
-			filtered = append(filtered, r)
+		meta, err := r.GetMeta()
+		if err != nil {
+			return resources, err
+		}
+		// TODO(droot): revisit this. There is an edge case where if relativePath(root, curr)
+		// is same as relativePath(curr, local-resource)
+		if relPath != "" && !strings.HasPrefix(meta.Annotations[kioutil.PathAnnotation], relPath) {
```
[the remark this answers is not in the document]
droot: Done.

internal/pipeline/executor.go
```diff
@@ -288,33 +245,35 @@ func hydrate(p *pkg, hctx *hydrationContext) (resources []*yaml.RNode, err error
...
-	hctx.hydrationStates[p.Path()] = Wet
-	p.resources = resources
-	return resources, err
 }
 
-// filterMetaData filters kpt metadata files such as pipeline,
-// Kptfile, permissions etc.
-func filterMetaData(resources []*yaml.RNode) []*yaml.RNode {
-	var filtered []*yaml.RNode
+// adjustRelPath updates the resources with given relative path.
```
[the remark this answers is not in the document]
droot: Done. Also added new method at the `pkg` level for `RelativePathTo` that takes another package.

internal/pipeline/testdata/app-with-generator/configs/db/Kptfile
```diff
@@ -0,0 +1,20 @@
+apiVersion: kpt.dev/v1alpha2
+kind: Kptfile
+metadata:
+  name: db
+pipeline:
+  mutators:
+  - image: gcr.io/sunilarora-sandbox/httpbin-fn:latest
```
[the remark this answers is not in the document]
droot: I plan to move this function to function catalog under `contrib` section. This will point to the function from the catalog before this PR is merged, so will keep this comment unresolved for now.

internal/pipeline/testdata/app-with-db/configs/resources.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: apps/v1
```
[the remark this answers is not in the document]
droot: Ack.

internal/pipeline/testdata/simplepackage/configs/Kptfile
```diff
@@ -4,9 +4,18 @@ metadata:
...
   mutators:
-  - image: gcr.io/kpt-fn/set-namespace
+  - image: gcr.io/kpt-fn/set-namespace:unstable
     configMap:
       namespace: staging
-  - image: gcr.io/kpt-fn/set-label
-    configMap:
-      tier: backend
+  - image: gcr.io/kpt-fn/set-label:unstable
+    config:
```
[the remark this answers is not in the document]
droot: Done.

internal/pkg/pkg.go
droot: Changed this one to `string(p.UniquePath)`.

internal/pkg/pkg.go
droot: Done.

internal/pkg/pkg.go
droot: Yes, done.

pkg/test/runner/runner.go
droot: Good point. Done.

internal/pipeline/testdata/app-with-generator/configs/db/Kptfile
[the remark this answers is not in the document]
droot: Initiated a PR for adding this function in function catalog: https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/182

internal/pkg/pkg.go
```diff
@@ -37,25 +52,25 @@ type Pkg struct {
...
 
 // New returns a pkg given an absolute or relative OS-defined path.
 // Use ReadKptfile or ReadPipeline on the return value to read meta resources from filesystem.
 func New(path string) (*Pkg, error) {
 	p := filepath.Clean(path)
 
-	u, err := filepath.EvalSymlinks(p)
+	// Note(droot): Calling EvalSymlinks before Abs causes a strange issue on OSX
+	// for directories under /tmp. If you have directory /tmp/a,
+	// evalSymlinks("a") !==> /private/tmp
```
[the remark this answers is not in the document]
droot: Yes.

internal/pkg/pkg.go
```diff
@@ -37,25 +52,25 @@ type Pkg struct {
...
+	kptfile *kptfilev1alpha2.KptFile
 }
 
 // New returns a pkg given an absolute or relative OS-defined path.
 // Use ReadKptfile or ReadPipeline on the return value to read meta resources from filesystem.
 func New(path string) (*Pkg, error) {
 	p := filepath.Clean(path)
 
-	u, err := filepath.EvalSymlinks(p)
+	// Note(droot): Calling EvalSymlinks before Abs causes a strange issue on OSX
```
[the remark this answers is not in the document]
droot: So did some digging:
On my laptop (OSX), `/tmp` and `/var` are symlinks pointing to `/private/tmp` and `/private/var` respectively and luckily, I ran into this issue during a manual test. So if `EvalSymlinks` is called on a path that contains a symlinks component for ex. calling `EvalSymlinks(`/tmp/some-dir`)` will resolve to `/private/tmp/some-dir` and eventually `Abs(`/private/tmp/some-dir`)` resulting in `/private/tmp/some-dir`).  Versus, assuming you are under `/tmp` dir, calling `EvalSymlinks(`some-dir`)` ==> `some-dir` and Abs(`some-dir`) resulting in `/tmp/some-dir`. 

The above scenario is not a problem if you don't have nested package hierarchy, this becomes a problem in a recursive hydration because subpackages will have absolute paths `/tmp/some-dir/subpkg-a` and they will resolve to `/private/tmp/some-dir/subpkg-a` resulting in an inconsistent filesystem hierarchy for the pkg.

So changing the order and calling `Abs(..)` first forces the symlinks to be resolved always (or calling Abs on the rootpkg before calling pkg.New() can avoid the issue).

Now, I expect to see same behavior on Linux as well, but haven't verified it.

internal/pipeline/executor.go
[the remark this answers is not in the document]
droot: Made another attempt today to use `pkg.UniquePath` type and worked pretty well this time. Tracked these changes in a separate commit to make easier to review.

internal/pkg/pkg.go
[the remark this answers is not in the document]
droot: Updated the comment with more details.

internal/pipeline/executor.go
```diff
@@ -288,33 +245,35 @@ func hydrate(p *pkg, hctx *hydrationContext) (resources []*yaml.RNode, err error
...
+func adjustRelPath(resources []*yaml.RNode, relPath string) ([]*yaml.RNode, error) {
 	for _, r := range resources {
-		meta, _ := r.GetMeta()
-		if !strings.Contains(meta.APIVersion, "kpt.dev") {
-			filtered = append(filtered, r)
+		meta, err := r.GetMeta()
+		if err != nil {
+			return resources, err
+		}
+		// TODO(droot): revisit this. There is an edge case where if relativePath(root, curr)
```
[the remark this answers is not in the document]
droot: Worked through an example today and verified that this edge case doesn't exist. resources will be stamped with correct relative path (root --> pkg) when they are either read or generated at that package. Subsequent call to `adjustRelPath` up in the chain (as we move from inward pkg to outer) will always have relative path prefixed. This use-case is exercised using two e2e test cases that we have `pkg-with-generator` and `pkg-with-subpkgs`, so we are good from testing standpoint.

I was initially thinking of using `pkgbuilder` under testutil to implement some more unit tests but that pkg requires more tweaking because it uses older `kptfile` format, so abandoned that idea.

droot:
@frankfarzan I believe I have addressed all the feedback. Pl. take a look. One thing I want to improve in a follow up PR:

- Error strings/context for end-users and debugging
- Possibly a way to implement lightweight unit tests which is a bit challenging given that the core logic depends heavily on filesystem calls and function-runner through docker, but I want to make an attempt because that can a long way.

internal/pkg/pkg.go
```diff
@@ -150,6 +154,15 @@ func (p *Pkg) String() string {
 	return string(p.DisplayPath)
 }
 
+// RelativePathTo returns current package's path relative to a given package.
```
frankfarzan, line 157: Explain the semantics of the output string. It's OS-specific path (not slash seperated).

frankfarzan (approved):
Regarding test coverage: In addition to increasing unit test coverage, #1548 will address conversion of function catalog examples.

PR #1484, opened 2021-02-21
add test for pipeline run using .krmignore file

### LINKED ISSUE #1414: Rationalize krmignore with the filtering resources in `pipeline run`
We are filtering out resources during pipeline execution for ex. Kptfile, function configs etc. We need to investigate and support `krmignore` use-cases as well. Creating this for us to investigate first what needs to be done and then implement it.

natasha41575:
adds a test for pipeline run using krmignore

same test as https://github.com/GoogleContainerTools/kpt/pull/1453 but ignores the resource file, which means there should be no diff 

I believe the functionality is already implemented (see comment on https://github.com/GoogleContainerTools/kpt/issues/1414)

close #1414


Shell32-Natsu (commented):
nit: No need to put all the files in `config` directory.

natasha41575:
@Shell32-Natsu moving the files results in unexpected behavior, looking into it now

Shell32-Natsu:
@natasha41575 I guess you need to add `.expect` to the `krmignore` as well.

natasha41575:
@Shell32-Natsu done

internal/pipeline/testdata/packagewithkrmignore/Kptfile
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: app
+pipeline:
+  mutators:
+  - image: gcr.io/kpt-fn/set-namespace
+    configMap:
+      namespace: staging
+  - image: gcr.io/kpt-fn/set-label
+    configMap:
+      tier: backend
```
Shell32-Natsu, line 12: Just a reminder, currently set-label function doesn't support this input from `configMap` field. Will fix it soon. https://github.com/GoogleContainerTools/kpt/issues/1487

Shell32-Natsu (approved):
LGTM

internal/pipeline/testdata/packagewithkrmignore/Kptfile
natasha41575: thanks!

PR #1491, opened 2021-02-23
add test for invalid configuration

natasha41575:
https://github.com/GoogleContainerTools/kpt/issues/1486


internal/pipeline/testdata/invalidconfig/.expected/config.yaml
```diff
@@ -0,0 +1,2 @@
+exitCode: 1
+network: false
```
Shell32-Natsu, line 2: I suppose the default value for network is false?

pkg/test/runner/runner.go
```diff
@@ -148,6 +148,12 @@ func (r *Runner) runPipeline() error {
 		fnErr = command.Execute()
 	}
 
+	// check for errors, because compareResult doesn't have access to the
+	// command's exit code for pipeline
+	if fnErr != nil && r.testCase.Config.ExitCode != 0 {
```
Shell32-Natsu, line 153: If the pipeline is expected to fail, we should not keep looping it in line 146.

`compareResult` can get the exit code from `fnErr`.

Shell32-Natsu (commented):
I have a question: In the case that a pipeline would fail, where will it print the error message?

internal/pipeline/testdata/invalidconfig/.expected/config.yaml
frankfarzan: +1, it should be omitted for brevity.

natasha41575: done

pkg/test/runner/runner.go
natasha41575: I changed it above so that it breaks when fnErr != nil.

For `compareResult`, it cannot do this for pipeline because I am running it with `cobra.Command.Execute` which does not return `exec.ExitError` as expected by `compareResult` (`fn run` test uses `exec.Command`). I could change pipeline to use `exec.Command` but that would require the developer to rebuild kpt locally every time they make changes, which doesn't seem ideal to me and could become tedious

natasha41575: If pipeline fails, the error message is printed in the terminal.

natasha41575:
@droot based on your feedback over chat, I made a top level folder for invalid configurations that has two different test cases with different invalid configurations.

Note that `configFileNotFound` passes with your PR https://github.com/GoogleContainerTools/kpt/pull/1483, but `multipleConfigs` doesn't. I assume that that is because the check for multiple configurations has not been implemented yet?

natasha41575:
> nit: Lets avoid camelcase for the directory names.
> 
> Also, pl. add config to skip a test so that we can skip tests those are expected to fail.

done @droot

internal/pipeline/testdata/invalidconfig/configfilenotfound/Kptfile
```diff
@@ -0,0 +1,8 @@
+apiVersion: kpt.dev/v1alpha2
+kind: Kptfile
+metadata:
+  name: app
+pipeline:
+  mutators:
+  - image: gcr.io/kpt-fn/set-namespace:unstable
+    configPath: configMap.yaml
```
droot, line 8: I am guessing `configMap.yaml` is missing. Can we add a comment here. Also pl. use lower case for filename.

pkg/test/runner/runner.go
```diff
@@ -45,6 +45,11 @@ func NewRunner(testCase TestCase, c string) (*Runner, error) {
 
 // Run runs the test.
 func (r *Runner) Run() error {
+	if r.testCase.Config.Skip {
```
droot, line 48: Not familiar with the runner implementation but if it has access to the `testing.T` object, then probably using `t.Skip()` will be better.

droot (commented):
Minor nits.

internal/pipeline/testdata/invalidconfig/configfilenotfound/Kptfile
natasha41575: done

pkg/test/runner/runner.go
natasha41575: runner does not have access to it but the test that uses runner does. I have added some logic to use t.Skip()

e2e/e2e_docker_test.go
```diff
@@ -44,6 +44,9 @@ func runPipelineTests(t *testing.T, path string) {
 			if err != nil {
 				t.Fatalf("failed to create test runner: %s", err)
 			}
+			if r.Skip() {
+				t.Skip()
```
droot, line 48: So `r.Run` will still be invoked below. I am curious what will be the `skip` behavior then ?

droot: We can jump on a call to to avoid back and forth on it.

droot: So I looked at the testing pkg https://golang.org/src/testing/testing.go?s=28687:28729#L810 and `t.Skip()` exits the `GoRoutine` that the test is running in, so that answers my question.

droot (approved):

Issue #1387, opened 2021-01-25
Enable Automated Kpt E2E Tests


runewake2:
Currently `kpt` is tested using a series of tests against a local `kind` cluster on a developers machine. This requires the tests to be run manually against the developers local machine and there is no automated way to get signals about the project health. We can automate this in our CI system to remove this manual step.

Completing this would allow e2e coverage alongside the existing unit and integration tests currently run against `kpt`.

* Ideally e2e tests would provide coverage over the `kpt` binaries we produce and validate that those binaries work.
* We may be able to continue using `kind` in ubuntu runtimes since Docker is available in these containers, however a different solution will be needed if we want to support tests over Windows or MacOS (currently Windows supports Windows based containers, MacOS does not support containers).
* Ideally each test run should occur in a hermetic environment. `kind` can provide this, but a static test environment likely will not. This accomplishes two things:
  * Avoids conflicts when 2 tests are triggered simultaneously (ex. 2 PR's).
  * Avoids conflicts from stale state from previous tests.
* Changes to [`end-to-end-test.sh`](https://github.com/GoogleContainerTools/kpt/blob/master/e2e/live/end-to-end-test.sh) may be required to provide clearer signals after a test is run.

PR #1492, opened 2021-02-24
E2E Testing Environment for Kpt Live Tests

runewake2:
Introduces a workflow that configures a KinD cluster for use with `end-to-end.sh` and runs those tests against the cluster. An example of the results from this execution is [here](https://github.com/runewake2/kpt/runs/1965686489?check_suite_focus=true).

Notes:

* This test is still triggered manually (navigate to Actions, select the workflow and choose to run it).
* This test may never fail. (the test script does not exit with an error so test failures are only displayed in workflow logs)

The intent is to resolve both of these in follow up PR's, in the meantime this will enable new tests to be run in this environment.

This uses https://github.com/engineerd/setup-kind to configure the KinD environment (downloading and installing KinD). I've fixed the action to a specific commit SHA (setup-kind v0.5.0) following [GitHub's recommendations](https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions).

Related to issue: #1387


runewake2:
I've added some additional changes that will cause `end-to-end-test.sh` to exit if a test does not pass. This should allow us to get feedback in these automated tests.

seans3 (approved):
/lgtm

mortent:
This looks good to me but why not run it on every PR?

runewake2:
I am running this manually right now to confirm it behaves well, once that's been done checks on pushes to master/next and pulls against those branches will be added. Doing this in stages should help prevent any potential errors from blocking unrelated PR's.

Issue #1352, opened 2021-01-11
Pull Requests Against kpt Build the Project Twice in CI


runewake2:
When a Pull Request is opened against `kpt` both [`verifyGuides.yml`](https://github.com/GoogleContainerTools/kpt/blob/master/.github/workflows/verifyGuides.yml#L29) and [`go.yml`](https://github.com/GoogleContainerTools/kpt/blob/master/.github/workflows/go.yml#L42) are triggered. Both of these jobs include a step which builds `kpt`. Since both binaries are built against the same source they should be identical. We should be able to reduce the work done in CI by removing this duplication, this should help increase the speed of automated feedback against PR's.

We may be able to use the https://github.com/actions/upload-artifact and https://github.com/actions/download-artifact tasks to share a single binary build with both child tasks.

PR #1497, opened 2021-02-26
Update e2e tests to run on pull/push

runewake2:
Fixes: #1387

This will cause pushes and pulls against kpt to run 2 copies of our e2e tests against Kubernetes versions 1.16 and 1.17 (versions may be added/removed by modifying the matrix).

Changes to the kpt docs/site will be ignored and not trigger tests.

Test results will appear like this: https://github.com/GoogleContainerTools/kpt/actions/runs/603827646

### DIFF
--- a/.github/workflows/e2eEnvironment.yml
+++ b/.github/workflows/e2eEnvironment.yml
@@ -1,18 +1,38 @@
 name: "Kpt Live - KinD Tests"
+
 on:
-  - workflow_dispatch
+  push:
+    branches:
+      - master
+      - next
+    paths-ignore: # do not run these tests when only changes to documentation are detected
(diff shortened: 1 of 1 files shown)


mortent:
How long does it take to run this test?

runewake2:
~5 minutes per environment (run in parallel). See: https://github.com/GoogleContainerTools/kpt/actions/runs/603827646
Total time to run the check is a bit higher but it is not on the longest path. Our build/test/lint job is the slowest job right now so this should have minimal impact on total CI runtime.

Introducing more tests may change this behavior, but there are other improvements we can make to improve test performance here (we are currently building kpt once per workflow for example #1352).

mortent (approved):

PR #1515, opened 2021-03-03
Add firebase hosting configs for docs site.

frankfarzan:


firebase.json
```diff
@@ -0,0 +1,16 @@
+{
+  "hosting": {
+    "public": "site",
+    "ignore": [
+      "firebase.json",
+      "**/.*",
+      "**/node_modules/**"
+    ],
+    "rewrites": [
```
etefera, line 9: Can we change this rewrite to the below:
``` 
"rewrites": [
      {
        "regex": "^[\/\\w\\.-]+[^\\.]..$",
        "destination": "/index.html"
      }
    ]
```
Can also have this done in another PR

etefera (approved):

runewake2 (approved):
Looks good, an optional comment on the action version.

👍  on the inclusion of a preview page. That will be really helpful during doc PR's.

frankfarzan:
@etefera the build_and_preview action is failing: https://github.com/GoogleContainerTools/kpt/runs/2033853095?check_suite_focus=true

runewake2:
[Is the `FIREBASE_SERVICE_ACCOUNT_KPT_DEV` secret configured](https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-a-repository)?

etefera:
> [Is the `FIREBASE_SERVICE_ACCOUNT_KPT_DEV` secret configured](https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-a-repository)?

The `firebase init hosting` command should've taken care of setting this, but it's worth checking for this repo. I'd thought maybe it would take some time to propagate according to https://github.com/FirebaseExtended/action-hosting-deploy/issues/40#issuecomment-729438938 but OP mentions that the the Firebase CLI might not have had access to the repo at the time it tried to write the secret

frankfarzan:
Per offline discussion, breakage is likely due to secrets not being passed to fork PR. @etefera to fast follow with a general fix.

PR #1528, opened 2021-03-05
 Trigger preview site builds on PRs.

etefera:


frankfarzan:
Have you read this:

https://securitylab.github.com/research/github-actions-preventing-pwn-requests

pull_request_target is dangerous since it will have access to secrets and write permission to the repo. But in this workflow you want to build and push a firebase deployment.

etefera:
> Have you read this:
> 
> https://securitylab.github.com/research/github-actions-preventing-pwn-requests
> 
> pull_request_target is dangerous since it will have access to secrets and write permission to the repo. But in this workflow you want to build and push a firebase deployment using the untrusted changes in the PR.

Yeah, +1 that `pull_request_target doesn't seem to be the right choice here

PR #1537, opened 2021-03-09
Don't index kpt.dev while in development

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1536
Adds KPT to page title.

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -2,6 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
+    <meta name="robots" content="noindex" />
     <title>KPT - Home</title>
     <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
     <meta name="description" content="Description" />
--- a/site/index.html
+++ b/site/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <title>Home</title>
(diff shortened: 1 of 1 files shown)


site/index.html
```diff
@@ -2,7 +2,8 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <title>Home</title>
+    <meta name="robots" content="noindex" />
```
droot, line 5: Is it enough ? or we need `robots.txt` at the root ?

etefera: This is the only HTML served from the domain so any page would be crawled with that attribute https://developers.google.com/search/docs/advanced/robots/robots_meta_tag

droot: ok.

droot (approved):

site/index.html
```diff
@@ -2,7 +2,8 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <title>Home</title>
+    <meta name="robots" content="noindex" />
+    <title>KPT - Home</title>
```
frankfarzan, line 6: we should use small cap `kpt` consistently everywhere.

frankfarzan (commented):
lgtm after fix

frankfarzan (approved):

site/index.html
etefera: SGTM, done.

PR #1540, opened 2021-03-11
Upgrade k8s.io dependencies.

etefera:
Removes a replace directive that breaks `go install`. Fixes https://github.com/GoogleContainerTools/kpt/issues/1535
Changes the namespace in a test to reflect a change in one of the dependencies.


Liujingfang1:
The change in `pkg/live/dual-delegating-loader_test.go` looks good to me.

mortent:
This looks good to me. @etefera, can you also update the libraries from k8s in https://github.com/kubernetes-sigs/cli-utils and make sure all the tests pass there before we merge this one?

etefera:
> This looks good to me. @etefera, can you also update the libraries from k8s in https://github.com/kubernetes-sigs/cli-utils and make sure all the tests pass there before we merge this one?

https://github.com/kubernetes-sigs/cli-utils/pull/337 is merged in and good to go

mortent:
@etefera I think we should just update the cli-utils version as part of this PR too. So we can use the version of cli-utils which is based on the same version of the k8s libraries. We will have to rely on an unreleased version, but that is ok and we will address that when we do a release.

etefera:
> @etefera I think we should just update the cli-utils version as part of this PR too. So we can use the version of cli-utils which is based on the same version of the k8s libraries. We will have to rely on an unreleased version, but that is ok and we will address that when we do a release.

Updated with the results of `go get sigs.k8s.io/cli-utils@70426d6` with tests still passing locally.

mortent (approved):

Issue #1506, opened 2021-03-02
create an image for kpt with gcloud and git included


mikebz:
### Is your feature request related to a problem? Please describe.
some functionality of kpt requires `gcloud` and `git` client to be installed on the Docker image.  The customer request is to provide an image with those already installed instead of having to install those manually.

### Describe the solution you'd like
There should be a docker image for kpt that people can use in their CI/CD pipeline that includes gcloud and git.  It should be an officially supported image.

### Describe alternatives you've considered
#1457 PR originally attempted upgrading the base image for kpt to go from alpine to `gcr.io/google.com/cloudsdktool/cloud-sdk:alpine` but the image size has gone from ~50MB to almost 200MB. 

### Additional context
@martinmaly - original requestor

martinmaly:
Thank you!

Would it work to have 2 images, one being the original, smaller kpt image and add the larger image which contains gcloud and git? that way customers using the kpt-only image won't be affected by addition of the other components?

mikebz:
that makes sense to me.  @mortent @frankfarzan feel free to chime in on this.

runewake2:
Right now I would recommend using some of the general purpose [`gcr.io/google.com/cloudsdktool/cloud-sdk`](https://github.com/GoogleCloudPlatform/cloud-sdk-docker) images (either the `latest` or `debian_component_based` versions). Right now `latest` does not include kpt but I'll see if we can get it added. The `debian_component_based` version does already include kpt and you can use that image right now.

These general purpose images do contain a number of features so are larger images (multiple GB). Would these images satisfy the use case you are trying to address here @martinmaly?

***

With the `debian_component_based` image kpt version `0.37.1` is currently available:

```sh
➜  docker run gcr.io/google.com/cloudsdktool/cloud-sdk:debian_component_based kpt version   
0.37.1
```

martinmaly:
Looking at the image, it may be initially sufficient. it seems that majority of the size is coming from gcloud (possibly some also from the base images). It may be good place to start and we can optimize if and as needed.

It appears that the image contains git so that should satisfy our requirements for now. 

Thank you!

runewake2:
To provide some context here outside our other conversations:

Version 0.39.0 of kpt introduces a new `kpt-gcloud` image that you can find here: [`gcr.io/kpt-dev/kpt-gcloud`](gcr.io/kpt-dev/kpt-gcloud)

[...]

PR #1553, opened 2021-03-16
Add Kpt + Gcloud Image Definition

runewake2:
See: #1506

This introduces a new container image that includes `gcloud` alongside other programs configured with `kpt`. The image will be released as `kpt-gcloud` into the same container registry as the kpt container (`kpt-dev`). The image is based off of gcr.io/google.com/cloudsdktool/cloud-sdk:alpine

This image is significantly larger than the `kpt` image already published at gcr.io/kpt-dev/kpt but smaller than the larger general purpose `cloud-sdk` images that are available with `kpt`.

***

This PR also updates the kpt releasing README with steps that will allow you to run and validate Cloud Build releases of `kpt` against your local fork.

Hardcoded github owner values have been removed from the kpt Goreleaser config to prevent publishing new releases to the `GoogleContainerTools/kpt` repo unintentionally. (note: this is still possible with wrong parameters and notes have been included to prevent this as best as possible)

> Note: A followup change to include the image on kpt.dev will follow the first release of these new images.


gcloud/Dockerfile
```diff
@@ -0,0 +1,34 @@
+# Copyright 2019 Google LLC
```
mortent, line 1: Would it be better to put this inside the `release` directory? It seems strange to create a new top-level `gcloud` directory.

release/tag/goreleaser.yaml
```diff
@@ -23,7 +23,7 @@ builds:
     goarch:
       - amd64
       - arm64
-    ldflags: -s -w -X github.com/GoogleContainerTools/kpt/run.version={{.Version}}
+    ldflags: -s -w -X github.com/{{ .Env._GITHUB_USER }}/kpt/run.version={{.Version}}
```
mortent, line 26: I don't think we should change this. The package path would be the same regardless of whether it run on the `GoogleContainerTools` repo or somewhere else right?

frankfarzan:
In the PR description, can you specify the size of the two container images?

release/gcloud/Dockerfile
```diff
@@ -0,0 +1,34 @@
+# Copyright 2019 Google LLC
```
frankfarzan, line 1: For consistency, the existing top-level Dockerfile should be moved to release directory as well. This can be a separate PR

release/gcloud/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+ARG KPT_VERSION=unknown
+
+WORKDIR /go/src/
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+COPY . .
+RUN go build -v -o /usr/local/bin/kpt -ldflags="-s -w -X github.com/GoogleContainerTools/kpt/run.version=$KPT_VERSION" ./
+
+FROM gcr.io/google.com/cloudsdktool/cloud-sdk:alpine
```
frankfarzan, line 28: Should probably use a pinned version instead. We don't want to have this change without our explicit consent.

frankfarzan (changes requested):

release/gcloud/Dockerfile
runewake2: Ack. Will open that PR after this is merged to avoid conflicts with the merge/squash of this change.

frankfarzan (approved):

release/gcloud/Dockerfile
runewake2: I've pinned this to `332.0.0-alpine` and updated our releasing steps to include updates to the base image versions prior to release.

mortent (approved):

PR #1554, opened 2021-03-16
use starlark based generator function for e2e test

droot:
This PR modifies an e2e test `pkg-with-generator` for `fn render` to use starlark based function instead of `Go` function (implemented in PR https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/182)

### DIFF
--- a/internal/pipeline/testdata/pkg-with-generator/configs/db/starlark-httpbin.yaml
+++ b/internal/pipeline/testdata/pkg-with-generator/configs/db/starlark-httpbin.yaml
@@ -1,0 +1,42 @@
+apiVersion: fn.kpt.dev/v1alpha1
+kind: StarlarkRun
+metadata:
+  name: httpbin-gen
+source: |
+  httpbin_deployment = {
+     "apiVersion": "apps/v1",
+     "kind": "Deployment",
+     "metadata": {
+        "name": "httpbin",
+     },
+     "spec": {
+        "replicas": 4,
+        "template": {
+           "spec": {
+              "containers": [
+                {
+                  "name": "httpbin",
+                  "image": "kennethreitz/httpbin",
+                  "ports": [
+                    {
+                       "containerPort": 9876
+                    }
+                  ]
+                }
+              ]
+           }
+        }
+    }
+  }
+  # filter to return if resource is HTTPBin resource
+  def is_httpbin(r):
+    return r["apiVersion"] == "apps/v1" and r["kind"] == "Deployment" and r["metadata"]["name"] == "httpbin"
+
+  def ensure_httpbin(resources):
+    for r in resources:
+      if is_httpbin(r):
+        return
+
+    resources.append(httpbin_deployment)
+
+  ensure_httpbin(ctx.resource_list["items"])
--- a/internal/pipeline/testdata/pkg-with-generator/configs/db/Kptfile
+++ b/internal/pipeline/testdata/pkg-with-generator/configs/db/Kptfile
@@ -4,7 +4,8 @@ metadata:
   name: db
 pipeline:
   mutators:
-  - image: gcr.io/sunilarora-sandbox/httpbin-fn:latest
+  - image: gcr.io/kpt-fn/starlark:unstable
+    configPath: starlark-httpbin.yaml
   - image: gcr.io/kpt-fn/set-namespace:unstable
     configMap:
       namespace: db


droot:
`starlark` is super powerful and is very nice to use. Comparing it with the `Go` implementation https://github.com/GoogleContainerTools/kpt/pull/1554 , I am questioning myself -- why would I use `Go` or `TS` for this ?


/cc @Shell32-Natsu @frankfarzan @mengqiy

droot:
I am thinking the `fn render` can support path to starlark scripts directly where it can wrap them into fnConfig directly so the user can specify it as follows:

```
apiVersion: kpt.dev/v1alpha2
...
pipeline:
 mutators:
  -image: gcr.io/kpt-fn/starlark:unstable
    configPath: <path/to/starlarkscript>
```

A nice way to use this will be platform teams can have a library of `starlark` scripts as kpt pkg and app teams can just include them as a subpackage in their top level pkg and point to the starlark scripts directly in the pipeline and be done!

/cc @frankfarzan @mengqiy @Shell32-Natsu @mortent

Shell32-Natsu (approved):
lgtm

internal/pipeline/testdata/pkg-with-generator/configs/db/starlark-httpbin.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: fn.kpt.dev/v1alpha1
+kind: StarlarkRun
+metadata:
+  name: httpbin-gen
```
Shell32-Natsu, line 4: Please update the name

internal/pipeline/testdata/pkg-with-generator/configs/db/starlark-httpbin.yaml
```diff
@@ -0,0 +1,34 @@
...
+                                                                "containerPort": 9876
+                                                        }
+                                                ]
+                                        }
+                                ]
+                        }
+                }
+        }
+  }
+  def ensureHTTPBin(resources):
```
mengqiy, line 32: You don't have to define a function if you don't need a for loop.

mengqiy:
> platform teams can have a library of starlark scripts as kpt pkg and app teams can just include them as a subpackage in their top level pkg

To reuse starlark scripts, a slightly different would be:
We introduce another client-side CRD `StarlarkScript` which wraps a starlark script. And `StarlarkRun` can reference it.
The platform team can have a list of `StarlarkScript` objects defined in different packages and the top level pkg can use them as sub-package.

mengqiy:
Starlark runtime function + starlark script is a lighter weight way to write a kpt function.

internal/pipeline/testdata/pkg-with-generator/configs/db/starlark-httpbin.yaml
droot: Made it idempotent and also using snake-cased syntax

mengqiy (approved):
LGTM

internal/pipeline/testdata/pkg-with-generator/configs/db/starlark-httpbin.yaml
mengqiy: That's great

internal/pipeline/testdata/pkg-with-generator/configs/db/starlark-httpbin.yaml
```diff
@@ -0,0 +1,41 @@
...
+           }
+        }
+    }
+  }
+  # filter to return if resource is HTTPBin resource
+  def is_httpbin(r):
+    return r["apiVersion"] == "apps/v1" and r["kind"] == "Deployment" and r["metadata"]["name"] == "httpbin"
+
+  def ensure_httpbin(resources):
+      httpbin_deployments = [r for r in resources if is_httpbin(r)]
```
frankfarzan, line 36: nit: you can optimize by returning on the first hit

frankfarzan (approved):

frankfarzan:
> `starlark` is super powerful and is very nice to use. Comparing it with the `Go` implementation #1554 , I am questioning myself -- why would I use `Go` or `TS` for this ?
> 
> /cc @Shell32-Natsu @frankfarzan @mengqiy

Like any other domain of computer science, choice of language/runtime depends on the use case. If you need to use ecosystem of libraries for reusability OR general purpose language constructs OR static typing OR ..., then Go or TS may be more appropriate.

internal/pipeline/testdata/pkg-with-generator/configs/db/starlark-httpbin.yaml
droot: Good pt, just fixed it. Was having too much fun with list comprehension.

PR #1556, opened 2021-03-17
Implement kpt fn doc command

mengqiy:
`kpt fn doc --image=IMAGE` will return the help text.

fixes: https://github.com/GoogleContainerTools/kpt/issues/1516

### DIFF
--- a/internal/cmdfndoc/cmdfndoc.go
+++ b/internal/cmdfndoc/cmdfndoc.go
@@ -1,0 +1,73 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package cmddesc contains the desc command
+package cmdfndoc
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/GoogleContainerTools/kpt/internal/docs/generated/fndocs"
+	"github.com/GoogleContainerTools/kpt/internal/util/cmdutil"
+	"github.com/spf13/cobra"
+)
+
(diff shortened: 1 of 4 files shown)


mengqiy:
To get the test passing, we need https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/189 to be merged first.

mengqiy:
The test is passing now.

mengqiy:
cc: @etefera

droot (approved):
Assuming we add `doc` subcommand for other groups as well, the change looks ok to me. Since its blocking other work, I am going to approve it and if something is not inline, we can change it in a follow up PR.

internal/docs/generated/fndocs/docs.go
```diff
@@ -96,6 +96,18 @@ var RunExamples = `
   kpt fn run DIR/
 `
 
+var DocShort = `Locally execuate a function to get the help text`
```
frankfarzan, line 99: Explain the WHAT not the HOW.

Something like:

`Displays the documentation for a function`

internal/cmdfndoc/cmdfndoc.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/GoogleContainerTools/kpt/internal/docs/generated/fndocs"
+	"github.com/GoogleContainerTools/kpt/internal/util/cmdutil"
+	"github.com/spf13/cobra"
+)
+
+func NewRunner(parent string) *Runner {
+	r := &Runner{}
+	c := &cobra.Command{
+		Use:     "doc [PKG_PATH] [flags]",
```
frankfarzan, line 32: What do you need PKG_PATH?

internal/cmdfndoc/cmdfndoc.go
```diff
@@ -0,0 +1,73 @@
...
+		Args:    cobra.MaximumNArgs(1),
+		Short:   fndocs.DocShort,
+		Long:    fndocs.DocShort + "\n" + fndocs.DocLong,
+		Example: fndocs.DocExamples,
+		RunE:    r.runE,
+	}
+	r.Command = c
+	c.Flags().StringVar(&r.Image, "image", "", "kpt function image name")
```
frankfarzan, line 40: Do you want to support executable similar to how eval does it?

frankfarzan, line 40: What happens if the image flag is not provided? Add a test for this.

mengqiy: No.
IMO it doesn't add much value if comparing with running `executable-name --help` directly.

PR #1583, opened 2021-03-19
Use Pkg type for Get, Fetch, and Update commands

mortent:
Updates the `Get`, `Fetch`, and `Update` functionality to use the `pkg.Pkg` structure. 

As part of this, it also makes some changes to the `pkg.Pkg` functionality:
 * The `pkg.Kptfile()` function returns an error if a `Kptfile` doesn't exist instead of returning an empty Kptfile. We no longer have the concept of implicit packages.
 * Updates the `pkg.Subpackages` function to include all subpackages, and not only subpackages in the direct subfolders. Also adds tests for this function.

Must merge after https://github.com/GoogleContainerTools/kpt/pull/1568

### DIFF
--- a/internal/pkg/pkg.go
+++ b/internal/pkg/pkg.go
@@ -24,7 +24,6 @@ import (
 	"strings"
 
 	kptfilev1alpha2 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1alpha2"
-	"github.com/GoogleContainerTools/kpt/pkg/kptfile/kptfileutil"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 	"sigs.k8s.io/kustomize/kyaml/sets"
@@ -167,23 +166,24 @@ func (p *Pkg) DirectSubpackages() ([]*Pkg, error) {
 			return nil
 		}
 
-		// Ignore anything inside the .git folder
-		if info.Name() == ".git" && info.IsDir() {
-			return filepath.SkipDir
-		}
-
 		// For every folder, we check if it is a kpt package
 		if info.IsDir() {
-			kfPath := filepath.Join(path, kptfilev1alpha2.KptFileName)
-			_, err := os.Stat(kfPath)
-			// If we get any other errors than IsNotExist, we bail out.
-			if err != nil && !os.IsNotExist(err) {
-				return fmt.Errorf("failed to check file at %s: %w", kfPath, err)
+			// Ignore anything inside the .git folder
+			// TODO: We eventually want to support user-defined ignore lists.
+			if info.Name() == ".git" {
+				return filepath.SkipDir
 			}
-			// If err is nil here, it means we did find the Kptfile. We add the
+
+			// Check if the directory is the root of a kpt package
+			isPkg, err := IsPackageDir(path)
+			if err != nil {
+				return err
+			}
+
+			// If the path is the root of a subpackage, add the
 			// path to the slice and return SkipDir since we don't need to
 			// walk any deeper into the directory.
-			if err == nil {
+			if isPkg {
 				packagePaths[path] = true
 				return filepath.SkipDir
 			}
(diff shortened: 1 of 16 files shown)


internal/pkg/pkg.go
```diff
@@ -159,29 +154,63 @@ func (p *Pkg) RelativePathTo(ancestorPkg *Pkg) (string, error) {
...
+
+		// Ignore anything inside the .git folder
+		rel := strings.TrimPrefix(path, pkgPath)
+		if copyutil.IsDotGitFolder(rel) {
+			return nil
+		}
+
+		// Any paths that are inside subpackages can be ignored since
+		// we only want the subpackages directly underneath the package.
+		for dir := range packagePaths {
```
frankfarzan, line 179: Can't you skip walking subpackages using filepath.SkipDir?

You should avoid doing path computations in favor of using just `info`

internal/pkg/pkg.go
```diff
@@ -159,29 +154,63 @@ func (p *Pkg) RelativePathTo(ancestorPkg *Pkg) (string, error) {
...
-		return subPkgs, fmt.Errorf("failed to read sub dirs for %s %w", p, err)
-	}
-	for _, f := range files {
-		if f.IsDir() {
-			dirPath := path.Join(pkgPath, f.Name())
-			isKptPkg, err := kptfileutil.HasKptfile(dirPath)
-			if err != nil {
-				return subPkgs, fmt.Errorf("failed to check kptfile in subpackage %w", err)
+	packagePaths := make(map[string]bool)
+	if err := filepath.Walk(pkgPath, func(path string, info os.FileInfo, err error) error {
```
frankfarzan, line 161: Walk does not follow symlinks.

frankfarzan (changes requested):

internal/pkg/pkg.go
```diff
@@ -159,29 +154,63 @@ func (p *Pkg) RelativePathTo(ancestorPkg *Pkg) (string, error) {
...
+		if err != nil {
+			return fmt.Errorf("failed to read package %s: %w", p, err)
+		}
+
+		// Ignore the root folder
+		if path == pkgPath {
+			return nil
+		}
+
+		// Ignore anything inside the .git folder
```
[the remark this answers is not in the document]
mortent: Yeah, we have an issue for this. We just haven't gotten to it yet.

internal/pkg/pkg.go
```diff
@@ -159,29 +154,63 @@ func (p *Pkg) RelativePathTo(ancestorPkg *Pkg) (string, error) {
 
 // SubPackages returns sub packages of a pkg.
 // A sub directory that has a Kptfile is considered a sub package.
-func (p *Pkg) SubPackages() (subPkgs []*Pkg, err error) {
+func (p *Pkg) SubPackages() ([]*Pkg, error) {
```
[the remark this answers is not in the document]
mortent: Renamed the function to `DirectSubpackages` for now. Not yet sure if we need to list all recursive packages. I will see when I try to phase out the `pkgutil` package if we need more.

internal/pkg/pkg.go
mortent: Added a TODO.

internal/pkg/pkg.go
```diff
@@ -157,31 +151,57 @@ func (p *Pkg) RelativePathTo(ancestorPkg *Pkg) (string, error) {
...
-			isKptPkg, err := kptfileutil.HasKptfile(dirPath)
-			if err != nil {
-				return subPkgs, fmt.Errorf("failed to check kptfile in subpackage %w", err)
+		// Ignore the root folder
+		if path == p.UniquePath.String() {
+			return nil
+		}
+
+		// Ignore anything inside the .git folder
+		if info.Name() == ".git" && info.IsDir() {
```
frankfarzan, line 171: nit move this inside the next if statement
```
if info.isDir() {
  if info.Name() == ".git" {}
  if isPackageDir(..) {}
 }
```

internal/pkg/pkg.go
```diff
@@ -157,31 +151,57 @@ func (p *Pkg) RelativePathTo(ancestorPkg *Pkg) (string, error) {
...
+			return filepath.SkipDir
+		}
+
+		// For every folder, we check if it is a kpt package
+		if info.IsDir() {
+			kfPath := filepath.Join(path, kptfilev1alpha2.KptFileName)
```
frankfarzan, line 177: Extract a function isPackageDir() for this

frankfarzan (commented):
lgtm for Package (I glossed over other files) with minor comments.

internal/pkg/pkg.go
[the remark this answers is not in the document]
frankfarzan: Add a TODO with the issue?

frankfarzan (approved):

Issue #1573, opened 2021-03-19
tutorial docs


frankfarzan:
These docs will replace existing Guides.

Issue #1574, opened 2021-03-19
Site infra: kpt book


frankfarzan:
For kpt v1, we are authoring the guide in a book format. To give a rough idea, see Rust book: https://doc.rust-lang.org/book/title-page.html

Requirements:

- [x] Page previous/next navigation. The reader should be guided sequentially through the book. This is different than the random access for command references we have today.
- [x] Chapter and page numbering. 
- [x] Organizing the `md` files. We should explore the best way to organize the `md` files. There doesn't necessary have to be a 1-to-1 mapping between an `md` and a rendered page (We can do post-processing if it makes sense). Also, need to decide how we name the directories and files as well.

frankfarzan:
I took a crack at organizing the book content in #1585. Note the following:

- The chapter and page numbers are encoded in the directory and file names.
- The directory and file name determines the title of the page. As such, the title is omitted in the `md` files.

frankfarzan:
As part of this, we need to figure out best way to formulate the URL for the rendered pages considering SEO, ability to rename chapter/pages, etc. See comments in #1585

PR #1585, opened 2021-03-19
Add initial outline of the kpt book

frankfarzan:
Rough outline of content for v1 guide as part of #1573


frankfarzan:
@etefera This should give an idea of what we're looking for in #1574

mikebz:
Question: if based on the user feedback we need to re-arrange the chapters what is your vision for updating that?  The URL changing is generally frowned upon for SEO purposes.  Other blogs and stackoverflow might be doing deep links to the chapters.

etefera (approved):
I like the {chapter}/{page} organization in this PR. It lends itself to simple pagination and generation of the sidebar. Speaking of which, the sidebar will be entirely replaced by the outline of the book, correct?

It's good to keep in mind that we have a lot of flexibility in the rendering of these pages and examples when they're siloed off like this as a book. Merging this in to experiment with such rendering for https://github.com/GoogleContainerTools/kpt/issues/1574 sounds good to me.

frankfarzan:
@mikebz You're correct that in general changing the URL is type of a breaking change. I'm not sure there's a better answer besides: try to add chapters and pages (additive) instead of deleting (subtractive). One mitigating factor is remove the page title from the URL:

`kpt.dev/book/01-concepts/01-packages` -> `kpt.dev/book/01/01`

Which allows us to rename the chapters and pages (It should still be about "concept of package" and not a totally different content).

@etefera Something to think about when rendering these pages.

frankfarzan:
> I like the {chapter}/{page} organization in this PR. It lends itself to simple pagination and generation of the sidebar. Speaking of which, the sidebar will be entirely replaced by the outline of the book, correct?

The book doesn't include such content as the command reference, contacts, faq. I think you're asking how does the navbar look when viewing the book? I think we can explore the options, but I think it makes sense to have navbar be only book content but have a backlink to top-level site somehow.




> 
> It's good to keep in mind that we have a lot of flexibility in the rendering of these pages and examples when they're siloed off like this as a book. Merging this in to experiment with such rendering for #1574 sounds good to me.

etefera:
> > I like the {chapter}/{page} organization in this PR. It lends itself to simple pagination and generation of the sidebar. Speaking of which, the sidebar will be entirely replaced by the outline of the book, correct?
> 
> The book doesn't include such content as the command reference, contacts, faq. I think you're asking how does the navbar look when viewing the book? I think we can explore the options, but I think it makes sense to have the navbar show only the book outline but have a backlink to top-level site somehow.
> 
> > It's good to keep in mind that we have a lot of flexibility in the rendering of these pages and examples when they're siloed off like this as a book. Merging this in to experiment with such rendering for #1574 sounds good to me.

Yeah, that was my question. That approach (similar to what's planned for the function catalog) makes sense to me, too.

etefera:
> @mikebz You're correct that in general changing the URL is type of a breaking change. I'm not sure there's a better answer besides: try to add chapters and pages (additive) instead of deleting (subtractive).
> 
> One mitigation may be to remove the page title from the URL:
> 
> `kpt.dev/book/01-concepts/01-packages` -> `kpt.dev/book/01/01`
> 
> Which allows us to rename the chapters and pages (It should still be about "concept of package" and not a totally different content).
> 
> @etefera Something to think about when rendering these pages.

It'd be nice to maintain the relationship between the path of the source markdown and the URL we present to users that has been consistent across the site so far. We should at least have high confidence in the chapter names/concepts for an outline like this. If that's the case, we can have `kpt.dev/book/concepts/01` representing the packages page above as well as the flexibility to change the ordering of the chapters and the content of the pages (at `/site/book/concepts/01.md` in this case) with minimal breakage.

mikebz:
From what I know about Google SEO (which isn't much) I think it's best to keep the title of the page in the page URL.  So for instance if if we have a page for kpt packages it would be best to keep the url as something along the lines of kpt.dev/..../packages

frankfarzan:
Let's follow discussion around URL in #1574

PR #1586, opened 2021-03-20
Add underline/bolding to selected sidebar item.

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1580
image


frankfarzan (commented):
While we're here, can you remove the leading dash:

` - blah` -> `blah`

etefera:
> While we're here, can you remove the leading dash:
> 
> ` - blah` -> `blah`
> 
> It doesn't add much value.

Done:
image

frankfarzan (approved):
Thanks

PR #1589, opened 2021-03-22
teach fn render to prune resources

### LINKED ISSUE #1524: `kpt fn render` should support resource pruning
Expected Behavior:
Pkg `App` has two resources `a.yaml` and `b.yaml` and it has a `pipeline` has one of the mutators that drops the resource `b.yaml`. So if user runs `pipeline run App`, `b.yaml` should be deleted from the local filesystem.

droot:
This PR adds support for pruning of resources in hydration.

fixes #1524 

Note: Error reporting is still not in-place. I will address all error reporting issues related to hydration in one PR after this PR.

### DIFF
--- a/internal/pipeline/executor.go
+++ b/internal/pipeline/executor.go
@@ -16,13 +16,16 @@ package pipeline
 
 import (
 	"fmt"
+	"os"
 	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/GoogleContainerTools/kpt/internal/pkg"
 	kptfilev1alpha2 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1alpha2"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
+	"sigs.k8s.io/kustomize/kyaml/sets"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
@@ -49,25 +52,42 @@ func (e *Executor) Execute() error {
 		return fmt.Errorf("failed to run pipeline in package %s %w", root.pkg, err)
 	}
 
+	if err = trackOutputFiles(hctx); err != nil {
+		return err
+	}
+
 	pkgWriter := &kio.LocalPackageReadWriter{PackagePath: string(root.pkg.UniquePath)}
 	err = pkgWriter.Write(resources)
 	if err != nil {
 		return fmt.Errorf("failed to save resources: %w", err)
 	}
+
+	if err = pruneResources(hctx); err != nil {
+		return err
+	}
+
 	return nil
 }
 
-//
 // hydrationContext contains bits to track state of a package hydration.
 // This is sort of global state that is available to hydration step at
 // each pkg along the hydration walk.
 type hydrationContext struct {
-	// root points to the root of hydration graph where we bagan our hydration journey
+	// root points to the root pkg of hydration graph
 	root *pkgNode
 
 	// pkgs refers to the packages undergoing hydration. pkgs are key'd by their
 	// unique paths.
 	pkgs map[pkg.UniquePath]*pkgNode
+
+	// inputFiles is a set of filepaths containing input resources to the
+	// functions across all the packages during hydration.
+	// The file paths are relative to the root package.
+	inputFiles sets.String
+
+	// outputFiles is a set of filepaths containing output resources. This
+	// will be compared with the inputFiles to identify files be pruned.
+	outputFiles sets.String
 }
 
 //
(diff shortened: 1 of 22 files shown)


internal/pipeline/executor.go
```diff
@@ -281,3 +313,46 @@ func fnChain(pl *kptfilev1alpha2.Pipeline, pkgPath pkg.UniquePath) ([]kio.Filter
...
+		}
+		outputSet.Insert(path)
+	}
+	hctx.outputFiles = outputSet
+	return nil
+}
+
+// pruneResources compares the input and output of the hydration and prunes
+// resources that are no longer present in the output of the hydration.
+func pruneResources(hctx *hydrationContext) error {
```
frankfarzan, line 350: presumably, this functionality exists in current `kpt fn run`. What's the plan for converging this? We need pruning for imperative functions as well.

droot: One thing to note is that during hydration, we read resources as we traverse the pkg hierarchy (and also gather resources by applying pipeline) and write it finally in the end. The read/write flow is very different from imperative run or just pkg level operation. For pruning, you want to track resources at reading time and writing time, so the flow will be different for hydration vs imperative workflow.

For ex, imperative runs etc. uses `kio.LocalPkgReaderWriter` https://github.com/kubernetes-sigs/kustomize/blob/kyaml/v0.10.15/kyaml/kio/pkgio_reader.go#L116 where it tracks the file while `.Read()` and prunes files while `Write(resources)` call. So that is useful at the pkg level, but isn't suited to be used for the hydration workflow.

What do you have in mind when you think of `converging` ?

frankfarzan: I'm not following why declarative and imperative are different when it comes to package io and prunning. There are 3 high level stages:

1.  Get the current set resources across all subpackages. This can be done in different ways. a) Aggregate `inputResources` as part of recursive algorithm b) `Read()` on top-level package with `includeSubpackages`
2. Runs functions as part recursive hydration to calculate the resultant resources across all packages. This is the `outputResources`.
3. Write the `outputResources` to disk across all packages with pruning (diff step 1 and 3).

The only difference between imperative and declarative is step 2, no? Pruning is a concern for package reader/writer (steps 1 and 3), and not the hydration process.  It would be weird to duplicate the logic in step 3.

internal/pipeline/executor.go
```diff
@@ -281,3 +313,46 @@ func fnChain(pl *kptfilev1alpha2.Pipeline, pkgPath pkg.UniquePath) ([]kio.Filter
...
+	resourcesTobePruned := hctx.inputFiles.Difference(hctx.outputFiles)
+	for f := range resourcesTobePruned {
+		if err := os.Remove(filepath.Join(string(hctx.root.pkg.UniquePath), f)); err != nil {
+			return fmt.Errorf("failed to delete file: %w", err)
```
frankfarzan, line 354: Pruning is not file-based by resource-based. In a multi-object files, a file is only deleted if all the resources in that file are deleted.

internal/pipeline/executor.go
droot: For pruning to work, we need to track the `input` resources (A) and `output` resources (B) so that we can compute the difference `A-B` at the end and perform the deletion from FS. Keeping track of `input` as we read resources during `imperative` or `declarative flow` is where we need the right abstraction so that we can de-dupe the logic for pruning. One way to do it will be have some sort of `resourceTracker` that you can provide to the abstractions `pkg.LocalReaderWriter()` or `hydrationContext` so that these methods could call `resourceTracker.Track(record)` as they read resource or feed to the function chains and at the end have some sort of `resourceTracker.Reconcile(output)` to get the difference between A-B and do the pruning.

Would like to have a chat about this. Scheduling 30 mins tomorrow to go over this.

internal/pipeline/executor.go
```diff
@@ -281,3 +313,46 @@ func fnChain(pl *kptfilev1alpha2.Pipeline, pkgPath pkg.UniquePath) ([]kio.Filter
...
+	hctx.outputFiles = outputSet
+	return nil
+}
+
+// pruneResources compares the input and output of the hydration and prunes
+// resources that are no longer present in the output of the hydration.
+func pruneResources(hctx *hydrationContext) error {
+	filesToBeDeleted := hctx.inputFiles.Difference(hctx.outputFiles)
+	for f := range filesToBeDeleted {
+		if err := os.Remove(filepath.Join(string(hctx.root.pkg.UniquePath), f)); err != nil {
```
frankfarzan, line 353: Same discussion as Donny's PR about semantics of path annotation.

internal/pipeline/executor.go
frankfarzan: Can you file an issue to follow up on this. I'm still missing something given current implementation of declarative flow dooesn't reimplement prunning.

droot: Filed an issue here https://github.com/GoogleContainerTools/kpt/issues/1609 to track this conversation.

internal/pipeline/executor.go
droot: We realized we need to address this problem across the codebase (especially hydration that deals with path annotation much more). This code uses the `kioutl.GetPathAnnotation` abstraction to get the path annotation. If this abstraction returns OS-specific path (by doing the conversion from OS-agnostic to OS-specific...eventually), rest of the code will be using OS-specific code as today ('filepath.Join(..)`).

@Shell32-Natsu filed an issue https://github.com/kubernetes-sigs/kustomize/issues/3749 to track this in `kyaml` library.

frankfarzan (approved):

Issue #1604, opened 2021-03-24
Correctly Handle Kpt Dependency License Versions


runewake2:
Kpt runs a [`check`](https://github.com/GoogleContainerTools/kpt/blob/master/scripts/create-licenses.sh) to validate kpt dependency licenses. This check can fail in the case where two dependencies exist:

```
foo/bar/
foo/bar/v2
```

If the `foo/bar/` module is not checked out with `go mod vendor` then the check will fail.

This behavior is currently present in Kpt's `posener` dependency.

runewake2:
Note: #1605 introduced a temporary workaround for this, but we still need a longer term solution for this.

runewake2:
Moving this back to the backlog. We have a temporary workaround in place. We should find a more permanent solution to this but that does not need to happen now. This issue should stick around until the workaround in [create-licenses](https://github.com/GoogleContainerTools/kpt/blob/master/scripts/create-licenses.sh) is removed.

Specifically we need a replacement for this code:

```sh
  if [[ "${PACKAGE}" == "github.com/posener/complete" ]]; then
    continue
  fi
```

PR #1605, opened 2021-03-24
Add workaround for dependency license scan.

runewake2:
This is a workaround for #1604 that adds an explicit exception for `github.com/posener/complete`.

To recreate the error this is addressing:

```sh
go mod vendor
go list -m -json all | grep posener
```

Because the source of v1 of `posener` is not in the `vendor` directory but `posener/v2` exists the check will ignore existing safeguards meant to catch this case. The `LICENSE` does not exist because the dependency doesn't exist.

**Looking for Feedback**: I was not able to find a more elegant way of solving this. If you know of one we will probably want to use that instead.

This is intended as a short term solution to unblock Kpt releases by specifically excluding the `github.com/posener/complete` project from the check.


seans3 (approved):
/lgtm

PR #1613, opened 2021-03-24
Fix for fix of license check

runewake2:
The previous fix was incorrect. This matches the PACKAGE to the exception.
Related to #1604


seans3 (approved):
/lgtm

PR #1623, opened 2021-03-25
book: chapter 1

frankfarzan:


package-examples/cassandra/statefulset.yaml
```diff
@@ -0,0 +1,101 @@
...
+              containerPort: 7199
+            - name: cql
+              containerPort: 9042
+          env:
+            - name: MAX_HEAP_SIZE
+              value: 512M
+            - name: HEAP_NEWSIZE
+              value: 100M
+            - name: CASSANDRA_SEEDS
+              value: cassandra-0.cassandra.default.svc.cluster.local
```
mortent, line 36: Does this example work? I would expect all replicas to be listed here. This is similar to the Zookeeper and Kafka examples.

package-examples/cassandra/statefulset.yaml
```diff
@@ -0,0 +1,101 @@
+apiVersion: apps/v1
```
mortent, line 1: Is this a good example for the intro? It contains quote a bit of yaml, most of which is pretty specific to Cassandra. Would it be better to start with a simpler example, maybe nginx?

package-examples/cassandra/statefulset.yaml
```diff
@@ -0,0 +1,101 @@
...
+        storageClassName: fast
+  serviceName: cassandra
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: fast
+parameters:
+  type: pd-ssd
+provisioner: k8s.io/minikube-hostpath
```
mortent, line 101: Does this work on non-minikube clusters?

site/book/01-getting-started/02-quickstart.md
```diff
@@ -0,0 +1,103 @@
...
+
+First, let's fetch the Cassandra package from Git to your local filesystem:
+
+```shell
+$ kpt pkg get https://github.com/GoogleContainerTools/kpt/package-examples/cassandra@next cassandra
+```
+
+kpt is fully integrated with Git and enables forking, rebasing and versioning a package of
+configuration using the underlying Git version control system. Since kpt is Git-native, any
+existing directory in Git containing Kubernetes resources is also a valid kpt package!
```
mortent, line 16: Might be splitting hairs here, but is it a valid kpt package (since that would imply kpt packages doesn't need to have a Kptfile), or should we frame this as any directory in git with KRM resources can be fetched into a kpt package?

site/book/01-getting-started/02-quickstart.md
```diff
@@ -0,0 +1,103 @@
...
+
+Alternatively, you may want to automatically customize and/or validate a package using _functions_.
+For example, you can set a label with key `env` on all the resources in the package:
+
+```shell
+$ kpt fn eval --image gcr.io/kpt-fn/set-label:v0.1 -- env=dev
+```
+
+This is an imperative one-time operation. It is also possible to declare a function as part of
+pipeline in the `Kptfile`. In this case, the author of a package has already declared a validator
```
runewake2, line 55: Missing word "a": "as part of **a** pipeline".

runewake2 (commented):
Some comments, most are questions/optional.

site/book/01-getting-started/01-installation.md
```diff
@@ -0,0 +1,12 @@
+This book requires kpt **version 1.0 or later**. Follow these steps:
+
+1. [Install the CLI][install].
+   TODO: Clean up installation page.
```
mikebz, line 4: I think we should either have the installation page or this page.

mikebz (approved):
Looks great.  One comment

mikebz:
Not sure if we ever reconciled the URL routing that will allow us to add and remove chapters without breaking URL links. Other numbered content in other languages don't necessarily have the numbering in the URLs: https://docs.python.org/3/using/index.html. - the navigation back and forth can be accomplished by TOC and _next_ and _previous_

My guess is that some of these books potentially have chapters because they believe that the book has gone through enough edits to have stable content.  Not sure if our project is there yet.

site/book/01-getting-started/02-quickstart.md
```diff
@@ -0,0 +1,103 @@
...
+function (`kubeval`) that validates the resources using their OpenAPI schema.
+
+Regardless of the how you choose to customize the package, you need to _render_ the package before
```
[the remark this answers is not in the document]
frankfarzan: Edited. Let me know if it's still ambiguous, open to suggestions how to best phrase this.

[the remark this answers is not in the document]
frankfarzan: Ditto. I'm leaning towards a linear walk through for now, linking can take a user down a depth-first rabbit hole. Don't feel strongly though, can revisit based on fishfooding.

site/book/01-getting-started/02-quickstart.md
```diff
@@ -0,0 +1,103 @@
...
+applying it the cluster:
+
+```shell
+$ kpt fn render
+```
+
+This runs the declared pipeline of functions and writes the fully rendered configuration in-place.
```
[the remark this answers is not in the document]
frankfarzan: remove in-place.

site/book/01-getting-started/02-quickstart.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Apply the Package
```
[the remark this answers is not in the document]
frankfarzan: Added an intro section. I also added as a sentence for introducing each command group. In general, I like us to focus more on the functionality than describing the CLI. For example, explaining what a function is using examples, is more important than talking about `kpt fn` command group.

site/book/01-getting-started/02-quickstart.md
```diff
@@ -0,0 +1,103 @@
...
+$ kpt live preview
+```
+
+Finally, apply the resources to the cluster:
+
+```shell
+$ kpt live apply --reconcile-timeout=15m
+```
+
+This waits for the resources to be reconciled on the cluster by monitoring their status.
```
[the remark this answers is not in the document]
frankfarzan: Good point. I added a TODO for adding the output the CLI to all the commands (pkg, fn, live) once we have a CLI build. The quick start guide will be progressively edited as part of m2 milestone.

site/book/01-getting-started/02-quickstart.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Clean up
+
+Delete the package from the cluster:
+
+```shell
+$ kpt live destroy
+```
+
+Congrats! You should now have a rough idea of what kpt is and what you can do with it.
```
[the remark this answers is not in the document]
frankfarzan: I get the "explain what you're gonna talk about, talk about it, and summarize what you talked about" approach. But I'm leaning towards keeping the quickstart short and sweet.

frankfarzan:
> Not sure if we ever reconciled the URL routing that will allow us to add and remove chapters without breaking URL links. Other numbered content in other languages don't necessarily have the numbering in the URLs: https://docs.python.org/3/using/index.html. - the navigation back and forth can be accomplished by TOC and _next_ and _previous_
> 
> My guess is that some of these books potentially have chapters because they believe that the book has gone through enough edits to have stable content. Not sure if our project is there yet.

Let's track that in a separate `area:site` issue.

site/book/01-getting-started/01-sytem-requirements.md
```diff
@@ -0,0 +1,27 @@
...
+[Install the kpt CLI][install] and ensure you are running **version 1.0 or later**:
+
+```shell
+$ kpt version
+```
+
+TODO: Clean up installation page.
+TODO: kpt 1.0 binaries needs to be available.
+
+## Docker
```
mikebz, line 15: git is also needed, and I am not sure it's a guarantee that it is installed.

mikebz (approved):

mikebz:
Overall I am in favor of merging this because I think more people will give feedback on a fully rendered text via kpt.dev rather than the PR.  Overall the approach to the book content before the release could be that we merge the chapters even with loose ends to make sure we get more eyes on it.  Prior to the release the downside of merging the content is fairly minimal in my opinion.

PR #1629, opened 2021-03-29
Add types for structured function results

### LINKED ISSUE #1627: Define structs for function structured result
Part of #1390

Shell32-Natsu:
close #1627


internal/pipeline/fn/fn_results.go
```diff
@@ -0,0 +1,56 @@
...
+	Exec string `yaml:"exec,omitempty"`
+	// ExitCode is the exit code from running the function
+	ExitCode int `yaml:"exitCode,omitempty"`
+	// Results is the list of results for the function
+	// TODO: The type has been changed to ResultItem in newer version of kyaml
+	Results []framework.Item `yaml:"results,omitempty"`
+}
+
+// FunctionResultList contains aggregated results from multiple functions
+type FunctionResultList struct {
```
frankfarzan, line 49: This is part of the kpt API and should be in `pkg/api`. Follow the ktpfile example.

internal/pipeline/fn/fn_results.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// FunctionResult contains the structured result from an individual function
+type FunctionResult struct {
+	// Image is the full name of the image that generates this result
+	// TODO: Image and Exec are mutually exclusive
```
frankfarzan, line 37: Don't add TODO to the schema, as it only concerns the implementation.

frankfarzan (changes requested):

internal/pipeline/fn/fn_results.go
Shell32-Natsu: Moved the file to `pkg/api/fnresult`

internal/pipeline/fn/fn_results.go
Shell32-Natsu: Removed TODOs

pkg/api/fnresult/types.go
```diff
@@ -0,0 +1,43 @@
...
+	Exec string `yaml:"exec,omitempty"`
+	// ExitCode is the exit code from running the function
+	ExitCode int `yaml:"exitCode,omitempty"`
+	// Results is the list of results for the function
+	Results []framework.Item `yaml:"results,omitempty"`
+}
+
+// FunctionResultList contains aggregated results from multiple functions
+type FunctionResultList struct {
+	// The GVK must be:
```
frankfarzan, line 37: kptfile defines constant for these, follow that pattern

pkg/api/fnresult/types.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// FunctionResult contains the structured result from an individual function
+type FunctionResult struct {
+	// Image is the full name of the image that generates this result
+	// mage and Exec are mutually exclusive
```
frankfarzan, line 25: typo

Shell32-Natsu: fixed

pkg/api/fnresult/types.go
Shell32-Natsu: Added constants for GVK

pkg/api/fnresult/types.go
```diff
@@ -0,0 +1,47 @@
...
+	// ExitCode is the exit code from running the function
+	ExitCode int `yaml:"exitCode,omitempty"`
+	// Results is the list of results for the function
+	Results []framework.Item `yaml:"results,omitempty"`
+}
+
+const (
+	FunctionResultListKind       = "FunctionResultList"
+	FunctionResultListGroup      = "config.kubernetes.io"
+	FunctionResultListVersion    = "v1beta1"
```
frankfarzan, line 38: why not v1alpha2?

Also follow the directory convention e.g.

kpt/pkg/api/kptfile/v1alpha2/types.go

Shell32-Natsu: > why not v1alpha2?

I cannot find a doc about what apiVersion should be used for List kinds so I copied the apiVersion from the KRM function spec. Does it mean the apiVersion for List kinds is not defined and we can choose what we want?

> Also follow the directory convention e.g.
> kpt/pkg/api/kptfile/v1alpha2/types.go

I didn't follow the pattern since I was not sure should we bind the function result list version with Kptfile version. Should we do this?

frankfarzan: You may be conflating two different things (I commented about this before):

1. The wire format as described in the Function Spec Standard.
2. The KRM resource outputted by the kpt CLI using --results flag. This is kpt-specific.

I assumed you're defining 2 here, not 1. These two do not necessarily need to be coupled to each other. `Results` field in `FunctonResult` is currently defiend as `framework.Item` for convinience, but they don't
have to be the same.

Shell32-Natsu: Update the api version to v1alpha1

pkg/api/fnresult/v1alpha1/types.go
```diff
@@ -0,0 +1,51 @@
...
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// Result contains the structured result from an individual function
+type Result struct {
+	// Image is the full name of the image that generates this result
+	// Image and Exec are mutually exclusive
+	Image string `yaml:"image,omitempty"`
+	// ExecPath is the the absolute path to the executable file
```
frankfarzan, line 27: wha type of path?

pkg/api/fnresult/types.go
frankfarzan: Why?

As mentioned previously, this is a kpt-specific kind, so should use the same group and version as Kptfile: https://github.com/GoogleContainerTools/kpt/blob/next/pkg/api/kptfile/v1alpha2/types.go#L26

pkg/api/fnresult/v1alpha1/types.go
Shell32-Natsu: It's an OS specific path. Updated the comment

pkg/api/fnresult/types.go
Shell32-Natsu: Use the same group and version with Kptfile now.

frankfarzan (approved):

Issue #1557, opened 2021-03-17
`kpt fn` error reporting and CLI output


droot:
`kpt fn render` should report errors with useful diagnostic info that helps the end-users as well as kpt developers. We should explore if we `kpt` can support error reporting explained in this [post](https://commandcenter.blogspot.com/2017/12/error-handling-in-upspin.html) by `Rob Pike`.  The approach in that post tries to strike the right balance with needs of both the personas.

In addition, being able to trace `input/output` for each function execution of the `pipeline` can assist in function level debugging as well.

droot:
This is addressed with #1634

frankfarzan:
Reopening this since we need to hash out the UX for error reporting and CLI output.

A list of all the data involved in running functions:

1. What function provides:
    1. ResourceList.results
    2. ResourceList.items
    3. stderr
    4. exit code
2. Orchestrator provides metadata:
    1. What function is running (Specially important when running a pipeline)

We need to figure out the best UX for providing all or subset of the following to the user across:

1. CLI output
    2. stdout
    3. stderr
4. FucntionResults files 

As part of this, need to consider happy paths and failure paths:

1. Happy path: The pipeline ran successfully
2. Failure paths:
    1. Function exceptional failures (e.g. function segfaulted)
    2. Validator failures which are WAI (kubeval found an issue)
    3. Orchestrator failed for some reason (e.g. IO failures or user input issues)

Some principles to follow:
- Don’t pollute the CLI output unnecessarily (Especially in the happy path)
- Provide realtime information about what’s happening, especially for long-running operations. The CLI should not appear stuck for long periods of time with no progress.
- Follow and add to the style guide: go/kpt-errors-style-guide

frankfarzan:
#1783 should be addressed as part of this.

frankfarzan:
As part this, need to add e2e test cases for both `render` and `eval`: go/kpt-test-cases

mortent:
Related to this, the current output while a pipeline is running is a little confusing. It prints
```
package ".": running function "gcr.io/kpt-fn/kubeval:v0.1": 
```
with the caret blinking at the end of the line. It looks like we expect some kind of user input.

frankfarzan:
Yes, I want to see a design doc iterating on the UI here. There are good prior art for CLI showing task progress.

natasha41575:
Adding that we should address the concerns raised here:
[...]

PR #1634, opened 2021-03-30
improve error handling in `fn render` command

droot:
This is an experimental PR to try out a new pattern for error handling (inspired by https://commandcenter.blogspot.com/2017/12/error-handling-in-upspin.html)

This PR:

- Adds a printer to print output to CLI.
- Introduces ctx parameter of type `context.Context` that makes it easy to make dependencies such as printer available in rest of the call stack
- Improve the error handling for `fn render`
- Prints basic progress for `fn render`

Note to reviewers: I think this is still a starting point that needs to be iterated upon, but much better than where we are today in error/progress reporting for `fn render`.

### DIFF
--- a/internal/cmdrender/executor.go
+++ b/internal/cmdrender/executor.go
@@ -15,13 +15,16 @@
 package cmdrender
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
 
+	"github.com/GoogleContainerTools/kpt/internal/errors"
 	"github.com/GoogleContainerTools/kpt/internal/pkg"
+	"github.com/GoogleContainerTools/kpt/internal/printer"
 	"github.com/GoogleContainerTools/kpt/internal/types"
 	kptfilev1alpha2 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1alpha2"
 	"sigs.k8s.io/kustomize/kyaml/kio"
@@ -33,13 +36,17 @@ import (
 // Executor hydrates a given pkg.
 type Executor struct {
 	PkgPath string
+	Printer printer.Printer
 }
 
 // Execute runs a pipeline.
-func (e *Executor) Execute() error {
+func (e *Executor) Execute(ctx context.Context) error {
+	op := errors.Op("fn.render")
+	pr := printer.FromContext(ctx)
+
 	root, err := newPkgNode(e.PkgPath, nil)
 	if err != nil {
-		return err
+		return errors.E(op, types.UniquePath(e.PkgPath), err)
 	}
 
 	// initialize hydration context
@@ -48,9 +55,9 @@ func (e *Executor) Execute() error {
 		pkgs: map[types.UniquePath]*pkgNode{},
 	}
 
-	resources, err := hydrate(root, hctx)
+	resources, err := hydrate(ctx, root, hctx)
 	if err != nil {
-		return fmt.Errorf("failed to run pipeline in package %s %w", root.pkg, err)
+		return errors.E(op, root.pkg.UniquePath, err)
 	}
 
 	if err = trackOutputFiles(hctx); err != nil {
@@ -67,6 +74,7 @@ func (e *Executor) Execute() error {
 		return err
 	}
 
+	pr.PkgPrintf(root.pkg.UniquePath, "rendered successfully\n")
 	return nil
 }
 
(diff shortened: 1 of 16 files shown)


droot:
This will address #1557

internal/cmdrender/executor.go
```diff
@@ -33,13 +36,17 @@ import (
 // Executor hydrates a given pkg.
 type Executor struct {
 	PkgPath string
+	Printer printer.Printer
 }
 
 // Execute runs a pipeline.
-func (e *Executor) Execute() error {
+func (e *Executor) Execute(ctx context.Context) error {
+	op := errors.Op("fn.render")
```
mortent, line 44: So I think we should standardize on the pattern used in upspin, which is to create these as constants: 
```
const op errors.Op = "fn.render"
```

I feel this makes it stand out from the rest of the code in the function and makes it clear that this is a constant.

internal/cmdrender/runtime/container.go
```diff
@@ -41,6 +44,8 @@ type ContainerFnPermission struct {
 // ContainerFn implements a KRMFn which run a containerized
 // KRM function
 type ContainerFn struct {
+	Ctx  context.Context
```
mortent, line 47: I think it would make more sense here to pass the context into the `Run` function.

internal/pkg/pkg.go
```diff
@@ -22,12 +22,14 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/GoogleContainerTools/kpt/internal/types"
 	kptfilev1alpha2 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1alpha2"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 	"sigs.k8s.io/kustomize/kyaml/sets"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
+
```
mortent, line 30: nit: Why space here?

internal/printer/printer.go
```diff
@@ -0,0 +1,81 @@
...
+// Helper functions to set and retrieve printer instance from a context.
+// Defining them here avoids the context key collision.
+
+// FromContext returns associated printer instance.
+func FromContext(ctx context.Context) Printer {
+	pr, ok := ctx.Value(printerKey).(*printer)
+	if ok {
+		return pr
+	}
+	return nil
```
mortent, line 74: I think we should just panic here instead of returning nil and call this `FromContextOrDie`. We don't really want to check for nil here every time we use this function. And a printer not being available on the context would be a programmer error.

internal/types/types.go
```diff
@@ -23,6 +29,22 @@ func (u UniquePath) String() string {
 	return string(u)
 }
 
+// RelativePath returns the relative path to current working directory.
+func (u UniquePath) RelativePath() string {
+	cwd, err := os.Getwd()
+	if err != nil {
+		return string(u)
```
mortent, line 36: Why not return error here?

internal/util/cmdutil/cmdutil.go
```diff
@@ -88,14 +89,17 @@ func ResolveAbsAndRelPaths(path string) (string, string, error) {
...
 	cmd := exec.Command("docker", "version")
-	var stderr bytes.Buffer
-	cmd.Stderr = &stderr
+	cmd.Stderr = buffer
 	err := cmd.Run()
 	if err != nil {
-		return fmt.Errorf(stderr.String() +
-			`Docker is required to run this command. 
-To install docker, follow the instructions at https://docs.docker.com/get-docker/`)
+		return errors.E(op, suggestedText, fmt.Errorf("%s", suggestedText))
```
mortent, line 102: This looks a bit strange. The way `errors.E` is currently implemented, we will drop either the error or the string.

internal/cmdrender/executor.go
droot: Good point. Done.

internal/cmdrender/runtime/container.go
droot: Yes, you are right. But in this case `Run` method is part of an interface that exists in `kyaml` library, so had to go with this.

internal/pkg/pkg.go
droot: fixed.

internal/printer/printer.go
droot: Makes sense. Done.

internal/types/types.go
droot: Done.

internal/util/cmdutil/cmdutil.go
droot: I was playing with `UserText` that I took it out and this is a residue from that iteration. Thanks for catching this. cleaned up.

droot (commented):
Thank you @mortent for the feedback. Addressed the feedback in a separate commit.

mortent (approved):

PR #1637, opened 2021-03-30
Add make command for sidebar

etefera:
Adds the make command as a build step before checking/deploying the site.
Also automatically renames 00.md special pages to README to avoid having the number in the URL.


Makefile
```diff
@@ -93,6 +93,10 @@ site-run-server:
 site-check:
 	./scripts/check-site.sh
 
+site-gen-sidebar:
```
frankfarzan, line 96: We shouldn't keep adding top-level make commands. What's the set of steps a developer needs to do locally to render the site?

There should be a single top-level command for the site.

scripts/generate_site_sidebar/generate_site_sidebar.go
```diff
@@ -0,0 +1,201 @@
...
+
+func getPage(pageFileName string, defaultName string, parentPath string) page {
+	// Split into page number and hyphenated name.
+	splitPageName := strings.SplitN(pageFileName, "-", 2)
+
+	pageName := defaultName
+	if pageFileName != introPage {
+		// Strip page number and extension from file name.
+		pageTitle := regexp.MustCompile(`^\d\d-?`).ReplaceAll([]byte(pageFileName), []byte(""))
+		pageName = titlecase.Title(strings.ReplaceAll(strings.ReplaceAll(string(pageTitle), ".md", ""), "-", " "))
```
frankfarzan, line 162: so is this logic duplicated in javascript and Go now? If you're preprocessing in Go, why do you need the runtime Javascript logic?

frankfarzan (changes requested):

.github/workflows/checkSite.yml
```diff
@@ -35,6 +35,11 @@ jobs:
         working-directory: href-checker
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
+      - name: Generate Sidebar
+        run: make site-gen-sidebar
+      - name: Normalize Book Introduction Pages
+        working-directory: site
+        run: find . -iname "00.md" -execdir mv {} README.md \; && sed -i s/00.md//g sidebar.md
```
frankfarzan, line 42: this kind of processing should be encapsulated in the make target. Otherwise, you have leaky abstractions.

scripts/generate_site_sidebar/generate_site_sidebar.go
etefera: Artifact of performing the logic at runtime first. Removed in favor of the preprocessing in Go.

.github/workflows/checkSite.yml
etefera: Consolidated into the make target.

Makefile
etefera: Keeping the site generation target separate for use with different methods of building the site, but added it to the site running target so a developer only needs to run `make site-run-server`.

etefera:
Demo is at https://kpt-dev-ee2a9--sidebar-generation-f9wtc6gd.web.app/book/

scripts/generate_site_sidebar/generate_site_sidebar.go
```diff
@@ -0,0 +1,202 @@
...
+
+	"github.com/igorsobreira/titlecase"
+)
+
+const markdownExtension = ".md"
+const introPage = "00.md"
+
+func main() {
+	fmt.Print(`<!--This sidebar was automatically generated by 'make site-generate'-->
+- [Installation](installation/)
```
frankfarzan, line 34: Adding/removing content should not require compiling a Go program.

You can put this in a md file and use Go templating to process it.

Makefile
```diff
@@ -87,7 +87,12 @@ gencatalog:
 	rm site/content/en/guides/consumer/function/catalog/*/_index.md
 	(cd site/content/en/guides/consumer/function/catalog/catalog && npm run gen-docs)
 
+site-generate:
+	go run ./scripts/generate_site_sidebar > site/sidebar.md
+	(cd site && find . -iname "00.md" -execdir mv {} README.md \; && sed -i s/00.md//g sidebar.md)
```
frankfarzan, line 92: Can this be encapsulated in the go program? Currently, how content is arranged sprinkled across Go, Bash, and Javascript.

scripts/generate_site_sidebar/generate_site_sidebar.go
etefera: Switched to Go templating.

Makefile
etefera: I don't think it would make much sense putting the renaming of `00.md` files to `README.md` in the go program. Ideally, that line would only be executed before deployment; perhaps that normalization should be added specifically as as a `run` line in the `firebase-hosting-merge.yml` action.

frankfarzan (approved):

frankfarzan:
I'm getting this:

```
» make site-run-server
make site-generate
go run ./scripts/generate_site_sidebar > site/sidebar.md
(cd site && find . -iname "00.md" -execdir mv {} README.md \; && sed -i s/00.md//g sidebar.md)
sed: 1: "sidebar.md": unterminated substitute pattern
make[1]: *** [site-generate] Error 1
make: *** [site-run-server] Error 2
```

etefera:
> (cd site && find . -iname "00.md" -execdir mv {} README.md \; && sed -i s/00.md//g sidebar.md)

Addressed in https://github.com/GoogleContainerTools/kpt/pull/1716

PR #1639, opened 2021-03-30
add errors pkg

droot:
This error implements `errors` type that will be used for error handling in the kpt codebase. This work is inspired by the design mentioned in https://commandcenter.blogspot.com/2017/12/error-handling-in-upspin.html

Here is a PR https://github.com/GoogleContainerTools/kpt/pull/1634 that uses this error pkg to improve error handling in `kpt fn render`.

Here is another https://github.com/GoogleContainerTools/kpt/pull/1602 that uses this error pkg in packaging commands.

### DIFF
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -22,18 +22,23 @@ import (
 	"github.com/GoogleContainerTools/kpt/internal/types"
 )
 
-// Error is an implementation of the error interface used in the kpt
-// codebase.
+// Error is the type that implements error interface used in the kpt codebase.
 // It is based on the design in https://commandcenter.blogspot.com/2017/12/error-handling-in-upspin.html
+// The intent is to capture error information in a structured format so
+// that we can display it differently to different users for ex. kpt developers
+// are interested in operational trace along with more diagnostic information while
+// kpt end-users may be just interested in a concise and actionable information.
+// Representing errors in structured format helps us decouple the error information
+// from how it is surfaced to the end users.
 type Error struct {
-	// Path is the path name of the object involved in Kpt operation.
+	// Path is the path of the object (pkg, file) involved in kpt operation.
 	Path types.UniquePath
 
 	// Op is the operation being performed, for ex. pkg.get, fn.render
 	Op Op
 
-	// Kind refers to classs of errors
-	Kind Kind
+	// Class refers to class of errors
+	Class Class
 
 	// Err refers to wrapped error (if any)
 	Err error
@@ -53,9 +58,9 @@ func (e *Error) Error() string {
 		b.WriteString(string(e.Path))
 	}
 
-	if e.Kind != 0 {
+	if e.Class != 0 {
 		pad(b, ": ")
-		b.WriteString(e.Kind.String())
+		b.WriteString(e.Class.String())
 	}
 
 	if e.Err != nil {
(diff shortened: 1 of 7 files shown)


internal/errors/errors.go
```diff
@@ -0,0 +1,170 @@
...
+		case error:
+			e.Err = a
+		case string:
+			e.Err = fmt.Errorf(a)
+		default:
+			panic(fmt.Errorf("unknown type %T for value %v in call to error.E", a, a))
+		}
+	}
+
+	wrappedErr, ok := e.Err.(*Error)
```
mortent, line 152: Since we are making this check on the concrete type, it isn't possible to create custom error types and make them behave as this error type. We can consider adding an interface here if we need it.

mortent (commented):
I think this is a good starting point. We should also consider how we want the `Is` function to behave, ref the implementation in Upspin: https://upspin.googlesource.com/upspin/+/5ddde7b8e6ff/errors/errors.go#484

internal/errors/errors.go
droot: Lifting (and suppressing) duplicate `path`, `op` or `kind` will require making it interface. We can add when we have concrete understanding of the behavior we want as we see more usage in the field.

internal/errors/errors.go
```diff
@@ -0,0 +1,170 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+// Error is an implementation of the error interface used in the kpt
+// codebase.
+// It is based on the design in https://commandcenter.blogspot.com/2017/12/error-handling-in-upspin.html
+type Error struct {
+	// Path is the path name of the object involved in Kpt operation.
```
frankfarzan, line 29: nit: it's lower-case `kpt`, both in code and docs.

frankfarzan, line 29: What's an "object" in this context?

internal/errors/errors.go
```diff
@@ -0,0 +1,170 @@
...
+	Path Path
+
+	// Op is the operation being performed, for ex. pkg.get, fn.render
+	Op Op
+
+	// Kind refers to classs of errors
+	Kind Kind
```
frankfarzan, line 36: Avoid `Kind` given it's overloaded in KRM

internal/errors/errors.go
```diff
@@ -0,0 +1,170 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Error is an implementation of the error interface used in the kpt
+// codebase.
+// It is based on the design in https://commandcenter.blogspot.com/2017/12/error-handling-in-upspin.html
+type Error struct {
```
frankfarzan, line 28: Maybe comment about the intended usage of this abstraction. Will this be shown the end-user of the CLI? Is it to facilitate better error handling/recovery as an implementation detail?

internal/errors/errors.go
```diff
@@ -0,0 +1,170 @@
...
+type Op string
+
+// Kind describes the class of errors encountered.
+type Kind int
+
+const (
+	Other        Kind = iota // Unclassified. Will not be printed.
+	Exist                    // Item already exists.
+	Internal                 // Internal error.
+	InvalidParam             // Value is not valid.
```
frankfarzan, line 100: These error types need to be well-documented in order for them to be used consistently.

What's a param? Is it user-facing or does this capture programmer error?
Or is programmer error "Internal"?

To disambiguate, provide a concrete example of each.

frankfarzan (commented):
I like the direction here.

internal/errors/errors.go
droot: Added some more context about the intention here. I would have liked to be more concrete here, but we haven't formed the metal model fully yet.

internal/errors/errors.go
droot: Done.

droot: Added examples to help here.

internal/errors/errors.go
droot: Renamed it to `class` for now.

internal/errors/errors.go
droot: I haven't really used this field much yet in hydration errors, @mortent is experimenting with it in packaging errors. I haven't formed a mental model yet about it. I can document it better once we are happy with it's usage (somewhere around m2 milestone).

frankfarzan (approved):
Thanks!

mortent (approved):

PR #1667, opened 2021-04-01
Don't refresh the whole page on pagination element click.

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1651
Demo is at https://kpt-dev-ee2a9--pagination-0mbqx470.web.app


frankfarzan (approved):
Much better.

PR #1674, opened 2021-04-02
Update idempotent test in e2e driver

Shell32-Natsu:
https://github.com/GoogleContainerTools/kpt/pull/1646#discussion_r605824173

Use `nonIdempotent` instead of `idempotent` because we cannot distinguish the `false` from explicit config and omitting it in config file.

### DIFF
--- a/pkg/test/runner/config.go
+++ b/pkg/test/runner/config.go
@@ -44,7 +44,8 @@ type EvalTestCaseConfig struct {
 type TestCaseConfig struct {
 	// ExitCode is the expected exit code from the kpt commands. Default: 0
 	ExitCode int `json:"exitCode,omitempty" yaml:"exitCode,omitempty"`
-	// NonIdempotent asks to not test idempotency of this test case. Default: false
+	// NonIdempotent indicates if the test case is not idempotent.
+	// By default, tests are assumed to be idempotent, so it defaults to false.
 	NonIdempotent bool `json:"nonIdempotent,omitempty" yaml:"nonIdempotent,omitempty"`
 	// Skip means should this test case be skipped. Default: false
 	Skip bool `json:"skip,omitempty" yaml:"skip,omitempty"`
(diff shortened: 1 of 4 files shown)


pkg/test/runner/config.go
```diff
@@ -44,9 +44,8 @@ type EvalTestCaseConfig struct {
 type TestCaseConfig struct {
 	// ExitCode is the expected exit code from the kpt commands. Default: 0
 	ExitCode int `json:"exitCode,omitempty" yaml:"exitCode,omitempty"`
-	// RunCount is how many times the test will run. Each iteration will be
-	// immediately run after previous one. Default: 1
-	RunCount int `json:"runCount,omitempty" yaml:"runCount,omitempty"`
+	// NonIdempotent asks to not test idempotency of this test case. Default: false
```
droot, line 47: Suggestion for comment:
```
// NonIdempotent indicates if the test case is not idempotent. By default, tests are assumed to be idempotent,
// so it defaults to false.
```

droot (approved):
looks good to me, minor suggest about the comment.

pkg/test/runner/runner.go
```diff
@@ -128,7 +128,7 @@ func (r *Runner) runFnEval() error {
 	var output string
 	var fnErr error
 	command := run.GetMain()
-	for i := 0; i < r.testCase.Config.RunCount; i++ {
+	for i := 0; i < r.testCase.Config.RunCount(); i++ {
```
frankfarzan, line 131: Should the expected results be tested after each iteration instead of at the end?

Example:

replica = 0
Function: replica++
Expected: replica = 2

This function is not idempotent.

pkg/test/runner/config.go
Shell32-Natsu: updated.

pkg/test/runner/runner.go
Shell32-Natsu: Update the implementation and now the result will be checked after each iteration.

Shell32-Natsu:
Will merge this and resolve the comments in new PR if any.

PR #1683, opened 2021-04-02
[DO NOT MERGE]add actions for macOS and Windows

Shell32-Natsu:
Run GitHub actions on windows, macos and linux.

This PR is used to demonstrate the issues we need to resolve to pass the macOS and Windows tests.

### DIFF
--- a/.github/workflows/go_linux.yml
+++ b/.github/workflows/go_linux.yml
@@ -1,0 +1,43 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


runewake2:
I think we could solve this with a matrix of the OS instead of creating multiple workflow files. More info and an example [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix).

Shell32-Natsu:
@runewake2 Thanks. Actually I create this PR to show the issues that we need to resolve to pass the windows and macos test. Should marked as do not merge.

PR #1685, opened 2021-04-05
Preview docs with no caching

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1680
Also normalizes docs previewing for site checking between local dev environment and GHA.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ gencatalog:
 	(cd site/content/en/guides/consumer/function/catalog/catalog && npm run gen-docs)
 
 site-run-server:
-	chmod o+rx -R site/
+	setfacl -Rd -m o::rx site/ && chmod o+rx -R site/
 	docker stop $$(docker ps -q --filter ancestor=kpt-site:latest) || docker build site/ -t kpt-site:latest
 	docker run -v `pwd`/site:/usr/share/nginx/html -p 3000:80 -d kpt-site:latest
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -92,9 +92,12 @@ gencatalog:
 	(cd site/content/en/guides/consumer/function/catalog/catalog && npm run gen-docs)
 
 site-run-server:
-	npx http-server site/ -p 3000
+	setfacl -Rd -m o::rx site/ && chmod o+rx -R site/
```
frankfarzan, line 95: this is a linux-specific command. Our developer infra should work on at least mac.

etefera: Added `chmod -R +a ...` for mac.

mikebz:
```
 ~/src/kpt » setfacl
zsh: command not found: setfacl
```
?

etefera: Yes, the `chmod` flags for ACLs don't exist on Linux. @mikebz That would be expected on Mac but the ACL would still be set due to the second half of the line. Since it's expected that one of the two commands will complete, suppressed the error of `setfacl` not existing so there's no error output on Mac.

mikebz: My question is what is the advantage of that command given that you are doing `o::rx` with it, basically doing a chmod anyway...

etefera: The advantage is to allow files created while the docker image is up to be picked up by the running site instance

frankfarzan (approved):

Issue #1671, opened 2021-04-02
Crashes with index out of range error


ashlineldridge:
### Expected behavior

Invalid command line invocations should produce appropriate error messages and not stack traces.

### Actual behavior

It's been a while between `kpt cfg set` calls and I ran the command wrong which produced the stack trace shown below.

```
> kpt cfg set account-id abc

panic: runtime error: index out of range [2] with length 2

goroutine 1 [running]:
sigs.k8s.io/kustomize/cmd/config/internal/commands.(*SetRunner).preRunE(0xc000970000, 0xc00096cb00, 0xc00093cd80, 0x2, 0x2, 0x0, 0x0)
	sigs.k8s.io/kustomize/cmd/config@v0.9.5/internal/commands/cmdset.go:78 +0x331
github.com/spf13/cobra.(*Command).execute(0xc00096cb00, 0xc00093cd40, 0x2, 0x2, 0xc00096cb00, 0xc00093cd40)
	github.com/spf13/cobra@v1.0.0/command.go:831 +0x523
github.com/spf13/cobra.(*Command).ExecuteC(0xc00094e000, 0x200000003, 0xc000000180, 0xc0007d7f78)
	github.com/spf13/cobra@v1.0.0/command.go:950 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
	github.com/spf13/cobra@v1.0.0/command.go:887
main.main()
	github.com/GoogleContainerTools/kpt/main.go:37 +0x65
```

### Information

Kpt version: 0.38.1

mikebz:
dupe of #1593

@phanimarupaka do we have a patch release of this?   I thought this was in .38.1
@runewake2 do you know if you grabbed the dependency update for this?

runewake2:
I believe we'd need a new `cmd/config` release to pick up the fix for this.

mikebz:
@runewake2 can you spin one up and create .38.2?

runewake2:
We could build `v0.39.1` after `cmd/config` is released. I am not able to perform releases in Kustomize though.

runewake2:
The fix for this should be available in version [`v0.39.1`](https://github.com/GoogleContainerTools/kpt/releases/tag/v0.39.1) of kpt.

PR #1686, opened 2021-04-05
Update cmd/config dependency

runewake2:
Pull in the updated cmd/config release (`v0.9.9`).

This should include the fix for #1671


mikebz:
LGTM but @frankfarzan , @mortent or @droot should do the final approval

mortent (approved):

PR #1708, opened 2021-04-07
Update CONTRIBUTING.md

etefera:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,7 +95,7 @@ in VSCode.
 This includes:
 
 - Lint docs with markdownlint to standardize them and make them easier to
-  update. `make lintdocs`
+  update.
 - Run the kpt website through the [W3 Link Checker] in recursive mode and fix
   warnings and errors.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,26 +54,22 @@ code changes and the commit message should start with `Docs:`.
 
 ### Run the docs locally
 
-Make docs changes and test them by using hugo to run the kpt site locally. Hugo
-continuously builds your website as you make changes.
-
-- install hugo
-- `make servedocs`
+Make docs changes and test them by running the site in a docker container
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -54,26 +54,22 @@ code changes and the commit message should start with `Docs:`.
 
 ### Run the docs locally
 
-Make docs changes and test them by using hugo to run the kpt site locally. Hugo
-continuously builds your website as you make changes.
-
-- install hugo
-- `make servedocs`
+Make docs changes and test them by running the site in a docker container
```
mikebz, line 57: why docker?

CONTRIBUTING.md
```diff
@@ -54,26 +54,22 @@ code changes and the commit message should start with `Docs:`.
...
 - Broken links
 - Rendering static content
-- Ordering of new pages using the "weight" key in the header
 
 ### Update docs
 
-Docs are under [site/] and use the [docsy] theme for hugo. Learn more about
-docsy from [docsy docs]. The site itself are generated from the markdown
-automatically after commit.
+Docs are under [site/] and use [docsify] to present the source markdown files.
```
mikebz, line 67: is docsify supposed to be a link?

etefera: Yes, a link to the docsify documentation.

CONTRIBUTING.md
etefera: We already have workflows that utilize `docker` and it allows for the most similar dev experience to the live site without relying directly on `firebase`.

mikebz (approved):

CONTRIBUTING.md
frankfarzan, line 67: Comments about the lines below:

- [ ] `make lintdocs`: this doesn't work.
- [ ] Let's remove the section about Omnigraffle. We're going to remove all these images because they're not readable since they pack too much in one image, and they're also stale.

PR #1712, opened 2021-04-07
add tag check to validator

### LINKED ISSUE #1709: pipeline validator doesn't accept function name with tag
https://github.com/GoogleContainerTools/kpt/blob/8b5daec4b33ca99d3618ce80db9ab68a8921c379/pkg/api/kptfile/v1alpha2/validation.go#L95-L109

Needs to add tag check.

Shell32-Natsu:
close #1709


droot (approved):
Thanks for quick turnaround :)

PR #1727, opened 2021-04-09
bump up go-yaml library version

droot:
This PR updates the `go-yaml` version to latest version (Jan, 2021). The latest version of `go-yaml` has fixes related to roundtripping of comment in the YAML docs and it also introduces a formatting change (indentation) for items part of a sequence (list). Items in the list are indented by two spaces relative to parent (as with other types fields).

Refer to [commit](https://github.com/go-yaml/yaml/commit/a5ece683394c3b88d90572e44d36c93aea492c2c) for more details on the formatting and roundtripping fixes.


e2e/testdata/fn-render/generator/db/Kptfile
```diff
@@ -4,7 +4,7 @@ metadata:
   name: db
 pipeline:
   mutators:
-  - image: gcr.io/kpt-fn/starlark:unstable
+  - image: gcr.io/sunilarora-sandbox/starlark-fn:0.1
```
droot, line 7: I will update it to point to `gcr.io/kpt-fn/...` once it is there.

e2e/testdata/fn-render/resource-has-pkgname-prefix/Kptfile
```diff
@@ -4,6 +4,6 @@ metadata:
   name: wordpress
 pipeline:
   mutators:
-  - image: gcr.io/kpt-fn/set-annotation:unstable
+  - image: gcr.io/sunilarora-sandbox/set-annotation:0.1
```
droot, line 7: I will update to the one from `gcr.io/kpt-fn/...` once it is there, using it for testing for now.

e2e/testdata/fn-render/resource-has-pkgname-prefix/wordpress-deployment.yaml
```diff
@@ -7,7 +8,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-apiVersion: v1 # Copyright 2019 Google LLC
+apiVersion: v1
```
droot, line 11: We can see the roundtripping fixes working here in this example.

e2e/testdata/fn-render/resource-has-pkgname-prefix/.expected/diff.patch
```diff
@@ -83,32 +84,47 @@ index cbbe597..6b307e8 100644
...
-index ee54311..60f1d52 100644
+index d762212..6837328 100644
 --- a/wordpress-deployment.yaml
 +++ b/wordpress-deployment.yaml
-@@ -10,12 +10,13 @@
- apiVersion: v1 # Copyright 2019 Google LLC
+@@ -8,18 +8,19 @@
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
```
droot, line 125: roundtripping fixes in action here. Note `Copyright 2019 Google....` is no longer appearing here right next to the apiVersion field.

mortent:
This looks good to me. But could we also include in the PR description the ways this change affects formatting? It looks like it only affects indentation for lists, but it would be great to have it listed.

droot:
> This looks good to me. But could we also include in the PR description the ways this change affects formatting? It looks like it only affects indentation for lists, but it would be great to have it listed.

Updated the description and also added the link to the commit in `go-yaml` repo. Good thing, all e2e tests are also passing that are using newer version of all `kpt-functions` that are also updated to latest version of `go-yaml`. So this PR is good to go.

.golangci.yml
```diff
@@ -29,7 +29,9 @@ linters:
     - govet
     - ineffassign
     - interfacer
-    - lll
+# lll fails for some tests (thirdparty/cmdconfig/cmdsource/_test.go)
+# Figure out how to exclude files with issues.
+#    - lll
```
mortent, line 34: I think we should disable this rule in the specific places where we have violations. The syntax is:
```
//nolint:lll
```

mortent:
I just found another issue in the current version of the yaml library that are fixed in the updated version. 
Marshalling of properties of type `yaml.Node` (as we are relying on [here](https://github.com/GoogleContainerTools/kpt/blob/722296a50a8d472a049efeed38f980852b290841/pkg/api/kptfile/v1alpha2/types.go#L242)) does not work with the current version of the library.

e2e/testdata/fn-render/generator/db/Kptfile
droot: Done.

e2e/testdata/fn-render/resource-has-pkgname-prefix/Kptfile
droot: Done.

.golangci.yml
droot: Good tip: fixed it. Thanks

droot:
> I just found another issue in the current version of the yaml library that are fixed in the updated version.
> Marshalling of properties of type `yaml.Node` (as we are relying on [here](https://github.com/GoogleContainerTools/kpt/blob/722296a50a8d472a049efeed38f980852b290841/pkg/api/kptfile/v1alpha2/types.go#L242)) does not work with the current version of the library.

That's great.

Shell32-Natsu (approved):
lgtm, but please wait for approval from @mortent

Shell32-Natsu:
I assume all changes are test results update right?

mortent (approved):

droot:
> I assume all changes are test results update right?

Yes.

PR #1730, opened 2021-04-09
Better error handling for get and update

mortent:
This updates the main code paths used for `kpt pkg get` and `kpt pkg update` to use the `internal/errors` package. It also improves the output during get/update to provide users with information about progress.

Some patters I see on how to use the `errors` package:
 * `Class` will mostly be used on errors returned from libraries outside of kpt. At a higher level, an error can be due to a number of different reasons, so it is hard to find a `Class` that matches.
 * Using the common error pattern `fmt.Errorf("something: %w", err)` is useful for errors returned from libraries. We should avoid doing this if the error returned is already of the `internal/errors.Error` type, since it will break the error hierarchy.
   * A challenge here is that something we want to add additional context to errors and the current structure doesn't allow us to do that while keeping the hierarchy.
 * UniquePath is something we should add only at the level where we are operating on a package. So low-level errors probably won't have this information.

### DIFF
--- a/internal/util/update/update.go
+++ b/internal/util/update/update.go
@@ -16,15 +16,19 @@
 package update
 
 import (
-	"errors"
+	"context"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/GoogleContainerTools/kpt/internal/errors"
 	"github.com/GoogleContainerTools/kpt/internal/gitutil"
 	"github.com/GoogleContainerTools/kpt/internal/pkg"
+	"github.com/GoogleContainerTools/kpt/internal/printer"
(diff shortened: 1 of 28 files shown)


internal/util/update/update.go
```diff
@@ -207,35 +215,41 @@ func (nrc *nilRepoClone) AbsPath() string {
...
+		return errors.E(op, p.UniquePath, err)
 	}
 
 	if kf.Upstream == nil || kf.Upstream.Git == nil {
 		return nil
 	}
 
 	g := kf.Upstream.Git
+
+	prntr := printer.FromContextOrDie(ctx)
```
droot, line 231: we should probably standardize on the local variable name for printer. I used `pr` but wasn't super happy about it, but found to be better than alternatives.

droot (approved):
Awesome!

PR #1734, opened 2021-04-12
Update live command on next branch

seans3:
* Updates `kpt live` command to default to `ResourceGroup` inventory object for new packages.
* `kpt live init` command now updates `Kptfile` instead of creating `inventory-template.yaml`.
* Adds `kpt live migrate` and `kpt live install-resource-group` commands by default (no need for env var).
* Update `Kptfile` version to `v1alpha2` for `e2e/live/testdata`.


pkg/live/inventoryrg.go
```diff
@@ -34,7 +34,7 @@ import (
 // resource used to store inventory.
 var ResourceGroupGVK = schema.GroupVersionKind{
 	Group:   "kpt.dev",
-	Version: "v1alpha1",
+	Version: "v1alpha2",
```
mortent, line 37: Why are we updating the version of the RG CRD? I don't think there is a need to keep this version aligned with the version in the Kptfile.

mortent (approved):

PR #1742, opened 2021-04-13
fix updating result in batches

Shell32-Natsu:
- fix `KPT_E2E_UPDATE_EXPECTED`
 - Update support for exec function in `fn eval`.
 - Use kpt executable instead of `GetMain`


mengqiy:
CI is failing

Shell32-Natsu:
Will split this PR to multiple ones to make them clearer.

PR #1743, opened 2021-04-13
last batch of samples migrating to the new format

mikebz:
part 3 of 3 for #1421 
### Description
This is the last batch of samples migration to kpt alpha1 format.  There is a mysql Helm chart doesn't have any Kpt related information at all so I am not sure what to do with it.

### Testing
Ran kpt pkg render on these files.  Everything checks out.


package-examples/mysql-kustomize/README.md
```diff
@@ -18,8 +18,7 @@ To connect to the database:
 
 To edit the package:
 
-    kpt cfg list-setters mysql-kustomize
-    kpt cfg set mysql-kustomize SETTER VALUE
+    kpt fn eval --image gcr.io/kpt-fn/apply-setters:VERSION mysql-kustomize --  [setter_name=setter_value]
```
natasha41575, line 21: I imagine people will be copy/pasting these commands. Would it be better to put an actual version here (and add a note that they can change the tag), rather than have the VERSION placeholder?

mikebz: Good feedback. I have updated the readme's based on @phanimarupaka 's template.

package-examples/function-export/Kptfile
```diff
@@ -0,0 +1,12 @@
+apiVersion: kpt.dev/v1alpha2
+kind: Kptfile
+metadata:
+  name: kpt-function-export-example
+pipeline:
+  mutators:
+    - image: gcr.io/kpt-functions/label-namespace
```
droot, line 7: I don't remember seeing this function `label-namespace` in the catalog. What is it expected to do (assuming it existed :)) [may be this is answered later in the PR..]

package-examples/function-export/README.md
```diff
@@ -0,0 +1,103 @@
+# function-export
+
+This is a package that has both a setter and a validation function.  It shows
+how you can use gatekeeper to perform checks that are run during rendering.
+Putting a validation function  into your package allows you to give package 
+consumers guidance on what acceptable configuration is while giving them 
+the freedom to customize.
```
droot, line 7: The name of the example pkg is misleading :). Why don't we call it `config-validation` ? and text can be reworded to say something like:

```
This example package demonstrates how you can modify the config and ensure that modifications are compliant with the policies. This package uses:

- a `mutator` function called `label-namespace` to customize (or modify) the config
- a `validator` function `gatekeeper` to ensure changes are inline with the policy 

```

package-examples/function-export/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Render the declared values
+
+Render the changes in the hydration pipeline by using `kpt fn render` command:
+
+  $ kpt fn render function-export/
+
+    package "function-export": running function "gcr.io/kpt-functions/label-namespace": SUCCESS
+    package "function-export": running function "gcr.io/kpt-functions/gatekeeper-validate": SUCCESS
+    package "function-export": rendered successfully
```
droot, line 71: Not related to this PR, but pl. give feedback on the output that you see `fn render`. This is just a starting point.

package-examples/function-export/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+
+  $ kpt fn render function-export/
+
+    kpt fn render function-export/ 
+    package "function-export": running function "gcr.io/kpt-functions/label-namespace": SUCCESS
+    package "function-export": running function "gcr.io/kpt-functions/gatekeeper-validate": FAILED
+    fn.render: pkg function-export:
+            pkg.render:
+            pipeline.run: Error: Found 1 violations:
```
droot, line 85: Error reporting in `fn render` still needs work :) but better than before.

package-examples/function-export/README.md
```diff
@@ -0,0 +1,103 @@
...
+    validators:
+      - image: gcr.io/kpt-functions/gatekeeper-validate
+
+
+
+### Render the declared values
+
+Render the changes in the hydration pipeline by using `kpt fn render` command:
+
+  $ kpt fn render function-export/
```
droot, line 67: btw, we can use user to `cd /path/to/example/pkg`, and then we can avoid repeating the pkg name in the subsequent commands.

package-examples/mysql-kustomize/README.md
```diff
@@ -0,0 +1,86 @@
+# mysql-kustomize
+
+This is a package that has several hydration functions in it's pipeline
```
droot, line 3: We should avoid using the term `hydration` in public facing docs/UX.

package-examples/mysql-kustomize/README.md
```diff
@@ -0,0 +1,86 @@
...
+as well as utilzing kustomize to patch an upstream configuration.  The 
+upstream configuration is saved in the /upstream folder and the /instance
+patch is configured using kpt functions.
+
+For the final configuration kustomize is used to build the patched config
+with `kpt live apply` doing the deployment.
```
droot, line 9: I really have not idea of mixing this `kustomize` and `kpt` pkg will do :) I know the intention is to migrate and ensure syntax is correct. But in this case, it feels like we will be better off by deleting this example and introduce it later to help user accomplish some particular task. This can get really confusing.

droot (commented):
Mike,
it's better to delete examples that don't make sense and introduce examples that helps user accomplish some task or demonstrate usage of some functionality.

package-examples/function-export/README.md
mikebz: I think for all the samples I followed @phanimarupaka wordpress example where the commands are executed from the parent directory.  I think it's OK either way and if we want to establish a standard I am happy to have a change for all the samples.

package-examples/function-export/README.md
mikebz: thanks for the suggestion.  Updating.  I am still going to add a little bit of contextual guidance after the steps

package-examples/function-export/README.md
mikebz: Overall I find this new output a big step forward!  I think previously it was hard to figure out what part of the pipeline you were failing on.   I am curious to see if there are flags that allow someone to see the input and output as well if the problem is hard to diagnose.

package-examples/function-export/README.md
mikebz: Indeed.  This is a step in the right direction.  I am sure users will give us good guidance on what else we can do to help.

package-examples/mysql-kustomize/README.md
mikebz: Sounds good.  I will remove the sample!

package-examples/function-export/Kptfile
mikebz: I am not sure what it does outside of seeing what it does to the config.  We can also just replace it with something else.  Could set-label with the right GVK filter.

mikebz:
@droot made some of the changes you were asking for.  I think one material question is about using this old function.  I am happy to use something else that @mengqiy just produced.  LMK what you think.

package-examples/function-export/Kptfile
droot: So this function exists in old catalog. @mengqiy might know about it. Overall, I think using the functions from function-catalog will help in users discovering these functions. `set-label` should work here. I am guessing the policy check will be related to use of invalid labels or something.

package-examples/function-export/README.md
droot: Ack.

package-examples/function-export/README.md
droot: Yes, standardizing the pattern for all examples make sense. Lets not do it in this PR.

package-examples/mysql-kustomize/README.md
droot: Thanks.

droot:
> @droot made some of the changes you were asking for. I think one material question is about using this old function. I am happy to use something else that @mengqiy just produced. LMK what you think.

Yeah, I commented in the PR about using `set-label` to achieve the same thing. other than that, looks good.

mikebz:
set-label is now used.  Other minor fixes as well.

droot (approved):

PR #1746, opened 2021-04-13
improve exec support in fn eval test

Shell32-Natsu:
Use `UniquePath` for `fnConfig` and `exec` internally.


pkg/test/runner/runner.go
```diff
@@ -113,15 +113,11 @@ func (r *Runner) runFnEval() error {
 	}
 	if r.testCase.Config.EvalConfig.Image != "" {
 		kptArgs = append(kptArgs, "--image", r.testCase.Config.EvalConfig.Image)
-	} else if r.testCase.Config.EvalConfig.ExecPath != "" {
-		kptArgs = append(kptArgs, "--exec-path", r.testCase.Config.EvalConfig.ExecPath)
+	} else if r.testCase.Config.EvalConfig.execUniquePath.Empty() {
```
droot, line 116: You probably want not-empty here ? `!...Empty()`

pkg/test/runner/runner.go
```diff
@@ -113,15 +113,11 @@ func (r *Runner) runFnEval() error {
...
+		kptArgs = append(kptArgs, "--exec-path", string(r.testCase.Config.EvalConfig.execUniquePath))
 	}
-	if r.testCase.Config.EvalConfig.FnConfig != "" {
-		fnConfigPath, err := filepath.Abs(filepath.Join(r.testCase.Path, r.testCase.Config.EvalConfig.FnConfig))
-		if err != nil {
-			return fmt.Errorf("failed to get absolute path to function config file: %w", err)
-		}
-		kptArgs = append(kptArgs, "--fn-config", fnConfigPath)
+	if r.testCase.Config.EvalConfig.fnConfigUniquePath.Empty() {
```
droot, line 119: !.Empty() ?

pkg/test/runner/runner.go
```diff
@@ -206,11 +202,8 @@ func (r *Runner) runFnRender() error {
 	for i := 0; i < r.testCase.Config.RunCount(); i++ {
 		command.SetArgs(kptArgs)
 		fnErr = command.Execute()
-		if fnErr != nil {
-			if r.testCase.Config.ExitCode != 0 {
-				return nil
-			}
-			break
+		if fnErr != nil && r.testCase.Config.ExitCode != 0 {
+			return nil
```
droot, line 206: The previous logic was stopping the for loop for the case where command returned the error and expected ExitCode was 0 (success), but now it will continue with the loop (which we don't want..right ?)

pkg/test/runner/runner.go
Shell32-Natsu: Yep, you are right. :)

pkg/test/runner/runner.go
Shell32-Natsu: I will mark this PR as draft. We will need to merge #1747 at first.

droot (approved):

PR #1747, opened 2021-04-13
use kpt binary in e2e test

Shell32-Natsu:
Use kpt binary instead of `GetMain`.


pkg/test/runner/runner.go
```diff
@@ -34,6 +32,22 @@ type Runner struct {
 	cmd           string
 	t             *testing.T
 	initialCommit string
+	kptBin        string
+}
+
+func getKptBin() (string, error) {
```
droot, line 38: I think taking an explicit path for `kpt binary` will be better (we can read from an env var).

I have noticed, people sometime install `kpt` using OS package manager etc and sometime that is installed in preferred paths and can cause confusion which binary is being used.

pkg/test/runner/runner.go
```diff
@@ -51,19 +65,26 @@ const (
...
-func NewRunner(t *testing.T, testCase TestCase, c string) (*Runner, error) {
+func NewRunner(t *testing.T, testCase TestCase, c, kptBin string) (*Runner, error) {
 	info, err := os.Stat(testCase.Path)
 	if err != nil {
 		return nil, fmt.Errorf("cannot open path %s: %w", testCase.Path, err)
 	}
 	if !info.IsDir() {
 		return nil, fmt.Errorf("path %s is not a directory", testCase.Path)
 	}
+	if kptBin == "" {
```
droot, line 76: Can this be optional ? I think current way of running test is also useful.

Is the kpt developer expected to build kpt first and then run the test might lead to human errors where test were invoked without building the kpt binary.

droot (commented):
Some questions.

Makefile
```diff
@@ -66,8 +66,8 @@ test:
 
 # This target is used to run Go tests that require docker runtime.
 # Some tests, like pipeline tests, need to have docker available to run.
-test-docker:	
-	go test -cover --tags=docker ./...
+test-docker: build
```
droot, line 69: We need to add it to the next target `test-fn-render` as well :)

pkg/test/runner/runner.go
```diff
@@ -34,13 +31,22 @@ type Runner struct {
...
+	}
+	return "kpt"
 }
 
 const (
 	// If this env is set to "true", this e2e test framework will update the
 	// expected diff and results if they already exist. If will not change
 	// config.yaml.
 	updateExpectedEnv string = "KPT_E2E_UPDATE_EXPECTED"
+	kptBinEnv         string = "KPT_E2E_BINARY"
```
droot, line 49: Can we shorten it to `KPT_E2E_BIN` ?

pkg/test/runner/runner.go
```diff
@@ -59,11 +65,14 @@ func NewRunner(t *testing.T, testCase TestCase, c string) (*Runner, error) {
 	if !info.IsDir() {
 		return nil, fmt.Errorf("path %s is not a directory", testCase.Path)
 	}
+	kptBin := getKptBin()
+	t.Logf("Using kpt binary: %s", kptBin)
```
droot, line 69: This is going to be life savor.

pkg/test/runner/runner.go
```diff
@@ -34,13 +31,22 @@ type Runner struct {
...
 	t             *testing.T
 	initialCommit string
+	kptBin        string
+}
+
+func getKptBin() string {
+	if p := os.Getenv(kptBinEnv); p != "" {
+		return p
+	}
+	return "kpt"
```
droot, line 41: May be we should panic here..?

pkg/test/runner/runner.go
```diff
@@ -200,17 +198,12 @@ func (r *Runner) runFnRender() error {
...
-		command.SetArgs(kptArgs)
-		fnErr = command.Execute()
-		if fnErr != nil {
-			if r.testCase.Config.ExitCode != 0 {
-				return nil
-			}
-			break
+		_, fnErr := runCommand("", r.kptBin, kptArgs)
+		if fnErr != nil && r.testCase.Config.ExitCode != 0 {
+			// TODO: compare render results
```
droot, line 205: what's this comment about ?

droot (changes requested):

pkg/test/runner/runner.go
```diff
@@ -262,7 +255,7 @@ func (r *Runner) compareResult(exitErr error, tmpPkgPath, resultsPath string) er
 		return fmt.Errorf("actual exit code %d doesn't match expected %d", exitCode, r.testCase.Config.ExitCode)
 	}
 
-	if exitCode != 0 {
+	if exitCode != 0 && expected.Results != "" {
```
mengqiy, line 258: This doesn't handle the case that the fn exit with 0 but it output some structured results (e.g. warning and info).

Makefile
Shell32-Natsu: done. All add a `test-fn-eval` target.

pkg/test/runner/runner.go
Shell32-Natsu: done

pkg/test/runner/runner.go
Shell32-Natsu: Will return an error is no env given

pkg/test/runner/runner.go
Shell32-Natsu: Updated the code.

pkg/test/runner/runner.go
Shell32-Natsu: Updated and now we will always compare results if there is any.

e2e/testdata/fn-render/path-index-duplicate/.expected/config.yaml
```diff
@@ -15,3 +15,4 @@
 # One of the functions in the pipeline fails resulting in
 # non-zero exit code and no changes in the resources.
 exitCode: 1
+skip: true
```
Shell32-Natsu, line 18: @droot @natasha41575 `kpt fn render` command will exit with 0 for this case.

pkg/test/runner/runner.go
```diff
@@ -35,13 +31,22 @@ type Runner struct {
 	cmd           string
 	t             *testing.T
 	initialCommit string
+	kptBin        string
+}
+
+func getKptBin() (string, error) {
+	if p := os.Getenv(kptBinEnv); p != "" {
```
droot, line 38: I realized we can take a commandline arg instead of environment variable (not sure about the benefits though or support both) https://blog.jbowen.dev/2019/08/using-go-flags-in-tests/

e2e/testdata/fn-render/path-index-duplicate/.expected/config.yaml
droot: Didn't understand. If `skip:true`, the test is not even invoked ...right, look for `t.Skip()` ? so not sure why would even be the command be executed.

droot (commented):
Some more drive by comments.

mengqiy:
CI is stilling failing. I guess that's because how we handle `skip`. Please fix.

Shell32-Natsu:
The error is because the kpt command doesn't properly return exit code. @droot Will create a quick fix.

e2e/testdata/fn-render/path-index-duplicate/.expected/config.yaml
Shell32-Natsu: This test can pass after the fix about the exit code. Remove the skip.

pkg/test/runner/runner.go
Shell32-Natsu: I think we can use env now and improve it later if we need.

droot (approved):

PR #1748, opened 2021-04-13
Add invalid repo test.

etefera:
Ensure error is displayed when repo doesn't exist.


internal/util/update/update_test.go
```diff
@@ -678,6 +678,46 @@ func TestCommand_Run_failInvalidPath(t *testing.T) {
 	}
 }
 
+// TestCommand_Run_failInvalidRepo verifies Run fails if the repo is invalid
+func TestCommand_Run_failInvalidRepo(t *testing.T) {
+	for _, strategy := range kptfilev1alpha2.UpdateStrategies {
```
mortent, line 683: We don't really need to test for each strategy here. This functionality is independent of which strategy is being used.

internal/util/update/update_test.go
```diff
@@ -678,6 +678,46 @@ func TestCommand_Run_failInvalidPath(t *testing.T) {
...
+		t.Run(string(strategy), func(t *testing.T) {
+			g := &testutil.TestSetupManager{
+				T: t,
+				ReposChanges: map[string][]testutil.Content{
+					testutil.Upstream: {
+						{
+							Data:   testutil.Dataset1,
```
mortent, line 690: We are trying to move more towards the `pkgbuilder` pattern, as using golden data sets becomes difficult when there are multiple repos where each has multiple commits. But for this test the content doesn't really matter, so it is ok here.

internal/util/update/update_test.go
```diff
@@ -678,6 +678,46 @@ func TestCommand_Run_failInvalidPath(t *testing.T) {
...
+				return
+			}
+
+			err := Command{
+				Pkg:      pkgtest.CreatePkgOrFail(t, g.LocalWorkspace.FullPackagePath()),
+				Repo:     "fake",
+				Strategy: strategy,
+			}.Run()
+			if !assert.Error(t, err) {
+				return
```
mortent, line 710: nit: We usually use `t.FailNow()` in these situations.

internal/util/update/update_test.go
etefera: Removed.

internal/util/update/update_test.go
etefera: I'll start basing other tests on that pattern.

internal/util/update/update_test.go
etefera: Replaced.

mortent (approved):

PR #1749, opened 2021-04-13
Add test for a kptfile without an upstream.

etefera:

### DIFF
--- a/internal/util/update/update_test.go
+++ b/internal/util/update/update_test.go
@@ -662,46 +662,27 @@ func TestCommand_ResourceMerge_NonKRMUpdates(t *testing.T) {
 }
 
 func TestCommand_Run_noUpstreamReference(t *testing.T) {
-	for i := range kptfilev1alpha2.UpdateStrategies {
-		strategy := kptfilev1alpha2.UpdateStrategies[i]
-		t.Run(string(strategy), func(t *testing.T) {
-			// Setup the test upstream and local packages
-			g := &testutil.TestSetupManager{
-				T: t,
-				ReposChanges: map[string][]testutil.Content{
-					testutil.Upstream: {
-						{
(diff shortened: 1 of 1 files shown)


internal/util/update/update_test.go
```diff
@@ -661,6 +661,49 @@ func TestCommand_ResourceMerge_NonKRMUpdates(t *testing.T) {
...
+					},
+				},
+			}
+			defer g.Clean()
+			if !g.Init() {
+				return
+			}
+
+			pkg := pkgtest.CreatePkgOrFail(t, g.LocalWorkspace.FullPackagePath())
+			kptfile, err := pkg.Kptfile()
```
mortent, line 689: So instead of fetching a package and then remove the Upstream information, I think the test will be simpler if you just create a Kptfile from scratch in a temp dir, and then just use the Update command on it. We have an example of this here: https://github.com/GoogleContainerTools/kpt/blob/786926a36f4a49872cb0fe986cc4a489ea67ab76/internal/util/update/update_test.go#L2855

internal/util/update/update_test.go
```diff
@@ -661,6 +661,49 @@ func TestCommand_ResourceMerge_NonKRMUpdates(t *testing.T) {
 	}
 }
 
+func TestCommand_Run_noUpstreamReference(t *testing.T) {
+	for i := range kptfilev1alpha2.UpdateStrategies {
```
mortent, line 665: No need to test for all strategies.

internal/util/update/update_test.go
etefera: Switched to creating a kptfile with no upstream.

internal/util/update/update_test.go
etefera: Removed strategies.

mortent (approved):

PR #1752, opened 2021-04-14
return non-zero exit code in case of error

droot:
`kpt` will return non-zero (1) exit code if it encounters error in running any command.


Shell32-Natsu (approved):
lgtm

PR #1753, opened 2021-04-14
Add tests for invalid UpstreamLock information.

etefera:
Note that there's no error for a bad ref in the UpstreamLock. The error message for a bad dir references missing subpackages.


internal/util/update/update_test.go
```diff
@@ -34,7 +34,9 @@ import (
 )
 
 const (
-	kptRepo = "github.com/GoogleContainerTools/kpt"
+	kptRepo         = "github.com/GoogleContainerTools/kpt"
+	mainBranch      = "master"
```
mortent, line 38: We should call this `masterBranch` instead of `mainBranch`. No branch is special, but most of the tests uses a mix of `main` and `master` as the name. This name would be a little confusing.

internal/util/update/update_test.go
```diff
@@ -678,6 +680,137 @@ func TestCommand_Run_failInvalidPath(t *testing.T) {
...
+		Pkg: pkgtest.CreatePkgOrFail(t, w.FullPackagePath()),
+	}.Run()
+
+	if assert.Error(t, err) {
+		assert.Contains(t, err.Error(), "does not appear to be a git repository")
+	}
+}
+
+func TestCommand_Run_badUpstreamLockRef(t *testing.T) {
+	repos, w, clean := testutil.SetupReposAndWorkspace(t, map[string][]testutil.Content{
```
mortent, line 728: For this situation I think we want to use the TestSetupManager so we actually get a fork in the workspace and then try to do the update with an incorrect ref. The way it is set up now, the repo is actually in a weird state since the `UpstreamLock` says the package has been fetched, but the resources doesn't actually exist on disk.

internal/util/update/update_test.go
```diff
@@ -678,6 +680,137 @@ func TestCommand_Run_failInvalidPath(t *testing.T) {
...
+			Ref:       "fake",
+		},
+	}
+	testutil.AddKptfileToWorkspace(t, w, kf)
+
+	// Update the local package
+	err := Command{
+		Pkg: pkgtest.CreatePkgOrFail(t, w.FullPackagePath()),
+	}.Run()
+	if !assert.NoError(t, err) {
```
mortent, line 765: This situation doesn't return an error? It is a bit of a strange state, but I would expect kpt to fail since the Ref doesn't exist.

internal/util/update/update_test.go
```diff
@@ -678,6 +680,137 @@ func TestCommand_Run_failInvalidPath(t *testing.T) {
...
+	}
+	testutil.AddKptfileToWorkspace(t, w, kf)
+
+	// Update the local package
+	err := Command{
+		Pkg: pkgtest.CreatePkgOrFail(t, w.FullPackagePath()),
+	}.Run()
+
+	if assert.Error(t, err) {
+		assert.Contains(t, err.Error(), "added in both upstream and local")
```
mortent, line 810: This error message doesn't seem related to what is actually being tested here. I'm not sure if a non-existent directory by itself should be considered an error.

internal/util/update/update_test.go
etefera: Renamed.

internal/util/update/update_test.go
etefera: Updated to use `TestSetupManager`

internal/util/update/update_test.go
etefera: Looks like it tries to clone using `Commit` instead of `Ref` https://github.com/GoogleContainerTools/kpt/blob/786926a36f4a49872cb0fe986cc4a489ea67ab76/internal/util/update/update.go#L230

mortent: Yeah, this is what it should do. We are using the `Commit` field to fetch the common ancestor for local and upstream so we can do a 3-way merge.
We don't actually use the `Ref` field in `GitLock`. I have been considering if it should be removed, but I haven't made a decision on that yet. We can just skip this test. We should test what happens if the `Commit` field has a value that doesn't exist in the repo.

internal/util/update/update_test.go
```diff
@@ -703,6 +704,144 @@ func TestCommand_Run_failInvalidPath(t *testing.T) {
 	}
 }
 
+func TestCommand_Run_badUpstreamLockRepo(t *testing.T) {
+	repos, w, clean := testutil.SetupReposAndWorkspace(t, map[string][]testutil.Content{
```
mortent, line 708: In all these tests I think we should use the `TestSetupManager`. This makes sure we run the tests on a fork that has been properly cloned from an upstream repo.

When we want to make changes to the local fork after the initial fetch, we should use the `LocalChanges` property on the `TestSetupManager`. Just remember that this specified a complete new state for the local fork. Creating the invalid Kptfile configurations should be pretty straightforward with the pkgbuilder.

If the tests start to look very similar, consider using table-driven tests. There are examples in this file.

etefera: It didn't seem like `LocalChanges` could be meaningfully changed after `TestSetupManager.Init()` is called, which was necessary to give `WithUpstreamLock()` a ref to the directory of the remote repo as well as a commit (and not just a commit index). I ended up using two instances of `TestSetupManager` to be able to get a remote repo path to give to the second instance which is actually being tested.

etefera:
@mortent OoO next business day so if the proposed test looks good to you, feel free to merge.

internal/util/update/update_test.go
```diff
@@ -703,6 +704,93 @@ func TestCommand_Run_failInvalidPath(t *testing.T) {
...
+		t.Run(tn, func(t *testing.T) {
+			upstreamContent := testutil.Content{
+				Pkg: pkgbuilder.NewRootPkg().
+					WithResource(pkgbuilder.DeploymentResource),
+				Branch: masterBranch,
+			}
+
+			// Using a TestSetupManager for just the remote repo to get a directory
+			// for use in the local UpstreamLock.
+			remote := &testutil.TestSetupManager{
```
mortent, line 740: We should only need to create one `TestSetupManager`. It goes through the steps of creating all the repos (as defined by the `ReposChanges` property), set up a local workspace, and then get the package from the repo name `upstream` into the workspace. The `TestSetupManager` keeps references to the created repos and workspace so we can use them.

I think in this case you should be able to basically just combine the two instances into one.

internal/util/update/update_test.go
```diff
@@ -703,6 +704,93 @@ func TestCommand_Run_failInvalidPath(t *testing.T) {
...
+			repo:             testutil.Upstream,
+			expectedErrorMsg: "unknown revision or path"},
+		"bad dir": {
+			dir:              "/fake",
+			commit:           "",
+			repo:             testutil.Upstream,
+			expectedErrorMsg: "no such file or directory"},
+		"bad repo": {
+			dir:              "/",
+			commit:           "",
```
mortent, line 726: Nit: Just leave them out instead of setting the nil value.

etefera: Done.

internal/util/update/update_test.go
```diff
@@ -703,6 +704,79 @@ func TestCommand_Run_failInvalidPath(t *testing.T) {
...
+
+			// If tc.repo exists in the testbed, use its repo directory for the UpstreamLock.
+			repo := g.Repos[tc.repo]
+			repoDir := tc.repo
+			if repo != nil {
+				repoDir = repo.RepoDirectory
+			}
+			localChanges := []testutil.Content{{
+				Pkg: pkgbuilder.NewRootPkg().WithKptfile(pkgbuilder.NewKptfile().
+					WithUpstreamRef(testutil.Upstream, tc.dir, masterBranch, "resource-merge").
```
mortent, line 759: We should probably avoid using the `tc.dir` value here so we know that the `Upstream` section is correct.

internal/util/update/update_test.go
```diff
@@ -703,6 +704,79 @@ func TestCommand_Run_failInvalidPath(t *testing.T) {
...
+			}
+
+			defer g.Clean()
+
+			if !g.Init() {
+				return
+			}
+
+			// If tc.repo exists in the testbed, use its repo directory for the UpstreamLock.
+			repo := g.Repos[tc.repo]
```
mortent, line 752: Nit: It would be better here to use the second return value to check whether the repo exists.

etefera: Done.

internal/util/update/update_test.go
etefera: Done, added a check in `update.go` to throw an error on faulty `UpstreamLock` paths.

internal/util/update/update.go
```diff
@@ -314,7 +314,12 @@ func (u Command) updatePackage(subPkgPath, localPath, updatedPath, originPath st
 	}
 
 	originExists := isRootPkg
-	if !isRootPkg {
+	if isRootPkg {
+		_, err := os.Stat(originPath)
```
mortent, line 318: Why is this needed?

etefera: Without this branch, the function does not check for the existence of the origin path before attempting to update the package later. The error that's surfaced in that case doesn't give any information about the bad path in the `UpstreamLock` unlike the error returned here.

mortent: Looking at this now, I actually think the original error here is the right one. In fact, there probably isn't such a thing as an invalid directory, it just happens to be a directory that doesn't exist. And if this is the case, the merge should treat this as an empty origin.
I think we can remove the `invalid dir` test case and this change. Then this PR should be good to merge.

etefera: Removed.

mortent (approved):

PR #1758, opened 2021-04-14
Add next button to the book cover.

etefera:
And removes previous button from last page.
Fixes https://github.com/GoogleContainerTools/kpt/issues/1720
image


site/index.html
```diff
@@ -98,12 +98,33 @@
...
         paginationFooters.forEach(
           (el) => (el.style.display = isBookPage ? "flex" : "none")
         );
+
+        // Don't show previous button on the book cover page.
+        const previousPaginationButtons = Array.from(
+          document.getElementsByClassName("pagination-item--previous")
+        );
+        const isBookCover =
+          isBookPage && document.location.pathname.toLowerCase().length < 7;
```
mikebz, line 113: length is a little magical here.  Is there a way to just firmly say that the book cover is path `x`?

site/index.html
```diff
@@ -98,12 +98,33 @@
...
+        previousPaginationButtons.forEach(
+          (el) => (el.style.display = isBookCover ? "none" : "flex")
+        );
+
+        // Don't show next button to non-book-pages.
+        const nextPaginationButtons = Array.from(
```
mikebz, line 119: is the strategy here to add the next/previous to everything and then hyde it via a style?  Is there a way to just have book pages derive from some template that has those and the rest of the pages don't have that?  It looks like the javascript in index.html is getting a bit much.

site/index.html
etefera: Added more precise matching.

site/index.html
etefera: Yeah, the next/previous is already added to the site by a plugin. This is additional styling to control where it appears based on the URL. It seems that's the preferred method of plugin management from Docsify. I agree the inline JS is getting a bit cluttered; we should move what we can to a separate JS file (which we can also automatically test) at a later time.

frankfarzan (approved):

Issue #1666, opened 2021-04-01
Functions e2e test harness should test CLI output


frankfarzan:
We want to test expected error messages and non-structured results.

frankfarzan:
One complication here is testing on different OS's if the output contains paths. One pragmatic option is to only test the output on select platforms.

droot:
Yes, for m2 milestone, we should target select platforms (Linux and OSX).

natasha41575:
#1762 checks the CLI output - but it does a substring match rather than an exact match (discussion was in [this comment thread](https://github.com/GoogleContainerTools/kpt/pull/1762#discussion_r614997518)). 

For example, for validate-resource-failure, the actual stdout is:
`package "/var/folders/l9/fkxgtdvn3435dzj2zj4gj9x000rk3j/T/kpt-pipeline-e2e-113726026/validate-resource-failure": running function "gcr.io/kpt-fn/starlark:unstable": FAILED`

but the test case here checks for a substring:

https://github.com/GoogleContainerTools/kpt/blob/a9589f16690bc84e4f26973f1aa413a96ac3a919/e2e/testdata/fn-render/validate-resource-failure/.expected/config.yaml#L17

For now we believe this is sufficient for our use case.

natasha41575:
reopening because the previous PR was just for fn render; I will update it for fn eval as well.

frankfarzan:
As part of this, please audit the error messages to ensure they're clear and actionable and conform to go/kpt-errors-style-guide. Many of the error message I'm seeing have poor UX.

natasha41575:
Added a comment https://github.com/GoogleContainerTools/kpt/issues/1783#issuecomment-825199107 listing the error messages that can be improved

PR #1762, opened 2021-04-15
fn render should check against expected stderr

### LINKED ISSUE #1755: `kpt fn render` e2e test cases should check content for stdout/stderr output messages
We introduced many negative e2e test cases for `kpt fn render` but we do not check content output on stdout/stderr for these test cases. Test harness is going to support assertions for content emitted on stdout/stderr and we should use that to ensure expected content is emitted in all the negative test cases.

/cc @Shell32-Natsu @natasha41575

natasha41575:
Fix #1755

#1666

Test harness compares standard error from the command to the expected standard error.


pkg/test/runner/config.go
```diff
@@ -54,7 +54,9 @@ type EvalTestCaseConfig struct {
 type TestCaseConfig struct {
 	// ExitCode is the expected exit code from the kpt commands. Default: 0
 	ExitCode int `json:"exitCode,omitempty" yaml:"exitCode,omitempty"`
-
+	// StdErr is the expected standard error output and should be checked
+	// when a nonzero exit code is expected. Default: ""
+	StdErr string `json:"stdErr,omitempty" yaml:"stdErr,omitempty"`
```
Shell32-Natsu, line 59: I think we should check stdout as well. And if content in stdout/stderr is pretty long or multiple lines, a string field may not very good.

pkg/test/runner/util.go
```diff
@@ -15,69 +15,74 @@
...
 
 func copyDir(src, dst string) error {
-	_, err := runCommand("", "cp", []string{"-r", src, dst})
+	_, _, err := runCommand("", "cp", []string{"-r", src, dst})
 	return err
 }
 
 func gitInit(d string) error {
-	o, err := runCommand(d, "git", []string{"init"})
+	o, s, err := runCommand(d, "git", []string{"init"})
```
Shell32-Natsu, line 43: Maybe stdout and stderr here are clearer.

pkg/test/runner/runner.go
```diff
@@ -206,7 +206,17 @@ func (r *Runner) runFnRender() error {
 	// run function
 	kptArgs := []string{"fn", "render", pkgPath}
 	for i := 0; i < r.testCase.Config.RunCount(); i++ {
-		output, fnErr := runCommand("", r.kptBin, kptArgs)
+		output, stderr, fnErr := runCommand("", r.kptBin, kptArgs)
+		if fnErr != nil {
```
Shell32-Natsu, line 210: I will suggest to not compare the results here but delegate it to `r.compareResult`.

pkg/test/runner/runner.go
```diff
@@ -206,7 +206,17 @@ func (r *Runner) runFnRender() error {
...
+			if r.testCase.Config.ExitCode != 0 {
+				expected := r.testCase.Config.StdErr
+				if !strings.Contains(stderr, expected) {
+					return fmt.Errorf("wanted %s, got %s", expected, stderr)
+				}
+				return nil
+			}
+			break
```
Shell32-Natsu, line 218: Don't break here. We need to compare the results even when there is an error.

Shell32-Natsu (changes requested):

Shell32-Natsu:
You may want to update the function which updates the expected results as well.

pkg/test/runner/config.go
natasha41575: what type for stdout/stderr would be good, if not string?

Shell32-Natsu: Maybe put them in `cli_output.txt` or `command_output.txt` in `.expected`? How do you think?

natasha41575: would we want two files (e.g. cli_stderr.txt, cli_stdout.txt), or combined into one?

Shell32-Natsu: I think 2 files are clearer. We should go this way. And also please remember to add them to batch update function.

droot: Two things here:

- YAML support [multiline ](https://stackoverflow.com/questions/3790454/how-do-i-break-a-string-in-yaml-over-multiple-lines) so specifying big text should be ok. Having separate files increases the burden on the test author, and reviewer. Having expected output in a single place as much as possible helps. So lets avoid multiple files here.
- Another thing I was thinking was to use using this as `substring` instead of exact match (haven't looked at the entire PR yet). In the future, we can treating this as Go template (that supports `PkgPath` as a variable) to allow test authors to include pkg-path also in the message.

pkg/test/runner/runner.go
```diff
@@ -206,7 +206,17 @@ func (r *Runner) runFnRender() error {
 	// run function
 	kptArgs := []string{"fn", "render", pkgPath}
 	for i := 0; i < r.testCase.Config.RunCount(); i++ {
-		output, fnErr := runCommand("", r.kptBin, kptArgs)
+		output, stderr, fnErr := runCommand("", r.kptBin, kptArgs)
```
droot, line 209: Can we rename `output` to `stdout` to make consistent with `stderr` ?

pkg/test/runner/util.go
droot: +1

pkg/test/runner/runner.go
```diff
@@ -206,7 +206,17 @@ func (r *Runner) runFnRender() error {
...
+		if fnErr != nil {
+			if r.testCase.Config.ExitCode != 0 {
+				expected := r.testCase.Config.StdErr
+				if !strings.Contains(stderr, expected) {
```
droot, line 213: we should do this comparison only if expectedStderr is != "". Also can we rename `expected` to `expectedStdErr` ?

e2e/testdata/fn-render/invalid-inline-fnconfig/.expected/config.yaml
```diff
@@ -12,4 +12,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-exitCode: 1
\ No newline at end of file
+exitCode: 1
+stdErr: "missing Resource metadata"
```
droot, line 16: ha ha ... this is terrible error message. Poor user has to go hunt....which metadata :)

e2e/testdata/fn-render/no-fnconfig/.expected/config.yaml
```diff
@@ -13,3 +13,4 @@
 # limitations under the License.
 
 exitCode: 1
+stdErr: "failed to configure function: function config must be a ConfigMap or SetLabelConfig"
```
droot, line 16: This is a poor error message, hope we improve it eventually.

e2e/testdata/fn-render/subpkg-has-invalid-kptfile/.expected/config.yaml
```diff
@@ -13,3 +13,4 @@
 # limitations under the License.
 
 exitCode: 1
+stdErr: "unable to parse \"Kptfile\": yaml: line 10: mapping values are not allowed in this context"
```
droot, line 16: morten has PR that changes this message, so you will have to change this after rebase.

e2e/testdata/fn-render/invalid-inline-fnconfig/.expected/config.yaml
natasha41575: The test harness is checking for a substring - the actual error output is much longer. I can update the PR to show the entire error message here

pkg/test/runner/config.go
Shell32-Natsu: Using a multiple line string in `config.yaml` has some disadvantages:
1. We cannot control the format when we automatically update the expected results. It maybe encoded to a single line string. (Not sure can we control the format)
2. If the output is very long, most parts of `config.yaml` will be the output string and it's also hard to read. Although not sure will we have a very long output in the test.

natasha41575: Output doesn't seem too long (and can be shorter if we use substrings). I will add checking stdout here and see what happens

pkg/test/runner/runner.go
natasha41575: done, thanks!

pkg/test/runner/runner.go
natasha41575: fixed!

pkg/test/runner/config.go
natasha41575: 1. I updated the PR to also check for stdout. Aligning with @droot's suggestion, I haven't moved it out to a new file for now but can do so if that is more desirable. 

2. @Shell32-Natsu If we move stdout/stderr into separate files I can add them to the batch update function.

However, with them in config.yaml I think it will be nontrivial to update. 

The batch update function doesn't update config.yaml currently - e.g. if it has the wrong expected exit code this needs to be changed manually AFAICT (though please correct me if I'm wrong). 

Updating config.yaml while preserving comments and not including defaults etc might take some time to figure out (and I think should be a separate issue/PR).

natasha41575: 1. I've updated to check the output of stdout as well. Aligning with @droot's suggestion, I've left it in config.yaml for now. 

2. @Shell32-Natsu is there an easy way to update config.yaml? Currently `updateExpected` doesn't seem to change it - e.g. if the expected exit code is wrong it has to be changed manually AFAICT (though please correct me if I'm wrong). I can write the output of yaml.Marshal(r.testCaseConfig) but that produces some unnecessary and unwanted output. Unless I'm missing an easy way to do the update, I suggest we file a separate issue for it because I don't think it is as high priority as some other tasks right now.

pkg/test/runner/util.go
natasha41575: done

pkg/test/runner/config.go
droot: Yes, if we plan to auto update like we do for `diff` and `results`, specifying them in separate files will be preferred.
I was thinking of using substring of the expected output (so expected them to be short) and avoiding regular updates, so wasn't thinking about batch auto update will be needed, but I could be wrong. If go down the route of supporting matching the exact output, then we will have to support template (pkgPath substitution might be needed) and keeping them in separate files.

WDYT ?

natasha41575: Yes, I am already using a substring of the expected output (checking if actual output contains the expected output). 

https://github.com/GoogleContainerTools/kpt/blob/c91da1d1cabef6cbe3809bca27f9c41b1a0813e6/pkg/test/runner/runner.go#L312

The actual output is longer (has multiple lines and includes the package name). If we check for exact match, separate files may be preferable as I was having some trouble with the multilines string. Personally I like this approach; it is simpler and easier to read, and as you've said, will avoid regular updates.

droot: I also think `substring` match is good enough for our use-case here. Lets go with this and if it doesn't work for us in the future, we can always address it by supporting separate files.

droot (approved):

pkg/test/runner/runner.go
```diff
@@ -268,6 +269,9 @@ func (r *Runner) compareResult(exitErr error, tmpPkgPath, resultsPath string) er
 	if exitCode != r.testCase.Config.ExitCode {
 		return fmt.Errorf("actual exit code %d doesn't match expected %d", exitCode, r.testCase.Config.ExitCode)
 	}
+	if exitCode != 0 {
+		return r.compareOutput(stdout, stderr)
+	}
```
Shell32-Natsu, line 274: Even the exit code is not 0, we still need to checkout results and diff. We have structured output in the failure. We also need to make sure the resources are not modified in failure. So no matter what the exit code is, we always need to check everything.

Shell32-Natsu (changes requested):
Only one comment, others LGTM.

e2e/testdata/fn-render/fn-failure/.expected/config.yaml
```diff
@@ -15,3 +15,5 @@
 # One of the functions in the pipeline fails resulting in
 # non-zero exit code and no changes in the resources.
 exitCode: 1
+stdErr: "httpbin-gen:4:1: got newline, want primary expression"
+stdOut: "running function \"gcr.io/kpt-fn/starlark:unstable\": FAILED"
```
frankfarzan, line 19: @Shell32-Natsu @droot 

What's the rationale for having this part of stdout vs stderror? What's the combined output shown to the user?

natasha41575: The combined output is as follows:
```
package ".": running function "gcr.io/kpt-fn/starlark:unstable": FAILED
fn.render: pkg .:
        pkg.render:
        pipeline.run: httpbin-gen:4:1: got newline, want primary expression 
```

The rationale is that stdout will have whether the function succeeded or failed, while stderr will have the error. 

In the case that a function succeeds, this stdout would be shown to the user: 
`package ".": running function "gcr.io/kpt-fn/starlark:unstable": SUCCESS`

frankfarzan: For the failure path, separating the debugging information between stdout and stderr seems like unnecessary complexity. Also note that we need to have a design on the desired UX tracked by https://github.com/GoogleContainerTools/kpt/issues/1557

frankfarzan: Also having separate stdout and stderror means, the final output can be interleaved differently. We actually want the exact aggregated output in the the order you have it there.

PR #1764, opened 2021-04-15
delete all files if all resources are deleleted

### LINKED ISSUE #1460: No change happens if delete all the resources through kpt ts functions
No change happens when we try to filter out all the resources through kpt function. 
kpt version: 0.36.1

We have a kpt ts function filtering resource not fit the conditions. Simply like below:
```
export async function filterResources(configs: Configs): Promise<void> {
  
  configs.delete(
    ...configs
      .getAll()
      .filter(
        ko =>
          condition1 || condition2 || condition3 ...
      ),
  );
```
If all resources in the `configs` match and should be deleted, then the input source directory keeps as unchanged. 

Expectation:
The input directory becomes empty. 

Reproduce:
It is not hard to reproduce the bug. 
*  Init a kpt work environment follow the [official document](https://googlecontainertools.github.io/kpt/guides/producer/functions/ts/quickstart/).
*  Add an extra ts function like below:
```
import { Configs } from "kpt-functions";

export async function filterInfraResources(configs: Configs): Promise<void> {
  configs.deleteAll(); 
}

filterInfraResources.usage = "";
```
* Build it and Run the function on the `example configs`. 
```
$ mkdir result
$ kpt fn source ../../example-configs/shipping-dev | node dist/filter_infra_resources_run.js | kpt fn sink result
$ ls result 
```
* Run `ls result`, which shows all the files from the original folder. 

But if we make the kpt function return at least one resource, this works. Modify the kpt function like below:
```
export async function filterInfraResources(configs: Configs): Promise<void> {
  configs.delete(
    ...configs.getAll().filter(ko => ko.metadata.name != "viewers"),
  );
}
```
Then, rebuild and re-run the above steps. I can get only `rolebinding_viewers.yaml` in the result folder. 

So the deletion only works when there is at least one config left, so we have to add a temporary config into the `configs` to make sure kpt deletes the rest of the resources. And delete the temp resources after `kpt fn run` finishes in our Java code. 

Any ideas? 

Thanks in advance!

Shell32-Natsu:
fix #1460

### DIFF
--- a/e2e/testdata/fn-eval/all-resource-deletion/pkg/.expected/config.yaml
+++ b/e2e/testdata/fn-eval/all-resource-deletion/pkg/.expected/config.yaml
@@ -1,0 +1,17 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+testType: eval
+image: gcr.io/kpt-fn/starlark:unstable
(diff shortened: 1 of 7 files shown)


droot (approved):
So this is a behavior change we need to record in our changelog for `v1`. Not sure how to track that.

Shell32-Natsu:
@droot I think we need to note it in the `fn eval` doc.

PR #1769, opened 2021-04-16
fn render can create resource in parent

### LINKED ISSUE #1744: kpt fn render generator functions duplicate resources generated in parent directories
See path-index-ancestor and path-index-outofpkg in https://github.com/GoogleContainerTools/kpt/pull/1699

@droot @mengqiy @Shell32-Natsu

natasha41575:
fix #1744 

This logic will have to change a little when we move the duplicate path detection closer to the functions, but for now it provides a fix for creating resources in the parent package. 

Also, this PR removes the duplicates before writing to output, but doesn't address why the duplicates are appearing.


internal/cmdrender/executor.go
```diff
@@ -375,25 +372,45 @@ func trackOutputFiles(hctx *hydrationContext) error {
...
 		fp = path.Clean(fp)
 		if pathIndexes[fp] == nil {
-			pathIndexes[fp] = make(map[string]bool)
+			pathIndexes[fp] = make(map[string]*yaml.RNode)
 		}
-		if _, ok := pathIndexes[fp][index]; ok {
-			return fmt.Errorf("resource at path %q and index %q already exists", fp, index)
+		if match, ok := pathIndexes[fp][index]; ok {
+			// path conflict, check if resource has the same content
+			if !duplicateNode(match, node) {
```
droot, line 393: This might encourage function authors to not worry about the function (especially generators) being idempotent. I am not sure if that is a good outcome.

internal/cmdrender/executor.go
```diff
@@ -375,25 +372,45 @@ func trackOutputFiles(hctx *hydrationContext) error {
...
+			}
+		} else {
+			result = append(result, node)
 		}
-		pathIndexes[fp][index] = true
+		pathIndexes[fp][index] = node
 	}
-	return nil
+	return result, nil
+}
```
droot, line 402: I am not able see the connection of this change with `fn render being able to create resource in parent` ? Can you elaborate more on that ?

droot (commented):
Some questions.

internal/cmdrender/executor.go
natasha41575: Maybe we can have a longer discussion about this since this doesn't seem like the right approach for this issue. 

Currently, fn render will create a duplicate of the resource when the path has prefix `../`. I think it is something related to being added to the input once as a transitive resource when processing the subpackage, and again when collecting all the resources in the current package. 

The fix here is to remove duplicates before outputting the resources. 

An alternate (and probably better) approach that I can work on is trying to make it not duplicate the resource in the first place, but that one will take a bit more time and investigation.

droot: > I think it is something related to being added to the input once as a transitive resource when processing the subpackage, and again when collecting all the resources in the current package.

You are right, we should root cause this issue that will help us address this correctly.

natasha41575: Closing this PR. I'll open a new one after further investigation.

PR #1775, opened 2021-04-19
kpt pipeline validation should use new format

### LINKED ISSUE #1754: kpt pipeline validation should return errors compatible with new error format.
We recently introduced `internal/errors` pkg and have partially used in `kpt fn render` command. The task here to use the same error format for the pipeline validations.

natasha41575:
fix #1754

### DIFF
--- a/pkg/api/kptfile/v1alpha2/validation.go
+++ b/pkg/api/kptfile/v1alpha2/validation.go
@@ -28,10 +28,9 @@ const (
 )
 
 func (kf *KptFile) Validate() error {
-	const op errors.Op = "kptfile.validate"
+	const op errors.Op = "v1alpha2.kptfile.validate"
 	if err := kf.Pipeline.validate(); err != nil {
-		err =  fmt.Errorf("pipeline is not valid: %w", err)
-		return errors.E(op, err)
+		return errors.E(op, fmt.Errorf("pipeline is not valid: %w", err))
 	}
 	// TODO: validate other fields
 	return nil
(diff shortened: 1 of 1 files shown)


pkg/api/kptfile/v1alpha2/validation.go
```diff
@@ -27,8 +28,10 @@ const (
 )
 
 func (kf *KptFile) Validate() error {
+	const op errors.Op = "kptfile.validate"
 	if err := kf.Pipeline.validate(); err != nil {
-		return fmt.Errorf("pipeline is not valid: %w", err)
+		err =  fmt.Errorf("pipeline is not valid: %w", err)
+		return errors.E(op, err)
```
Shell32-Natsu, line 34: Can we change this to `return errors.E(op, fmt.Errorf("pipeline is not valid: %w", err))`?

pkg/api/kptfile/v1alpha2/validation.go
```diff
@@ -27,8 +28,10 @@ const (
 )
 
 func (kf *KptFile) Validate() error {
+	const op errors.Op = "kptfile.validate"
```
Shell32-Natsu, line 31: Not sure how should we construct this `op` string. Should it be `[pkgName].[structName].[funcName]` or any string that can represent current operation?

natasha41575: Checked with Morten and he said the convention for now is <package name>.<function name>. I will update the PR.

pkg/api/kptfile/v1alpha2/validation.go

pkg/api/kptfile/v1alpha2/validation.go
natasha41575: updated it to include the package name

Shell32-Natsu: We may need to standardize the format of this string. And it will be good to document how to use this `errors` package. @droot @mortent

Shell32-Natsu (approved):

pkg/api/kptfile/v1alpha2/validation.go
natasha41575: I agree, this should be addressed as part of #1783 and #1557

PR #1780, opened 2021-04-21
inject openapi schema into kpt functions

natasha41575:
Combined with https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/236, this demonstrates a working POC for a version of the set-label function that gets its configuration from an augmented OpenAPI schema provided by kpt. 

It will help inform the design for https://github.com/GoogleContainerTools/kpt/issues/1608.


natasha41575:
@mengqiy

internal/util/openapi/augments/builtin.go
```diff
@@ -0,0 +1,105 @@
+package augments
+
+const JsonPatchBuiltin = `
+[
+  {"op": "add",
+    "path": "/definitions/io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta/properties/annotations/x-kubernetes-field-meaning",
```
mengqiy, line 6: It seems we need to add these kind of augment for every GVK.
Doing it manually doesn't seems to be good, since human may accidentally forget to do it for some types.
We may want to have a small tool to do these cross-cutting changes (e.g. .metadata.annotations, .metadata.labels and .metadata.namespace) for us eventually.

internal/util/openapi/openapi.go
```diff
@@ -77,14 +88,89 @@ func ReadOpenAPISchemaFromDisk(path string) ([]byte, error) {
...
+}
+
+func GetSchemaKRM() (*kyaml.RNode, error) {
+	schema, err := kyaml.Parse(`
+apiVersion: kpt.dev/v1
+kind: OpenAPI
+metadata:
+  name: openAPI
+  annotations:
+    config.kubernetes.io/path: "openapi_openAPI.yaml"
```
mengqiy, line 129: This is just a fake name right?

natasha41575: Yes, AFAICT this name doesn't matter since kpt will remove this resource from the output before writing anything out. We will need to have a regression test with a resource with the same file name.

internal/util/openapi/augments/builtin.go
natasha41575: Agreed, a tool for cross-cutting changes will be better than doing it manually.

natasha41575: A note: for anything in metadata, we only need to add it once. There is one definition for ObjectMeta referenced by all GVKs.

internal/cmdrender/executor.go
```diff
@@ -274,6 +274,12 @@ func (pn *pkgNode) runPipeline(ctx context.Context, input []*yaml.RNode) ([]*yam
 		return nil, errors.E(op, pn.pkg.UniquePath, err)
 	}
 
+	schemaKRM, err := openapi.GetSchemaKRM()
```
aodinokov, line 277: @natasha41575 
I'm monitoring this because I'm interested in something similar: I want to create validator that would have a shared info with kustomize about OpenApi Schema that was defined in kustomization.yaml. I saw your recent work on OpenApi field in kustomize and this looks like a follow up work. I even wanted to create a feature request and propose implementation with that, but now I see that you're working on that already :)

just a thought that this implementation may lead to incompatibility with the previous versions for KRM-functions that don't expect to see in the input:

```
apiVersion: kpt.dev/v1
kind: OpenAPI
metadata:
  name: openAPI
  annotations:
    config.kubernetes.io/path: "openapi_openAPI.yaml"
```

Shouldn't it be an additional annotation in FunctionConfig rather than items?
If it's a annotation it's not necessary to convert it to yaml, it can be json.
Another thought:
not all of functions require that, so it's possible to agree that kpt:next or kustomize would check that kpt-function config has that annotation that is empty and inject there OpenApi.json, e.g.

```
apiVersion: v1
kind: ConfigMap
metadata:
   name: kubevalConfig
   annotations:
      krm/OpenApiSchema: ""
...
```
would be converted to the config where krm/OpenApiSchema has a value that includes the whole OpenApi json. That json could be extracted by the krm-function and used for validation.

If this annotation is empty when everything works as usual.
I think this approach may allow to resolve the potential issues with unexpected documents in pipeline as well as unexpected data in FunctionConfig for functions that already exist...

natasha41575: @aodinokov Thank you for taking the time to look at this! We are considering a few different approaches to do this injection and this solution is just one of the options we are considering. I agree that having an unexpected resource in the input is not ideal. 

Another option we are considering is adding the schema as a new field in ResourceList (e.g. ResourceList.OpenAPI) rather than a new resource itself. We are also considering starting up an http server that the function container can curl in order to get the schema. 

The other thing you mentioned is whether we want to inject OpenAPI into every function, or just the ones that need it; we are giving this some thought as well. The issue with the latter is that it's the function developer that knows whether the function requires OpenAPI schema, not the package composer. It's problematic to give the package composer control over this. 

We are still weighing the pros and cons of each of these options, and we appreciate any other feedback you might have. Another thing to note is that this implementation will only affect kpt. Were you interested in seeing something similar in kustomize?

aodinokov: @natasha41575 Thank you for the quick response and providing so much information!

>starting up an http server that the function container can curl in order to get the schema

I've looked into that PR as well, but haven't looked deeper because from the first glance its architecture looks a bit too complicated to me :) : It reminds in some way k8s architecture - where you have API server + engine that runs containers  and that containers can access the server itself :) Anyway it's a good pattern (at least for servers) and maybe I'm wrong that it's an overcomplication. Need to think about it more.

Speaking more about OpenApi server -  will it be possible to run 2+ instances of kpt simultaneously? How krm function will know where to connect if there are 2+ servers? Another question - I guess for ip socket it will be necessary to allow network for container - seems like a disadvantage. BTW, I guess there can be also an option with **unix socket** rather than IP socket (I see it's now works on [8080 port](https://github.com/GoogleContainerTools/kpt/pull/1868/files#diff-d0e1ce0f7af25713f66ada2d783193b985ad98b1502855bf28e7d19a9e804b02R133)). In case of unix socket it's possible to add an additional 'bind' mount to docker run cmd. It should be a bit more secure than the current implementation and it doesn't break the API. It will just extend it. The question will be - how this will affect other types of functions (exec, starlark). Most likely starlark won't support that as-is. will need to add some function to access that. Exec will require that the path to the unix socket/address:port would be the same. So if we create a socket per RunFs instance or per Kpt instance we will have a unique name/path and we'll need to pass that path/(address:port in case of ip socket) to the function. There will be a question - how? :) Env-variable? Additional field in ResourceList? Still the good thing about OpenApiServer - is that it pull-model - only functions that need OpenApi will use that API.

>new field in ResourceList (e.g. ResourceList.OpenAPI)

I think this may look even better than the option with annotation. Old functions won't unmarshall that value and ignore that.
New functions will get that value. The bad thing - it's push-model. But we already have some functions fine-tuning things like `networking: true` for containers. I know that there were some inconvenience for users - some people didn't know how to turn it on, or used incorrect path (this API has changed couple of times). But anyway this may be done in the same way - just extend the function configuration with additional field `openAPI: true`. The next step - maybe it would be a good idea to generalize that to pass an arbitrary composer context and call it 'contextFilter: ["openApi"]' and the field in that case ResourceList.Context: {OpenAPI: "<json>",...} ... Interesting that under that angle it resembles the idea with additional annotations, but without additional ResourceList field. The composer will have a list of allowed annotations, it will scan the Function config looking for them and if found it with e.g. empty value - sets its value to the actual. The disadvantage is - when an user creates a function config he must set that annotations. If he forgot - nothing would work. BUT, the function implementation may check if all annotations required for work is set and if something is missed - just fail with the appropriate error.

>Another thing to note is that this implementation will only affect kpt. Were you interested in seeing something similar in kustomize?

Yes, I think once you decide on the interface for `kpt:next`, it would be great to backport that interface extension to kyaml.RunFn and use it for krm-based plugins to share more information with plugins. There is no need in 2 different approaches: KRM functions should work in the same/compatible way if possible.

My use-case what I would like to have was related to static analysis of yaml-documents using kubeval-based validator.
The current kustomize implementation introduced openApi field to override OpenApi schema and validator field where I can put kubeval krm function, but kubeval also requires some form of OpenApi schema. Right now it would be necessary to synchronize openApi field with configuration of kubeval krm-function, which is error-prone. I think it would be useful if kustomize could share OpenApi scheme with not only builtin plugins, but external ones as well. That would allow to get openApi schema in kubeval krm-function where it could be converted to the format needed for kubeval and in that form it can be used for validation. This could be a logical continuation of improving approach with krm-function. E.g. the same implementation would allow also to write analog of SMP plugin, but not as builtin but as krm function. That means that the difference between builtin and external plugins will be minimized, which in my opinion is a good thing.

natasha41575: >I've looked into that PR as well, but haven't looked deeper because from the first glance its architecture looks a bit too complicated to me :) : It reminds in some way k8s architecture - where you have API server + engine that runs containers and that containers can access the server itself :) Anyway it's a good pattern (at least for servers) and maybe I'm wrong that it's an overcomplication. Need to think about it more.

I'm inclined to agree with you here. We are still thinking about it, but adding the schema to ResourceList seems like a much simpler option to me. Need to weigh the pros and cons a bit more.

>will it be possible to run 2+ instances of kpt simultaneously? How krm function will know where to connect if there are 2+ servers? 

This is an excellent point. Thank you. I will bring it up in design discussion. The more I think about it, the more complications I see with starting a server. Unix socket is an interesting idea too, though you're right that we will have to pass the path/address:port to the function. If we did that, it would probably end up being a field in ResourceList. 

> But anyway this may be done in the same way - just extend the function configuration with additional field openAPI: true.

We are seriously considering this to avoid the massive size of stdin being sent to functions that don't require it, but we are still hesitant to give package composers this knob when it's the function developer who knows whether the schema is needed, not the package composer. Although I suppose the same could be said about the `networking: true` option. If the latency of injecting OpenAPI into functions is sufficiently high, we will have to take a step back and reevaluate if we really want to do it for every single function whether it is required or not. 

>  The next step - maybe it would be a good idea to generalize that to pass an arbitrary composer context and call it 'contextFilter: ["openApi"]' and the field in that case ResourceList.Context: {OpenAPI: "",...}

This is an interesting idea too. I will bring it up in our design discussions. This has the same problem as I mentioned above (the package composer will have to know if a function needs OpenAPI), but if we are willing to have them set a field for it then this is worth looking into as well. A downside: the function developer would have to fail with the appropriate error, or the poor user will have no idea what he did wrong. If we can, we should have kpt catch errors like this if it can rather than depending on function developers to do so. 

You've given me a lot to think about, I really appreciate it. I will follow up when we have a final design.

PR #1789, opened 2021-04-22
overview page

### LINKED ISSUE #1681: kpt.dev landing page
We have a stop-gap overview page:
https://kpt.dev/?id=overview

The end state should have a high production value. For comparison, see these landing pages:

https://www.terraform.io/
https://kustomize.io/
https://www.pulumi.com/

mikebz:
Draft of the overview page.  Intent here is to focus on end user problems and position kpt as the right tool for specific use cases that tend to resonate with platform teams.

image

Closes #1681

### DIFF
--- a/site/README.md
+++ b/site/README.md
@@ -7,32 +7,40 @@
 
 #### Problem space
 
-`kpt` is primarily targeted at engineers who are looking to do configuration
-and policy enforcement of their kubernetes infrastructure at scale.  `kpt` 
-design is aimed to solve the following problems:
+kpt is targeted at configuration and policy enforcement of kubernetes 
+infrastructure at scale.  kpt design is aimed to solve the 
+following problems:
 
-1. Quick creation of blueprints from existing configuration without 
-introducing a template or domain specific language into your infrastructure 
-stack
-2. Updating downstream forks of configuration packages with upstream 
-changes with minimal effort and repetitive work
-3. A command line (imperative) and configuration file (imperative) way to 
-customize the configuration and enforce policy
-4. Packaging up parts of customization and validation into reusable building
-blocks (KRM functions) and sharing them within your enteprise.
-5. Applying and pruning a group of resources in a cluster.
+1. Quick creation of configuration blueprints from existing YAML files 
+without introducing a template or domain specific language into your infrastructure stack.
+1. Updating downstream forks of configuration packages with upstream 
+changes with minimal effort and repetitive work.
+1. A command line (imperative) and configuration file (declarative) way to 
+customize the configuration and enforce policy.
+1. Packaging up parts of customization or validation pipeline into 
+reusable building blocks (KRM functions) and sharing them within your 
+enteprise.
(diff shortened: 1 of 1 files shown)


site/README.md
```diff
@@ -1,3 +1,46 @@
 ### Overview
 
-TODO(#1681)
+#### What is kpt?
```
frankfarzan, line 3: I was thinking overview shouldn't be a long description like this (Honestly it's a turn off for a casually interested user). That long explanation belongs in the chapter 2 of the book: https://kpt.dev/book/02-concepts.

This overview can either be:

a) Quick pointers to different sections of the page (i.e. an updated version of landing page at https://googlecontainertools.github.io/kpt/)
b) A quick demo video

frankfarzan: We can start with a) and add or replace it with b) later.

mikebz: I am happy to shorten this but I think what we are missing is "why read more".  if we look at npm or helm websites very quickly they tell you what problems they solve, and these tools are already known quantities for engineers.  I don't think that a video of a CLI really gets that across.  My thought is that the first page below the fold needs to get the "why" across in as little text as possible but position the tool against or with other config tools out there.

frankfarzan: NPM landing page is couple of splashy claims:

https://www.npmjs.com/

Terraform landing page is demoes and screenshots interleaved with brief value-add statements:

https://www.terraform.io/

I find Helm's landing page to be cluttered and confusing TBH, not the best example to follow.

If you want the WHY in the long-form instead of marketing material like that, then we should just move content from https://kpt.dev/book/02-concepts/ to this page for now.

mikebz: I think the overview page does need to be somewhat of a positioning and benefits page.  Examining the TF page below the fold (which maps to this overview page.  It does explain 3 things: write, plan and apply:
- concise language
- checking the configuration before applying
- support for hundreds of cloud providers for application

npm has a slightly different angle because they are already an established tool and therefore very quickly they go into the fact that they are "Gratefully serving everyone from solo devs to the Fortune 500" "Relied upon by more than 11 million developers worldwide, npm is committed to making JavaScript development elegant, productive, and safe."

We can shrink the overview but somehow we need to convince someone to read a book.

Above the fold we need to have this:
image

and below this:
image

The concepts part of the 'Book' is more detailed and takes time to understand.  I think it speaks to the user who has already decided to take the voyage.

frankfarzan: I don't disagree, instead of having 9 bullet points like this, we need something more concise and also visually appealing. BTW, we don't need to go to far, kustomize landing page has been nicely revamped:

https://kustomize.io/

Specifically: `#Overview` and `#Tutorial` (Book in our case).

mikebz: I think the best way to avoid having a TODO on the front page is just to push this in and then come up with a new version after.

frankfarzan: I'm going to put in a place-holder soon. It's also not a big deal to have TODOs.

PR #1792, opened 2021-04-22
Ensures ResourceGroup CRD applied before preview/apply on next branch

seans3:
* Ensures the `ResourceGroup` CRD is applied before the `preview` or `apply` commands are run. Fixes: https://github.com/GoogleContainerTools/kpt/issues/1787
* Fixes `apply` and `init` commands to default to current directory if there are no args (apply currently hangs): Fixes https://github.com/GoogleContainerTools/kpt/issues/1786
* TODO: Resolve duplicate RESTMapper reset code here and in `cli-utils`

### DIFF
--- a/commands/applycmd.go
+++ b/commands/applycmd.go
@@ -4,8 +4,6 @@
 package commands
 
 import (
-	"os"
-
 	"github.com/GoogleContainerTools/kpt/pkg/live"
 	"github.com/GoogleContainerTools/kpt/thirdparty/cli-utils/apply"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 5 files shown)


mikebz:
question - seems like no tests were updated.  Do we have a test for this scenario?

commands/applycmd.go
```diff
@@ -42,18 +40,22 @@ func (w *ApplyRunnerWrapper) Command() *cobra.Command {
...
-// "AlreadyExists" error for CRD installation.
+// RunE runs the ResourceGroup CRD installation as a pre-step.
 func (w *ApplyRunnerWrapper) RunE(cmd *cobra.Command, args []string) error {
-	if _, exists := os.LookupEnv(resourceGroupEnv); exists {
-		klog.V(4).Infoln("wrapper applyRunner detected environment variable")
-		err := live.ApplyResourceGroupCRD(w.factory)
-		if err != nil && !apierrors.IsAlreadyExists(err) {
+	// Install ResourceGroup CRD if it does not already exist, and reset
+	// the RESTMapper to make the CRD available.
+	err := live.ApplyResourceGroupCRD(w.factory)
```
mortent, line 47: This doesn't seem to wait for the CRD to become `Established`. Without it, the command might still report that the resource type doesn't exist: https://github.com/GoogleContainerTools/kpt/issues/1612

mortent (commented):
Can we add a TODO that we want to fix the duplicate code for resetting the RESTMapper?

commands/applycmd.go
frankfarzan: Somewhat related: Given this is a long-running operation, is there a user-facing message saying kpt is installing CRD?

seans3: @frankfarzan Currently, the CRD application is silent for first-time `preview` and `apply`, and it does not give any feedback during this operation. Should I add a message?

mortent: I think we should think about whether we want to install the RG CRD when someone is running the `preview` command. If I do preview, I don't expect any resources to be applied.

seans3:
Closed in favor of: https://github.com/GoogleContainerTools/kpt/pull/1817

PR #1797, opened 2021-04-22
use strict for kubeval in nginx example

mengqiy:


frankfarzan (approved):

package-examples/nginx/Kptfile
```diff
@@ -6,3 +6,5 @@ pipeline:
   validators:
   # TODO: This currently fails https://github.com/GoogleContainerTools/kpt/issues/1621
```
frankfarzan, line 7: remove

PR #1802, opened 2021-04-22
Remove unused flags for the update command

mortent:
This removes some unused flags from the `kpt pkg update` command:

 * repo: This would allow users to update from a different repo than the package was originally forked from. This is a difficult situation for merge and unlikely to be used much. 
 * dry-run: This was only used for the git-merge-patch strategy which we no longer support. We could consider a dry-run flag, but currently users can accomplish this easily by just doing the actual update and then remove the changes with git.
 * verbose: If we want to have more output, we should do it either using the klog library or have a solution that works across all kpt commands.

### DIFF
--- a/commands/livecmd.go
+++ b/commands/livecmd.go
@@ -18,8 +18,8 @@ import (
 	"os"
 
 	"github.com/GoogleContainerTools/kpt/internal/cmdfetchk8sschema"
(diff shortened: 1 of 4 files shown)


frankfarzan:
Does reference docs need to be updated?

mortent:
> Does reference docs need to be updated?

No, the reference doc PR already assumes these flags will go away.

frankfarzan (approved):

PR #1803, opened 2021-04-23
moving to alpha2

### LINKED ISSUE #1645: Convert or delete all Kptfile kpt.dev/v1alpha1 from the next branch
This includes either deleting or converting examples and docs to use v1alpha2

mikebz:
### What
Removing all the kpt.dev/alpha1 and replacing them with kpt.dev/alpha2

### Testing 
unit test and e2e tests executed

Closes #1645


internal/testutil/testdata/updateMergeConflict/Kptfile
```diff
@@ -1,4 +1,4 @@
-apiVersion: kpt.dev/v1alpha1
+apiVersion: kpt.dev/v1alpha2
```
frankfarzan, line 1: Similar comment. The schema has changed. There's no `upstream.git.commit` any more.

@mortent Please review the use of these testdata.

demos/07-sync-run-0/Kptfile
```diff
@@ -1,4 +1,4 @@
-apiVersion: kpt.dev/v1alpha1
+apiVersion: kpt.dev/v1alpha2
```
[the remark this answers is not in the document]
frankfarzan: https://github.com/GoogleContainerTools/kpt/issues/1654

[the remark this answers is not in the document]
mikebz: Yes, I was a bit conflicted about whether to remove this content or move it.  I am happy removing it but migrating all the content is a fairly big job so I think I'd rather prune rather than migrate and then prune.  Does that seem like an OK plan to you?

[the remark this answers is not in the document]
frankfarzan: yep, delete it please. I commented on that issue as well.

internal/testutil/testdata/updateMergeConflict/Kptfile
mikebz: I want to make sure we understand the scope of what is deleted.  helloworld-set folder is deleted but it seems like there are no tests failing.  That leads me to believe that this test data was not used in any of the current tests, just something that needed to be cleaned up for a while.

We might argue that whenever we deleted the test that was using this test data that we might have lost some coverage, but that's not related to this change.

frankfarzan:
I agree, this doesn't block this PR.

internal/pkg/pkg_test.go
```diff
@@ -129,7 +129,7 @@ kind: Custom
 spec:
   image: nginx:1.2.3`,
 				`
-apiVersion: kpt.dev/v1alpha1
+apiVersion: kpt.dev/v1alpha2
```
frankfarzan, line 132: the schema here needs to change, it's using `setterDefinitions`

mikebz: ahhh... good catch, just saw this comment. Ignore the re-request.

mikebz: fixed now and I checked other places and didn't find any out of date definitions.

internal/pkg/pkg_test.go
```diff
@@ -129,18 +129,18 @@ kind: Custom
...
+pipeline:
+  mutators:
+    - image: gcr.io/kpt-fn/apply-setters:unstable
+      configMap:
+        replicas: 5
+`,
 				`
-apiVersion: kpt.dev/v1alpha1
+apiVersion: kpt.dev/v1alpha2
 kind: Pipeline
```
frankfarzan, line 144: we don't have a `Pipeline` kind anymore, maybe file an issue for clean up here?

frankfarzan (approved):
there's some cruft requiring cleanup, can be separate issue

internal/pkg/pkg_test.go
mikebz: sorry merged this before seeing this comment.  We should do a cleanup overall.

PR #1806, opened 2021-04-23
update fn render path-index test cases

### LINKED ISSUE #1744: kpt fn render generator functions duplicate resources generated in parent directories

natasha41575:
Close #1744

These tests were failing with the error that there were duplicate http-bin resources. This is because the functions that generate resources outside of their current package can't be idempotent; they have no means of checking if those resources already exist. As a result, running the test twice resulted in a path annotation conflict on the second run; the function tried to create a resource with the same path and index as the resource created in the first run. 

This PR updates the test cases to reflect the non-idempotency of these functions.


natasha41575:
@Shell32-Natsu PTAL

Shell32-Natsu (approved):
lgtm

frankfarzan:
This tests shouldn't be passing in the first place. Next time please wait for @droot or myself to review such changes.

PR #1814, opened 2021-04-25
Resolve errors to user-facing messages at the top level

mortent:
This is a POC for how we can keep the general error handling solution (which prints the trace to the user) with a way to have better actionable and informative errors for "expected" error situations. This solution allows us to handle errors at the cobra level (so avoid printing of error message scattered through the codebase), provide clear error messages when possible, and fall back to the full trace for errors that we "don't expect".

### DIFF
--- a/internal/errors/resolver/resolvers.go
+++ b/internal/errors/resolver/resolvers.go
@@ -1,0 +1,93 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package resolver
+
+import (
+	goerrors "errors"
+	"fmt"
+	"strings"
+
+	"github.com/GoogleContainerTools/kpt/internal/gitutil"
+)
+
+//nolint:gochecknoinits
+func init() {
+	AddErrorResolver(&gitExecErrorResolver{})
+}
+
+const (
+	genericGitExecError = `
+Error executing git command {{ printf "%q " .gitcmd }}
+{{- if gt (len .repo) 0 -}}
+against repo {{ printf "%q " .repo }}
+{{- end }}
+{{- if gt (len .ref) 0 -}}
+for reference {{ printf "%q " .ref }}
+{{- end }}
+
+{{- if gt (len .stdout) 0 }}
+{{ printf "\nStdOut:" }}
+{{ printf "%s" .stdout }}
+{{- end }}
+
+{{- if gt (len .stderr) 0 }}
+{{ printf "\nStdErr:" }}
+{{ printf "%s" .stderr }}
+{{- end }}
+`
+
+	unknownRefGitExecError = `
+Unknown ref {{ printf "%q" .ref }}. Please verify that the reference exists in upstream repo {{ printf "%q" .repo }}.
+
+{{- if gt (len .stdout) 0 }}
+{{ printf "\nStdOut:" }}
+{{ printf "%s" .stdout }}
+{{- end }}
+
+{{- if gt (len .stderr) 0 }}
+{{ printf "\nStdErr:" }}
+{{ printf "%s" .stderr }}
+{{- end }}
+`
+)
+
+// gitExecErrorResolver is an implementation of the ErrorResolver interface
+// that can produce error messages for errors of the gitutil.GitExecError type.
+type gitExecErrorResolver struct{}
+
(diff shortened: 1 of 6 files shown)


droot:
After looking at the code, an alternative approach (slightly tweaked) hit me:

Error can implement `Resolver` interface to give a hint to top level error handler that this is how this error needs to be resolved for the CLI purpose etc. Top level default error handler can inspect the entire error chain and if it finds an error of `Resolver` type, then it calls `Resolve` method and prints output to the user.

We can provide some syntax sugar to make it easy to construct errors of `Resolver` type. [In @Shell32-Natsu PR where it is printing `fn error` inline, it can return an error with NoOp essentially signalling the default handler that no need to do other than non-zero exitcode]

So in this case, the the `gitError` can implement the `Resolve` method.

Last time, when I experimented with `UserText` type in the `errors`, it was sort of a way to short-circuit (or resolve) the error. and I found it convenient to be able to control the CLI output closer to where the error happened.  

Happy to chat offline.

frankfarzan:
Can you provide some sample output in the PR description. For example:
- function failure in `eval`
- git-related failure in `get`

mortent:
I have only added support for git-related errors in this PR. It mostly just puts in place the plumbing to translate from a kpt error to a more user-friendly message. 
For the situation where a user provides an unknown ref with the `kpt pkg get` command, the output looks as follows:

```
$ kpt pkg get https://github.com/GoogleContainerTools/kpt/package-examples/nginx@nex
cloning https://github.com/GoogleContainerTools/kpt@nex

Unknown ref "nex". Please verify that the reference exists in upstream repo "https://github.com/GoogleContainerTools/kpt".

StdErr:
fatal: ambiguous argument 'nex': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]' 
```

internal/errors/resolver/resolvers.go
```diff
@@ -0,0 +1,93 @@
...
+	switch {
+	// TODO(mortent): Checking the content of the output at this level seems a bit awkward. We might
+	// consider doing this the the gitutil package and use some kind of error code to signal
+	// the different error cases to higher levels in the stack.
+	case strings.Contains(gitExecErr.StdErr, " unknown revision or path not in the working tree"):
+		return ExecuteTemplate(unknownRefGitExecError, tmplArgs)
+	default:
+		return ExecuteTemplate(genericGitExecError, tmplArgs)
+	}
+}
```
droot, line 93: Can we add a resolver for the cases where error output is already handled as part of the command ? This will be needed by the PR @Shell32-Natsu is working on related to `fn` output.

I am guessing two things will be required for that:
```
// An error type to wrap the original error
type NoOpErr struct{ err Error }

// Resolver that handles NoOpErr
type NoOpErrResolver struct{}// other possible names EmptyErrResolver, ZeroErrResolver

func (er NoOpErrResolver) Resolve (err error) (...) {
  // checks if its NoOpErr, marks it resolved
}
```

droot (approved):
Looks good to me. I suggested adding a NoOpErr and a resolver, but its okay to add in a follow up PR as well.

frankfarzan:
> I have only added support for git-related errors in this PR. It mostly just puts in place the plumbing to translate from a kpt error to a more user-friendly message.
> For the situation where a user provides an unknown ref with the `kpt pkg get` command, the output looks as follows:
> 
> ```
> $ kpt pkg get https://github.com/GoogleContainerTools/kpt/package-examples/nginx@nex
> fetching package package-examples/nginx from https://github.com/GoogleContainerTools/kpt to /Users/<user>/nginx
> cloning https://github.com/GoogleContainerTools/kpt@nex
> 
> Unknown ref "nex". Please verify that the reference exists in upstream repo "https://github.com/GoogleContainerTools/kpt".
> 
> StdErr:
> fatal: ambiguous argument 'nex': unknown revision or path not in the working tree.
> Use '--' to separate paths from revisions, like this:
> 'git <command> [<revision>...] -- [<file>...]' 
> ```

I would tweak this for readability:

```
$ kpt pkg get https://github.com/GoogleContainerTools/kpt/package-examples/nginx@nex
cloning https://github.com/GoogleContainerTools/kpt@nex

Error: Unknown ref "nex". Please verify that the reference exists in upstream repo "https://github.com/GoogleContainerTools/kpt".

Details:
fatal: ambiguous argument 'nex': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]' 
```

mortent:
> I would tweak this for readability:
> 
> ```
> $ kpt pkg get https://github.com/GoogleContainerTools/kpt/package-examples/nginx@nex
> fetching package package-examples/nginx from https://github.com/GoogleContainerTools/kpt to /Users/<user>/nginx
> cloning https://github.com/GoogleContainerTools/kpt@nex
> 
> Error: Unknown ref "nex". Please verify that the reference exists in upstream repo "https://github.com/GoogleContainerTools/kpt".
> 
> Details:
> fatal: ambiguous argument 'nex': unknown revision or path not in the working tree.
> Use '--' to separate paths from revisions, like this:
> 'git <command> [<revision>...] -- [<file>...]' 
> ```

Adopted these changes in the PR.

PR #1817, opened 2021-04-26
Apply ResourceGroup CRD using a Tasks

seans3:
* Apply `ResourceGroup` CRD using the following steps:
  1. Apply the `ResourceGroup` CRD
  2. Wait for the `ResourceGroup` CRD to become `Established` on the API Server. Ensures there is not a race condition.
  3. Resets the `RESTMapper` so it will have the `ResourceGroup` CRD.
* Adds this new function to `apply`, `migrate`, and `install-resource-group` commands.
* TODO: Add end-to-end tests to validate functionality.


pkg/live/apply-crd-task.go
```diff
@@ -0,0 +1,78 @@
...
+	"k8s.io/cli-runtime/pkg/resource"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util"
+	"sigs.k8s.io/cli-utils/pkg/apply/taskrunner"
+)
+
+// ApplyCRDTask encapsulates information necessary to apply a
+// Custom Resource Definition (CRD) as a task within a task queue.
+// Implements the Task interface.
+type ApplyCRDTask struct {
```
mortent, line 28: Do we need a new Task type for this? I would think that just using the existing `ApplyTask` should work.

seans3: I tried that, but it assumes an inventory object (for InventoryPolicy) which is not involved in this operation.

commands/applycmd.go
```diff
@@ -24,7 +24,12 @@ func GetApplyRunner(provider provider.Provider, loader manifestreader.ManifestLo
 	w := &ApplyRunnerWrapper{
 		applyRunner: applyRunner,
 		factory:     provider.Factory(),
+		ioStreams:   ioStreams,
+		applyCRD:    false,
 	}
+	cmd := applyRunner.Command
+	cmd.Flags().BoolVar(&w.applyCRD, "apply-inventory-crd", false,
```
mortent, line 31: I think something like `install-resource-group` would be a better name for the flag here as it mirrors the name of the command we already have.

commands/applycmd.go
```diff
@@ -35,25 +40,56 @@ func GetApplyRunner(provider provider.Provider, loader manifestreader.ManifestLo
 type ApplyRunnerWrapper struct {
```
mortent, line 40: Do we still need the wrapper now that we have moved the cobra commands from cli-utils into the kpt repo?

commands/applycmd.go
```diff
@@ -35,25 +40,56 @@ func GetApplyRunner(provider provider.Provider, loader manifestreader.ManifestLo
...
-			return err
-		}
+	if w.applyCRD {
+		// Force the apply of the inventory ResourceGroup CRD before applying.
+		// This does nothing (no error) if the CRD already exists.
+		w.ApplyResourceGroupCRD()
+	} else if !live.ResourceGroupCRDApplied(w.factory) {
+		// Otherwise, report the inventory ResourceGroup if missing.
+		fmt.Fprintln(w.ioStreams.Out, "inventory ResourceGroup CRD is missing")
+		fmt.Fprintln(w.ioStreams.Out, "run 'kpt live install-resource-group' to remedy")
```
mortent, line 65: Should there be a return here? Also, should we handle this error through the centralized error handler?

commands/applycmd.go
```diff
@@ -35,25 +40,56 @@ func GetApplyRunner(provider provider.Provider, loader manifestreader.ManifestLo
...
+			} else {
+				succeeded = false
+			}
+		}
+	}
+	// Print success or failure output for ResourceGroup apply. A failure
+	// must NOT stop the following 'applyRunner.RunE'.
+	if firstApply {
+		if succeeded {
+			fmt.Fprintf(w.ioStreams.Out, "applying inventory ResourceGroup CRD...success\n")
```
mortent, line 89: Nit: This looks a bit strange as the output suggests we are printing this before we actually make the change.

thirdparty/cli-utils/preview/cmdpreview.go
```diff
@@ -169,3 +185,32 @@ func (r *PreviewRunner) RunE(cmd *cobra.Command, args []string) error {
 	printer := printers.GetPrinter(r.output, r.ioStreams)
 	return printer.Print(ch, drs)
 }
+
+// ApplyResourceGroupCRD applies the inventory ResourceGroup CRD, printing
+// output if this is the first time the CRD apply happened.
+func (r *PreviewRunner) ApplyResourceGroupCRD() {
```
mortent, line 191: This seems like the same code as in the `ApplyRunnerWrapper`. Can we avoid the duplication?

commands/applycmd.go
seans3: I think we can get rid of the wrapper. I will do that in a follow-on PR.

commands/applycmd.go
seans3: Since the `Run` will error out immediately, we can continue or return here. Let me know what you'd like.

commands/applycmd.go
seans3: I don't quite understand this comment.

commands/applycmd.go
seans3: Done

commands/applycmd.go
seans3: After running e2e tests, I realized returning here would break legacy `ConfigMap` inventory applies. So we can not return here.

commands/applycmd.go
seans3: Fixed.

thirdparty/cli-utils/preview/cmdpreview.go
seans3: Fixed by moving more code into `inventoryrg.go`. The printing is still left to the commands instead of the library.

commands/applycmd.go
mortent: That sounds strange. Does that mean that if we use the legacy inventory, kpt will complain about missing ResourceGroup CRD?

seans3: Yes.

mortent: Can we fix that? It sounds like a pretty confusing UX.

seans3: Yes. I was hoping to fix it in a follow-on PR.

mortent (approved):

PR #1830, opened 2021-04-28
book: chapter 3

frankfarzan:
Part of https://github.com/GoogleContainerTools/kpt/issues/1573


site/book/03-packages/01-getting-a-package.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+The `Kptfile` contains two sections to keep track of the upstream package:
+
+1. The `upstream` section contains the user-specified reference to the upstream package. This
+   contains three pieces of information:
+   - `repo`: The git repository where the package can be found
+   - `directory`: The directory within the git repository where this package can be found
+   - `ref`: The git reference for the package. This can be either a brach, tag, or commit SHA.
+2. The `upstreamLock` section contains the resolved upstream reference. This section is managed by
```
mortent, line 60: I think the word `resolved` here is a bit ambiguous. Could we phrase this around the fact that this defines the point at which the local fork and the upstream source package diverged?

site/book/03-packages/01-getting-a-package.md
```diff
@@ -0,0 +1,118 @@
...
+   kpt and should not be changed manually.
+
+So essentially, the `upstream` section defines the "desired state" while the `upstreamLock` section
+defines the “current state”.
```
mortent, line 64: Similar to above, I think we need to explain in more details here what it means the the `upstreamLock` section defines the current state.

site/book/03-packages/01-getting-a-package.md
```diff
@@ -0,0 +1,118 @@
...
+      configMap:
+        ms-image: mysql
+        ms-tag: 5.6
+```
+
+As you can see, this `Kptfile` doesn't have the `upstream` and `upstreamLock` sections.
+This is because there are two different package types in kpt:
+
+- **Independent package:** A package where the `Kptfile` has `upstream` defined.
+- **Dependent package:** A package where the `Kptfile` doesn’t have `upstream` defined.
```
mortent, line 89: There is also the situation where a package created from scratch will not have any `upstream` defined. I do like the explanation here for the situation where the package is a subpackage of an independent package.

site/book/03-packages/02-exploring-a-package.md
```diff
@@ -0,0 +1,35 @@
...
+
+```shell
+$ kpt pkg tree wordpress/
+PKG: wordpress
+├── [Kptfile]  Kptfile wordpress
+├── [service.yaml]  Service wordpress
+├── deployment
+│   ├── [deployment.yaml]  Deployment wordpress
+│   └── [volume.yaml]  PersistentVolumeClaim wp-pv-claim
+└── PKG: mysql
```
mortent, line 17: Not directly related to this, but it would be useful if the `tree` command showed whether a subpackage is dependent or independent.

site/book/03-packages/04-updating-a-package.md
```diff
@@ -0,0 +1,48 @@
+Since a forked package contains a reference to its upstream package, kpt can fetch any update to the
```
mortent, line 1: Not only the upstream package, but the exact commit at which point the local fork and the upstream package diverged (i.e. the common ancestor).

site/book/03-packages/06-composing-a-package.md
```diff
@@ -0,0 +1,59 @@
...
+# author resources in mysql
+```
+
+This creates a [dependent package].
+
+### Get an existing package
+
+```shell
+$ cd wordpress
+$ kpt pkg get https://github.com/GoogleContainerTools/kpt.git/package-examples/wordpress/mysql@package-examples/wordpress/v0.1
```
mortent, line 48: I'm not sure if this is the best example here. Would it be better if we pulled in a subpackage from a completely separate repo?

site/book/02-concepts/02-package-lifecycle.md
```diff
@@ -19,8 +19,8 @@ First, you get a package from upstream. Then, you explore the content of the pac
 it better. Then you typically want to customize the package for you specific needs. Finally,
 you render the package which produces the final resources that can be directly applied to the
 cluster. Render is a required step even if you may not have declared a function in the package
-yourself since the upstream package hierarchy may contain a function declaration. The package may
-also have packages that have declared a function.
+yourself since the upstream package hierarchy may contain a function declaration. In addition,
```
mikebz, line 22: is it *yourself* or *itself*

site/book/03-packages/01-getting-a-package.md
```diff
@@ -0,0 +1,118 @@
+Packaging in kpt based on Git forking. Packages are published by committing them to a Git repository
+and packages are consumed by creating a new fork of the package.
```
mikebz, line 2: A lot of passive voice here.  Suggestion:
> Producers publish packages by committing them to a Git repository.  Consumers fork the package to use it.

site/book/03-packages/01-getting-a-package.md
```diff
@@ -0,0 +1,118 @@
...
+Let's revisit the Wordpress example:
+
+```shell
+$ kpt pkg get https://github.com/GoogleContainerTools/kpt.git/package-examples/wordpress@v0.1
+```
+
+A package on a Git repo can be fetched by specifying a branch, tag, or a commit SHA. In this case,
+we are specifying tag `v0.1`.
+
+> Refer to the [command reference][get-doc] for more details.
```
mikebz, line 13: is this a quote?  Using quoting markdown.  Not sure what is quoted.

frankfarzan: This is a hack for generating a Notes/Info box.  We need a proper solution here @etefera. Hugo supported this. e.g. see `pageinfo` directive in 

https://github.com/GoogleContainerTools/kpt/blob/master/site/content/en/guides/consumer/set/_index.md

What's the equivalent way to do it with the new site?

etefera: We can simulate the `pageinfo` directive if need be, but this is an example of the lock-in we may want to avoid. If the approach here is *too* hacky, it may be preferable to use `<div class="pageinfo">` directly so that the rendered markdown is more portable. We can then style it independently or process it however else we want.

site/book/03-packages/03-editing-a-package.md
```diff
@@ -0,0 +1,49 @@
...
+
+At the end of the day, editing a package will result in a Git commit that fully specifies
+the package. This process can be manual or automated depending on your use case.
+
+We will cover package composition later in this chapter. For now, let's focus on editing resources
+_within_ a package.
+
+## Manual edits
+
+As mentioned earlier, you can manually edit or _author_ KRM resources using your favorite editor.
```
mikebz, line 21: these terms are highlighted but I am not sure why.  Are these concepts, that are defined else where?  Do we want the user to somehow focus on these?  It's not a big deal but just trying to map this pattern to other documents that I have seen.

site/book/03-packages/04-updating-a-package.md
```diff
@@ -0,0 +1,48 @@
...
+upstream package and merge it with local changes.
+
+The package can be updated with changes from upstream in a declarative way by updating the
+`upstream` section in the `Kptfile` then running the `update` command.
+
+For example, change the value of `upstream.git.ref` to `v0.2`:
```
mikebz, line 7: Not a blocker but I'd go with the CLI update first.  In other package managers the file format is a black box and most of the people probably expect to update with the CLI first.

mikebz (commented):
A few comments but in general I would not hold up this chapter for the comments that I have.  Given that the doc update on an unpublished site is easy and the number of people who can easily read, comment and file bugs increases my strategy on these is to push and then make adjustments after.

Morten might have other ideas on how to approach this and he is actively reviewing this. So I am leaving the approval to him.

LGTM

site/book/03-packages/01-getting-a-package.md
```diff
@@ -0,0 +1,118 @@
+Packaging in kpt based on Git forking. Packages are published by committing them to a Git repository
```
droot, line 1: `is` missing

site/book/03-packages/01-getting-a-package.md
```diff
@@ -0,0 +1,118 @@
...
+and packages are consumed by creating a new fork of the package.
+
+Let's revisit the Wordpress example:
+
+```shell
+$ kpt pkg get https://github.com/GoogleContainerTools/kpt.git/package-examples/wordpress@v0.1
+```
+
+A package on a Git repo can be fetched by specifying a branch, tag, or a commit SHA. In this case,
```
droot, line 10: s/on/in

site/book/03-packages/01-getting-a-package.md
```diff
@@ -0,0 +1,118 @@
...
+the following fetches the packages to a directory names `mywordpress`:
+
+```shell
+$ kpt pkg get https://github.com/GoogleContainerTools/kpt.git/package-examples/wordpress@v0.1 mywordpress
+```
+
+The _name_ of a package is given by its directory name. Since the Kptfile is a KRM resource and
+follows the familiar structure of KRM resources, the name of the package is also available from the
+`metadata.name` field. This must always be the same of the directory name and kpt will update it
+automatically when forking a package. In this case, `metadata.name` is set `mywordpress`.
```
droot, line 108: This makes me wonder if end users have to really worry at all about the `metadata.name` ? It seems to be completely internal to `kpt` right ?

droot (approved):
Reading the book content first time, its reading pretty well, I like the easy conversation style.
Just a few minor typos.

site/book/03-packages/01-getting-a-package.md
mortent: I don't think we want to force users into a flow that involves a `publish` command, so I think we need a solution that doesn't solely rely on enforcement during publishing.

I do think we want a `validate` command that can handle this and any other issues in the structure/configuration of a package. This would also be useful for users as a git commit hook or as part of a CI pipeline.
The validation can also be part of other commands, we just have to decide if a violation here is fatal or we just provide a warning.

Another alternative would be that we just stop setting the name in the Kptfile altogether as part of the init, get and update commands, and we just say that the name of a package is given by the folder. If users want to set it themselves, they can but we simply ignore it. Forcing them to be the same seemed like a good idea at the time, but maybe it is easier if we just completely ignore the `metadata.name` field.

mortent (approved):

site/book/03-packages/01-getting-a-package.md
frankfarzan: One thing to think about: what other checks would be part of the validation? My intuition is to make `render` the one-stop shop for ensuring package is in a valid and reconciled state, but let's discuss if there's a use case to have a "validate" or "vet" command that's separate.

PR #1839, opened 2021-04-29
update test fn-render/multiple-fnconfig

natasha41575:
This test case passes so we don't need to skip it

### DIFF
--- a/e2e/testdata/fn-render/multiple-fnconfig/.expected/config.yaml
+++ b/e2e/testdata/fn-render/multiple-fnconfig/.expected/config.yaml
@@ -13,4 +13,4 @@
 # limitations under the License.
 
 exitCode: 1
-skip: true # TODO: enable once the testcase is supported
+stdErr: "following fields are mutually exclusive: 'config', 'configMap', 'configPath'. Got \"configPath, configMap\""


mikebz:
LGTM

e2e/testdata/fn-render/multiple-fnconfig/.expected/config.yaml
```diff
@@ -13,4 +13,4 @@
 # limitations under the License.
 
 exitCode: 1
-skip: true # TODO: enable once the testcase is supported
+stdErr: "following fields are mutually exclusive: 'config', 'configMap', 'configPath'. Got \"configPath, configMap\""
```
droot, line 16: We should address the following in a followup PR:

Improve the error message to fit our error-style ?

Wondering how to convey `oneOf` fields in more user friendly way: One option could be:
"function config must have only `config` or `configMap` or `configPath` input"

We also need the pkg and function involved in this error to make it really actionable. I have some ideas on introducing a new `validationError`, let have a chat offline about that.

droot (approved):
Approving this to enable the testcase, but I would really like the error message improvement as a followup PR.

e2e/testdata/fn-render/multiple-fnconfig/.expected/config.yaml
frankfarzan: Yes, every message needs to meet go/kpt-errors-style-guide.

In this case, I see the following issues:
1. We should tell users why they CAN do. They can specify only one of the 3 fields.
2. Quotation issue: Why is `"configMap, configMap"`

natasha41575: Added a comment about this to #1557; let me know if there should be a separate issue instead.

Shell32-Natsu: For a specific error message I think we should create a separate issue otherwise #1557 will be longer and longer.

PR #1854, opened 2021-04-30
Update 01-getting-a-package.md

mikebz:
wordpress was not mentioned anywhere prior, revisit doesn't make sense.


site/book/03-packages/01-getting-a-package.md
```diff
@@ -1,7 +1,7 @@
 Packaging in kpt is based on Git forking. Producers publish packages by committing them to a Git
 repository. Consumers fork the package to use it.
 
-Let's revisit the Wordpress example:
+Let's start with a wordpress example:
```
frankfarzan, line 4: It was first introduced here:

https://kpt.dev/book/02-concepts/01-packages

PR #1871, opened 2021-05-01
Better errors for git failures

mortent:
Improve the way we handle errors from git by resolving the particular situation as soon as possible after executing the command. The resolver just determines the error text based on the type. This also adds a few new cases so we can provide a more user-friendly error message for additional cases involving git issues.

Finally, this makes sure we invoke all git commands with the `GIT_TERMINAL_PROMPT` env variable set to `0`. This prevents git from prompting the user for username/password when accessing private repos using the https protocol.

Fixes: https://github.com/GoogleContainerTools/kpt/issues/1846

### DIFF
--- a/internal/errors/resolver/git.go
+++ b/internal/errors/resolver/git.go
@@ -55,7 +55,7 @@ Error: No git executable found. kpt requires git to be installed and available i
 `
 
 	httpsAuthRequired = `
-Error: Repository {{ printf "%q" .repo }} requires authentication. Kpt does not support this for the 'https' protocol. Please use the 'git'' protocol instead.
+Error: Repository {{ printf "%q" .repo }} requires authentication. Kpt does not support this for the 'https' protocol. Please use the 'git' protocol instead.
 
 {{- template "ExecOutputDetails" . }}
 `
@@ -67,7 +67,7 @@ Error: Unable to access repository {{ printf "%q" .repo }}.
 `
 
 	repositoryNotFound = `
-Error: Repository {{ printf "%q" .repo }} is unavailable.
+Error: Repository {{ printf "%q" .repo }} not found.
 
 {{- template "ExecOutputDetails" . }}
 `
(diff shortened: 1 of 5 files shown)


frankfarzan:
For such PR(s), please provide before and after output in the description.

mortent:
Here are some examples of what the output will look like with this change:

Fetching from a repository that doesn't exist:
```
$ kpt pkg get https://google.com/foobar.git                                                           

Error: Repository "https://google.com/foobar" not found.

Details:
fatal: repository 'https://google.com/foobar/' not found
```

Fetching from a private repo with the https protocol:
```
$ kpt pkg get https://github.com/mortent/kpt-packages/kafka

Error: Repository "https://github.com/mortent/kpt-packages" requires authentication. Kpt does not support this for the 'https' protocol. Please use the 'git' protocol instead.

Details:
fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

Fetching while offline:
```
$ kpt pkg get https://github.com/GoogleContainerTools/kpt/package-examples/nginx@next              
cloning https://github.com/GoogleContainerTools/kpt@next

Error: Unable to access repository "https://github.com/GoogleContainerTools/kpt".

Details:
fatal: unable to access 'https://github.com/GoogleContainerTools/kpt/': Could not resolve host: github.com
```

internal/errors/resolver/git.go
```diff
@@ -39,33 +39,37 @@ against repo {{ printf "%q " .repo }}
...
+Error: No git executable found. kpt requires git to be installed and available in the path.
 
-{{- if gt (len .stderr) 0 }}
-{{ printf "%s" .stderr }}
-{{- end }}
+{{- template "ExecOutputDetails" . }}
+`
+
+	httpsAuthRequired = `
+Error: Repository {{ printf "%q" .repo }} requires authentication. Kpt does not support this for the 'https' protocol. Please use the 'git' protocol instead.
```
frankfarzan, line 58: D5 https://docs.google.com/document/d/1DGN_1VWechH0agxExQxDYpKVSLqE5r1yKd95m_rE3JQ/edit?hl=en#heading=h.2h2qg2c12zy

mortent: Fixed

frankfarzan (approved):

PR #1875, opened 2021-05-03
remove extra new line before error messages

Shell32-Natsu:
The diff format is updated according to #1813


e2e/testdata/fn-render/success-stdout/.expected/diff.patch
```diff
@@ -1,15 +1,14 @@
 diff --git a/resources.yaml b/resources.yaml
-index 7a494c9..c77c357 100644
+index 7a494c9..a9dd224 100644
```
mortent, line 2: Why this change? There doesn't seem to be anything in the rest of the PR that would change the format of the output.

Shell32-Natsu: I think the output format is changes in #1813

mortent (approved):

PR #1879, opened 2021-05-03
Fix bad generated docs errors.

etefera:


droot:
@etefera Can we run `make generate` and update the generated files. It will also help us verify the change.

etefera:
> @etefera Can we run `make generate` and update the generated files. It will also help us verify the change.
Pushed

internal/docs/generated/fndocs/docs.go
```diff
@@ -42,8 +28,7 @@ var FnExamples = `
 
 var DocShort = `Display the documentation for a function`
 var DocLong = `
-` + "`" + `kpt fn doc` + "`" + ` invokes ` + "`" + `docker run` + "`" + ` to extract the help text. It requires the
-image to set the default entrypoint which supports ` + "`" + `--help` + "`" + ` flag.
+` + "`" + `kpt fn doc` + "`" + ` invokes the function container with ` + "`" + `--help` + "`" + ` flag.
 
   kpt fn doc --image=IMAGE
```
droot, line 34: why is the license header gone ? Did you use `make generate` because that invokes add-license to add the license ?

droot (changes requested):

internal/docs/generated/fndocs/docs.go
etefera: Yes, this is a bug tracked here https://github.com/GoogleContainerTools/kpt/issues/1403
People have been working around it by manually adding the license header, which we can do here if that's preferred.

droot: Got it. In that case, let's move away from license requirement for generated files.

droot (approved):

droot (approved):
Awesome!

frankfarzan (approved):

PR #1887, opened 2021-05-04
Improve error messages to users for kpt pkg update

mortent:
This PR updates the error message shown to users when the local package is either not in a git repository, or the git repository has uncommitted files.

Output when no git repo:
```
$ kpt pkg update kafka
updating package /Users/<user>/kafka
Package "/Users/<user>/kafka" is not within a git repository. Please initialize a repository using 'git init' and then commit the changes using 'git commit -m "<commit message>"'. 
```

Output when repo has uncommitted files:
```
$ kpt pkg update kafka
updating package /Users/<user>/kafka
Package "/Users/<user>/kafka" contains uncommitted changes. Please commit the changes using 'git commit -m "<commit message>"'. 
```

Fixes: https://github.com/GoogleContainerTools/kpt/issues/1852

### DIFF
--- a/internal/errors/resolver/update.go
+++ b/internal/errors/resolver/update.go
@@ -1,0 +1,66 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


internal/errors/resolver/update.go
```diff
@@ -0,0 +1,66 @@
...
+
+//nolint:gochecknoinits
+func init() {
+	AddErrorResolver(&updateErrorResolver{})
+}
+
+var (
+	//nolint:lll
+	pkgNotGitRepo = `
+Package {{ printf "%q" .repo }} is not within a git repository. Please initialize a repository using 'git init' and then commit the changes using 'git commit -a -m "<commit message>"'.
```
frankfarzan, line 30: should we just leave it at using `git commit`. Since `git commit -a` won't always work if there are new files (Need to use `git add`). I rather have this more general and correct than more specific and problematic.

frankfarzan (approved):

PR #1891, opened 2021-05-04
Move cluster-related flags to live subcommand

mortent:
This moves the construction of the `util.Factory` into the `live` subcommand. This means that all flags related to a k8s cluster will now be exposed on the `live` command instead of the top level `kpt` command.

The current solution for letting users provide custom openapi schemas used the factory to allow users to fetch the k8s schema from the cluster. This PR removes this functionality completely. Depending on how we want to support this going forward, we might want to keep the functionality for reading openapi schemas from disk.

Fixes: https://github.com/GoogleContainerTools/kpt/issues/1886


commands/livecmd.go
```diff
@@ -113,3 +116,21 @@ func GetLiveCommand(name string, f util.Factory) *cobra.Command {
 
 	return liveCmd
 }
+
+func newFactory(cmd *cobra.Command, version string) util.Factory {
```
frankfarzan, line 120: `util.Factory` name is problematic on many levels:
a) What is this Java in 90s? :)
b) `util` itself is a meaningless term. "util for X" should just be called "X", otherwise everything is a util.

Feel free to address this in a separate PR and/or create a package alias to stop the bleeding in the call sites for now.

mortent: `util.Factory` comes from the kubernetes codebase, so the name is not something we can change (other than creating an import alias).

frankfarzan: Let's create an alias

mortent: Renamed `util` to `cluster`

frankfarzan (approved):

PR #1896, opened 2021-05-04
`kpt fn render` ensures validators don't mutate resources 

### LINKED ISSUE #1860: `fn render` should forbid validators to modify resources

droot:
With this PR, `kpt fn render` ensures that validators in a pipeline don't mutate the resources. 

This is achieved by splitting the execution of `mutators` from `validators`. Now we run validators on a copy of mutated resources.

Note: It will be nice if `kpt fn render` can flag a validator function that mutates the resources. I think that is doable, but need be investigated how best to do that.

fixes #1860

### DIFF
--- a/internal/cmdrender/executor.go
+++ b/internal/cmdrender/executor.go
@@ -301,8 +301,6 @@ func (pn *pkgNode) runPipeline(ctx context.Context, hctx *hydrationContext, inpu
 
 // runMutators runs a set of mutators functions on given input resources.
 func (pn *pkgNode) runMutators(ctx context.Context, hctx *hydrationContext, input []*yaml.RNode) ([]*yaml.RNode, error) {
-	const op errors.Op = "pipeline.runMutators"
-
 	if len(input) == 0 {
 		return input, nil
 	}
@@ -318,7 +316,7 @@ func (pn *pkgNode) runMutators(ctx context.Context, hctx *hydrationContext, inpu
 
 	mutators, err := fnChain(ctx, pn.pkg.UniquePath, pl.Mutators)
 	if err != nil {
-		return nil, errors.E(op, pn.pkg.UniquePath, err)
+		return nil, err
 	}
 
 	output := &kio.PackageBuffer{}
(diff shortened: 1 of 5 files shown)


droot:
Ignore this PR for now, @Shell32-Natsu identified an issue with this, so I am going to fix that.

internal/cmdrender/executor.go
```diff
@@ -286,30 +286,99 @@ func (pn *pkgNode) runPipeline(ctx context.Context, hctx *hydrationContext, inpu
...
+		return nil, errors.E(op, pn.pkg.UniquePath, err)
+	}
+	// print a new line after a pipeline running
+	pr.Printf("\n")
+	return mutatedResources, nil
+}
+
+// runMutators runs a set of mutators functions on given input resources.
+func (pn *pkgNode) runMutators(ctx context.Context, hctx *hydrationContext, input []*yaml.RNode) ([]*yaml.RNode, error) {
+	const op errors.Op = "pipeline.runMutators"
```
Shell32-Natsu, line 304: For this `op`, since we don't want to expose internal implementation details to users, I suspect does it really have a use case? Is it only used in debug? Is there a debug mode for kpt? And why don't we directly add a real stack trace here?

internal/cmdrender/executor.go
```diff
@@ -286,30 +286,99 @@ func (pn *pkgNode) runPipeline(ctx context.Context, hctx *hydrationContext, inpu
...
+	mutatedResources, err := pn.runMutators(ctx, hctx, input)
+	if err != nil {
+		return nil, errors.E(op, pn.pkg.UniquePath, err)
+	}
+
+	if err = pn.runValidators(ctx, hctx, mutatedResources); err != nil {
+		return nil, errors.E(op, pn.pkg.UniquePath, err)
+	}
+	// print a new line after a pipeline running
+	pr.Printf("\n")
```
Shell32-Natsu, line 298: Why do we need this?

droot: This is the original code.

internal/cmdrender/executor.go
droot: This is no longer useful, we took out the stacktrace printing. Taking it out.

internal/cmdrender/executor.go
Shell32-Natsu: Aha, I remember that I added this. :0

frankfarzan (approved):

PR #1898, opened 2021-05-04
make eval to use kpt container & exec runtime

Shell32-Natsu:
`fn eval` is using container and exec runtime from kyaml. These runtimes are [hardcoding](https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/fn/runtime/exec/exec.go#L34) the command error outputs to `os.Stderr` so we cannot get the stderr output from the runtime.

One alternative is refactoring the kyaml implementation to accept a writer for stderr. However, given we have had our own [container runtime](https://github.com/GoogleContainerTools/kpt/blob/next/internal/fnruntime/container.go) in kpt and is used by `render`. It's better to make `eval` and `render` to use the same runtime.

This PR is modifying the container runtime in kpt to support the functionalities required by `fn eval` and adding an exec runtime for `fn eval`.

### DIFF
--- a/e2e/testdata/fn-eval/missing-fn-image/.expected/config.yaml
+++ b/e2e/testdata/fn-eval/missing-fn-image/.expected/config.yaml
@@ -17,5 +17,4 @@ exitCode: 1
 image: gcr.io/kpt-fn/dne # non-existing image
 args:
   namespace: staging
-stdErr: "Failed to fetch \"latest\" from request \"/v2/kpt-fn/dne/manifests/latest\""
-
+stdErr: 'failed to check function image existence: function image "gcr.io/kpt-fn/dne" doesn''t exist'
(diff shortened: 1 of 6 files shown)


e2e/testdata/fn-eval/missing-fn-image/.expected/config.yaml
```diff
@@ -17,5 +17,4 @@ exitCode: 1
 image: gcr.io/kpt-fn/dne # non-existing image
 args:
   namespace: staging
-stdErr: "Failed to fetch \"latest\" from request \"/v2/kpt-fn/dne/manifests/latest\""
-
+stdErr: 'failed to check function image existence: function image "gcr.io/kpt-fn/dne" doesn''t exist'
```
droot, line 20: Not this PR: Can we remove this `failed to check function image existence` part from the error ?

droot (approved):

e2e/testdata/fn-eval/missing-fn-image/.expected/config.yaml
Shell32-Natsu: We probably need to define a struct for this error?

PR #1902, opened 2021-05-05
Update live reference docs

mortent:
This covers the `kpt live` commands.


site/reference/live/README.md
```diff
@@ -10,14 +10,7 @@ description: >
...
-
 <!--mdtogo:Long-->
-| Reads From              | Writes To                |
-|-------------------------|--------------------------|
-| local files             | cluster                  |
-| cluster                 | stdout                   |
-
-Live contains the next-generation versions of apply related commands for
-deploying local configuration packages to a cluster.
+`live` contains apply functionality for deploying local configuration
```
frankfarzan, line 14: "apply functionality" -> "functionality"

site/reference/live/apply/README.md
```diff
@@ -1,298 +1,36 @@
...
-# apply resources and specify how often to poll the cluster for resource status
-kpt live apply --reconcile-timeout=15m --poll-period=5s my-dir/
-```
-<!--mdtogo-->
-
 ### Synopsis
 <!--mdtogo:Long-->
 ```
-kpt live apply DIR [flags]
+kpt live apply [DIR|-] [flags]
```
frankfarzan, line 18: We're using `PKG_PATH` for package directories. `DIR` for normal directories. e.g.:

https://kpt.dev/reference/pkg/update/

Same for other commands.

site/reference/live/apply/README.md
```diff
@@ -1,298 +1,36 @@
...
 
 #### Flags
-
 ```
 --poll-period:
   The frequency with which the cluster will be polled to determine
-  the status of the applied resources. The default value is every 2 seconds.
+  the status of the applied resources. The default value is 2 seconds.
 
 --reconcile-timeout:
```
frankfarzan, line 35: Currently, all usage in the book uses `--reconcile-timeout=15m`. Should the default for this flag be changed?

site/reference/live/apply/README.md
```diff
@@ -1,298 +1,36 @@
...
+  of the path to a directory cause kpt to read from stdin.
 ```
 
 #### Flags
-
 ```
 --poll-period:
   The frequency with which the cluster will be polled to determine
-  the status of the applied resources. The default value is every 2 seconds.
+  the status of the applied resources. The default value is 2 seconds.
```
frankfarzan, line 33: General question: what's the ordering of the flag here? It should be alphabetical but it's not (Applies to other commands as well)

site/reference/live/status/README.md
```diff
@@ -1,111 +1,81 @@
 ---
-title: "Status"
+title: "`status`"
```
frankfarzan, line 2: @phanimarupaka are we changing this command to be a verb i.e. `observe`? Same question for `tree`, there are the only two noun commands.

site/reference/live/apply/README.md
frankfarzan: We should also have a consistent description for the flag. "Path to a directory containing KRM resources and a Kptfile" is a awkward (The critical thing is Kptfile, otherwise not having any other KRM resources is a valid case).

frankfarzan (commented):
Some of the comments apply to commands in `pkg` as well.

site/reference/live/apply/README.md
```diff
@@ -1,298 +1,36 @@
...
-DIR:
-  Path to a package directory.  The directory must contain exactly
-  one ConfigMap with the inventory object annotation.
+DIR|-:
+  Path to a directory containing KRM resources and a Kptfile with inventory
+  information. The path can be relative or absolute. Providing '-' instead
+  of the path to a directory cause kpt to read from stdin.
 ```
 
 #### Flags
```
frankfarzan, line 29: should the ex-global flags (your other PR) be listed here?

site/reference/live/apply/README.md
mortent: I have updated the reference docs for all the live commands. I added some new language to the description, but I'm open to suggestions here.

site/reference/live/apply/README.md
mortent: So we could use another default value here, the challenge is that the correct value varies widely depending on what is being applied. For regular k8s resources like Deployments, a minute or two should be sufficient in almost all cases unless it involves a large number of replicas. But for resources that represent out-of-cluster resources, like Ingress or KCC, the appropriate timeout would be far more than that. The requirement for users to use `--reconcile-timeout=0` to avoid waiting also feels a little bit awkward.
If we wanted to chose another default, something like 5m would probably be appropriate.

site/reference/live/apply/README.md
mortent: Fixed this for the live reference docs. Will do another pass at the pkg docs when we have merged this one.

site/reference/live/apply/README.md
mortent: Those apply to all commands in `live` command group, so I think those should be included in the `live` page. Similar, we probably want to document the global flags on the top-level group. 
Alternatively we can look into ways to include subtemplates so we can get the complete set of flags for every command, but this will lead to a pretty long list of flags.

mortent: If this looks good otherwise, I can do it as a follow-up.

frankfarzan (approved):

PR #1913, opened 2021-05-06
better handle multi-line stderr

mengqiy:
An example of multi-line stderr:
```
$ kpt fn render
Package ".": 

[RUNNING] "gcr.io/kpt-fn/kubeval:unstable"
[FAIL] "gcr.io/kpt-fn/kubeval:unstable"
  Stderr:
    "[ERROR] Additional property templates is not allowed in object 'v1/ReplicationController//bob' in file resources.yaml in field templates"
    "[ERROR] Invalid type. Expected: [integer,null], given: string in object 'v1/ReplicationController//bob' in file resources.yaml in field spec.replicas"
    ""
  Exit Code: 1
```

There are leading whitespaces in each line.
If we put this multi-line string in yaml like
```
exitCode: 1
stderr: |
      "[ERROR] Additional property templates is not allowed in object 'v1/ReplicationController//bob' in file resources.yaml in field templates"
      "[ERROR] Invalid type. Expected: [integer,null], given: string in object 'v1/ReplicationController//bob' in file resources.yaml in field spec.replicas"
```
the leading whitespace will be ignored and it makes it unable to test these scenarios.

### DIFF
--- a/e2e/testdata/fn-eval/invalid-fn-config-file/pkg/.expected/config.yaml
+++ b/e2e/testdata/fn-eval/invalid-fn-config-file/pkg/.expected/config.yaml
@@ -16,4 +16,5 @@ testType: eval
 exitCode: 1
 image: gcr.io/kpt-fn/set-namespace:v0.1
 fnConfig: ../../config.yaml
-stdErr: "wrong Node Kind for  expected: MappingNode was ScalarNode: value: {I am not a valid config file}"
+stdErr:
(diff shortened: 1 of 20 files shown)


mengqiy:
@Shell32-Natsu @droot PTAL

Shell32-Natsu:
Can you try to use [Block Indentation Indicator](https://yaml.org/spec/1.2/spec.html#id2793979) to preserve the leading spaces?

mengqiy:
That works! Thanks!

PR #1914, opened 2021-05-06
Introduce Additional Diff Tests

### LINKED ISSUE #1776: Improve Test Coverage for kpt Package Diff
- [x] Kptfile has diff when pipeline changes are made (#1777)
- [x] Running a diff against a directory which does not contain a Kptfile (empty dir) (#1914)
- [x] Running a diff against a directory which does not exist (#1914)
- [x] Running a diff when an invalid git ref is provided (#1914)
- [x] Running a 3 way diff provides three parameters to the diff command (#1914)

runewake2:
This introduces several tests over kpt's `pkg diff` command:

* Running a diff against a directory which does not contain a Kptfile (empty dir)
* Running a diff against a directory which does not exist
* Running a diff when an invalid git ref is provided
* Running a 3 way diff provides three parameters to the diff command

This also disables these tests when run on a Windows OS (the `diff` command used is not available).

Fixes #1776

### DIFF
--- a/internal/util/diff/diff_test.go
+++ b/internal/util/diff/diff_test.go
@@ -22,6 +22,8 @@ import (
 	"bufio"
 	"bytes"
 	"io"
+	"io/ioutil"
+	"log"
 	"strings"
 	"testing"
 
@@ -38,6 +40,8 @@ func TestCommand_Diff(t *testing.T) {
 		fetchRef     string
 		diffRef      string
 		diffType     DiffType
+		diffTool     string
+		diffOpts     string
 		expDiff      string
 	}{
 
@@ -63,6 +67,8 @@ func TestCommand_Diff(t *testing.T) {
 			fetchRef: "v2",
 			diffRef:  "master",
 			diffType: DiffTypeRemote,
+			diffTool: "diff",
+			diffOpts: "-r -i -w",
 			expDiff: `
 39c39
 <             - containerPort: 80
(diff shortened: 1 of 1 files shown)


internal/util/diff/diff_test.go
```diff
@@ -242,6 +254,115 @@ func TestCommand_Diff(t *testing.T) {
...
+	diffOutput := &bytes.Buffer{}
+	err := (&Command{
+		Path:         g.LocalWorkspace.FullPackagePath(),
+		Ref:          "hurdygurdy", // ref should not exist in upstream
+		DiffType:     DiffTypeCombined,
+		DiffTool:     "diff",
+		DiffToolOpts: "-r -i -w",
+		Output:       diffOutput,
+	}).Run(fake.CtxWithNilPrinter())
+	assert.Error(t, err)
```
mortent, line 291: Can we do some kind of verification here that we actually got the error we expected?

internal/util/diff/diff_test.go
```diff
@@ -242,6 +254,115 @@ func TestCommand_Diff(t *testing.T) {
...
+			diffOutput := &bytes.Buffer{}
+			cmdErr := (&Command{
+				Path:         tc.directory,
+				Ref:          "master",
+				DiffType:     DiffTypeCombined,
+				DiffTool:     "diff",
+				DiffToolOpts: "-r -i -w",
+				Output:       diffOutput,
+			}).Run(fake.CtxWithNilPrinter())
+			assert.Error(t, cmdErr)
```
mortent, line 361: Same as above. When we expect errors, we should try to verify that it fails for the expected reason.

mortent (approved):

PR #1918, opened 2021-05-06
kpt fn: refactored function runner

droot:
This PR refactors components related to running functions. With this change, `fn render` and `fn eval` will be able to share a common function runner.

### DIFF
--- a/internal/fnruntime/runner.go
+++ b/internal/fnruntime/runner.go
@@ -12,48 +12,100 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package pipeline provides struct definitions for Pipeline and utility
-// methods to read and write a pipeline resource.
-package cmdrender
+package fnruntime
 
 import (
 	"context"
+	goerrors "errors"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 
 	"github.com/GoogleContainerTools/kpt/internal/errors"
-	"github.com/GoogleContainerTools/kpt/internal/fnruntime"
+	"github.com/GoogleContainerTools/kpt/internal/printer"
 	"github.com/GoogleContainerTools/kpt/internal/types"
+	fnresult "github.com/GoogleContainerTools/kpt/pkg/api/fnresult/v1alpha2"
 	kptfilev1alpha2 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1alpha2"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
 	"sigs.k8s.io/kustomize/kyaml/fn/runtime/runtimeutil"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
-// newFnRunner returns a fnRunner from the image and configs of
-// this function.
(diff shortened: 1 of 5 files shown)


droot:
This PR should simplify (avoid the function wrapper code) in the the other PR @Shell32-Natsu https://github.com/GoogleContainerTools/kpt/pull/1907 is working on.

internal/fnruntime/runner.go
```diff
@@ -12,48 +12,100 @@
...
+func (fr *FunctionRunner) do(input []*yaml.RNode) (output []*yaml.RNode, err error) {
+	var results []framework.ResultItem
+
+	output, err = fr.filter.Filter(input)
+
+	// TODO(droot): read the structured result fr.filer.Result()
+	// once kyaml changes to runtimeutil.FunctionFilter are merged
+	fnResult := fnresult.Result{
+		Image:   fr.containerRunner.Image,
+		Results: results,
```
Shell32-Natsu, line 96: Exit code should be in the fn result as well. But can be addressed in original structured results PR.

internal/fnruntime/runner.go
```diff
@@ -12,48 +12,100 @@
...
+		},
+	}, nil
+}
+
+// FunctionRunner wraps FunctionFilter and implements kio.Filter interface.
+type FunctionRunner struct {
+	ctx             context.Context
+	fnResults       *fnresult.ResultList
+	containerRunner *ContainerFn
+	filter          *runtimeutil.FunctionFilter
```
Shell32-Natsu, line 66: Use `kio.Filter` here.

internal/fnruntime/runner.go
```diff
@@ -12,48 +12,100 @@
...
+			FunctionConfig: config,
+		},
+	}, nil
+}
+
+// FunctionRunner wraps FunctionFilter and implements kio.Filter interface.
+type FunctionRunner struct {
+	ctx             context.Context
+	fnResults       *fnresult.ResultList
+	containerRunner *ContainerFn
```
Shell32-Natsu, line 65: We cannot use `ContainerFn` here because in `eval` it can be `ExecFn`. I think we don't need this field? Looks like what we need from it are `Image`/`ExecPath` and `Ctx`. `Ctx` can be fetched from runner itself. We can have a `Image`/`ExecPath` in the `FunctionRunner`.

internal/fnruntime/runner.go
```diff
@@ -12,48 +12,100 @@
...
+
+func (fr *FunctionRunner) do(input []*yaml.RNode) (output []*yaml.RNode, err error) {
+	var results []framework.ResultItem
+
+	output, err = fr.filter.Filter(input)
+
+	// TODO(droot): read the structured result fr.filer.Result()
+	// once kyaml changes to runtimeutil.FunctionFilter are merged
+	fnResult := fnresult.Result{
+		Image:   fr.containerRunner.Image,
```
Shell32-Natsu, line 95: For exec function this should be `ExecPath`. But can be addressed in original structured results PR.

droot:
Had a chat with @Shell32-Natsu and he is going to merge this PR with #1907 to ensure `FunctionRunner` meets the need of `fn eval` and `fn render`.

@frankfarzan pl. don't review this PR.

PR #1965, opened 2021-05-11
remove eval flags

Shell32-Natsu:
Remove flags from `fn eval`:
 - `log-steps`: No need to have this since we have our own CLI output
 - `include-subpackages`: Not used in `fn eval` and `fn run`. The flag was added but the code which uses it in `fn run` is missing for some reason. If we want to support it we can reintroduce it as a new feature.


Shell32-Natsu:
> we can re-introduce `include-subpackages` when we introduce filtering resources functionality.
> 
> @Shell32-Natsu , we should start documenting the changes for `fn run` --> `fn eval` in a markdown file that we can link to the changelog for 1.0.

https://github.com/GoogleContainerTools/kpt/issues/1971

PR #1968, opened 2021-05-11
Improve the docs for kpt live preview

mortent:
Updates the reference doc for the `kpt live preview` command to be more accurate about what it actually does, in particular the differences between running preview client-side and server-side.

It does raise some questions around how deep we want to go in the description of the commands in the reference doc. This is somewhat related to a single flag, but it doesn't seem like we want to go into too much detail in the description of each flag. Open for input here.

### DIFF
--- a/site/reference/live/preview/README.md
+++ b/site/reference/live/preview/README.md
@@ -9,8 +9,27 @@ description: >
     Preview the changes apply would make to the cluster
 -->
 
-`preview` shows the operations that will be done when running `apply` or 
-`destroy` against a cluster.
+`preview` validates the resources in the package against the cluster and shows
+which resources will be applied and pruned.
+
+If the `preview` command is used without specifying the `server-side` flag,
+validation and dry-run will primarily happen client-side. `kpt` will display the 
+operations that will be performed against the apiserver for each resource when
+running `apply`. The operations are:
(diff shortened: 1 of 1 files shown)


site/reference/live/preview/README.md
```diff
@@ -9,8 +9,27 @@ description: >
...
+`preview` validates the resources in the package against the cluster and shows
+which resources will be applied and pruned.
+
+If the `preview` command is used without specifying the `server-side` flag,
+validation and dry-run will primarily happen client-side. `kpt` will display the 
+operations that will be performed against the apiserver for each resource when
+running `apply`. The operations are:
+ * `created`: The resource doesn't currently exist in the cluster and will be created.
+ * `configured`: The resource exists in the cluster and will be updated or remain unchanged.
+ * `failed`: The resource can't be applied.
```
frankfarzan, line 21: How about other categories: {pruned, skipped}?

Also why "failed" appears twice?

```
6 resource(s) applied. 0 created, 0 unchanged, 6 configured, 0 failed (preview)
0 resource(s) pruned, 0 skipped, 0 failed (preview)
```

Which of these are mutually exclusive?

frankfarzan (commented):
Thanks.

In terms of what should be covered in reference docs: If you're providing the reference for the output of the command, it should be in the reference docs.

PR #1972, opened 2021-05-11
Simplify apply cobra command

mortent:
This PR addresses several issues:
 * Removes support for the ConfigMap inventory objects.
 * Consolidates the two nested cobra commands that were a leftover from when we imported cobra commands from the cli-utils library. This simplifies the code and mean all the flags are defined in one place
 * Validates the value of the `--output` flag and returns an error if it is not one of the supported values.
 * The allowed values for the `--output` flag are printed in quotes to make it easier to separate them from the rest of the text.
 * Uses the printer for messages to the terminal.
 * Error messages for missing RG CRD or error when applying the RG CRD are handled and displayed the same way we do with other kpt errors. 

Missing RG CRD: 
```
Error: The ResourceGroup CRD was not found in the cluster. Please install it either by using the '--install-resource-group' flag or the 'kpt live install-resource-group' command.
```

Error installing RG CRD (ignore the actual error, I introduced an issue to make it fail):
```
installing inventory ResourceGroup CRD.
Error: Unable to install the ResourceGroup CRD.

Details:
CustomResourceDefinition.apiextensions.k8s.io "resourcegroups.kpt.dev" is invalid: [metadata.name: Invalid value: "resourcegroups.kpt.dev": must be spec.names.plural+"."+spec.group, spec.group: Required value] 
```

The plan is to do the same to the remaining `kpt live` commands.


seans3:
I think we need to remove the `ConfigMap` inventory object tests in `e2e/live/end-to-end-test.sh`

mortent:
I have updated the e2e tests. This involved:
 * Removing the tests using the ConfigMap inventory object.
 * Use the latest version of kpt from the master branch for setting up the ConfigMap inventory for testing the migrate command.

seans3 (approved):
/lgtm

PR #1977, opened 2021-05-12
Improve error messages when failing to read/parse Kptfile

mortent:
Improve the error messages for situations where kpt fails to read or parse the Kptfile. 

Error message when Kptfile doesn't exist:
```
$ kpt fn render      
Error: No Kptfile found at "/tmp/kpttest".
```

Error message for other issues involving the Kptfile (in this example the Kptfile has an unrecognized property):
```
$ kpt fn render
Error: Kptfile at "/tmp/kpttest" can't be read.

Details:
yaml: unmarshal errors:
  line 7: field foo not found in type v1alpha2.KptFile 
```

Fixes: https://github.com/GoogleContainerTools/kpt/issues/1976

### DIFF
--- a/e2e/testdata/fn-render/missing-kptfile/.expected/config.yaml
+++ b/e2e/testdata/fn-render/missing-kptfile/.expected/config.yaml
@@ -13,4 +13,4 @@
 # limitations under the License.
 
 exitCode: 1
-stdErr: "missing-kptfile/Kptfile: no such file or directory"
+stdErr: "No Kptfile found at"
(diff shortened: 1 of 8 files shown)


e2e/testdata/fn-render/missing-kptfile/.expected/config.yaml
```diff
@@ -13,4 +13,4 @@
 # limitations under the License.
 
 exitCode: 1
-stdErr: "missing-kptfile/Kptfile: no such file or directory"
\ No newline at end of file
+stdErr: "No Kptfile found at"
```
frankfarzan, line 16: incomplete message

mortent: This is intentional. The check does a substring search and the error message includes the path to the package, which is not possible to predict before the test runs. We could resolve the full path when the test runs and put it into the expected message when the test runs, but not sure if that is worth the complexity.

frankfarzan (approved):

PR #1982, opened 2021-05-12
automated test to validate auto image pull

runewake2:
This adds an automated test that verifies images are automatically pulled when kpt functions are executed.


natasha41575:
This LGTM but should wait for lead approval

@droot @frankfarzan This is for line 11 of fn-render in go/kpt-test-cases

Shell32-Natsu (commented):
Thanks for this test. Maybe you can update the implementation here
https://github.com/GoogleContainerTools/kpt/blob/next/internal/fnruntime/container.go#L146-L156 
to use `docker image inspect` instead of `docker image ls`.

runewake2:
Is that the behavior we want @Shell32-Natsu? The commands behave slightly different. `docker image ls` will succeed if you do not provide a tag for example but inspect may fail.

```
$ docker image ls gcr.io/kpt-fn/search-replace
REPOSITORY                     TAG                 IMAGE ID            CREATED             SIZE
gcr.io/kpt-fn/search-replace   v0.1                187c912bbdb5        2 weeks ago         19MB

$ docker image inspect gcr.io/kpt-fn/search-replace                                                                                      
Error: No such image: gcr.io/kpt-fn/search-replace
```

Shell32-Natsu:
> Is that the behavior we want @Shell32-Natsu? The commands behave slightly different. `docker image ls` will succeed if you do not provide a tag for example but inspect may fail.
> 
> ```
> $ docker image ls gcr.io/kpt-fn/search-replace
> REPOSITORY                     TAG                 IMAGE ID            CREATED             SIZE
> gcr.io/kpt-fn/search-replace   v0.1                187c912bbdb5        2 weeks ago         19MB
> 
> $ docker image inspect gcr.io/kpt-fn/search-replace                                                                                      
> Error: No such image: gcr.io/kpt-fn/search-replace
> ```

I think this is the expected. For our functions in the catalog, they don't have `latest` tag so pulling it without tag is not possible. In most cases user needs to provide the image name as well as tag.

runewake2:
I'll take a look, but I don't think that change is straightforward. `docker image ls random-image` returns successfully but `inspect` would fail. We would need to update our logic to make this behave how we expect.

internal/fnruntime/container_test.go
```diff
@@ -69,3 +70,39 @@ func TestContainerFn(t *testing.T) {
...
+	output := &bytes.Buffer{}
+	err := instance.Run(input, output)
+	assert.NoError(t, err)
+
+	// Inspect fails with an exit code of 1 if no image is present
+	// We expect this command to succeed (the image exists)
+	cmdImageInspect := exec.Command("docker", "image", "inspect", image)
+	errImageInspect := cmdImageInspect.Run()
+	assert.NoError(t, errImageInspect)
+}
```
droot, line 108: It's not clear to me what exactly are we testing here ?

droot: When you invoke `instance.Run()`:
1. Image could be pulled using the `prepareImage()`
2. Or it could be pulled using the `docker run`

so either case, `docker image inspect` will succeed.. right ? or may be I am missing something.

runewake2: This is specifically targeting the container functions and that executing the container Command will pull the image. It is not testing **how** that image is pulled (yet). This is meant as a proxy for manually testing this functionality.

`docker image inspect` exits with non-zero exit code if the image does not exist. A quick example:

```
$ docker pull postgres:latest
latest: Pulling from library/postgres
Status: Downloaded newer image for postgres:latest
docker.io/library/postgres:latest

$ docker image rm postgres:latest
Untagged: postgres:latest
...

$ docker image inspect postgres:latest
[]
Error: No such image: postgres:latest
```

This test is meant to:

1. Remove any existing local image reference (if one exists)
2. Execute the function as a no-op
3. Validate that the function image is available locally

The test case I had in mind was:

> Given an environment that does not have a function image locally, when kpt executes that function, then the function image is checked out locally.

This could be refactored to target specifically covering `prepareImage` if we'd prefer but is currently broader than that. If you'd like I could append a test to this PR that introduces that coverage.

droot: > This could be refactored to target specifically covering prepareImage if we'd prefer but is currently broader than that. If you'd like I could append a test to this PR that introduces that coverage.

No, not in favor of testing internal functions. Infact, I would like the test case to be:

> Given an environment that does not have a function image locally, when kpt _should be able to execute_ that function successfully.

Ideally having it in e2e test case would be better but we don't have a way to setup the test-case today where we ensure the function image isn't present before running the test. 

But this is fine for now.

droot (approved):

internal/fnruntime/container_test.go
droot: Today @Shell32-Natsu added support for running `setup.sh` script in e2e test framework https://github.com/GoogleContainerTools/kpt/pull/2032

So, we can implement this test as e2e test. That will be better. So copy one of the simple tests and add `setup.sh` that deletes the docker images being used in that test and successful run of the test will achieve the same outcome as here.

droot (changes requested):
Suggest to implement this as e2e test since support for `setup.sh` and `teardown.sh` has been added.

internal/fnruntime/container_test.go
```diff
@@ -19,6 +19,7 @@ package fnruntime_test
 import (
```
frankfarzan, line 19: @droot are these supposed to be unit tests or e2e tests?

droot: Actually these were added before we had `e2e` tests. I have suggested @runewake2 to implement this test in `e2e` now because with `setup.sh` support in e2e harness. We can safely remove these tests after that is done. @Shell32-Natsu Can you pl. confirm.

droot (approved):
Pl. update the row corresponding to this test green in the test sheet and also, mark it green.

Thank you!

PR #1984, opened 2021-05-12
Only copy root packages for remote diffing.

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1606
Normalize usage of package copying to use a single function depending on the depth of subpackage copying using the `Subpackage` function.
All callers of `CopyPackage` or `CopyPackageWithSubpackages` are replaced with `CopyPackage` using the appropriately copied subpackages.


internal/pkg/pkg_test.go
```diff
@@ -424,6 +424,16 @@ func TestSubpackages(t *testing.T) {
 						"local-sub1",
 					},
 				},
+				{
+					matcher:   []SubpackageMatcher{None},
```
mortent, line 428: It is good that we add tests for this. Is there a real use-case for this, or is it just for completeness? Asking for a list of subpackages and then also specify that we don't want any subpackages returned seems a bit strange.

internal/util/diff/diff_test.go
```diff
@@ -246,9 +340,14 @@ func TestCommand_Diff(t *testing.T) {
 				DiffToolOpts: tc.diffOpts,
 				Output:       diffOutput,
 			}).Run(fake.CtxWithNilPrinter())
-			assert.NoError(t, err)
+			if !assert.NoError(t, err) {
+				t.FailNow()
+			}
 
 			filteredOutput := filterDiffMetadata(diffOutput)
+			if tc.hasLocalSubpackageChanges {
```
mortent, line 348: Do we need to check this separately from the other output?

internal/util/pkgutil/pkgutil.go
```diff
@@ -63,10 +63,15 @@ func WalkPackage(src string, c func(string, os.FileInfo, error) error) error {
...
 }
 
-// CopyPackage copies the content of a single package from src to dst. It
-// will not copy resources belonging to any subpackages.
-func CopyPackage(src, dst string, copyRootKptfile bool) error {
-	return WalkPackage(src, func(path string, info os.FileInfo, err error) error {
+// CopyPackage copies the content of a single package from src to dst. If includeSubpackages
+// is true, it will copy resources belonging to any subpackages.
+func CopyPackage(src, dst string, copyRootKptfile bool, matcher pkg.SubpackageMatcher) error {
+	subpackagesToCopy, err := pkg.Subpackages(src, matcher, true)
```
mortent, line 69: How does this handle deeply nested subpackages? Remember there might be several layers of subpackages, both local and remote.

internal/util/pkgutil/pkgutil_test.go
```diff
@@ -105,6 +106,181 @@ func TestWalkPackage(t *testing.T) {
...
 }
 
+func TestCopyPackage(t *testing.T) {
+	testCases := map[string]struct {
+		pkg               *pkgbuilder.RootPkg
+		copyRootKptfile   bool
+		subpackageMatcher pkg.SubpackageMatcher
+		expected          []string
+	}{
+		"subpackages without root kptfile": {
```
mortent, line 116: Add some tests here around deeply nested packages.

mortent (commented):
Can you add some more details about the PR in the description? Just explain the high-level approach so it is easier to understand the code.

internal/util/diff/diff_test.go
etefera: There's certainly no other way to reconcile the dynamically generated local subpackage location with the diff output.

internal/pkg/pkg_test.go
etefera: This would be for completeness. I agree that internally we should be smart about how we deal with copying subpackages, but we should probably make sure this utility function handles it as expected, which is to say not copying them at all.

internal/pkg/pkg.go
```diff
@@ -198,6 +198,8 @@ const (
 	Local SubpackageMatcher = "LOCAL"
 	// remote means only remote subpackages will be returned.
 	Remote SubpackageMatcher = "REMOTE"
+	// None means that no subpackages will be returned.
+	None SubpackageMatcher = "NONE"
```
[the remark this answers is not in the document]
etefera: We could either change the callers to have a call this function differently when there are no subpackages desired, but this seems like the most robust solution regarding this function. Happy to switch to an alternative instead, if that's preferred.

etefera:
> Can you add some more details about the PR in the description? Just explain the high-level approach so it is easier to understand the code.

Added some more details about what's changed.

internal/util/pkgutil/pkgutil.go
etefera: I was under the impression that local packages (whatever that would mean) under remote subpackages should be handled as if the entirety was a remote subpackage, which this should hopefully do (will address in a test case as mentioned below). Is that not the case?

internal/util/pkgutil/pkgutil_test.go
etefera: Added positive cases surrounding nested packages and their types. Let me know if more are desired.

internal/util/diff/diff_test.go
mortent: Ah, the printed path will be an issue.

mortent (approved):

PR #1995, opened 2021-05-13
Remove Draft Flag from Next Releases

runewake2:
Kpt vNext is no longer released as a draft.


mortent (approved):

runewake2:
@googlebot I fixed it.

PR #2001, opened 2021-05-13
updated reference for fn source and sink

Shell32-Natsu:
Some other Go doc files are also updated after running `make generate`.

https://github.com/GoogleContainerTools/kpt/pull/1981#discussion_r631498688


Shell32-Natsu:
Dup with #2013

PR #2007, opened 2021-05-13
Update 02-imperative-function-execution.md

mengqiy:


site/book/04-using-functions/02-imperative-function-execution.md
```diff
@@ -108,7 +108,7 @@ page.
 For example, `kubeval` function can consume a JSON schema file:
 
 ```shell
-$ kpt fn eval --image gcr.io/kpt-fn/kubeval:v0.1 --mount type=bind,src="/path/to/schema-dir",dst=/schema-dir --as-current-user wordpress -- additional_schema_locations=/schema-dir
+$ kpt fn eval --image gcr.io/kpt-fn/kubeval:v0.1 --mount type=bind,src="/path/to/schema-dir",dst=/schema-dir --as-current-user wordpress -- schema_location=file:///schema-dir
```
frankfarzan, line 111: Docs here, show additional_schema_location ving value of `file:///`

https://catalog.kpt.dev/kubeval/v0.1/

It doesn't show schema_location with this scheme

mengqiy: Actually we are showing the file scheme:
image

mengqiy: `additional_schema_locations` will only be used when the primary one `schema_location` doesn't have the desired schema.
My point is that a user can't really verify if `--mount` is working when their provided schemas are not used.

frankfarzan (approved):

PR #2013, opened 2021-05-14
fn sink and source reference docs

droot:
This PR does the following:

- Add reference (help) docs for `kpt fn source` and `kpt fn sink`
- Drops `wrap-kind` `wrap-apiversion` flags from `kpt fn source`. I don't think they are useful. We can introduce them when we see the need. Locking the API surface area.
- Renamed `function-config` flag in `fn source` to `fn-config` to make it consistent with `fn eval`

### DIFF
--- a/site/reference/fn/eval/README.md
+++ b/site/reference/fn/eval/README.md
@@ -22,13 +22,13 @@ Refer to the [Imperative Function Execution] for detailed overview.
 
 <!--mdtogo:Long-->
 
-```shell
+```
 kpt fn eval [DIR|-] [flags] [-- fn-args]
 ```
 
 #### Args
 
-```shell
+```
 DIR|-:
   Path to the local directory containing resources. Defaults to the current
   working directory. Using '-' as the directory path will cause `eval` to
@@ -37,14 +37,14 @@ DIR|-:
 
   1. Multi object YAML where resources are separated by `---`.
 
-  2. [Function Specification] wire format where resources are wrapped in an object
-     of kind ResourceList. Refer to [Function Specification] format for more details.
+  2. `Function Specification` wire format where resources are wrapped in an object
+     of kind ResourceList.
  
   If the output is written to `stdout`, resources are written in multi object YAML
   format where resources are separated by `---`.
 ```
 
-```shell
+```
 fn-args:
   function arguments to be provided as input to the function. These must be
   provided in the `key=value` format and come after the separator `--`.
@@ -52,7 +52,7 @@ fn-args:
 
 #### Flags
 
-```shell
+```
 --as-current-user:
   Use the `uid` and `gid` of the kpt process for container function execution.
   By default, container function is executed as `nobody` user. You may want to use
@@ -107,55 +107,55 @@ fn-args:
 ## Examples
 <!--mdtogo:Examples-->
 
-```shell
+```
 # execute container my-fn on the resources in DIR directory and
 # write output back to DIR
 $ kpt fn eval DIR --image gcr.io/example.com/my-fn
 ```
 
-```shell
+```
 # execute container my-fn on the resources in DIR directory with
 # `functionConfig` my-fn-config
 $ kpt fn eval DIR --image gcr.io/example.com/my-fn --fn-config my-fn-config
 ```
 
-```shell
+```
 # execute container my-fn with an input ConfigMap containing `data: {foo: bar}`
 $ kpt fn eval DIR --image gcr.io/example.com/my-fn:v1.0.0 -- foo=bar
 ```
 
-```shell
+```
 # execute executable my-fn on the resources in DIR directory and
 # write output back to DIR
 $ kpt fn eval DIR --exec-path ./my-fn
 ```
 
-```shell
+```
 # execute container my-fn on the resources in DIR directory,
 # save structured results in /tmp/my-results dir and write output back to DIR
 $ kpt fn eval DIR --image gcr.io/example.com/my-fn --results-dir /tmp/my-results-dir
 ```
 
-```shell
+```
 # execute container my-fn on the resources in DIR directory with network access enabled,
 # and write output back to DIR
 $ kpt fn eval DIR --image gcr.io/example.com/my-fn --network 
 ```
 
-```shell
+```
 # execute container my-fn on the resource in DIR and export KUBECONFIG
 # and foo environment variable
 $ kpt fn eval DIR --image gcr.io/example.com/my-fn --env KUBECONFIG -e foo=bar
 ```
 
-```shell
+```
 # execute kubeval function by mounting schema from a local directory on wordpress package
 $ kpt fn eval --image gcr.io/kpt-fn/kubeval:v0.1 \
   --mount type=bind,src="/path/to/schema-dir",dst=/schema-dir \
   --as-current-user wordpress -- additional_schema_locations=/schema-dir
 ```
 
-```shell
+```
 # chaining functions using the unix pipe to set namespace and set labels on
 # wordpress package
 $ kpt fn source wordpress \
--- a/site/reference/fn/eval/README.md
+++ b/site/reference/fn/eval/README.md
@@ -32,7 +32,16 @@ kpt fn eval [DIR|-] [flags] [-- fn-args]
 DIR|-:
   Path to the local directory containing resources. Defaults to the current
   working directory. Using '-' as the directory path will cause `eval` to
-  read resources from `stdin` and write the output to `stdout`.
+  read resources from `stdin` and write the output to `stdout`. When resources are
+  read from `stdin`, they must be in one of the following input formats:
+
+  1. Multi object YAML where resources are separated by `---`.
+
+  2. [Function Specification] wire format where resources are wrapped in an object
+     of kind ResourceList. Refer to [Function Specification] format for more details.
+ 
+  If the output is written to `stdout`, resources are written in multi object YAML
+  format where resources are separated by `---`.
 ```
 
 ```shell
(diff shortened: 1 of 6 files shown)


site/reference/fn/sink/README.md
```diff
@@ -1,51 +1,45 @@
...
 
 ```shell
-kpt fn sink [DIR]
-
-DIR:
-  Path to a package directory.  Defaults to stdout if unspecified.
+# read resources from DIR directory, execute my-fn on them and write the
+# output to DIR directory.
+$ kpt fn source DIR |
+  kpt fn run --image gcr.io/example.com/my-fn |
```
mengqiy, line 38: eval

site/reference/fn/source/README.md
```diff
@@ -1,71 +1,64 @@
...
+$ kpt fn source DIR
+```
 
-DIR:
-  Path to a package directory.  Defaults to stdin if unspecified.
+```shell
+# read resources from DIR directory, execute my-fn on them and write the
+# output to DIR directory.
+$ kpt fn source DIR |
+  kpt fn run --image gcr.io/example.com/my-fn |
```
mengqiy, line 57: eval

site/reference/fn/sink/README.md
```diff
@@ -1,51 +1,45 @@
...
 
 ```shell
-# run a function using explicit sources and sinks
-kpt fn source DIR/ |
-  kpt fn run --image gcr.io/example.com/my-fn |
-  kpt fn sink DIR/
+kpt fn sink DIR [flags]
+
+DIR:
+  Path to a local directory to resources to. Directory must exist.
```
mengqiy, line 25: > Path to a local directory to resources to.

This doesn't read well.

Is `DIR` required? If so, this behavior is different

site/reference/fn/source/README.md
droot: Done.

site/reference/fn/sink/README.md
droot: missed `write` :) fixed.

Yes, `DIR` is required (probably was done as part of CLI usability work).

>if so, this behavior is different

what behavior are you referring to ? and what is different ?

site/reference/fn/sink/README.md
droot: done.

site/reference/fn/sink/README.md
```diff
@@ -1,51 +1,45 @@
...
+   Write resources to a local directory
 ---
 
 <!--mdtogo:Short
-    Specify a directory as an output sink package
+   Write resources to a local directory
 -->
 
-Implements a [sink function] by reading STDIN and writing configuration.
+`sink` reads resources from `stdin` and writes them to a local directory.
```
Shell32-Natsu, line 13: If there is no directory specified, `sink` will output to stdout. Will we change this in v1?

site/reference/fn/source/README.md
```diff
@@ -1,71 +1,64 @@
...
 
 <!--mdtogo:Short
-    Specify a directory as an input source package
+    Source resources from a local directory
 -->
 
-Implements a [source function] by reading configuration and writing to STDOUT.
+`source` reads resources from a local directory and writes them in `functionSpec`
+wire format to `stdout`. The output of the `source` can be pipe'd to commands
+such as `kpt fn eval` that accepts `functionSpec` format. It is useful for
```
Shell32-Natsu, line 15: Maybe use the link to function spec here.

site/reference/fn/sink/README.md
mengqiy: `DIR` was optional before. If unspecified, it prints to stdout.

site/reference/fn/sink/README.md
frankfarzan, line 13: The format of stdin for `sink` and stdout for `source` are currently underspecified.

site/reference/fn/source/README.md
```diff
@@ -1,71 +1,64 @@
...
+   Source resources from a local directory
 ---
 
 <!--mdtogo:Short
-    Specify a directory as an input source package
+    Source resources from a local directory
 -->
 
-Implements a [source function] by reading configuration and writing to STDOUT.
+`source` reads resources from a local directory and writes them in `functionSpec`
```
frankfarzan, line 13: Compare:

```
kpt fn source wordpress \
  | less
```

This is wrapped in ResourceList.

```
kpt fn source wordpress \
  | kpt fn eval --image gcr.io/kpt-fn/kubeval:v0.1 - \
  | less
```

This is multi-object YAML.


Is there a reason, all wire format as passed through pipes should not be the same:

```
kpt fn source wordpress \       
  | kpt fn eval --image gcr.io/kpt-fn/set-namespace:v0.1 - -- namespace=mywordpress \
  | kpt fn sink wordpress2
```

Need to also improve testing for ths.

frankfarzan: Related to whether we want flexibility of both wrapped and unwrapped with flags:

1. Source can emit output wrapped or unwrapped
2. Eval can accept input as wrapped or unwrapped
3. Eval can emit output as wrapped or unwrapped
4. Sink can accept input as wrapped or unwrapped

site/reference/fn/sink/README.md
```diff
@@ -1,51 +1,45 @@
...
 ```shell
-kpt fn sink [DIR]
-
-DIR:
-  Path to a package directory.  Defaults to stdout if unspecified.
+# read resources from DIR directory, execute my-fn on them and write the
+# output to DIR directory.
+$ kpt fn source DIR |
+  kpt fn eval --image gcr.io/example.com/my-fn - |
+  kpt fn sink DIR
```
natasha41575, line 39: All of the examples here use `source` and `sink` together. Do they always have to be used in conjunction? What happens if one is used without the other? It might be helpful to explicitly answer these questions in the documentation or have examples

site/reference/fn/source/README.md
natasha41575: For the OpenAPI scenario specifically: if we choose not to wrap the output in ResourceList, another option we'd discussed was wrapping the OpenAPI schema in its own KRM resource. 

I think either unwrapped or wrapped is fine, but I agree that it makes sense to have consistency across commands.

site/reference/fn/sink/README.md
droot: DIR is required argument now. This is changed in `v1`. Not sure output to `stdout` makes any sense. If that is the intention, output from `source` or `eval` should be good enough, no need of `sink`. Right ?

droot: Added brief note about the input for `sink` and output for `source`.

site/reference/fn/sink/README.md
droot: `source`, 'eval' and `sink` can be used to create your own little pipeline where `source` acts as a generator (builtin) function, followed by set of `eval` functions and then final `sink` function that writes the output. So far, that is the only use-case I know of. I want to see more use-cases of these in the field.

site/reference/fn/source/README.md
droot: Done.

site/reference/fn/source/README.md
droot: > Is there a reason, all wire format as passed through pipes should not be the same:

I am not familiar with the reasoning. But when it comes to reading resources from a byte stream such as 'stdin`, all these command use same `kio.ByteReader` that supports resource in both (wrapped/unwrapped) format without specifying the format, so `eval` and `sink` supports both the formats. And when writing the output, they use `kio.ByteWriter` that writes in multi-yaml object format by default unless specified to wrap. For `fn source`, it always write the output in wrapped format.

So looking at the use case of creating your own pipeline using `source, eval, sink`, `source` needs to support `ResourceList` output format. `eval` needs to support `ResourceList` input as default (that is does) and it SHOULD output `ResourceList` by default (and that's where the confusion arises). and `sink` accepts `ResourceList` as input.
So only change of default output of `eval` to `ResourceList` will bring consistency and will bring it closer to the `pipeline` execution.

Now output to `multi-yaml` object has its benefits of being able to pipe to `kubectl` etc., so that can be supported with a flag.

droot (commented):
Addressed the feedback. I am refraining from making behavioral changes to `source`, `sink` or `eval` as part of this PR. I have updated the reference to document current behavior. Can follow up with another PR for behavior change.

site/reference/fn/eval/README.md
```diff
@@ -32,7 +32,16 @@ kpt fn eval [DIR|-] [flags] [-- fn-args]
 DIR|-:
   Path to the local directory containing resources. Defaults to the current
   working directory. Using '-' as the directory path will cause `eval` to
-  read resources from `stdin` and write the output to `stdout`.
+  read resources from `stdin` and write the output to `stdout`. When resources are
+  read from `stdin`, they must be in one of the following input formats:
+
+  1. Multi object YAML where resources are separated by `---`.
+
+  2. [Function Specification] wire format where resources are wrapped in an object
```
frankfarzan, line 40: nit: you're linking twice

site/reference/fn/source/README.md
```diff
@@ -1,71 +1,64 @@
...
-# print to stdout configuration from DIR/ formatted as an input source
-kpt fn source DIR/
+DIR:
+  Path to the local directory containing resources. Defaults to the current
+  working directory.
 ```
 
+#### Flags
+
 ```shell
```
frankfarzan, line 37: nit: other commands don't use `shell` directive, so there's little bit of inconsistency in rendering

frankfarzan (approved):
some nits

site/reference/fn/eval/README.md
droot: Fixed. Removed the linking here because links are not being rendered in this section (may be because it's a code block)

site/reference/fn/source/README.md
droot: Thanks for point it out. Cleaned it up.

PR #2016, opened 2021-05-14
Use `set-label` instead of `apply-setters` in wordpress package.

frankfarzan:


phanimarupaka (changes requested):

phanimarupaka (approved):

site/book/04-using-functions/01-declarative-function-execution.md
```diff
@@ -20,21 +20,17 @@ metadata:
...
 
 This declares two functions:
 
-- `apply-setters` is an example mutator function which provides a form of parameterization. You
-  don't need to understand the nitty-gritty of how this function works except that it mutates
-  resources that have been parameterized with a given key (e.g. `wp-tag`) by applying the
-  corresponding value (e.g. `4.8-apache`).
-- `kubeval` is an example validator function which you've seen before. It validates the resources
+- `set-label` is an example mutator function which adds a set of labels to resources.
+- `kubeval` is an example validator function which validates the resources
```
droot, line 33: We should remove `example` from these lines. That gives reader an impression that these functions are ready for production use.

droot (approved):

Issue #1960, opened 2021-05-11
Structured results use 4 spaces indentation in some places


mengqiy:
Structured results use 4 spaces indentation in some places
Here is the code: https://github.com/GoogleContainerTools/kpt/blob/d2dc33fe6c0185df83d82e885907ceefbc62445f/internal/fnruntime/utils.go#L32

</details>
For example:

```
apiVersion: kpt.dev/v1alpha2
kind: FunctionResultList
metadata:
    name: fnresults
exitCode: 1
items:
    - image: gcr.io/kpt-fn/kubeval:v0.1
      stderr: |
        [ERROR] selector is required in object 'apps/v1/Deployment//nginx-deployment' in file resources.yaml in field selector
        [ERROR] template is required in object 'apps/v1/Deployment//nginx-deployment' in file resources.yaml in field template
        [ERROR] Invalid type. Expected: [integer,null], given: string in object 'apps/v1/Deployment//nginx-deployment' in file resources.yaml in field spec.replicas
        [ERROR] Failed to parse raw kubeval output:
        Unexpected token E in JSON at position 0

        ERR  - stdin: Failed initializing schema file:///tmp/master-standalone-strict/custom-custom-v1.json: open /tmp/master-standalone-strict/custom-custom-v1.json: no such file or directory
         in object 'custom.io/v1/Custom//custom' in file resources.yaml
      exitCode: 1
      results:
        - message: selector is required
          severity: error
          resourceRef:
            apiVersion: apps/v1
            kind: Deployment
          field:
            path: selector
          file:
            path: resources.yaml
        - message: template is required
          severity: error
          resourceRef:
            apiVersion: apps/v1
            kind: Deployment
          field:
            path: template
          file:
            path: resources.yaml
    ...
```
</details>
Even though it's valid yaml, this behavior is not consistent with other yaml formatting in kpt.

droot:
https://github.com/GoogleContainerTools/kpt/pull/2174 addressed this.

PR #2017, opened 2021-05-14
Add test for eval structured output

Shell32-Natsu:


frankfarzan (approved):

droot:
@Shell32-Natsu I think the PR is running into this issue https://github.com/GoogleContainerTools/kpt/issues/1960 Will be good to root cause this. I think this is some processing the testrunner does.

Shell32-Natsu:
> @Shell32-Natsu I think the PR is running into this issue #1960 Will be good to root cause this. I think this is some processing the testrunner does.

I think this is different issue. Looks like the order of the output results will be changed case-by-case. Not sure is this a function issue or orchestrator issue.

PR #2019, opened 2021-05-15
Correct cross origin link typo.

etefera:
Also provides a workaround to anchors sometimes not loading in the right target.
https://kpt-dev-ee2a9--catalog-link-c9c8ly0r.web.app/


site/static/js/plugins.js
```diff
@@ -134,6 +134,20 @@ function processAsciinemaTags(content) {
   );
 }
 
+// Workaround for https://github.com/docsifyjs/docsify/pull/1468
+function defaultLinkTargets() {
```
runewake2, line 138: I'm not sure I understand why this is needed. Can you help me understand?

runewake2 (commented):
Looks good but idk what `defaultLinkTargets` is solving here. Some tests might help define the expected behavior.

frankfarzan:
Does this generally solve the issue of any external link not opening? e.g. Links in https://kpt.dev/contact/

site/static/js/plugins.js
etefera: External links (such as those in kpt.dev/contact/) open in a new frame if you first navigate directly to the referrer, then click them, as expected. However, there's currently a bug where going to another page first (say, kpt.dev/faq/) then the referrer will drop the target, changing it to _self as described in the commented PR. Without explicitly marking those links as safe for cross-origin, links targeting the same frame as a Docsify page won't open in the routing mode we use. This function resets the target of each external page link that isn't marked as cross-origin safe to whatever Docsify thinks it should've been normally (blank in our case)

etefera:
> Does this generally solve the issue of any external link not opening? e.g. Links in https://kpt.dev/contact/

Yes, observe going to https://kpt.dev/faq/ then https://kpt.dev/contact/ then clicking an external link.

frankfarzan (approved):

PR #2023, opened 2021-05-16
Organize site directory

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1678
Moves inline JS and CSS to separate static files.
Removes remaining hugo files.
Removes `node_modules` from the site's directory in the repo.


mikebz:
what tests were performed to ensure that the site was not broken?  5000 file purge is hard to review so trying to make sure we do some before/after checks.

etefera:
> what tests were performed to ensure that the site was not broken? 5000 file purge is hard to review so trying to make sure we do some before/after checks.

Manually ran locally and at https://kpt-dev-ee2a9--clean-site-dir-r3q5fcok.web.app/

mikebz (approved):

mikebz:
Thank you.  Approving now but for large changes like that I'd recommend doing a few things:
1) different browser checks
2) pages that are different than other pages (maybe they have different widgets)
3) mobile rendering

Thank you!

etefera:
> Thank you. Approving now but for large changes like that I'd recommend doing a few things:
> 
> 1. different browser checks
> 2. pages that are different than other pages (maybe they have different widgets)
> 3. mobile rendering
> 
> Thank you!

Ah yes, meant that 1&2 had happened both ran locally and in the above link

PR #2027, opened 2021-05-17
enforce `fn render` path invariants

### LINKED ISSUE #1958: fn render invariant: functions should not be able to modify resources outside of their current package
When running `kpt fn render`, a function in a package should only be able to modify and generate resources within the current package. If it tries to write outside of its package, kpt should fail with error.

i.e. fn-render/path-index-ancestor and fn-render/path-index-outofpkg should fail with exit code 1. 

cc @frankfarzan @droot @Shell32-Natsu 

ref: https://github.com/GoogleContainerTools/kpt/issues/1559

natasha41575:
close #1958 

This PR does 4 things: 

1. If a function in a pipeline tries to write to a file outside of its current package, `kpt fn render` will throw the error: 

`function must not modify resources outside of package: resource has path <filepath>`

2. Moves detection of path conflicts to the `fnruntime` package

3. Adds a test `generator-absolute-path` to show current behavior of a path beginning with `/` (it is treated as a relative path)  

4. Updates the test `path-index-current` to be run without a subpackage.


internal/cmdrender/executor.go
```diff
@@ -425,6 +425,9 @@ func adjustRelPath(resources []*yaml.RNode, relPath string) ([]*yaml.RNode, erro
 				return resources, err
 			}
 		}
+		if strings.HasPrefix(path.Clean(currPath), "../") {
```
droot, line 428: couple of drive by comments:

shouldn't we do this check before updating the path annotation ? (above this code).

I am still noodling over another approach i.e. doing this in the `FunctionRunner` we have in pkg `fnruntime` instead of here. That's right after executing the function. I would like to enforce invariant as close as possible to the function execution.

is checking for `../` enough ?

e2e/testdata/fn-render/path-index-ancestor/.expected/config.yaml
```diff
@@ -12,8 +12,5 @@
...
 # limitations under the License.
 
-# This package is nonIdempotent because functions in the subpackage (db)
-# do not have knowledge of the resources in the parent package. Therefore
-# when it runs twice, it tries to recreate the same resource in the parent,
-# resulting in a filepath conflict and throwing an error.
-nonIdempotent: true
\ No newline at end of file
+exitCode: 1
+stdErr: "Error: adjust relative path: cannot modify resources outside of package at path ../deployment_httpbin.yaml"
```
frankfarzan, line 16: "adjust relative path" seems out of palce

natasha41575: "adjust relative path" no longer shows up in the error message

internal/cmdrender/executor.go
natasha41575: moved the logic to `fnruntime` package

also - I'm using path.Clean which should resolve things like `a/b/../../../c/d` to `../c/d`. Added a few unit tests for this.  Hopefully in this case checking for `../` is enough. I was trying to think of other ways that someone would be able to write outside of the package and none come to mind.

droot (changes requested):

internal/fnruntime/runner.go
```diff
@@ -208,6 +213,23 @@ func printFnExecErr(ctx context.Context, fnErr *ExecError) {
...
+// to a file outside the package
+func checkResourcePaths(resources []*yaml.RNode) error {
+	for _, r := range resources {
+		currPath, _, err := kioutil.GetFileAnnotations(r)
+		if err != nil {
+			return err
+		}
+		if strings.HasPrefix(path.Clean(currPath), "../") {
+			return fmt.Errorf("cannot modify resources outside of package: resource has path %s", currPath)
+		}
```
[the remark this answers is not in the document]
natasha41575: Both with and without a subpackage has the same output: 
```
Package "my-kpt-package": 
[RUNNING] "gcr.io/kpt-fn/starlark:unstable"
[FAIL] "gcr.io/kpt-fn/starlark:unstable"
Error: function must not modify resources outside of package: resource has path ../deployment_httpbin.yaml
```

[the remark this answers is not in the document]
natasha41575: Without subpkg: 

```
Package "my-kpt-package": 
[RUNNING] "gcr.io/kpt-fn/starlark:unstable"
[FAIL] "gcr.io/kpt-fn/starlark:unstable"
Error: function must not modify resources outside of package: resource has path ../deployment_httpbin.yaml
```

With subpkg:
```
Package "my-kpt-package/subpackage": 
Package "my-kpt-package": 
[RUNNING] "gcr.io/kpt-fn/starlark:unstable"
[FAIL] "gcr.io/kpt-fn/starlark:unstable"
Error: function must not modify resources outside of package: resource has path ../deployment_httpbin.yaml 
```

internal/fnruntime/runner.go
```diff
@@ -208,6 +213,23 @@ func printFnExecErr(ctx context.Context, fnErr *ExecError) {
...
+// metadata.annotation field. checkResourcePaths ensures that those paths
+// only point to files within the package, throwing an error if there is a path
+// to a file outside the package
+func checkResourcePaths(resources []*yaml.RNode) error {
+	for _, r := range resources {
+		currPath, _, err := kioutil.GetFileAnnotations(r)
+		if err != nil {
+			return err
+		}
+		if strings.HasPrefix(path.Clean(currPath), "../") {
```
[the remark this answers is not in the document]
natasha41575: current behavior is "/directory" is treated the same as "./directory"

internal/fnruntime/runner.go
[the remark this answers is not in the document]
droot: ok. this is fine.

e2e/testdata/fn-render/path-index-ancestor/.expected/config.yaml
```diff
@@ -12,8 +12,5 @@
...
 # limitations under the License.
 
-# This package is nonIdempotent because functions in the subpackage (db)
-# do not have knowledge of the resources in the parent package. Therefore
-# when it runs twice, it tries to recreate the same resource in the parent,
-# resulting in a filepath conflict and throwing an error.
-nonIdempotent: true
\ No newline at end of file
+exitCode: 1
+stdErr: "Error: function must not modify resources outside of package: resource has path ../deployment_httpbin.yaml"
```
droot, line 16: Can't believe that we managed to get the previous testcase merged.

internal/fnruntime/runner.go
```diff
@@ -208,6 +213,33 @@ func printFnExecErr(ctx context.Context, fnErr *ExecError) {
 	pr.OptPrintf(printOpt.Stderr(), "  Exit code: %d\n\n", fnErr.ExitCode)
 }
 
+// path (location) of a KRM resources is tracked in a special key in
+// metadata.annotation field. checkResourcePaths throws an error if there is a path
+// to a file outside the package, or if the same index/path is on multiple resources
+func checkResourcePaths(nodes []*yaml.RNode) error {
```
droot, line 219: can we rename this function to `enforcePathInvariants`

internal/fnruntime/runner_test.go
```diff
@@ -163,3 +167,317 @@ func TestMultilineFormatter(t *testing.T) {
...
+				if !assert.Error(t, err) {
+					t.FailNow()
+				}
+				assert.Equal(t, c.expectedErr, err.Error())
+			}
+		})
+	}
+}
+
+func TestDetectPathConflicts(t *testing.T) {
```
droot, line 243: You can probably combine these two test functions into `TestEnsurePathInvariants()`. They seem to be inconsistent right now (the implementation logic got merged, but these tests are still divergent :)

droot (changes requested):
Looking good, some minor suggestions.

droot (approved):
Thanks for working on this. Really happy with the final result! 🚢  (ship it)

PR #2031, opened 2021-05-17
Temporarily disable gatekeeper function

Shell32-Natsu:
We discovered that `ensure-gatekeeper` function's structured output is not stable (result ordering is random) that cause the test to be flaky. This PR `skips` this flaky tests until we the fix lands in `ensure-gatekeeper`.


droot (approved):
Thanks @Shell32-Natsu 

nit: always have a brief description of the PR to help other understand what this PR does and why is it needed. I added for this one.

mengqiy (approved):

PR #2034, opened 2021-05-17
Refactor kpt live init and kpt live migrate commands

mortent:
Align the UX for `kpt live` more with the rest of kpt. Also, update the code to follow the same conventions as the rest of the kpt codebase. This PR updates the `kpt live init` and `kpt live migrate` commands.


seans3 (approved):
/lgtm

PR #2035, opened 2021-05-17
Run Kind e2e tests on next branch

mortent:
This is almost the same as the GA that runs on the main branch. There are two changes:
 * This only runs on PRs (so not on push)
 * It tests versions 1.18 and 1.19, which seems to be the most widely used version right now.


seans3 (approved):
/lgtm

PR #2036, opened 2021-05-18
Fix breaking diff test

mortent:
This fixes a failing test for `diff`. We should look into a better solution here, the current approach seems fragile.

### DIFF
--- a/internal/util/diff/diff_test.go
+++ b/internal/util/diff/diff_test.go
@@ -346,7 +346,7 @@ locally changed: foo
 
 			filteredOutput := filterDiffMetadata(diffOutput)
 			if tc.hasLocalSubpackageChanges {
-				filteredOutput = regexp.MustCompile("Only in /tmp.+:").ReplaceAllString(filteredOutput, "locally changed:")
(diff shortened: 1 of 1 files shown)


phanimarupaka (approved):

internal/util/diff/diff_test.go
```diff
@@ -346,7 +346,7 @@ locally changed: foo
 
 			filteredOutput := filterDiffMetadata(diffOutput)
 			if tc.hasLocalSubpackageChanges {
-				filteredOutput = regexp.MustCompile("Only in /tmp.+:").ReplaceAllString(filteredOutput, "locally changed:")
+				filteredOutput = regexp.MustCompile("Only in /(tmp|var).+:").ReplaceAllString(filteredOutput, "locally changed:")
```
mikebz, line 349: Seems like an OS specific value.  Specifically on Mac Big Sur it starts with /var
```
mktemp
/var/folders/vc/54chww3n1fb9c81mq8mz973h00rl2k/T/tmp.gXZR59Y9
```

on CloudShell (linux)
```
mktemp
/tmp/tmp.6WtSIciYBv
```

so we can technically predict which one we should be looking for if we care.  Using  tmp|var is OK for these two platforms though.

etefera (approved):

PR #2045, opened 2021-05-19
Support fetching subpackages using tag partials.

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1829
Consolidates usage of the `CreatePkgOrFail` test helper function within `fetch_test.go`

### DIFF
--- a/internal/util/fetch/fetch_test.go
+++ b/internal/util/fetch/fetch_test.go
@@ -439,30 +439,51 @@ func TestCommand_Run_tag(t *testing.T) {
 
 func TestCommand_Run_subdir_at_tag(t *testing.T) {
 	testCases := map[string]struct {
-		dir string
+		dir         string
+		tag         string
+		upstreamPkg *pkgbuilder.RootPkg
 	}{
 		"reads subdirectory": {
 			dir: "/java/expected",
+			tag: "java/v2",
+			upstreamPkg: pkgbuilder.NewRootPkg().
+				WithSubPackages(pkgbuilder.NewSubPkg("java").
+					WithResource("deployment").
+					WithSubPackages(pkgbuilder.NewSubPkg("expected").
+						WithFile("expected.txt", "My kptfile and I should be the only objects"))),
 		},
 		"reads subdirectory with no leading slash": {
 			dir: "java/expected",
+			tag: "java/v2",
+			upstreamPkg: pkgbuilder.NewRootPkg().
+				WithSubPackages(pkgbuilder.NewSubPkg("java").
+					WithResource("deployment").
+					WithSubPackages(pkgbuilder.NewSubPkg("expected").
+						WithFile("expected.txt", "My kptfile and I should be the only objects"))),
+		},
+		"reads specific subdirectory": {
+			dir: "/java/not_expected/java/expected",
+			tag: "java/expected/v2",
+			upstreamPkg: pkgbuilder.NewRootPkg().
+				WithSubPackages(pkgbuilder.NewSubPkg("java").
+					WithResource("deployment").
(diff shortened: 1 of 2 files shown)


internal/util/fetch/fetch_test.go
```diff
@@ -433,6 +427,54 @@ func TestCommand_Run_tag(t *testing.T) {
 	})
 }
 
+func TestCommand_Run_subdir_at_tag(t *testing.T) {
+	g := &testutil.TestSetupManager{
```
mortent, line 431: We shouldn't use the `TestSetupManager` here. It creates both an upstream repo and then fetches it into the local workspace. This is useful for testing update scenarios.

For testing the fetch functionality, we should just set up an upstream repo and an empty workspace, and then explicitly use the fetch command to download the package into the workspace. This means we have to write the Kptfile into the workspace before using the fetch command. There are some other tests here that uses this pattern.

internal/util/fetch/fetch_test.go
```diff
@@ -433,6 +427,54 @@ func TestCommand_Run_tag(t *testing.T) {
...
+	}.Run(fake.CtxWithNilPrinter())
+	if !assert.NoError(t, err) {
+		t.FailNow()
+	}
+
+	files, err := ioutil.ReadDir(actualPkg.UniquePath.String())
+	if !assert.NoError(t, err) {
+		t.FailNow()
+	}
+	fileNames := make([]string, 0)
```
mortent, line 469: I think it might be easier to just use the pkgbuilder to set up the expected content of the fetched package, and then use the `testutil.Assert*` command to verify.

internal/util/fetch/fetch_test.go
etefera: Set up repos with `SetupReposAndWorkspace` instead.

internal/util/fetch/fetch_test.go
etefera: Ended up using `AssertEqual` and `AssertKptfile` here.

internal/util/fetch/fetch_test.go
```diff
@@ -433,6 +426,69 @@ func TestCommand_Run_tag(t *testing.T) {
 	})
 }
 
+func TestCommand_Run_subdir_at_tag(t *testing.T) {
```
mortent, line 429: We should test additional cases here. Examples are:
 * Verify that the most specific version is chosen if there are multiple possible matches.
 * What happens if there are no match?
 * What happens if the provided `dir` doesn't have the leading `/`?

internal/util/fetch/fetch_test.go
```diff
@@ -433,6 +426,69 @@ func TestCommand_Run_tag(t *testing.T) {
...
+
+	actualPkg := pkgtesting.CreatePkgOrFail(t, rw.FullPackagePath())
+	err = Command{
+		Pkg: actualPkg,
+	}.Run(fake.CtxWithNilPrinter())
+	if !assert.NoError(t, err) {
+		t.FailNow()
+	}
+
+	g.AssertEqual(t, rw.WorkspaceDirectory, actualPkg.UniquePath.String(), false)
```
mortent, line 462: I think this could be a more compact way to check both that the content of the package is as expected and the content of `UpstreamLock` is correct:
https://github.com/GoogleContainerTools/kpt/blob/029b577ab66de50f4a988c1e2d5bee2ee68c53f6/internal/util/get/get_test.go#L1379-L1393

etefera: Switched.

internal/util/fetch/fetch_test.go
etefera: Added tests.

mortent (approved):

PR #2049, opened 2021-05-19
Remove preview and diff

mortent:
This removes the `kpt pkg diff` and `kpt pkg preview` commands. The `preview` functionality is available in the `kpt live apply` and `kpt live destroy` commands using the `--dry-run` flag.

Unlike kubectl, the `--dry-run` flag here is just a boolean. We choose between server-side and client-side dry-run based on the `--server-side` flag.


frankfarzan:
How about the doc changes to the book and the references?

mortent:
> How about the doc changes to the book and the references?

Planning to do that in a follow-up PR. I forgot to mention that in the PR description.

frankfarzan (approved):

mortent:
@googlebot I signed it!

PR #2059, opened 2021-05-20
e2e test cases for fnconfig for fn-render

droot:
This PR adds following e2e tests related to `functionConfig` for `kpt fn render`:

- fn config is not in the current package `This closes a major security attack vector`
- fn config is not a relative path
- fn config should not be mutated by the functions in the ancestor package

The PR also adds support for passing a test matcher that enables invoking a subset of e2e tests that is very handy during test development. An example below:

```
# invoke fnconfig related tests
make test-fn-render T=fnconfig

# invoke subpkgs related tests for fn-eval
make test-fn-eval T=subpkg

# it uses Go test's `-run=PATTERN` feature to enable
```


e2e/testdata/fn-render/fnconfig-outside-package/Kptfile
```diff
@@ -0,0 +1,10 @@
+apiVersion: kpt.dev/v1alpha2
+kind: Kptfile
+metadata:
+  name: app-with-db
+pipeline:
+  mutators:
+    - image: gcr.io/kpt-fn/set-label:unstable
+      configPath: labelconfig.yaml
+    - image: gcr.io/kpt-fn/set-label:unstable
+      configPath: some/../../home/.kube/config # config pointing outside the package
```
frankfarzan, line 10: do you restrict that function config must be in the *same directory* as the Kptfile?

pkg/api/kptfile/v1alpha2/validation.go
```diff
@@ -147,13 +147,21 @@ func IsNodeZero(n *yaml.Node) bool {
...
+		return fmt.Errorf("path must not be empty")
+	}
+	p = path.Clean(p)
 	if path.IsAbs(p) {
-		return fmt.Errorf("path is not relative")
+		return fmt.Errorf("path must be relative")
 	}
-	if strings.TrimSpace(p) == "" {
-		return fmt.Errorf("path cannot have only white spaces")
+	if strings.HasPrefix(p, "..") {
```
frankfarzan, line 159: HasPrefix -> Contains would be a stronger check

pkg/api/kptfile/v1alpha2/validation.go
```diff
@@ -147,13 +147,21 @@ func IsNodeZero(n *yaml.Node) bool {
...
+		// fn config must not live outside the package directory
+		// Allowing outside path opens up an attack vector that allows
+		// reading any YAML file on package consumer's machine.
+		return fmt.Errorf("path must not be outside the package directory")
+
 	}
 	if p != sourceAllSubPkgs && strings.Contains(p, "*") {
 		return fmt.Errorf("path contains asterisk, asterisk is only allowed in './*'")
```
frankfarzan, line 167: what's `./*` in the context of `configPath`?

e2e/testdata/fn-render/fnconfig-outside-package/Kptfile
frankfarzan: Discussed offline. We should restrict the fnconfig to be in the same package (but potentially a nested directory), but not in subpackages. Referring to wordpress package, it can be in:

`wordpress`
`wordpress/deployment`

but not any other directory.

pkg/api/kptfile/v1alpha2/validation.go
natasha41575: similar to the concern here https://github.com/GoogleContainerTools/kpt/pull/2027#discussion_r633937293 is just checking for ".." enough? Should we check for HasPrefix(path.Clean(p)) to account for cases like `a/../../b`?

pkg/api/kptfile/v1alpha2/validation_test.go
```diff
@@ -276,7 +276,7 @@ func TestValidatePath(t *testing.T) {
 		},
 		{
 			"./*",
-			true,
+			false,
 		},
```
natasha41575, line 280: we should have a test case for a path like `a/../../b`

e2e/testdata/fn-render/fnconfig-pkgfn-refers-subdir/Kptfile
```diff
@@ -0,0 +1,8 @@
+apiVersion: kpt.dev/v1alpha2
+kind: Kptfile
+metadata:
+  name: app-with-db
+pipeline:
+  mutators:
+    - image: gcr.io/kpt-fn/set-label:unstable
+      configPath: confs/labelconfig.yaml
```
droot, line 8: @frankfarzan here it refers to a file in subdirectory `confs`.

droot (commented):
Thanks for the review. Addressed the feedback.

pkg/api/kptfile/v1alpha2/validation.go
droot: Using `strings.Contains` now.

@natasha41575 path is already cleaned above. So contains will catch `../b` that you described. Also added more unit test cases to test this case.

pkg/api/kptfile/v1alpha2/validation.go
droot: Removed the `./*` bit. This code was originally for validating `pipeline.Sources` which we took it out. Cleaned it up.

pkg/api/kptfile/v1alpha2/validation_test.go
droot: Added.

e2e/testdata/fn-render/fnconfig-not-relative/.expected/config.yaml
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+exitCode: 1
+stdErr: |
+  Kptfile is invalid:
+  Field: `pipeline.mutators[0].configPath`
```
frankfarzan, line 18: I'm not a fan of these structured printout, it seems inconsistent with how we present the function result item in fluid english sentence

https://kpt.dev/book/04-using-functions/03-function-results

`Invalid type. Expected: integer, given: string in object "v1/Service/" in file "service.yaml" in field "spec.ports.0.port"`

Not blocking this, but should follow up.

e2e/testdata/fn-render/fnconfig-outside-package/.expected/config.yaml
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+exitCode: 1
+stdErr: |
+  Kptfile is invalid:
+  Field: `pipeline.mutators[1].configPath`
+  Value: "some/../../home/.kube/config"
+  Reason: path must not be outside the package directory
```
frankfarzan, line 20: Drop the word "directory". Otherwise, it implies a 1-to-1 correspondence between a package and a directory.

e2e/testdata/fn-render/missing-fnconfig/.expected/config.yaml
```diff
@@ -13,7 +13,11 @@
 # limitations under the License.
 
 exitCode: 1
-stdErr: 'Error: missing function config "labelconfig1.yaml"'
+stdErr: |
+  Kptfile is invalid:
+  Field: `pipeline.mutators[1].configPath`
+  Value: "labelconfig1.yaml"
+  Reason: functionConfig must exist in current package
```
frankfarzan, line 20: nit: "in the current package"

e2e/testdata/fn-render/fnconfig-not-relative/.expected/config.yaml
droot: Ack. I agree, printing english sentence is much better experience.

e2e/testdata/fn-render/fnconfig-outside-package/.expected/config.yaml
droot: Done.

e2e/testdata/fn-render/missing-fnconfig/.expected/config.yaml
droot: Done.

frankfarzan (approved):

PR #2062, opened 2021-05-20
add test for eval stdout

Shell32-Natsu:
Add a test case for `fn eval` stdout output. Also make test runner to always check stdout and stderr.

### DIFF
--- a/e2e/testdata/fn-eval/output-to-stdout/.expected/config.yaml
+++ b/e2e/testdata/fn-eval/output-to-stdout/.expected/config.yaml
@@ -1,0 +1,47 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+stdOut: |
+  # Copyright 2021 Google LLC
+  #
+  # Licensed under the Apache License, Version 2.0 (the "License");
+  # you may not use this file except in compliance with the License.
+  # You may obtain a copy of the License at
+  #
+  #      http://www.apache.org/licenses/LICENSE-2.0
+  #
+  # Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


e2e/testdata/fn-eval/output-to-stdout/.expected/config.yaml
```diff
@@ -0,0 +1,47 @@
...
+  # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  # See the License for the specific language governing permissions and
+  # limitations under the License.
+  apiVersion: apps/v1
+  kind: Deployment
+  metadata:
+    name: nginx-deployment
+    namespace: staging
+    annotations:
+      config.kubernetes.io/path: resources.yaml
```
natasha41575, line 35: do we want these annotations appearing in the output?

frankfarzan: you do not, seems like a bug

frankfarzan: @droot This is about use cases we talked about today and expected behavior of:

$ kpt fn source HERE | kpt fn eval | kpt fn sink HERE
$ kpt fn source HERE | kpt fn eval | kpt fn sink THERE
$ kpt fn eval HERE | kubectl apply -f

Shell32-Natsu: I don't think this is a bug. The content in eval stdout may be used by sink and sink needs these information to write to file system.

droot: @natasha41575 As @Shell32-Natsu stated, the output of `eval` in this case can be piped to `sink` that needs path annotation etc. to be able to write to filesystem. So this is expected, however, this causes problem for the 3) usecase @frankfarzan mentioned because `kubectl` will apply those annotation to the cluster which we don't want. 

@frankfarzan I will drafting a separate issue enumerating these use-cases, so that we have a better handle (and plan) for these use-cases.

For the scope of this PR, this output is correct.

frankfarzan: It's not necessarily a bug, but depending on the usage, it may not be desirable:

It's needed to preserve the directory structure here:

$ kpt fn source HERE | kpt fn eval | kpt fn sink THERE

but if we want to apply to cluster:

$ kpt fn eval HERE | kubectl apply -f

It will appear as annotations in the cluster state.

frankfarzan: Sounds good. Let's follow up. Ultimately all these use cases need to be documented in reference doc examples.

frankfarzan (approved):

PR #2073, opened 2021-05-21
Refactor kpt live status command

mortent:
Updates the `kpt live status` command to be more aligned with the other `kpt live` commands. Also updates the tests to use the inventory info in the Kptfile rather than testing with inventory templates.


seans3 (approved):
/lgtm

Issue #1662, opened 2021-04-01
Visually distinguish new and old sites


frankfarzan:
There will be two kpt sites co-existing for a while:

1. https://googlecontainertools.github.io/kpt  for master (0.39)
2. https://kpt.dev for next (1.0+)

Both sites should make it very obvious what kpt version they're targeting. If a user does a Google search and lands in one of these, there should be a prominent way to quickly figure this out what version the doc is talking about.

Related: Figure out a docs deprecation plan, considering SEO.

mikebz:
in backlog grooming, moving to M3 to focus on functionality

frankfarzan:
Should consider having kpt.dev use a different theme as well, currently it uses the same color palette as existing site. I'm personally a fan of dark themes (you can use an extension to see how it would look), but don't have a strong opinion.

mikebz:
I think a different theme is a good idea.  Dark is good but also since this is a "Google Cloud Tools" production we might want to have something that aligns more with https://cloud.google.com/docs

Either way this is probably something we can save for M4.

frankfarzan:
Yes, there are multiple aspects to this issue which can be separated:

- [ ] New site should show the CLI version (1.0+) as docs are versioned (Visible on every page)
- [ ] Old site should show it's targeting pre 1.0 version. Do not need to make major surgery to the old site, probably a banner  on every page suffices.
- [ ] In addition, using a different theme to as an additional differentiator.
- [ ] After the release, new site should be indexed instead of the old one and need to optimize SEO

etefera:
Sent a request to temporarily remove site from index and merged a change that should indicate the docs are for an in-development version of kpt.

PR #2077, opened 2021-05-21
Address issues in version dropdown.

etefera:
Addresses #1662 for the new site.
Preview at https://kpt-dev-ee2a9--version-dropdown-98kk6v2g.web.app

### DIFF
--- a/site/static/js/plugins.js
+++ b/site/static/js/plugins.js
@@ -50,7 +50,7 @@ async function addVersionDropdown() {
                 <button class="btn btn-primary dropdown-toggle" type="button" data-toggle="dropdown">v1.0.0-alpha.3
                 <span class="caret"></span></button>
                 <ol class="dropdown-menu">
-                <li><a href="/">v1.0.0-alpha.3</a></li>
+                <li><a href="/installation/">v1.0.0-alpha.3</a></li>
                 <li><a href="https://googlecontainertools.github.io/kpt/installation/" target="_self">v0.39</a></li>
                 </ol>
               </div>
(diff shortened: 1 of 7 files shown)


site/static/js/plugins.js
```diff
@@ -43,6 +43,23 @@ function convertFromHugo(content) {
...
 }
 
+async function addVersionDropdown() {
+  const sidebar = document.getElementsByClassName("sidebar").item(0);
+  const versionDropdown = `
+  <div class="dropdown">
+                <button class="btn btn-primary dropdown-toggle" type="button" data-toggle="dropdown">v1.0.0-alpha.3
+                <span class="caret"></span></button>
+                <ol class="dropdown-menu">
+                <li><a href="/installation/">v1.0.0-alpha.3</a></li>
```
mikebz, line 53: this is OK for now but I think the full implementation of a version drop down would preserve the current page and maybe just modify the URL or something along those lines.

mikebz (approved):
approving as a temporary solution.

site/static/js/plugins.js
mikebz: I would also maybe name it v1 beta1 or something more durable.

etefera: Will do for now
We should decide on the branch/tag naming scheme for multiple versions of the documentation and tool we want to show publicly. That'll drive the dropdown entries in the future

frankfarzan (approved):

site/static/js/plugins.js
frankfarzan: I missed this. What's `v1beta1`? Please use the release semver.

PR #2079, opened 2021-05-21
add test for env in eval

Shell32-Natsu:
The test case will build a simple image to print the envs.


e2e/testdata/fn-eval/function-env/.expected/setup.sh
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+
+set -eo pipefail
+
+# Build a docker image to print env
+IMAGE_TAG="gcr.io/kpt-fn/local-printenv:unstable"
+docker build -t $IMAGE_TAG - <<EOF
+FROM busybox:stable
+CMD ["printenv"]
+EOF
```
droot, line 23: hmm..why not have this image hosted in `gcr.io` ? It's better to not depend on images from `dockerhub` because we might run in to some rate-limiting issues (I remember they introduced some in recently).

droot (commented):
The test itself looks good. I have a question about hosting the test image in gcr.io.

e2e/testdata/fn-eval/function-env/.expected/setup.sh
Shell32-Natsu: Working with @mengqiy to have a GCR for test/demo functions

droot: Good thing is once that image is in `gcr`, all we have to do will be just drop the `setup.sh`. You can change `gcr.io/kpt-fn` to `gcr.io/kpt-demo-fn` (whatever the correct name is).

Shell32-Natsu: done

droot (approved):

PR #2081, opened 2021-05-21
add test for out-of-place sink

Shell32-Natsu:
This test case works in this way:

1. Create a temporary directory
2. Run eval and sink to the temp dir
3. Copy back the results

### DIFF
--- a/e2e/testdata/fn-eval/out-of-place/.expected/exec.sh
+++ b/e2e/testdata/fn-eval/out-of-place/.expected/exec.sh
@@ -23,6 +23,7 @@ kpt fn source \
 | kpt fn sink $TEMP_DIR
 
 # copy back the resources
+rm -r ./*
 cp $TEMP_DIR/* .
 
 # remove temporary directory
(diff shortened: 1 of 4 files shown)


e2e/testdata/fn-eval/out-of-place/.expected/exec.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# create a temporary directory
+TEMP_DIR=$(mktemp -d)
+
+kpt fn source \
+| kpt fn eval - --image gcr.io/kpt-fn/set-namespace:v0.1 -- namespace=staging \
+| kpt fn sink $TEMP_DIR
+
+# copy back the resources
+cp $TEMP_DIR/* .
```
droot, line 26: I think `sink` does not prune resources. So `cp` may not capture the entire behavior here.

Shell32-Natsu: updated.

droot (approved):

PR #2083, opened 2021-05-21
Test behavior of kpt live init when there is not Kptfile

mortent:
This adds an additional unit test to verify the behavior when a Kptfile can't be found when running the `kpt live init` command. It also updates the error handling so we get the same error message as with the `pkg` and `fn` commands.


seans3 (approved):
/lgtm

PR #2086, opened 2021-05-21
support both ts and go sdk formats for fn results

natasha41575:
workaround for https://github.com/GoogleContainerTools/kpt/issues/2056

There is a discrepancy between the results definition in functions SDK and the results definition in kpt. 

This PR adds some extra work in kpt to get the name/namespace from the functions result. 

It would be better to have a consistent schema between kpt and the functions sdk.

### DIFF
--- a/internal/fnruntime/runner.go
+++ b/internal/fnruntime/runner.go
@@ -174,6 +174,45 @@ func parseStructuredResult(yml *yaml.RNode, fnResult *fnresult.Result) error {
 	if err != nil {
 		return err
 	}
+
+	return parseNameAndNamespace(yml, fnResult)
+}
+
+func parseNameAndNamespace(yml *yaml.RNode, fnResult *fnresult.Result) error {
+	items, err := yml.Elements()
+	if err != nil {
+		return err
+	}
+
+	for i := range items {
+		r, err := items[i].Pipe(yaml.Lookup("resourceRef"))
+		if err != nil {
+			return err
+		}
+		if r == nil {
+			return nil
+		}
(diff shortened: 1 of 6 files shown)


pkg/api/fnresult/v1alpha2/types.go
```diff
@@ -73,3 +73,20 @@ func NewResultList() *ResultList {
 		Items: []Result{},
 	}
 }
+
+type ResultItem struct {
```
droot, line 77: Add a brief note about this struct. It will also be good to highlight why we are duplicating this here instead of using `framework.ResultItem` from the kyaml.

internal/fnruntime/runner.go
```diff
@@ -174,6 +173,50 @@ func parseStructuredResult(yml *yaml.RNode, fnResult *fnresult.Result) error {
...
+	for i := range items {
+		if err := getResourceRefMetadata(items[i], fnResult, i); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func getResourceRefMetadata(item *yaml.RNode, fnResult *fnresult.Result, i int) error {
```
droot, line 195: we don't need the `i` arg.

Better name for the function would be: `populateResourceRef`

droot (changes requested):
Looking good. Minor comments.

internal/fnruntime/runner.go
natasha41575: without `i`, how will we know which fnResult item this corresponds to? It's used on line 217

droot (approved):

PR #2088, opened 2021-05-24
Update cli-utils dependency and code in thirdparty

mortent:
Update the cli-utils dependency and the code from cli-utils in thirdparty.


seans3 (approved):
/lgtm

Issue #2010, opened 2021-05-13
output of `eval` with Unix pipe


frankfarzan:
```
kpt fn source wordpress \
  | kpt fn eval --image gcr.io/kpt-fn/set-namespace:v0.1 - \
  | kpt fn eval --image gcr.io/kpt-fn/set-label:v0.1 -
error:   Stderr:
    "failed to configure function: input namespace cannot be empty"
  Exit Code: 1
error:   Stderr:
    "failed to configure function: input label list cannot be empty"
  Exit Code: 1
```

Related comment: https://github.com/GoogleContainerTools/kpt/pull/1981#discussion_r632026952

Shell32-Natsu:
Some discussions in another issue.

https://github.com/GoogleContainerTools/kpt/issues/2006#issuecomment-841393567

https://github.com/GoogleContainerTools/kpt/issues/2006#issuecomment-841397475

Shell32-Natsu:
#2090 adds test cases for this behavior and slightly improves the output. Detail design should be done in #1915. Close this one and use #1915 to track.

PR #2090, opened 2021-05-24
Improve output when function fails in pipe

Shell32-Natsu:
When there is any error in function:

1. The function failure message `[FAIL] "gcr.io/kpt-fn/xxx"` will be output to stderr.
2. Structured results will be output to stderr.
3. Content in stderr and exit code will be output to stderr.

The output of piped eval command
```
kpt fn source \
| kpt fn eval - --image gcr.io/kpt-fn/set-namespace:v0.1 \
| kpt fn eval - --image gcr.io/kpt-fn/dne -- foo=bar \
| kpt fn sink .
```
changed from 
```
error: Stderr:
    "failed to configure function: input namespace cannot be empty"
  Exit Code: 1
error: Function image "gcr.io/kpt-fn/dne" doesn't exist.
```
to
```
[FAIL] "gcr.io/kpt-fn/set-namespace:v0.1"
  Results:
    [ERROR] failed to configure function: input namespace cannot be empty
  Stderr:
    "failed to configure function: input namespace cannot be empty"
  Exit code: 1


[FAIL] "gcr.io/kpt-fn/dne"
error: Function image "gcr.io/kpt-fn/dne" doesn't exist.
```

#2010

### DIFF
--- a/e2e/testdata/fn-eval/error-in-pipe/.expected/config.yaml
+++ b/e2e/testdata/fn-eval/error-in-pipe/.expected/config.yaml
@@ -13,8 +13,14 @@
 # limitations under the License.
 
 exitCode: 1
-stdErr: |
-  error:   Stderr:
+stdErr: |2
+  [FAIL] "gcr.io/kpt-fn/set-namespace:v0.1"
+    Results:
+      [ERROR] failed to configure function: input namespace cannot be empty
+    Stderr:
       "failed to configure function: input namespace cannot be empty"
-    Exit Code: 1
+    Exit code: 1
+
+   
+  [FAIL] "gcr.io/kpt-fn/dne"
   error: Function image "gcr.io/kpt-fn/dne" doesn't exist.
--- a/e2e/testdata/fn-eval/error-in-pipe/.expected/config.yaml
+++ b/e2e/testdata/fn-eval/error-in-pipe/.expected/config.yaml
@@ -1,0 +1,20 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+exitCode: 1
+stdErr: |
+  error:   Stderr:
+      "failed to configure function: input namespace cannot be empty"
+    Exit Code: 1
+  error: Function image "gcr.io/kpt-fn/dne" doesn't exist.
--- a/e2e/testdata/fn-eval/missing-fn-config/.expected/config.yaml
+++ b/e2e/testdata/fn-eval/missing-fn-config/.expected/config.yaml
@@ -18,4 +18,3 @@ image: gcr.io/kpt-fn/set-namespace:v0.1
 stdErr: "failed to configure function: input namespace cannot be empty"
 stdOut: |
   [RUNNING] "gcr.io/kpt-fn/set-namespace:v0.1"
-  [FAIL] "gcr.io/kpt-fn/set-namespace:v0.1"
--- a/e2e/testdata/fn-eval/error-in-pipe/.expected/exec.sh
+++ b/e2e/testdata/fn-eval/error-in-pipe/.expected/exec.sh
@@ -1,0 +1,21 @@
+#! /bin/bash
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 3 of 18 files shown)


e2e/testdata/fn-eval/error-in-pipe/.expected/config.yaml
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+exitCode: 1
+stdErr: |
+  error:   Stderr:
```
frankfarzan, line 17: What's the equivalent output when not piping?
Why/how this differs from that output?

Shell32-Natsu: If there is only the first function in pipe, the output will be:
```
Stderr:
  "failed to configure function: input namespace cannot be empty"
Exit Code: 1
```
I guess we should output structured result to stderr when there is an error in the function in this case?

droot: Had an offline chat with @Shell32-Natsu about it. I think we can bring in consistency in the error output in `piped` and `non-piped` scenario. So if we have multiple `fn eval` chained, the error output need to have enough context such as "which function" and all the details such as structured-result output, stderr output dumped on `stderr` (something we already do in case of `non-piped`). He is going to share the proposed output. It will require minor tweak in our printing logic (`stderr` vs `stdout`).

Shell32-Natsu: Expected output for kubeval:
```
[FAIL] "gcr.io/kpt-fn/kubeval:unstable"
  Results:
    [ERROR] selector is required in object "apps/v1/Deployment/" in file "resources.yaml" in field "selector"
    [ERROR] template is required in object "apps/v1/Deployment/" in file "resources.yaml" in field "template"
    [ERROR] Invalid type. Expected: [integer,null], given: string in object "apps/v1/Deployment/" in file "resources.yaml" in field "spec.replicas"
    [ERROR] Validating arbitrary CRDs is not supported yet. You can skip them by setting ignore_missing_schemas or skip_kinds in the function config
: ERR  - stdin: Failed initializing schema file:///tmp/master-standalone/custom-custom-v1.json: open /tmp/master-standalone/custom-custom-v1.json: n
o such file or directory  in object "custom.io/v1/Custom/" in file "resources.yaml"
  Exit code: 1
```

For set-namespace:
```
[FAIL] "gcr.io/kpt-fn/set-namespace:v0.1"
  Results:
    [ERROR] failed to configure function: input namespace cannot be empty
  Stderr:
    "failed to configure function: input namespace cannot be empty"
  Exit code: 1
```

WDYT? @frankfarzan @droot

e2e/testdata/fn-eval/error-in-pipe/.expected/config.yaml
```diff
@@ -0,0 +1,26 @@
...
+  [FAIL] "gcr.io/kpt-fn/set-namespace:v0.1"
+    Results:
+      [ERROR] failed to configure function: input namespace cannot be empty
+    Stderr:
+      "failed to configure function: input namespace cannot be empty"
+    Exit code: 1
+
+   
+  [FAIL] "gcr.io/kpt-fn/dne"
+  error: Function image "gcr.io/kpt-fn/dne" doesn't exist.
```
frankfarzan, line 26: This seems inconsistent compared to above. I understand it's the orchestrator throwing the error, but we should think of a way to present this better.

e2e/testdata/fn-eval/missing-fn-config/.expected/config.yaml
```diff
@@ -18,4 +18,3 @@ image: gcr.io/kpt-fn/set-namespace:v0.1
 stdErr: "failed to configure function: input namespace cannot be empty"
 stdOut: |
   [RUNNING] "gcr.io/kpt-fn/set-namespace:v0.1"
-  [FAIL] "gcr.io/kpt-fn/set-namespace:v0.1"
```
frankfarzan, line 21: Why this change?

Also, I've commented on this in couple of other PRs. What's the rationale for what's printed to stderr vs stdout? Why not just use stderr like when piping? Also if you print to different streams, then they may be presented out of order.

frankfarzan (changes requested):

e2e/testdata/fn-eval/missing-fn-config/.expected/config.yaml
Shell32-Natsu: > Why this change?

I had a discussion with @droot and we agreed that the line `[FAIL] ...` is an error message and should be printed to stderr to provide enough information about which function fails in a command pipe.

> Also, I've commented on this in couple of other PRs. What's the rationale for what's printed to stderr vs stdout? Why not just use stderr like when piping? Also if you print to different streams, then they may be presented out of order.

The detail design for what should be printed to stderr and what should be printed to stdout should be done in #1915. We can have tests to cover the behavior before #1915 is done.

e2e/testdata/fn-eval/error-in-pipe/.expected/config.yaml
Shell32-Natsu: I think we should address all these issue in #1915

Shell32-Natsu: There is another issue https://github.com/GoogleContainerTools/kpt/issues/2084 which is relative to this error message.

e2e/testdata/fn-eval/missing-fn-config/.expected/config.yaml
frankfarzan: ok. I think using two channels is not the way to go for the reasons I stated. So consider it designed. Won't block this PR.

frankfarzan (approved):

PR #2093, opened 2021-05-25
Fix cli usage output for kpt live commands

mortent:
This PR cleans up some issues with the `kpt live` commands:
 * Adds missing docs for `kpt live install-resource-group` command.
 * Updates all `kpt live` commands to use the same structure for `Short`, `Long`, and `Example`.
 * Remove the `match-server-version` flag.


site/reference/live/install-resource-group/README.md
```diff
@@ -0,0 +1,29 @@
...
+linkTitle: "install-resource-group"
+type: docs
+description: >
+Install the ResourceGroup CRD in the cluster.
+---
+<!--mdtogo:Short
+    Install the ResourceGroup CRD in the cluster.
+-->
+
+`install-resource-group` installs the ResourceGroup CRD in the cluster. `kpt`
```
frankfarzan, line 12: explain why/when you should run this command?

mortent: Changed the language a bit here. Ideally we should link to the book here, since a thorough explanation of why this is needed becomes a description of the overall design for `kpt live`.

site/reference/live/install-resource-group/README.md
```diff
@@ -0,0 +1,30 @@
+---
```
frankfarzan, line 1: we don't really need these hugo directives anymore, a simple title suffices. Should do a cleanup at some point.

frankfarzan (approved):

PR #2025, opened 2021-05-17
added github pre-commit hook for make generate

NagaHarshita:
Added `make generate` to the pre-commit hook similar to the [GoogleContainerTools/kpt-functions-catalog#287](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/287). This is to be enabled manually from the client-side by using following commands 
- `git config core.hooksPath .github/hooks`
- `chmod +x .github/hooks`

PR #2110, opened 2021-05-26
Check docs synchronization

runewake2:
This wraps the pre-commit hook introduced in #2025 with a GitHub Action. This can help catch documentation desync if the pre-commit hook is not installed on the contributors machine and helps provide visibility to the state of documentation for reviewers.

There is potential future improvement to automatically resolve these issues with documentation desync by automatically committing make generate results but that is a larger task.

### DIFF
--- a/.github/workflows/checkDocSync.yml
+++ b/.github/workflows/checkDocSync.yml
@@ -14,10 +14,6 @@
 
 name: Check Documentation Synchronized
 on:
-  push:
-    branches:
-      - master
-      - next
   pull_request:
     branches:
       - master
--- a/.github/workflows/checkDocSync.yml
+++ b/.github/workflows/checkDocSync.yml
@@ -28,4 +28,4 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: |
-          ./.github/pre-commit
+          ./.github/hooks/pre-commit
--- a/.github/workflows/checkDocSync.yml
+++ b/.github/workflows/checkDocSync.yml
@@ -1,0 +1,31 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


mikebz:
LGTM, Morten can be the final approver though.

runewake2:
One note about merging this that we should keep in mind if we move forward with this change. It will cause some changes made only in GitHub to fail, if an edit is submitted to update our docs that includes content that is generated that PR will be blocked by this check. This will affect our reference docs more than others but may be inconvenient for new contributors trying to submit quick fixes. Only changes made within the `mdtogo` tags would encounter this.

Depending on the impact of that limitation we may want to prioritize automating the doc generation synchronization differently (if PRs are staying open for extended periods or abandoned due to this limitation).

We have a few options for that automation:

1. Have `make generate` be a dependency of `make build`. This should cause docs to be generated prior to any build and cause our builds to pull those updated docs in. I'm not sure what other effects that approach might have (we may be able to delete the entire generated codebase if it works well or there may be other complications). I believe this approach may create issues in dev tools (for example inconsistent results between `make test` and `go test` because `go test` does not generate docs).
2. Have an automated job to create new PRs with doc updates or append commits when docs are updated. I've done something like this in the past in other projects.
3. There may be an opportunity to leverage `go generate` to accomplish something like option 1 but I need to learn more about how to use that feature.

I'll be pulling this info into an issue to track automation of this in the future once this is merged.

mikebz:
@runewake2 thanks for explaining the downside.  I think the downside here might outweight the upside.  if someone can't quickly update the docs (because they are in the section that gets compared) that I believe is a serious downside.  Two questions:
1) if we are already running a pre-commit hook does that alleviate the issues upstream?
2) can github just run this check like a cron job and just notify the contributors when the docs are out of sync?

Thank you!

runewake2:
The pre-commit check needs to be manually installed locally to be effective, this change would eliminate needing to do that.

If we just want a warning we can move this to trigger on pushes. The status will not be visible in PRs anymore but that will result in code owners getting an alert when desync occurs instead of blocking pr merge.

I don't think it'd be much work to have those failures create prs themselves either but idk how that'd interact with our CLA.

mortent:
I think this solution is fine for now. Longer-term I think we want this to happen automatically either as part of the PR jobs or just something that we generate on every merge.

mortent (approved):

PR #2111, opened 2021-05-26
Surface docker pull output for fn images.

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1948

Example output if image doesn't exist locally:
```Package "basicpipeline": 
[RUNNING] "gcr.io/kpt-fn/set-namespace:unstable"
unstable: Pulling from kpt-fn/set-namespace
339de151aab4: Already exists 
f3760b887006: Pull complete 
Digest: sha256:0406befc8abf98e5d597f0bd7cd2d92a514f060b7d05c492f3834218ae447e28
Status: Downloaded newer image for gcr.io/kpt-fn/set-namespace:unstable
gcr.io/kpt-fn/set-namespace:unstable
[PASS] "gcr.io/kpt-fn/set-namespace:unstable"
[RUNNING] "gcr.io/kpt-fn/set-label:unstable"
[PASS] "gcr.io/kpt-fn/set-label:unstable"

Successfully executed 2 function(s) in 1 package(s).```


natasha41575:
Could you provide an example of the output of kpt fn eval if the image does not exist locally?

etefera:
> Could you provide an example of the output of kpt fn eval if the image does not exist locally?

Added the output when complete. Progress bar disappears a bit too quickly when trying to copy it.

natasha41575:
Maybe keep an eye on Sam's test here https://github.com/GoogleContainerTools/kpt/pull/1982. When it is merged I think we can use that test to check the stdout here

Shell32-Natsu:
I am hesitating to merge this PR before #1915 is finished. And this UX change may need a doc.

mikebz:
@droot @phanimarupaka @mengqiy do we want to have this?  If not then should we close it?

droot:
The way this is implemented currently clobbers the kpt output with docker output and isn't a great UX. Will close this because in the short term, this isn't needed. Eventually, we can support verbosity flag for debugging function runtime specific logs.

PR #2113, opened 2021-05-26
Clean up demo videos.

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1654


mikebz:
when I press play the video resizes.

etefera:
> when I press play the video resizes.

Addressed and the demo has been updated.

mikebz:
LGTM

mortent:
The video looks good to me. Do we have a way to run e2e tests on the original script for the video so we can make sure it doesn't become stale? We can do that separately, but we should probably keep the original issue open until we have that in place.

etefera:
> The video looks good to me. Do we have a way to run e2e tests on the original script for the video so we can make sure it doesn't become stale? We can do that separately, but we should probably keep the original issue open until we have that in place.

No tests for the script yet but I do agree that should be tracked. It should probably be in a different issue, however, tracking an enhancement for these casts.

frankfarzan:
preview page doesn't work.

etefera:
> preview page doesn't work.

Updated URL.

mortent (approved):

frankfarzan:
Several issues:
1. https://screenshot.googleplex.com/4F8bwgMjVYHWXSK
2. IMO, The video is too long and slow. Some suggestions:
   - We don't need to demo the entire quick start guide
   - Speed up playback speed
3. Consider moving into a separate page with a call-to-action link from the overview

frankfarzan:
FYI, this is not a high priority issue (it's an good-to-have M4), so don't spend too much time on it in M3.

mikebz:
can we merge what we have and just create an issue for other follow ups or keep this issue open?

frankfarzan:
This requires more work, IMO it's not emphasizing the right aspects. e.g.:
- render is missing
- we're not showing the Kptfile pipeline
- eval is glossed over
- etc.

Given it's not a blocker, I suggest putting the issue on the backlog.

mikebz:
can we just merge this and iterate?

frankfarzan:
This is a good-to-have so we don't need to rush it before addressing the comments.

etefera:
Reopening and changing scope to just deleting unused videos.

droot (approved):

PR #2116, opened 2021-05-27
Fix reading from stdin for kpt live

mortent:
This PR updates the ManifestReader implementation we use to find the resources that should be applied and the inventory information. The main changes are:
 * It removes the use of the `Loader` interface. This was introduced to allow reusing the cli commands in `cli-utils` in kpt where we use the Kptfile and the ResourceGroup. We no longer reuse the cli commands from `cli-utils`, so we can simplify the code by removing the abstraction.
 * It also removes the DualDelegation implementation of the Provider and Loader interfaces. 
 * It fixes the issue where reading resources from stdin didn't work.
 * When reading resources from disk, it ignores files that are referenced as function config in Kptfiles.

The error messages regarding the inventory is not good enough. I plan to address those in a follow-up (since this is already a huge PR).


commands/livecmd.go
```diff
@@ -61,28 +61,25 @@ func GetLiveCommand(ctx context.Context, _, version string) *cobra.Command {
...
 
 	// Add the migrate command to change from ConfigMap to ResourceGroup inventory
-	// object. Also add the install-resource-group command.
+	// object.
 	klog.V(2).Infoln("adding kpt live migrate command")
-	migrateCmd := cmdmigrate.NewCommand(ctx, cmProvider, rgProvider, cmLoader, rgLoader, ioStreams)
-	installRGCmd := GetInstallRGRunner(f, ioStreams).Command
-	liveCmd.AddCommand(migrateCmd, installRGCmd)
+	cmProvider := provider.NewProvider(f)
+	cmLoader := manifestreader.NewManifestLoader(f)
```
seans3, line 79: We might want to add a `TODO` to remove this `ConfigMap Loader` once we get move past the inventory migration.

seans3 (commented):
Looks good. One small nit.

/lgtm

seans3 (approved):

PR #2122, opened 2021-05-27
remove `config` in function spec

natasha41575:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1661


e2e/testdata/fn-eval/include-meta-resources-v1alpha1/Kptfile
```diff
@@ -12,13 +12,8 @@ dependencies:
...
     functions:
       - image: gcr.io/kpt-fn/my-func:v0.1
-        config:
-          apiVersion: v1
-          kind: ConfigMap
-          metadata:
-            name: function-config
-          data:
-            param1: value1
+        configMap:
```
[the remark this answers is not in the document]
natasha41575: You're right, I don't think we do. I'll change that back.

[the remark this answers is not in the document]
natasha41575: Thanks!

pkg/api/kptfile/v1alpha2/validation.go
```diff
@@ -79,21 +79,10 @@ func (f *Function) validate(fnType string, idx int) error {
...
-			}
-		}
-		configFields = append(configFields, "config")
-	}
 	if len(configFields) > 1 {
 		return &ValidateError{
 			Field: fmt.Sprintf("pipeline.%s[%d]", fnType, idx),
-			Reason: fmt.Sprintf("only one of 'config', 'configMap', 'configPath' can be specified. Got %q",
-				strings.Join(configFields, ", ")),
+			Reason: "only one of 'configMap', 'configPath' can be specified; got both",
```
Shell32-Natsu, line 85: We can use avoid using slice `configFields` but a simple counter here.

natasha41575: Instead of a counter, I just an if statement. The code is much simpler with only 2 fields to check. Thanks for pointing this out

Shell32-Natsu (approved):
lgtm

pkg/api/kptfile/v1alpha2/validation.go
```diff
@@ -74,27 +73,12 @@ func (f *Function) validate(fnType string, idx int) error {
...
-	}
-	if !IsNodeZero(&f.Config) {
-		config := yaml.NewRNode(&f.Config)
-		if _, err := config.GetMeta(); err != nil {
+		if len(f.ConfigMap) != 0 {
 			return &ValidateError{
-				Field:  fmt.Sprintf("pipeline.%s[%d].config", fnType, idx),
-				Reason: "functionConfig must be a valid KRM resource with `apiVersion` and `kind` fields",
+				Field: fmt.Sprintf("pipeline.%s[%d]", fnType, idx),
+				Reason: "only one of 'configMap', 'configPath' can be specified; got both",
```
droot, line 79: An alternative messaging that is more inline with http://go/kpt-errors-style-guide

functionConfig must not specify `configMap` and `configPath` both at the same time.

pkg/api/kptfile/v1alpha2/validation.go
```diff
@@ -74,27 +73,12 @@ func (f *Function) validate(fnType string, idx int) error {
...
 		}
-		configFields = append(configFields, "configPath")
-	}
-	if len(f.ConfigMap) != 0 {
-		configFields = append(configFields, "configMap")
-	}
-	if !IsNodeZero(&f.Config) {
-		config := yaml.NewRNode(&f.Config)
-		if _, err := config.GetMeta(); err != nil {
+		if len(f.ConfigMap) != 0 {
```
droot, line 76: I think having an explicit check for this condition outside this the `if` will be better because user may not get this error if they have an invalid `functionConfig` and also `configMap` at the same time, right ?

```
 if len(f.ConfigMap) != 0 && f.ConfigPath != "" {
   // throw error
 }
```

droot (changes requested):

pkg/api/kptfile/v1alpha2/validation.go

droot (approved):

PR #2127, opened 2021-05-27
Improve error messages when inventory information is missing or invalid

mortent:
This PR improves the error messages displayed to users concerning the inventory information:
 * If there are multiple `Kptfile` resources with inventory information, we report that as an error to the user. This only applies to when we are reading resources from stdin where we have no way to decide which Kptfile resource we should use. When reading from disk, we only look for this information in the Kptfile in the root package.
 * If any of `name`, `namespace`, or `inventory-id` are not set when we are trying to apply/destroy, we report an error.

Examples:

Multiple Kptfiles with inventory:
```
Error: Multiple Kptfile resources with inventory information found. Please make sure at most one Kptfile resource contains inventory information.
```

Inventory name is missing:
```
Error: The inventory information is not valid. Please update the information in the Kptfile or provide information with the command line flags. To generate the inventory information the first time, use the 'kpt live init' command.

Details:
"inventory.name" must not be empty
```

Must be merged after https://github.com/GoogleContainerTools/kpt/pull/2116

### DIFF
--- a/internal/cmdapply/cmdapply.go
+++ b/internal/cmdapply/cmdapply.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/GoogleContainerTools/kpt/internal/cmdutil"
+	"github.com/GoogleContainerTools/kpt/internal/util/strings"
 	"github.com/GoogleContainerTools/kpt/pkg/live"
 	"github.com/GoogleContainerTools/kpt/thirdparty/cli-utils/flagutils"
 	"github.com/GoogleContainerTools/kpt/thirdparty/cli-utils/printers"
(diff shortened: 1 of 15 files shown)


frankfarzan:
> This PR improves the error messages displayed to users concerning the inventory information:
> 
> * If there are multiple `Kptfile` resources with inventory information, we report that as an error to the user. This only applies to when we are reading resources from stdin where we have no way to decide which Kptfile resource we should use. When reading from disk, we only look for this information in the Kptfile in the root package.
> * If any of `name`, `namespace`, or `inventory-id` are not set when we are trying to apply/destroy, we report an error.
> 
> Examples:
> 
> Multiple Kptfiles with inventory:
> 
> ```
> Error: Multiple Kptfile resources with inventory information found. Please make sure at most one Kptfile resource contains inventory information.

Should print the the path to Kptfile(s).

> ```
> 
> Inventory name is missing:
> 
> ```
> Error: The inventory information is not valid. Please update the information in the Kptfile or provide information with the command line flags. To generate the inventory information the first time, use the 'kpt live init' command.

https://github.com/GoogleContainerTools/kpt/blob/next/docs/style-guides/errors.md

Be more precise. Reference the field in question.

> 
> Details:
> "name" must not be empty
> ```
> 
> Must be merged after #2116

mortent:
> Should print the the path to Kptfile(s).

This is when reading files from Stdin, so there is no filepath. This is also the reason why the error message references Kptfile resources. This can't happen when reading from disk, since we only consider inventory information in the Kptfile in the root package.

> Be more precise. Reference the field in question.

Fixed.

frankfarzan (approved):

PR #2129, opened 2021-05-27
Fix test

Shell32-Natsu:
- Fix the test failure due to new function release
- Pin the functions version used in tests.


runewake2 (approved):
Thanks for putting this together!

Issue #1502, opened 2021-03-01
High level comparison of kpt hydration with kustomize


droot:
The intention here is to do high level comparison of the kpt hydration features with kustomize. It will help us identify the gaps as well as help us understand in which scenario one is better than the other.

mikebz:
I think the FAQ is a first stab of addressing this.  I am going to wait for customer questions to see what else people want to know about.

Issue #1545, opened 2021-03-13
[Question] clarifications on `live` commands relationship with kustomize


marshall007:
We have recently adopted `kpt`/`kustomize` as our preferred mechanism for maintaining and distributing a consistent baseline cluster configuration. The tooling is all great, but we are still trying to understand best practices and continue developing our own.

Discussion in issues like #1447 make it seem like we're going about it all wrong... our understanding of `kpt`'s primary role is that it provides strong advantages over referencing remotes from your `kustomization.yaml` files inline:

- more suitable for gitops in terms of being able to track and pull in upstream changes
- ability to create/apply setters
- confidently make arbitrary modifications to the upstream directory (when necessary) and still track the remote changes with 3-way merges

In practice, there is little consistently in terms of whether the upstream remote is:

- a well-maintained `kustomize` package with good overlays/patches/components
- a poorly maintained `kustomize` package that only provides you a "kitchen sink" deployment
- just a set of static manifests

Since our goal is to provide a consistent set of sub-packages that play well together it seemed like our best option was to wrap the upstream of each sub-package with overlays to support consumer use cases + a `kustomization.yaml` at the root representing a default installation. We also create an `upstream/kustomization.yaml` if it doesn't exist so that (1) it can be easily referenced from another `kustomization` and (2) that we can include necessary `patches`/`nameReference` overrides/etc so that everything works as expected when further kustomizations are applied.

This feels right as both a publisher and consumer, but only because we are putting the extra work in to guarantee to the consumer that `kustomize build <dir> | kpt live apply` always works to deploy any sub-package or all of them at once.

It would be much easier for us to use, easier to explain to consumers, and safer default behavior if all `kpt live` commands were aware of `Kustomization` resources and built them appropriately.

I am still getting caught up to speed on the project but I've seen several issues (like https://github.com/GoogleContainerTools/kpt/issues/407#issuecomment-743704436) where it seems like this is loosely planned but it is not actually mentioned in the roadmap and similar issues have been closed as won't fix.

Thanks in advance for any feedback on our approach and/or clarifications around future integration with `kustomize`!

mikebz:
Hi @marshall007, first of all thank you for being an active user of our toolchain.  There is a lot to unpack here.

## upstream remotes 

First of all you are correct that remotes have a stability gap over pulling packages locally.  We have seen that from customers and that is one of the key motivations for kpt (even though remotes are going to be hard to unwind because of how popular kustomize is).  

[...]

PR #2136, opened 2021-05-29
new kustomize sample

mikebz:
This is a new kustomize example which is more in line with how we want people to think about kustomize rendering with kpt live apply and packages.

It replaces the helloworld-kustomize but was started from scratch.  The modifications to the files are just git trying to be smart.

related to #1502


package-examples/kustomize-pkg/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+
+### View the package contents
+
+List the package contents in a tree structure.
+
+```shell
+$ kpt pkg tree kustomize-pkg/
+
+PKG: kustomize-pkg
```
mortent, line 31: While the output of `kpt pkg tree` here is accurate, it is kinda misleading. It hides the `bases` and `overlays` directories. It seems like we remove any directories that doesn't have any KRM resources inside them.

package-examples/kustomize-pkg/README.md
```diff
@@ -0,0 +1,163 @@
...
+│   ├── [kustomization.yaml]  Kustomization 
+│   └── [pass-patch.yaml]  Deployment deployment-patch
+└── prod
+    ├── [kustomization.yaml]  Kustomization 
+    └── [pass-patch.yaml]  Deployment deployment-patch
+```
+
+### kustomize the config
+
+We recommend that you keep the kustomize instructions to rendering only such as adding a namespace, transforming or applying a patch.  The second you mix kpt packages and remote bases you will be missing out on a big advantage of having a guaranteed stable base.
```
mortent, line 47: Can we make this clearer? It is not clear to me how using kustomize remote bases would necessarily be bad here, other than challenges around mixing packages fetched with kpt and remote bases with kustomize.

package-examples/kustomize-pkg/README.md
```diff
@@ -0,0 +1,163 @@
...
+    commit: 4d2aa98b45ddee4b5fa45fbca16f2ff887de9efb
+    directory: package-examples/nginx
+    ref: package-examples/nginx/v0.2
+    repo: https://github.com/GoogleContainerTools/kpt
+  type: git
+```
+
+### Apply the package
+
+It's possible to use kustomize build and kpt live apply, but it does require passing the inventory to `kpt live apply` from `kustomize build` output.  One of the base Kptfiles needs to be initialized with the inventory object.  We will use the nginx base, but in real life exployment it might be better to initialize inventory on your root package.
```
mortent, line 138: Initializing the inventory information in the base is not the recommended way. It means that every variant created (so `dev` and `prod` in this example) would have the same inventory, and thus would collide if installed in the same cluster.
Also, I think we should refer to the information added to the Kptfile about the inventory as `inventory information`. It is a bit vague and not a great name, but we usually refer to the actual resource in the cluster as the inventory object.

package-examples/kustomize-pkg/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+We recommend that you keep the kustomize instructions to rendering only such as adding a namespace, transforming or applying a patch.  The second you mix kpt packages and remote bases you will be missing out on a big advantage of having a guaranteed stable base.
+
+You can edit the patches and kustomize files we have created in the overlay folders.
+
+```yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../../bases/nginx
```
mortent, line 55: This structure is probably ok for an example, but I think the recommended way to do this would be to have each env have separate git references. The structure in this example makes it a bit hard to make changes in the base and then do progressive rollouts into different environments.

package-examples/kustomize-pkg/README.md
frankfarzan: That gets convoluted though as you're effectively using two package managers concurrently. It's hard even for me to reason about it. Instead, you want to have a clean conceptual separation of concerns. The price is that you cannot just drop any arbitrary existing kustomization resource since it may use remote bases.

+1 to making this intentional limiting of kustomize features more precise. Explicitly state what features of kustomize is not recommended?

package-examples/kustomize-pkg/README.md
```diff
@@ -0,0 +1,163 @@
...
+    repo: https://github.com/GoogleContainerTools/kpt
+  type: git
+```
+
+### Apply the package
+
+It's possible to use kustomize build and kpt live apply, but it does require passing the inventory to `kpt live apply` from `kustomize build` output.  One of the base Kptfiles needs to be initialized with the inventory object.  We will use the nginx base, but in real life exployment it might be better to initialize inventory on your root package.
+
+```shell
+$ kpt live init bases/nginx
```
frankfarzan, line 141: I don't see a `kpt live apply` command below.

Also, is the assumption that the they want to use `kpt live apply`? Or does the user want to use packaging but not hydration and live? Providing both options makes sense.

package-examples/kustomize-pkg/README.md
mikebz: it's a good observation and I think we should absolutely see if we can fix the tree output.

package-examples/kustomize-pkg/README.md
mikebz: great feedback, would you recommend having a Kptfile in every overlay?

package-examples/kustomize-pkg/README.md
mikebz: What would be the layout of a solution like that?  Every overlay then gets a separate base?  Why create an overlay then?  We can just have the final config in every folder.  I haven't seen a pattern like that, even though I am not opposed to showing it.

package-examples/kustomize-pkg/README.md
mikebz: Live apply from STDIN currently doesn't work so that's why we don't have the output.  What I have seen is that people who are using kustomize and see the value in kpt usually like both areas: packaging and apply.  They generally already know what to do with `kubectl -k` or `kustomize build` and piping to `kubectl -f -`.  Providing them an example of doing that probably doesn't add any value in addition to what they can already gather from kustomize docs.

package-examples/kustomize-pkg/README.md
mikebz: I added stronger language, but I think enumerating every feature of kustomize is going to create a wackamole game.  There are new features added and deprecated as we speak and it's best to have a larger umbrella like hydration and packaging in order to avoid updating the sample language every time something changes in kustomize.

package-examples/kustomize-pkg/README.md
mikebz: I moved the kpt inventory to ovrelays. PTAL

mikebz:
also worth noting that this sample will help with addressing #1545

package-examples/kustomize-pkg/README.md
```diff
@@ -0,0 +1,158 @@
...
+4. [Render the kustomization](#render-the-kustomization)
+5. [Apply the package](#apply-the-package)
+6. [Clean up resources](#clean-up-resources)
+
+### Fetch the package
+
+Get the example package on to local using `kpt pkg get`. Note that this package is for this example, wihin that package we are also using the nginx sub-package which is an alternative to having a remote base. 
+
+```shell
+$ kpt pkg get https://github.com/GoogleContainerTools/kpt.git/package-examples/kustomize-pkg
```
frankfarzan, line 19: don't you want to specify the `next` branch?
Does these commands work currently?

package-examples/kustomize-pkg/README.md
```diff
@@ -0,0 +1,158 @@
...
+### View the package contents
+
+List the package contents in a tree structure.
+
+```shell
+$ kpt pkg tree kustomize-pkg/
+
+PKG: kustomize-pkg
+├── [Kptfile]  Kptfile kustomize-pkg
+├── PKG: nginx
```
frankfarzan, line 34: Part of the tree has changed @phanimarupaka

package-examples/kustomize-pkg/README.md
frankfarzan: you can use `tree` command for now

package-examples/kustomize-pkg/README.md
mikebz: this is a copy from the template and the rest of them are using the default branch which will be main soon.  I don't think we should single this sample out.

frankfarzan: are you saying no one should try this example before we make `next` default?

mikebz: We have all the rest of the samples that do not point to a branch.  I am fine making this point to next just to move forward but until now no one has complained about the other samples. 

Two theories:
1) people looking at samples already understand how to point at the right branches and are in tune with the product lifecycle
2) people are looking at the samples to just understand the flow and read the pattern.

mikebz: I will change it to next, please update it to point to default branch when you change the branches over.

frankfarzan: Two wrongs don't make a right. People are consuming wordpress and nginx through the book which have correct versioning using tags. Just add the next branch to make this work in the meantime.

mikebz: all the existing samples are updated.

frankfarzan (approved):

PR #2145, opened 2021-06-02
Add details about the merge strategies to the Kpt book

mortent:
This describes the resource-merge strategy in some more details. I'm uncertain how it fits within the flow of the book as it seems like focused on specific examples. We can consider whether it fits better somewhere else on the site.

### DIFF
--- a/site/book/03-packages/05-updating-a-package.md
+++ b/site/book/03-packages/05-updating-a-package.md
@@ -1,3 +1,4 @@
+
 An independent package records the exact commit where the local fork and the
 upstream package diverged. This enables kpt to fetch any update to the upstream
 package and merge it with local changes.
@@ -53,7 +54,6 @@ The `update` command updates the local `wordpress` package and the dependent
 2. New upstream commit
 3. Local (edited) package
 
-<<<<<<< HEAD
 Several different strategies are available to handle the merge. By default, the
 `resource-merge` strategy is used which performs a structural comparison of the
 resource using OpenAPI schema.
--- a/site/book/03-packages/05-updating-a-package.md
+++ b/site/book/03-packages/05-updating-a-package.md
@@ -117,7 +117,7 @@ the three different sources as local, upstream and origin.
 In the discussion, we will be referring to non-associative and associative lists. A
 non-associative list either has elements that are scalars or another list, or it has elements
 that are mappings but without an associative key. An example of this in the kubernetes
-API is the command property on containers:
+API is the `command` property on containers:
 ```yaml
 apiVersion: v1
 kind: Pod
@@ -159,6 +159,8 @@ associative keys:
 * `name`
 * `containerPort`
 
+The 3-way merge algorithm operates both on the level of each resource and on
+each individual field with a resource. The rules are:
 
 | Resource level |
 | --- |
--- a/site/book/03-packages/05-updating-a-package.md
+++ b/site/book/03-packages/05-updating-a-package.md
@@ -76,15 +76,15 @@ $ git add .; git commit -m "Updated wordpress to v0.4"
 
 ### Resource-merge
 
-The `resource-merge` strategy performs a structural comparison of each resource using the
-OpenAPI schema. So rather than performing a text-based merge, `kpt` leverages the
+The resource-merge strategy performs a structural comparison of each resource using the
+OpenAPI schema. So rather than performing a text-based merge, kpt leverages the
 common structure of KRM resources.
 
 #### Resource identity
-In order to perform a per-resource merge, `kpt` needs to be able to match a resource in
+In order to perform a per-resource merge, kpt needs to be able to match a resource in
 the local package with the same resource in the upstream version of the package. It does
 this matching based on the identity of a resource, which is the combination of group,
-kind, name and namespace. So in our `wordpress` example, the identity of the`Deployment`
+kind, name and namespace. So in our wordpress example, the identity of the`Deployment`
 resource is:
 ```
 group: apps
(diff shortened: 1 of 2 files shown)


site/book/03-packages/05-updating-a-package.md
```diff
@@ -65,4 +65,97 @@ Once you have successfully updated the package, commit the changes:
...
+## Merge strategies
+
+### Resource-merge
+
+The `resource-merge` strategy performs a structural comparison of each resource using the
+OpenAPI schema. So rather than performing a text-based merge, `kpt` leverages the
+common structure of KRM resources.
+
+#### Resource identity
+In order to perform a per-resource merge, `kpt` needs to be able to match a resource in
```
frankfarzan, line 77: https://docs.google.com/document/d/1DGN_1VWechH0agxExQxDYpKVSLqE5r1yKd95m_rE3JQ/edit?hl=en

Rule D5

Same below

site/book/03-packages/05-updating-a-package.md
```diff
@@ -65,4 +65,97 @@ Once you have successfully updated the package, commit the changes:
...
+```
+
+#### Merge rules
+`kpt` performs a 3-way merge for every resource. This means it will use the resource
+in the local package, the updated resource from upstream, as well as the resource
+at the version where the local and upstream package diverged (i.e.
+common ancestor). When discussing the merge rules in detail, we will be referring to
+the three different sources as local, upstream and origin.
+
+In the discussion, we will be referring to non-associative and associative lists. A
```
frankfarzan, line 110: Prefer to use examples over lengthy descriptions like this. It's more effective and succint.

site/book/03-packages/05-updating-a-package.md
```diff
@@ -65,4 +65,97 @@ Once you have successfully updated the package, commit the changes:
...
+associative keys:
+* `mountPath`
+* `devicePath`
+* `ip`
+* `type`
+* `topologyKey`
+* `name`
+* `containerPort`
+
+On the resource level, the merge rules are as follows:
```
frankfarzan, line 126: This is probably better represented as as a table. Same for other rules below.

site/book/03-packages/05-updating-a-package.md
mortent: Added some examples here. I think the descriptions need to still be there.

site/book/03-packages/05-updating-a-package.md
mortent: I tried to turn this into tables. I'm not sure how it works.

site/book/03-packages/05-updating-a-package.md
```diff
@@ -65,4 +65,132 @@ Once you have successfully updated the package, commit the changes:
...
+kpt performs a 3-way merge for every resource. This means it will use the resource
+in the local package, the updated resource from upstream, as well as the resource
+at the version where the local and upstream package diverged (i.e.
+common ancestor). When discussing the merge rules in detail, we will be referring to
+the three different sources as local, upstream and origin.
+
+In the discussion, we will be referring to non-associative and associative lists. A
+non-associative list either has elements that are scalars or another list, or it has elements
+that are mappings but without an associative key. An example of this in the kubernetes
+API is the command property on containers:
```
frankfarzan, line 113: Style guide:

command -> `command`

site/book/03-packages/05-updating-a-package.md
```diff
@@ -65,4 +65,132 @@ Once you have successfully updated the package, commit the changes:
...
+* `mountPath`
+* `devicePath`
+* `ip`
+* `type`
+* `topologyKey`
+* `name`
+* `containerPort`
+
+
+| Resource level |
```
frankfarzan, line 156: add a sentence or two describing this table, it comes out of nowhere

site/book/03-packages/05-updating-a-package.md
```diff
@@ -65,4 +65,132 @@ Once you have successfully updated the package, commit the changes:
...
+| --- |
```
frankfarzan, line 157: This may be missing the suggestion to use table (Single column table is just a list).

Can you formulate this using 4 columns as follows:

C1: Origin
C2: Upstream
C3: Local
C4: Result

Same for other tables below. Otherwise, I still find this hard to read.

frankfarzan (changes requested):

site/book/03-packages/05-updating-a-package.md
mortent: I agree we should look at the best way to explain this, but I'm not convinced adding tables is the right approach here. It would effectively mean enumerating all the possible options instead of the explaining the fundamental rules which is encoded in the algorithm. There is considerable number of possible permutations, so the result will be four pretty large tables.

site/book/03-packages/05-updating-a-package.md
```diff
@@ -53,9 +53,14 @@ The `update` command updates the local `wordpress` package and the dependent
 2. New upstream commit
 3. Local (edited) package
 
+<<<<<<< HEAD
```
frankfarzan, line 56: merge artifact

site/book/03-packages/05-updating-a-package.md
frankfarzan: I think we have two options:

1- either simplify this section by presenting the same information in a better way (original comment here)
2- remove these details (Specifically the `merge rules` content) from the book as it's disrupts the linear flow of a user reading the book. This is really more of a appendix/reference for merge strategy and is unlike other chapters.

Since 1 doesn't sound feasible, I suggest going with 2. These tables can be moved to reference doc for update and linked from here. WDYT?

mortent: Yeah, I agree with that. Regardless of how we present this data, it probably gets into more detail than we want in the book. I think we will also run into other topics that are too nitty gritty for the book, but also doesn't fit in the reference docs.

mortent: I have moved the content to the `update` reference docs. I put it in a separate **Details** section at the end of the doc. I don't want to put too much details early so users have to scroll in order to see the support args and flags.

Another challenge is the way we use the markdown headers. In the reference section, even top-level section use header 3 (`###`), so the available header hierarchy in these pages are very limited. Also, our site style makes header 5 text smaller than the main body, which looks weird. Created a separate issue for this: https://github.com/GoogleContainerTools/kpt/issues/2303

frankfarzan (approved):

PR #2146, opened 2021-06-02
`fn render` will fail if the function config has multiple resources.

### LINKED ISSUE #1959: Fn render should error when there are multiple resources in a fn config file
Current behavior: Only the first one is respected and the rest are ignored.

Shell32-Natsu:
close #1959

### DIFF
--- a/thirdparty/kyaml/runfn/runfn.go
+++ b/thirdparty/kyaml/runfn/runfn.go
@@ -402,7 +402,8 @@ func (r *RunFns) getFunctionConfig() (*yaml.RNode, error) {
 		return nil, fmt.Errorf("failed to read function config %q: %w", r.FnConfigPath, err)
 	}
 	if len(nodes) > 1 {
-		return nil, fmt.Errorf("function config file %q must not contain more than 1 config", r.FnConfigPath)
+		return nil, fmt.Errorf("function config file %q must not contain more than 1 config, got %d",
+			r.FnConfigPath, len(nodes))
 	}
 	return nodes[0], nil
 }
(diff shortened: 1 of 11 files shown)


thirdparty/kyaml/runfn/runfn.go
```diff
@@ -386,23 +386,24 @@ func getUIDGID(asCurrentUser bool, currentUser currentUserFunc) (string, error)
 	return fmt.Sprintf("%s:%s", u.Uid, u.Gid), nil
 }
 
-// etFunctionConfig returns yaml representation of functionConfig that can
+// getFunctionConfig returns yaml representation of functionConfig that can
 // be provided to a function as input.
 func (r *RunFns) getFunctionConfig() (*yaml.RNode, error) {
```
natasha41575, line 391: can this function be moved to fnruntime/runner.go and shared between `eval` and `render`? What's the reason for keeping them separate?

natasha41575:
also please keep an eye on https://github.com/GoogleContainerTools/kpt/pull/2133, I think there is some overlap and may affect where you want to do this check. I guess it depends on which one we want to merge first

Shell32-Natsu:
> also please keep an eye on #2133, I think there is some overlap and may affect where you want to do this check. I guess it depends on which one we want to merge first

I think it's probably better to check this during validating. One of my concern is we will need to read the config file twice, one for validating one for using. @droot WDYT?

Shell32-Natsu:
Another problem is `fn eval` doesn't validate the content in function config file. We should probably keep consistent with `render`.

natasha41575:
@Shell32-Natsu I just updated #2133 to share a function to validate the function config file. Please take a look. I think it should also cover #1959.

Shell32-Natsu:
> @Shell32-Natsu I just updated #2133 to share a function to validate the function config file. Please take a look. I think it should also cover #1959.

Right, the method in #2133 is better. Close this one.

PR #2149, opened 2021-06-03
Adds a KptPackageWalker that supports ignore patterns

mortent:
This adds `KptPackageWalker` that can walk a kpt package and invoke the provided callback function for every file/directory in the package that matches the provided criteria. It also supports ignore patterns that may exist in the Kptfile. This replaces the previous `WalkPackage` function that was part of the `pkgutil` package.

This will be leveraged in a follow-up PR to implement a `KptPackageReader` that implements the `kio.Reader` interface.

Must merge after https://github.com/GoogleContainerTools/kpt/pull/2148


internal/pkg/walker.go
```diff
@@ -0,0 +1,148 @@
...
+		return true
+	}
+	return false
+}
+
+// KptfileYamlFileMatchFunc is a matcher func that matches the Kptfile and files
+// with either .yaml or .yml extensions.
+var KptfileYamlFileMatchFunc FileMatchFunc = func(s string, info os.FileInfo) bool {
+	if info.IsDir() {
+		return true
```
droot, line 52: same comment as above.

internal/pkg/walker.go
```diff
@@ -0,0 +1,148 @@
...
+	if info.Name() == kptfilev1alpha2.KptFileName {
+		return true
+	}
+
+	ext := filepath.Ext(info.Name())
+	if ext == ".yaml" || ext == ".yml" {
+		return true
+	}
+	return false
+}
```
droot, line 64: probably we can just do `return YamlFilesMatch(..) && (info.Name() == 'kptfile const')

internal/pkg/walker.go
```diff
@@ -0,0 +1,148 @@
...
+
+// WalkFunc defines the type for the callback function used by the
+// Walker.Walk function.
+type WalkFunc func(string, os.FileInfo, error) error
+
+// FileMatchFunc is the type of a function that can be provided to the
+// Walker to filter the files that will be provided to the caller.
+type FileMatchFunc func(string, os.FileInfo) bool
```
droot, line 72: Adding the behavior for `dir` here will help the reader.

internal/pkg/walker.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	// If the Kptfile could be read and it has at least one ignore pattern,
+	// turn it into a string.
+	var ignorePatterns string
+	if err == nil && len(kf.Ignore) > 0 {
+		ignorePatterns = strings.Join(kf.Ignore, "\n")
+	}
+
+	pkgPath := pkg.UniquePath.String()
```
droot, line 106: We should prefer `string(pkg.UniquePath)` over `pkg.UniquePath.String()` because `.String()` may not always return what we want here simply because it can be changed for pretty display.

internal/pkg/walker.go
```diff
@@ -0,0 +1,148 @@
...
+func (p *Walker) Walk(pkg *Pkg, cb WalkFunc) error {
+	kf, err := pkg.Kptfile()
+
+	// Any other errors than os.ErrNotExist means we should error out.
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+
+	// If the Kptfile could be read and it has at least one ignore pattern,
+	// turn it into a string.
```
droot, line 100: I think explaining `why` is more useful than `how` here ( I am guessing this is because the `gitignore` pkg needs ignore-pattern in that form).

internal/pkg/walker.go
```diff
@@ -0,0 +1,148 @@
...
+// that will be provided to the caller through the callback function.
+type Walker struct {
+	// FileMatchFunc defines the function that will be used by the walker to
+	// filter which files should be provided to the callback. By default, all
+	// files will be provided.
+	FileMatchFunc FileMatchFunc
+
+	// HonorKptfileIgnore defines whether the walker should honor the ignore
+	// list from the package Kptfile.
+	HonorKptfileIgnore bool
```
droot, line 85: We probably need the inverse of this to make default behavior to make the default to `honor`. May be `IgnoreKptIgnore` :)

internal/pkg/walker.go
```diff
@@ -0,0 +1,148 @@
...
+type FileMatchFunc func(string, os.FileInfo) bool
+
+// Walker provides functionality for walking the content of a
+// kpt package. It allows for customization of the files and directories
+// that will be provided to the caller through the callback function.
+type Walker struct {
+	// FileMatchFunc defines the function that will be used by the walker to
+	// filter which files should be provided to the callback. By default, all
+	// files will be provided.
+	FileMatchFunc FileMatchFunc
```
droot, line 81: Recently @Shell32-Natsu introduced a function to skip files in kyaml Readers, have to see how we can use that to implement kptignore aware resource reader.

droot (changes requested):
Looks good. Have some minor comments.

internal/pkg/walker.go
```diff
@@ -0,0 +1,148 @@
...
+// AllFileMatchFunc is a matcher func that matches all files.
+var AllFileMatchFunc FileMatchFunc = func(string, os.FileInfo) bool {
+	return true
+}
+
+// YamlFileMatchFunc is a matcher func that matches all files with .yaml or
+// .yml extensions.
+var YamlFileMatchFunc FileMatchFunc = func(_ string, info os.FileInfo) bool {
+	if info.IsDir() {
+		return true
```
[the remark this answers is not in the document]
mortent: Added a comment here and clarified in the function comment.

internal/pkg/walker.go
```diff
@@ -0,0 +1,148 @@
...
+	ext := filepath.Ext(info.Name())
+	if ext == ".yaml" || ext == ".yml" {
+		return true
+	}
+	return false
+}
+
+// KptfileYamlFileMatchFunc is a matcher func that matches the Kptfile and files
+// with either .yaml or .yml extensions.
+var KptfileYamlFileMatchFunc FileMatchFunc = func(s string, info os.FileInfo) bool {
```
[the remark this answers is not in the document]
mortent: Renamed this to `KptfileYamlMatcher`

internal/pkg/walker.go
```diff
@@ -0,0 +1,148 @@
...
+	"strings"
+
+	"github.com/GoogleContainerTools/kpt/internal/errors"
+	kptfilev1alpha2 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1alpha2"
+	"github.com/monochromegane/go-gitignore"
+	"sigs.k8s.io/kustomize/kyaml/copyutil"
+)
+
+// AllFileMatchFunc is a matcher func that matches all files.
+var AllFileMatchFunc FileMatchFunc = func(string, os.FileInfo) bool {
```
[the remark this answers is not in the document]
mortent: Changed the Walker to use an interface instead of a function type and then just provide a function implementation of the interface. This also leads to better names here, so this implementation of the `FileMatcher` interface is `AllMatcher`.

internal/pkg/walker.go
mortent: Simplified this somewhat. But don't want to be too clever here.

internal/pkg/walker.go
mortent: Clarified this in the comment.

internal/pkg/walker.go
```diff
@@ -0,0 +1,148 @@
...
+			if err != nil && !errors.Is(err, os.ErrNotExist) {
+				return err
+			}
+			if err == nil {
+				return filepath.SkipDir
+			}
+		}
+
+		// If the file doesn't meet the matcher criteria, we can just return.
+		if !p.FileMatchFunc(path, info) {
```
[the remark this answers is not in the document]
mortent: Fixed.

internal/pkg/walker.go
mortent: Renamed this to `IgnoreKptfileIgnorePatterns`. It is a bit long, but I think works better than something like `IgnoreKptIgnore`.

internal/pkg/walker.go
mortent: This implementation combines the glob pattern and the skip file function in the upstream kyaml pkg reader into the `FileMatcher` interface. It is harder to include the Ignore pattern here since it requires checking the Kptfile content.
The idea is to have a `KptPackageReader` implementation of the `kio.Reader` interface that we can use in kpt instead of the `kio.LocalPackageReader`.

droot (approved):

PR #2151, opened 2021-06-03
Update cli-utils and klog dependencies

mortent:
This PR pulls in the latest version of cli-utils, which has been updated to use the v1.21 version of the kubernetes libraries. This also means we no longer get old versions of kustomize pulled in through the k8s libraries. This allows us to remove the rewrite rules in the `go.mod` file that breaks `go install`.

The k8s libraries are meant to be used with go 1.16, so this updates the project to use go 1.16 consistently across all CI jobs.

It also updates the project to use klog v2.


droot (approved):

seans3 (approved):
/lgtm

PR #2154, opened 2021-06-03
Install RG CRD with kpt live apply by default unless dry-run

mortent:
This PR changes the default value of the `--install-resource-group` flag for `kpt live apply` to `true`, but changes it to `false` for dry-run unless the user has explicitly used the flag.

After this change, the behavior is:
 * kpt does _not_ install the resource-group CRD by default during dry-run.
 * Users can override this by adding the `install-resource-group` flag. So `kpt live apply --dry-run --install-resource-group` will install the resource-group CRD.
 * `kpt live apply` (without dry-run) will install the resource-group CRD by default.

Fixes: https://github.com/GoogleContainerTools/kpt/issues/2092


seans3:
Looks like we need a small change in `end-to-end-test.sh`.

seans3 (approved):
/lgtm

PR #2155, opened 2021-06-03
check for docker before running `fn eval`

natasha41575:
fixes https://github.com/GoogleContainerTools/kpt/issues/2084

Adds a check for docker before running `fn eval`.

For example:
```
$ kpt fn eval simple-function --image gcr.io/kpt-fn/set-namespace:v0.1.3 -- namespace=staging

error: docker must be running to use this command
To install docker, follow the instructions at https://docs.docker.com/get-docker/.
```

```
$ kpt fn render basic-pipeline

Package "basic-pipeline":
error: docker must be running to use this command
To install docker, follow the instructions at https://docs.docker.com/get-docker/.
```

### DIFF
--- a/internal/util/cmdutil/cmdutil.go
+++ b/internal/util/cmdutil/cmdutil.go
@@ -22,7 +22,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/GoogleContainerTools/kpt/internal/errors"
 	"github.com/GoogleContainerTools/kpt/internal/fnruntime"
 	"github.com/spf13/cobra"
 )
@@ -80,10 +79,8 @@ func ResolveAbsAndRelPaths(path string) (string, string, error) {
 // DockerCmdAvailable runs `docker ps` to check that the docker command is
 // available, and returns an error with installation instructions if it is not
 func DockerCmdAvailable() error {
-	const op errors.Op = "docker.check"
-
-	suggestedText := `Docker is required to run this command.
-To install docker, follow the instructions at https://docs.docker.com/get-docker/
+	suggestedText := `docker must be running to use this command
+To install docker, follow the instructions at https://docs.docker.com/get-docker/.
 `
 	buffer := &bytes.Buffer{}
 
@@ -91,7 +88,7 @@ To install docker, follow the instructions at https://docs.docker.com/get-docker
 	cmd.Stderr = buffer
 	err := cmd.Run()
 	if err != nil {
-		return errors.E(op, fmt.Errorf("%s", suggestedText))
+		return fmt.Errorf("%s", suggestedText)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


Shell32-Natsu:
Please add the example output in PR description.

natasha41575:
> Please add the example output in PR description.

Done

internal/fnruntime/container.go
```diff
@@ -80,6 +81,10 @@ type ContainerFn struct {
 // It reads the input from the given reader and writes the output
 // to the provided writer.
 func (f *ContainerFn) Run(reader io.Reader, writer io.Writer) error {
+	// check for docker availability
+	if err := dockerCmdAvailable(); err != nil {
```
droot, line 85: I understand the motivation to move it to a common layer here but the approach has downsides:

- the check is performed on every function run now, so if pipeline has 10 functions, `docker version` will be invoked 10 times so it adds to the latency
- Since the check is done at the lowest layer instead of the first thing when `fn eval` or `fn render` is called, so it doesn't fail fast.

So having a common utility function to check availability of docker and invoking it in the `fn eval` and `fn render` command runners as pre-run step seems to avoid both the above problem with a little downside of repeating the call in both the places.

internal/fnruntime/container.go
```diff
@@ -110,6 +115,25 @@ func (f *ContainerFn) Run(reader io.Reader, writer io.Writer) error {
 	return nil
 }
 
+// dockerCmdAvailable runs `docker version` to check that the docker command is
+// available, and returns an error with installation instructions if it is not.
+func dockerCmdAvailable() error {
+	const op errors.Op = "docker.check"
```
droot, line 121: let's remove this. This approach didn't turn out to be useful from UX perspective.

internal/fnruntime/container.go
```diff
@@ -110,6 +115,25 @@ func (f *ContainerFn) Run(reader io.Reader, writer io.Writer) error {
...
+	suggestedText := `Docker must be running to use this command.
+To install docker, follow the instructions at https://docs.docker.com/get-docker/
+`
+	buffer := &bytes.Buffer{}
+
+	cmd := exec.Command(dockerBin, "version")
+	cmd.Stderr = buffer
+	err := cmd.Run()
+	if err != nil {
+		return errors.E(op, fmt.Errorf("%s", suggestedText))
```
droot, line 132: don't use `errors.E` here, instead just return the `fmt.Errorf(...)`

droot (changes requested):

internal/fnruntime/container.go
natasha41575: got it, thanks. Moved it out.

internal/fnruntime/container.go
natasha41575: done

internal/fnruntime/container.go
natasha41575: done

droot (approved):

PR #2161, opened 2021-06-04
Display helpful error when encountering old Kptfile format

mortent:
Adds verification that the Kptfile is using the `v1alpha2` format. If a file with the legacy `v1alpha1` version is found, an error message is shown to the user.

If the legacy version if found in a file that is already in the local fork, we display an error that includes the path to the package on the local filesystem. If a Kptfile with the legacy version is found when fetching/updating a package from an upstream repo, we display an error message that includes the git reference (instead of the path to some temporary directory).

To make sure we only have to do this check in one place, this PR removes the `ReadFile` function from the `kptfileutil` package and updates all code to read Kptfiles using the function in the `pkg` package. It also makes sure we pass around the Kptfile struct as a pointer rather than a value. Both are valid, but this makes sure we are consistent.

Error message when running `kpt fn render` on an old package: 
```
$ kpt fn render helloworld-set 
Error: Kptfile at "/Users/mortent/Development/mortent/temp/kpttest/helloworld-set" has an old version ("v1alpha1") of the Kptfile schema. Please update the package to the latest format. See https://kpt.dev/installation/migration for more details
```

Error message when using `kpt pkg get` to fetch an old package:
```
$ kpt pkg get https://github.com/GoogleContainerTools/kpt.git/package-examples/helloworld-set

Package "helloworld-set":
Fetching https://github.com/GoogleContainerTools/kpt@master.
From https://github.com/GoogleContainerTools/kpt
 * branch              master     -> FETCH_HEAD
Error: Kptfile at "https:/github.com/GoogleContainerTools/kpt/package-examples/helloworld-set@master" has an old version ("v1alpha1") of the Kptfile schema. Please update the package to the latest format. See https://kpt.dev/installation/migration for more details. 
```

Error message when running `kpt fn render` on a package where the Kptfile has an unknown GVK:
```
Error: Kptfile at "/Users/mortent/Development/mortent/temp/kpttest/helloworld-set" has an unknown resource type ("kpt.com/v1beta1, Kind=Kptfile").
```

### DIFF
--- a/.github/workflows/e2eEnvironment.yml
+++ b/.github/workflows/e2eEnvironment.yml
@@ -28,6 +28,10 @@ jobs:
       matrix:
         version: [ 1.18, 1.19 ]
     steps:
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
       - uses: actions/checkout@master
       # Pinned to Commit to ensure action is consistent: https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions
       # If you upgrade this version confirm the changes match your expectations
--- a/e2e/live/end-to-end-test.sh
+++ b/e2e/live/end-to-end-test.sh
@@ -144,6 +144,8 @@ function downloadPreviousKpt {
 function buildKpt {
     set -e
     if [ -z $BUILD_DEPS_AT_HEAD ]; then
+	echo "checking go version"
+	go version
 	echo "Building kpt locally..."
 	go build -o $BIN_DIR -v . > $OUTPUT_DIR/kptbuild 2>&1
 	echo "Building kpt locally...SUCCESS"
--- a/internal/cmdliveinit/cmdliveinit.go
+++ b/internal/cmdliveinit/cmdliveinit.go
@@ -192,7 +192,7 @@ func updateKptfile(p *pkg.Pkg, inv *kptfilev1alpha2.Inventory, force bool) error
 	}
 	// Finally, set the inventory parameters in the Kptfile and write it.
 	kf.Inventory = inv
-	if err := kptfileutil.WriteFile(p.UniquePath.String(), *kf); err != nil {
+	if err := kptfileutil.WriteFile(p.UniquePath.String(), kf); err != nil {
 		return errors.E(op, p.UniquePath, err)
 	}
 	return nil
--- a/internal/cmdliveinit/cmdliveinit_test.go
+++ b/internal/cmdliveinit/cmdliveinit_test.go
@@ -10,10 +10,10 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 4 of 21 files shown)


frankfarzan:
Just commenting on the PR description for now:

> Adds verification that the Kptfile is using the `v1alpha2` format. If a file with the legacy `v1alpha1` version is found, an error message is shown to the user.

To be future proof, how do you want to handle `v1apha2` and `v1`? The major change will be ignore stuff, but there may others. We don't need to decide now, but assume there may be `v1alpha2` packages in the wild when we release `v1`.

> 
> If the legacy version if found in a file that is already in the local fork, we display an error that includes the path to the package on the local filesystem. If a Kptfile with the legacy version is found when fetching/updating a package from an upstream repo, we display an error message that includes the git reference (instead of the path to some temporary directory).
> 
> To make sure we only have to do this check in one place, this PR removes the `ReadFile` function from the `kptfileutil` package and updates all code to read Kptfiles using the function in the `pkg` package. It also makes sure we pass around the Kptfile struct as a pointer rather than a value. Both are valid, but this makes sure we are consistent.
> 
> Error message when running `kpt fn render` on an old package:
> 
> ```
> $ kpt fn render helloworld-set 
> Error: Kptfile at "/Users/mortent/Development/mortent/temp/kpttest/helloworld-set" is using and old version of the Kptfile format used by pre-1.0 versions of kpt. Please update the package to use the new format. See https://kpt.dev for more details on how to do the migration.

- Save some searching by linking to https://kpt.dev/installation/migration
- "format" -> "schema"

> ```
> 
> Error message when using `kpt pkg get` to fetch an old package:
> 
> ```
> $ kpt pkg get https://github.com/GoogleContainerTools/kpt.git/package-examples/helloworld-set
> 
> Package "helloworld-set":
> Fetching https://github.com/GoogleContainerTools/kpt@master.
> From https://github.com/GoogleContainerTools/kpt
>  * branch              master     -> FETCH_HEAD
> Error: Kptfile at "https:/github.com/GoogleContainerTools/kpt/package-examples/helloworld-set@master" is using and old version of the Kptfile format used by pre-1.0 versions of kpt. Please update the package to use the new format. See https://kpt.dev for more details on how to do the migration. 
> ```

mortent:
> To be future proof, how do you want to handle `v1alpha2` and `v1`? The major change will be ignore stuff, but there may others. We don't need to decide now, but assume there may be `v1alpha2` packages in the wild when we release `v1`.

Updated the PR to handle this better. We now have a list of deprecated versions that will result in an error message asking users to upgrade the package. It also introduces a custom error message used whenever the GVK in the Kptfile doesn't match what we expect or is one of the known old formats.


> 
> * Save some searching by linking to https://kpt.dev/installation/migration
> * "format" -> "schema"

Fixed

frankfarzan (approved):

PR #2163, opened 2021-06-04
out of place support in test harness

droot:
This PR demonstrates how test harness can be modified to test out-of-place workflow.

@phanimarupaka the change is very minimal and can be added in your out of place support PR because that is required to test this support, hence, keeping this PR in draft mode.

/cc @phanimarupaka @Shell32-Natsu

### DIFF
--- a/pkg/test/runner/runner.go
+++ b/pkg/test/runner/runner.go
@@ -52,6 +52,7 @@ const (
 	expectedResultsFile string = "results.yaml"
 	expectedDiffFile    string = "diff.patch"
 	expectedConfigFile  string = "config.yaml"
+	outDir              string = "out"
 	setupScript         string = "setup.sh"
 	teardownScript      string = "teardown.sh"
 	execScript          string = "exec.sh"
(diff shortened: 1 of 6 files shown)


pkg/test/runner/runner.go
```diff
@@ -275,6 +282,10 @@ func (r *Runner) runFnRender() error {
 		}
 	}
 
+	if r.IsOutOfPlace() {
+		destDir = filepath.Join(pkgPath, outDir)
```
Shell32-Natsu, line 286: The output dir is a subdirectory in the package? I am hesitating is this a good idea.

droot: So the only downside to this approach (or pattern) is that your rendered config lives side by side with original source config and that can be painful for `git diff` (review and approval workflows). But that exactly is the use-case `git diff` for the test harness in this case.

I realized we can extend this to support `function results` as well and get away from the special handling of them.

droot:
@frankfarzan @Shell32-Natsu @phanimarupaka I don't plan to merge this PR because it requires `-o` flag which is in the `out-of-place` PR, but, if we are in agreement on the approach, then @phanimarupaka can pick these changes in the `out-of-place` PR.

Also, same approach can be leveraged by the `e2e` tests using the shell scripts (function chaining especially) by using another special directory and adding that to `krmignore` file.

PR #2181, opened 2021-06-07
upgrade dependencies in v0

natasha41575:
This PR is to see how easy it is to upgrade kyaml in v0. We will want to do this after the next release of kyaml.


droot:
@natasha41575 Do we know why the CI tests (`kpt live`) are failing. It seems to be hitting timeout.

natasha41575:
@droot I was trying to investigate but running into some trouble running it locally. I'll let you know if I can figure it out.

natasha41575:
@droot I was trying to investigate but running into some trouble running it locally. I'll let you know if I can figure it out.

go.mod
```diff
@@ -4,23 +4,27 @@ go 1.14
...
-	k8s.io/kubectl v0.20.4
-	sigs.k8s.io/cli-utils v0.25.0
-	sigs.k8s.io/kustomize/cmd/config v0.9.10
-	sigs.k8s.io/kustomize/kyaml v0.10.17
+	k8s.io/kube-openapi v0.0.0-20210421082810-95288971da7e
+	k8s.io/kubectl v0.21.1
+	sigs.k8s.io/cli-utils v0.25.1-0.20210603052138-670dee18a123
+	sigs.k8s.io/kustomize/api v0.8.10 // indirect
+	sigs.k8s.io/kustomize/cmd/config v0.9.12
+	sigs.k8s.io/kustomize/kyaml v0.10.20
```
droot, line 27: We should do this latest kyaml that includes the fix we are interested in.

internal/util/update/resource-merge.go
```diff
@@ -92,8 +92,6 @@ func (u ResourceMergeUpdater) Update(options UpdateOptions) error {
 		OriginalPath: original.AbsPath(),
 		UpdatedPath:  updated.AbsPath(),
 		DestPath:     options.PackagePath,
-		// TODO: Write a test to ensure this is set
-		MergeOnPath: true,
```
droot, line 96: What are the implications for this ? @mortent

droot (commented):
@mortent not sure how confident we can be about this update. Can you also pl. take a look at it.

go.mod
droot: So `v0.10.21` is available. Look at this PR https://github.com/GoogleContainerTools/kpt/pull/2169/files

droot (changes requested):

go.mod
```diff
@@ -4,23 +4,27 @@ go 1.14
...
+	k8s.io/cli-runtime v0.21.1
+	k8s.io/client-go v0.21.1
 	k8s.io/klog v1.0.0
-	k8s.io/kubectl v0.20.4
-	sigs.k8s.io/cli-utils v0.25.0
-	sigs.k8s.io/kustomize/cmd/config v0.9.10
-	sigs.k8s.io/kustomize/kyaml v0.10.17
+	k8s.io/kube-openapi v0.0.0-20210421082810-95288971da7e
+	k8s.io/kubectl v0.21.1
+	sigs.k8s.io/cli-utils v0.25.1-0.20210603052138-670dee18a123
```
mortent, line 24: We should use released versions of the libraries where possible.

go.mod
```diff
@@ -4,23 +4,27 @@ go 1.14
...
+	sigs.k8s.io/kustomize/api v0.8.10 // indirect
+	sigs.k8s.io/kustomize/cmd/config v0.9.12
+	sigs.k8s.io/kustomize/kyaml v0.10.20
 )
+
+replace gopkg.in/yaml.v3 => gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c
```
mortent, line 30: Do we have to do this? We should try to avoid adding replace directives ref https://github.com/GoogleContainerTools/kpt/issues/1535

go.mod
```diff
@@ -4,23 +4,27 @@ go 1.14
 
 require (
 	github.com/cpuguy83/go-md2man/v2 v2.0.0
-	github.com/go-errors/errors v1.0.1
+	github.com/go-errors/errors v1.4.0
```
mortent, line 7: This updates a lot of dependencies. Do we need to update all of this for kyaml? In particular, I'm not sure we need to update the cli-utils library here.

internal/util/update/resource-merge.go
mortent: This is safe. The interface for Merge3 has changed a bit, but `MergeOnPath` defaults to `true`.

go.mod
natasha41575: For cli-utils, we need this or we get the following error:
```
../../../../pkg/mod/sigs.k8s.io/kustomize@v2.0.3+incompatible/pkg/transformers/config/factorycrd.go:71:47: cannot use api.Schema.SchemaProps.Properties (type map[string]"k8s.io/kube-openapi/pkg/validation/spec".Schema) as type myProperties in argument to looksLikeAk8sType
```
I believe it's the same reason we have this on the next branch.

go.mod
natasha41575: same as above comment, we need to update cli-utils or we get a build error

go.mod
natasha41575: I also updated the PR so that it updates as few dependencies as possible while still building

go.mod
natasha41575: updated to the latest kyaml

yhrn:
I opened another PR (#2205) that I think has fewer problems in `go.mod`. I had most of this stuff lying around. But like I wrote in there the tests don't go through even though it builds just fine. Feel free to take anything from that PR that might be useful.

go.mod
natasha41575: Edit: Never mind, seems like the build error only occurred for kyaml v0.10.20, we don't need to update cli-utils for 10.21

natasha41575: Edit: After updating it to kyaml v0.10.21 I get a different error if I don't use this version of cli-utils: 

```
No license could be found for sigs.k8s.io/kustomize - aborting.
```

PR #2182, opened 2021-06-07
Print error message to users when encountering a Kptfile with an unknown version of the Kptfile resource

mortent:
This verifies that the Kptfile resource within a Kptfile actually uses the `v1alpha1` version. If a different version is found, it will return an error asking users to upgrade kpt. 

Examples:

Fetching a package with a v1alpha2 version of the Kptfile:
```
$ kpt pkg get https://github.com/GoogleContainerTools/kpt.git/package-examples/wordpress@next .
fetching package "/package-examples/wordpress" from "https://github.com/GoogleContainerTools/kpt" to "wordpress"
error: failed to clone git repo: package at "https:/github.com/GoogleContainerTools/kpt/package-examples/wordpress@next" is using a newer version ("v1alpha2") of the Kptfile schema. Please try updating kpt.
```

Running `kpt fn run` on a package with v1alpha2 version of the Kptfile:
```
$ kpt fn run wordpress
error: package at "<path>/wordpress" is using a newer version ("v1alpha2") of the Kptfile schema. Please try updating kpt.
```

### DIFF
--- a/pkg/kptfile/kptfileutil/util.go
+++ b/pkg/kptfile/kptfileutil/util.go
@@ -44,7 +44,7 @@ func (e *UnknownKptfileVersionError) Error() string {
 	} else {
 		source = e.PkgPath
 	}
-	return fmt.Sprintf("package at %q is using an unknown version of the Kptfile schema. It might be for a newer version of kpt. Please try updating kpt.", source)
+	return fmt.Sprintf("package at %q is using a newer version (%q) of the Kptfile schema. Please try updating kpt.", source, e.Version)
 }
 
 // ReadFile reads the KptFile in the given directory
--- a/pkg/kptfile/kptfileutil/util.go
+++ b/pkg/kptfile/kptfileutil/util.go
@@ -15,16 +15,37 @@
 package kptfileutil
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/GoogleContainerTools/kpt/internal/util/git"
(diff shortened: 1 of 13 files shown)


pkg/kptfile/kptfileutil/util.go
```diff
@@ -15,17 +15,38 @@
...
+}
+
+func (e *UnknownKptfileVersionError) Error() string {
+	var source string
+	if e.RepoSpec != nil {
+		source = e.RepoSpec.RepoRef()
+	} else {
+		source = e.PkgPath
+	}
+	return fmt.Sprintf("package at %q is using an unknown version of the Kptfile schema. It might be for a newer version of kpt. Please try updating kpt.", source)
```
droot, line 47: 1. We should include the version (when available) in the message. Actually semver pkgs helps in comparing versions, so saying something like `pkg at "x" is using newer version "v1.2.2" of Kptfile. Please try updating kpt.` will read much better. 
2. I think we can remove the `schema` word.

I realized we will need something similar for kpt CLI 1.0 encountering older versions of `Kptfile` that are incompatible. There we need to point them to migration guide.

mortent: Updated the PR to print the version as part of the message. Also update the wording of the error message, but I think we should keep the `schema` word.

PR for handling kpt 1.0 encountering older versions of the Kptfile schema is here: https://github.com/GoogleContainerTools/kpt/pull/2161

frankfarzan (approved):

pkg/kptfile/kptfileutil/util.go
droot: Ack.

droot (approved):

PR #2183, opened 2021-06-07
Add copy to clipboard button to shell blocks.

etefera:
Fixes https://github.com/GoogleContainerTools/kpt/issues/1867
Adds a copy to clipboard button when hovering over shell blocks.
image
Demo at https://kpt-dev-ee2a9--copy-button-iuaxzgwx.web.app


mikebz:
Looking at what the precedent is - it could be better to have a button that just has a "copy"
image

etefera:
> Looking at what the precedent is - it could be better to have a button that just has a "copy"
> image

Updated as below 
image

natasha41575:
@etefera can the clipboard button be on the right, where you originally put the text?

etefera:
> @etefera can the clipboard button be on the right, where you originally put the `copy to clipboard` text?

Should we keep the `shell` text? If so, do we keep the icon on the left or right of it?

natasha41575:
@etefera how does it look if you put it below shell?

frankfarzan:
Thanks for the quick turn around. Comments:

1. Agree that it should be on the right, example from GCP docs:
https://kpt-dev-ee2a9--copy-button-iuaxzgwx.web.app/book/03-packages/06-creating-a-package
2. Match the opacity of other icons.
3. Doesn't handle blocks with multiple commands e.g. https://kpt-dev-ee2a9--copy-button-iuaxzgwx.web.app/book/03-packages/06-creating-a-package
4. Doesn't handle multi-line commands e.g. last block on https://kpt-dev-ee2a9--copy-button-iuaxzgwx.web.app/book/04-using-functions/02-imperative-function-execution

etefera:
> @etefera how does it look if you put it below shell?

Updated to look like this 
image

natasha41575:
> Updated to look like this

Never mind, I think it will look better either to the left or right of `shell`.

etefera:
> Thanks for the quick turn around. Comments:
> 
> 1. Agree that it should be on the right, example from GCP docs:
>    https://kpt-dev-ee2a9--copy-button-iuaxzgwx.web.app/book/03-packages/06-creating-a-package
> 2. Match the opacity of other icons.
> 3. Doesn't handle blocks with multiple commands e.g. https://kpt-dev-ee2a9--copy-button-iuaxzgwx.web.app/book/03-packages/06-creating-a-package Each line that has a `$` should have a copy button (alternatively, need to limited one command per block somehow).
> 4. Doesn't handle multi-line commands e.g. last block on https://kpt-dev-ee2a9--copy-button-iuaxzgwx.web.app/book/04-using-functions/02-imperative-function-execution

Updated with below
image

1. Done, moved next to language indicator.
2. Done.
3. I couldn't repro this across multiple browsers. Seemed like it was copying every line with a `$` leading the block, which is expected behavior to me. If we prefer having a copy button for every line, I recommend adding "One command per code block" to the style guide but it wouldn't make sense to disallow copying multiple lines in a block here.
4. Done.


As for the unrelated changes, I believe adding those two points to the style guide is sufficient for them to be tracked when we add linting across the site.

frankfarzan:
https://github.com/GoogleContainerTools/kpt/pull/2192/files

frankfarzan:
You can rebase now.

etefera:
> You can rebase now.

Rebased

site/static/js/plugins.js
```diff
@@ -152,6 +152,38 @@ function processAsciinemaTags(content) {
...
+    copyButton.appendChild(copyIcon);
+
+    copyButton.addEventListener("click", (el) =>
+      navigator.clipboard.writeText([
+        el
+          .composedPath()
+          .find((el) => el.classList.contains(buttonClassName))
+          .previousElementSibling.textContent.split("\n")
+          .map((s) => s.trimLeft())
+          .filter((s) => s.startsWith("$") || s.startsWith("|"))
```
frankfarzan, line 177: - Multi-line should be based on line continuation character not pipes (make sure to test this).
- If a shell block doesn't have '$', don't display the icon. This is defense in-depth even if we have a linter (which we don't today).
- This also needs to be done for catalog site after auditing it for correct usage of shell blocks
- We should enable docs linter ASAP

etefera: - Automated test for this.
- Automated test for this.
- We should be able to refer to plugins in production from the catalog site. Happy to test it after this PR is merged.
- I agree; it looks like an m4 task https://github.com/GoogleContainerTools/kpt/issues/1652 Since it'll take some time, it should be bumped up to m3 if we want to prioritize

frankfarzan (approved):
Thanks. This is a major improvement to docs usability.

frankfarzan:
Minor hiccup is HEREDOC:

https://kpt-dev-ee2a9--copy-button-iuaxzgwx.web.app/book/04-using-functions/02-imperative-function-execution?id=fn-config-flag

mikebz:
I would just shrink the button by 35% it's increasing the height of the text box for no good reason and the 'shell' text look small next to it.  Beyond that looks much better.

etefera:
> Minor hiccup is HEREDOC:
> 
> https://kpt-dev-ee2a9--copy-button-iuaxzgwx.web.app/book/04-using-functions/02-imperative-function-execution?id=fn-config-flag

I'm thinking we should probably avoid using HEREDOC and instead specify the file in its own `yaml` block in this case. My vote is for banning it in the style guide in favor of explicit blocks.

etefera:
> I would just shrink the button by 35% it's increasing the height of the text box for no good reason and the 'shell' text look small next to it. Beyond that looks much better.

Shrunk the icon

PR #2185, opened 2021-06-07
shorten timeout for some docker commands

### LINKED ISSUE #2005: kpt fn should time out instead of hanging when Docker is not available
When `docker ps` hangs, `kpt fn render` also hangs. Kpt fn should never hang, and should time out and throw an error.

The same should be true generally, e.g. if git is unresponsive, etc.

cc @droot 

Somewhat related: kpt should show docker progress https://github.com/GoogleContainerTools/kpt/issues/1948

Shell32-Natsu:
1. docker image inspect: 5s
2. docker image pull: 5min
3. docker run: 5min
4. docker version: 5s

close #2005

### DIFF
--- a/internal/fnruntime/container.go
+++ b/internal/fnruntime/container.go
@@ -33,10 +33,11 @@ import (
 type containerNetworkName string
 
 const (
-	networkNameNone containerNetworkName = "none"
-	networkNameHost containerNetworkName = "host"
-	defaultTimeout  time.Duration        = 5 * time.Minute
-	dockerBin       string               = "docker"
+	networkNameNone     containerNetworkName = "none"
+	networkNameHost     containerNetworkName = "host"
+	defaultLongTimeout  time.Duration        = 5 * time.Minute
+	defaultShortTimeout time.Duration        = 5 * time.Second
+	dockerBin           string               = "docker"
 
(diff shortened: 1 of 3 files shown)


natasha41575 (approved):
this LGTM

droot (commented):
@Shell32-Natsu I have looked at it. I am contemplating if we should abstract out all the docker command execution in a separate pkg first. I am also noticing that `ctx` is not wired all the way from top, wondering if we should do all that first and then adjust the timeouts.

Shell32-Natsu:
> @Shell32-Natsu I have looked at it. I am contemplating if we should abstract out all the docker command execution in a separate pkg first. I am also noticing that `ctx` is not wired all the way from top, wondering if we should do all that first and then adjust the timeouts.

I am not sure how should `ctx` be passed all the way from top. For every docker command we need to invoke `cancel`. Can these commands share a same `ctx`?

Shell32-Natsu:
Have an offline discussion with @droot, it's important for us to add a short timeout for internal check like `docker version` and `docker image inspect`. But for long operation like `image pull` and `docker run` we should carefully consider the default timeout and how to expose this time to user. So in this PR I will only add time out for `version` and `image inspect` commands.

droot (approved):

PR #2189, opened 2021-06-08
Improve the error message when apply can't find type

mortent:
This improves the error message when a user tries to apply a resource and the type can't be found in the cluster or as a CRD among the applied resources.

Example output:
```
$ kpt live apply
Error: 1 resource types could not be found in the cluster or as CRDs among the applied resources.

Resource types:
TraitDefinition.core.oam.dev 
```


seans3 (approved):
/lgtm

PR #2194, opened 2021-06-08
sink default to current directory

### LINKED ISSUE #2159: `fn sink` should write to current directory by default
Currently the command `fn sink` requires an argument `DIR` to point to the sink destination. To keep consistent with other commands, the argument should be optional and sink should write to PWD by default.

Shell32-Natsu:
close #2159 

`fn sink` will use `.` by default.

### DIFF
--- a/thirdparty/cmdconfig/commands/cmdsink/cmdsink.go
+++ b/thirdparty/cmdconfig/commands/cmdsink/cmdsink.go
@@ -19,7 +19,6 @@ func GetSinkRunner(name string) *SinkRunner {
 		Long:    fndocs.SinkShort + "\n" + fndocs.SinkLong,
 		Example: fndocs.SinkExamples,
 		RunE:    r.runE,
-		Args:    cobra.MinimumNArgs(1),
 	}
 	r.Command = c
 	return r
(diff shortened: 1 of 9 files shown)


thirdparty/cmdconfig/commands/cmdsink/cmdsink.go
```diff
@@ -35,7 +34,11 @@ type SinkRunner struct {
 }
 
 func (r *SinkRunner) runE(c *cobra.Command, args []string) error {
-	outputs := []kio.Writer{&kio.LocalPackageWriter{PackagePath: args[0]}}
+	dir := "."
```
frankfarzan, line 37: do you assume POSIX?

Shell32-Natsu: I believe this should also work on Windows. I notice that in other commands [some](https://github.com/GoogleContainerTools/kpt/blob/next/internal/cmdinit/cmdinit.go#L72) are using `pkg.CurDir` which is basically a `.` and [some](https://github.com/GoogleContainerTools/kpt/blob/next/internal/cmdliveinit/cmdliveinit.go#L84) are using `os.Getwd`.

frankfarzan: Should be consistent. Let's standardize on pkg.CurDir across the board.

Shell32-Natsu: done

droot (approved):

Issue #2039, opened 2021-05-18
`fn render/eval/source/sink` should generate proper error message when there are non-KRM resources


Shell32-Natsu:
Example error:
```
Input resource list must not contain non-KRM resource:
...
{resource yaml string}
...
```

PR #2198, opened 2021-06-08
proper error message for fn eval/render when there are non-KRM resources

natasha41575:
partially addresses #2039, improves error message for `fn eval` and `fn render` when there are non-KRM resources

Before:
```
[FAIL] "gcr.io/kpt-fn/set-namespace:v0.1.3"
  Results:
    [ERROR] failed to run function: error unmarshaling JSON: while decoding JSON: Object 'Kind' is missing in '{"Non-krm-resource":true,"metadata":{"annotations":{"config.k8s.io/id":"1","config.kubernetes.io/index":"0","config.kubernetes.io/path":"nonkrm.yaml"}}}'
  Stderr:
    "[error] /// : failed to run function: error unmarshaling JSON: while decoding JSON: Object 'Kind' is missing in '{\"Non-krm-resource\":true,\"metadata\":{\"annotations\":{\"config.k8s.io/id\":\"1\",\"config.kubernetes.io/index\":\"0\",\"config.kubernetes.io/path\":\"nonkrm.yaml\"}}}'"
  Exit code: 1
```

After:
```
Error: input resource list must contain only KRM resources:
non-krm.yaml: resource must have `apiVersion`
```

Will do the same for `fn source` and `fn sink` in a separate PR.


pkg/api/kptfile/v1alpha2/validation.go
```diff
@@ -183,6 +183,20 @@ func IsKRM(n *yaml.RNode) error {
 	return nil
 }
 
+//nolint
+func AreKRM(nodes []*yaml.RNode) error {
+	for i := range nodes {
+		if err := IsKRM(nodes[i]); err != nil {
+			nodes[i].PipeE(yaml.ClearAnnotation("config.kubernetes.io/index"))
```
frankfarzan, line 190: why do we do this? Add a comment.

pkg/api/kptfile/v1alpha2/validation.go
```diff
@@ -183,6 +183,20 @@ func IsKRM(n *yaml.RNode) error {
...
+			nodes[i].PipeE(yaml.ClearAnnotation("config.kubernetes.io/path"))
+			yaml.ClearEmptyAnnotations(nodes[i])
+			str := nodes[i].MustString()
+			return fmt.Errorf("%s:\n...\n%s...", err.Error(), str)
```
frankfarzan, line 194: Instead of printing the entire resource like this (which is not truncated and could be large), provide the path to the file.

thirdparty/kyaml/runfn/runfn.go
```diff
@@ -146,6 +150,38 @@ func (r RunFns) getNodesAndFilters() (
...
+		// can read them upon execution
+		input := &bytes.Buffer{}
+		err = kio.ByteWriter{Writer: input}.Write(n)
+		if err != nil {
+			return err
+		}
+		result = append(result, []kio.Reader{&kio.ByteReader{Reader: input}}...)
+		nodes = append(nodes, n...)
+	}
+	nodes, err := (&filters.IsLocalConfig{
```
frankfarzan, line 172: add a comment explaining how this interacts with local config or meta resources (these are distinct concepts).

droot:
I am actually thinking of another approach. It's important to enforce this invariant on input/output resources of each function execution(doing it in fnruntime can cover both `eval` and `render`). Use the `path/index` to derive a user-friendly location (2nd resource in <some.yaml>) to report an error (avoid dumping the yaml content). One caveat here: for generated resources, the path will be auto generated one, but that's okay I think.

Doing above is not sufficient to ensure the package is valid because if a package doesn't have any pipeline, function runner will not be invoked, we can have an explicit call for such cases (I think , in function named `runPipeline`, we return if `pipeline.IsEmpty()`, we can plug call for KRM check in that case). [This is not an issue for `fn eval`]

So, tests should include case where non-KRM resources with no pipeline as well.

pkg/api/kptfile/v1alpha2/validation.go
```diff
@@ -183,6 +183,20 @@ func IsKRM(n *yaml.RNode) error {
 	return nil
 }
 
+//nolint
```
droot, line 186: curious what linter is complaining about here ?

natasha41575: that the error message ends in punctuation, and that I am ignoring the error of PipeE. IMO if PipeE throws an error we can safely ignore it.

natasha41575:
> I am actually thinking of another approach. It's important to enforce this invariant on input/output resources of each function execution(doing it in fnruntime can cover both eval and render). Use the path/index to derive a user-friendly location (2nd resource in <some.yaml>) to report an error (avoid dumping the yaml content). One caveat here: for generated resources, the path will be auto generated one, but that's okay I think.

Will do

pkg/api/kptfile/v1alpha2/validation.go

natasha41575:
@droot I moved the check to `fnruntime` and added a test case for no pipeline, thanks!

frankfarzan (approved):
please wait for @droot

droot (approved):
New approach is neat! Thank you.

PR #2205, opened 2021-06-09
Update kyaml and related deps

yhrn:
Alternative to #2181 with no replace directive. Everything builds fine and `make build tidy` does not modify anything, however the test do not go through and I won't have time to look at that in a while.


go.mod
```diff
@@ -5,22 +5,23 @@ go 1.14
...
 	github.com/go-errors/errors v1.0.1
-	github.com/go-openapi/spec v0.19.5
 	github.com/olekukonko/tablewriter v0.0.4
 	github.com/pkg/errors v0.9.1
 	github.com/posener/complete/v2 v2.0.1-alpha.12
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.6.1
-	gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c
+	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
```
natasha41575, line 14: using this version results in sequence nodes to be indented, e.g. 
```
    --- FAIL: TestSearchCommand/search_replace_by_array_objects_path (0.01s)
        cmdsearch_test.go:116: 
                Error Trace:    cmdsearch_test.go:116
                Error:          Not equal: 
                                expected: "apiVersion: apps/v1\nkind: Deployment\nmetadata:\n  name: nginx-deployment\nspec:\n  replicas: 3\n  foo:\n  - c: thing0\n  - c: thing-new\n  - c: thing2"
                                actual  : "apiVersion: apps/v1\nkind: Deployment\nmetadata:\n  name: nginx-deployment\nspec:\n  replicas: 3\n  foo:\n    - c: thing0\n    - c: thing-new\n    - c: thing2"
                                
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -7,4 +7,4 @@
                                   foo:
                                -  - c: thing0
                                -  - c: thing-new
                                -  - c: thing2
                                +    - c: thing0
                                +    - c: thing-new
                                +    - c: thing2
                Test:           TestSearchCommand/search_replace_by_array_objects_path
```

We need gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c, but running `go mod tidy` will change it, hence why I had a replace directive. 

@mortent @droot is it okay to upgrade gopkg.in/yaml.v3 in kpt alpha and have the indentations change?

go.mod
```diff
@@ -5,22 +5,23 @@ go 1.14
...
-	k8s.io/client-go v0.20.4
+	k8s.io/api v0.21.1
+	k8s.io/apimachinery v0.21.1
+	k8s.io/cli-runtime v0.21.1
+	k8s.io/client-go v0.21.1
 	k8s.io/klog v1.0.0
-	k8s.io/kubectl v0.20.4
+	k8s.io/kube-openapi v0.0.0-20210421082810-95288971da7e
+	k8s.io/kubectl v0.21.1
 	sigs.k8s.io/cli-utils v0.25.0
```
natasha41575, line 23: AFAICT the license check will fail unless we use `sigs.k8s.io/cli-utils v0.25.1-0.20210603052138-670dee18a123`

go.mod
yhrn: I see, I didn't understand from the discussion in the other PR that this was the reason you needed this specific version. Is there no way to configure the indentation behavior?

natasha41575:
Thank you for the PR! Will add a couple of things from here to mine.

go.mod
droot: Unfortunately the indentation behavior in the `yaml.v3` isn't configurable, so the update can cause inconvenience. So we are avoiding the update to keep the impact minimal.

@natasha41575 we should add a comment explaining the reasoning in the other PR.

natasha41575: > @natasha41575 we should add a comment explaining the reasoning in the other PR.
will do

mortent: It seems like being able to control the version of the yaml library in kpt separately from kyaml is something we are going to run into in the future too. Adding a replace directive in the kpt go.mod file causes problems for `go install`, so we should look into ways around this. I would guess that adding a layer of indirection here might work, so can use the replace directive in a modules imported by the kpt module. 
Would it work if we just set up a separate go module that just imports kyaml and adds the replace directive? We can then import that module and in this way allow kpt to update kyaml and keep the old version of the yaml library. It seems like something we could check pretty quickly.

go.mod
mortent: Yeah, the the master branch of cli-utils now uses the latest version of the kubectl libraries, so we no longer pull in the old version of kustomize.

go.mod
natasha41575: @mortent what would that mean for releasing? Would we have to release the new go module first and then release kpt for each release?

mortent: Yeah, the solution (if it works) I'm suggesting here would mean that we would have a new module (something like kyaml-kpt) that we would have to release every time we want to bump the kyaml version in kpt.

yhrn: I think it only has a chance of working if the newer version of `yaml.v3` isn't pulled in transitively by some other dependency but unfortunately I think it is pulled in by `github.com/googleapis/gnostic` that is pulled in by a bunch of k8s deps. Have you already looked into fixing this directly in `yaml.v3`, i.e. making the behavior configurable? To me that feels like the more maintainable solution.

natasha41575: We're going to do a little investigation and see what else we can do here. Thanks @yhrn, will keep you posted!

liamfallon:
This PR is being closed because the changes in this PR have been:

1. Superseded by https://github.com/kptdev/kpt/pull/4196 the dependency updates
2. The other changes are on files that have since been separately updated or are no longer in the codebase

Issue #2091, opened 2021-05-24
fix schema for `resourceRef` in kyaml


natasha41575:
The [ResourceRef schema in the kpt function SDKs](https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/6d56c218eb8d72b0a56366b7b3927105d4068ec3/ts/kpt-functions/src/types.ts#L383) is 

```
resourceRef:
    apiVersion: apps/v1
    kind: Deployment
    namespace: ''
    name: nginx-deployment
```

and [the ResourceRef schema in kyaml](https://github.com/kubernetes-sigs/kustomize/blob/0f614e92f72f1b938a9171b964d90b197ca8fb68/kyaml/yaml/types.go#L139) which is used by kpt [here](https://github.com/GoogleContainerTools/kpt/blob/b6ad66da7fb3afd457ae551f5d3eeac707e9eaf0/pkg/api/fnresult/v1alpha2/types.go#L39) is:

```
resourceRef:
    apiVersion: apps/v1
    kind: Deployment
    metadata: 
      namespace: ''
      name: nginx-deployment
```

Our current workaround (https://github.com/GoogleContainerTools/kpt/pull/2086) is to duplicate the `ResultItem` type from kyaml, but change the type used for `resourceRef`. Instead of duplicating this type, we should fix it upstream in kyaml.

We will also need to make sure both function orchestrators (kpt and kustomize) are still backwards compatible with the old schema. 

cc @droot

mikebz:
@natasha41575 is this 2 weeks indeed?

natasha41575:
@mikebz that was a very conservative overestimate because I'm not sure yet what how much will need to change in kustomize. Maybe 1 week is a better estimate. I'll change it.

natasha41575:
The PR is here: https://github.com/kubernetes-sigs/kustomize/pull/3962

natasha41575:
This is fixed in kyaml. After it is released and kpt's dependency is upgraded all that's left to do here is some cleanup.

mikebz:
how much work is left on this?  Can we close this out soon?

natasha41575:
> how much work is left on this? Can we close this out soon?

To close this, https://github.com/GoogleContainerTools/kpt/pull/2169 has to merge and then I will make a small PR that will finish the work for this. In terms of work there is very little left but it can't be done until #2169 is merged.

PR #2207, opened 2021-06-09
use kyaml resultItem

natasha41575:
Cleanup for #2091 

Now that `ResultItem` has been fixed in kyaml, we can remove the duplicate introduced in https://github.com/GoogleContainerTools/kpt/pull/2086.

At the time of writing, `parseNameAndNamespace` https://github.com/GoogleContainerTools/kpt/blob/c198ed4c27fc570429f822903cd520c22fa7bd57/internal/fnruntime/runner.go#L187 is still used to support the old resourceRef schema with a metadata field.


droot (approved):
Awesome!

PR #2208, opened 2021-06-09
book: chapter 6

frankfarzan:
Part of #1573


site/book/06-deploying-packages/01-initializing-a-package-for-apply.md
```diff
@@ -0,0 +1,36 @@
...
+uses heuristics to automatically choose the namespace. In this example, all the
+resources in `wordpress` package were in the `default` namespace, so it chose
+`default` for the namespace. Alternatively, you can manually configure the name
+and namespace of the `ResourceGroup` resource.
+
+?> Refer to the [init command reference][init-doc] for usage.
+
+!> Once a package is applied to the cluster, you do not want to change the
+metadata in the `inventory` section as it severs the association between the
+package and the `ResourceGroup` leading to destructive operations.
```
droot, line 34: I am thinking about a common CUJ:

I have a package A which I tested by `kpt live init` and then applying to my cluster. Everything looked good, so I uprevved (or tagged) the package A. So inventory section is in the Kptfile. 

Downstream consumer of package A (let's call it A') has already applied the package in their cluster, so they already have an inventory section in the Kptfile of A'. And, now they run `pkg update A'`

What is going to happen here ? Will `pkg update A'` going to ignore the `inventory` section from the package A ?
Ignoring the `inventory` section from upstream seems reasonable ? right ?

/cc @mortent 

Or may be, we need to add a section in the book about `publishing a package` where we can talk about the best practices about publishing a package that can eventually be codified in the `kpt pkg publish` command.

droot (commented):
Looks good to me, I will let @mortent approve.

site/book/README.md
```diff
@@ -1,15 +1,24 @@
...
 
-- Chapter 1 is a quickstart to get you oriented with kpt and sets up the context for the rest
-  of the book.
-- Chapter 2 introduces the fundamental concepts in kpt.
-- Chapter 3 covers how to work with kpt packages.
-- Chapter 4 covers how to use kpt functions to automate configuration changes.
-- Chapter 5 guides you through developing custom functions.
-- Chapter 6 covers how to apply a package to a Kubernetes cluster.
+- [Chapter 1] is a quickstart to get you oriented with kpt and sets up the
+  context for the rest of the book.
```
mikebz, line 10: not sure if the newline is needed here, but if it renders fine - it's all good.

mikebz:
LGTM as well

site/book/06-deploying-packages/00.md
```diff
@@ -0,0 +1,68 @@
...
+
+Then imagine the package is updated to contain the following resources:
+
+```
+service-1 (Service)
+deployment-1 (Deployment)
+config-map-2 (ConfigMap)
+```
+
+When the updated package is applied, `config-map-1` is automatically pruned from
```
mortent, line 35: Maybe use removed/deleted instead of pruned here just to avoid explaining pruning by saying resources are pruned.

site/book/06-deploying-packages/00.md
```diff
@@ -0,0 +1,68 @@
...
+would be reported as successful as soon as the resource has been created in the
+API server. With reconcile status, `live apply` will wait until the desired
+number of pods have been created and become available.
+
+For core kubernetes types, reconcile status is computed using hardcoded rules.
+For CRDs, the status computation is based on recommended [convention for status
+fields] that needs to be followed by custom resource publishers. If CRDs follow
+these conventions, `live apply` will be able to correctly compute status.
+
+Once you know the reconcile status of a single resource, then you really want to
```
mortent, line 61: I'm not sure about this section. We used to have the concept of aggregate status in kpt live, but we eventually moved away from it. For `kpt live apply`, what we really care about isn't a single aggregate status, but that each resource reach its desired end state. If the package contains a combination of new resources and updates to existing resources, as well as lacking some resources that should be pruned, what we actually want to wait for is all new and updated resources to be reconciled and all pruned resources to be fully removed from the cluster (i.e. all finalizers have completed).

site/book/06-deploying-packages/01-initializing-a-package-for-apply.md
```diff
@@ -0,0 +1,36 @@
+Before you can apply the package to the cluster, it needs to be initialized
+using `live init`. This is a one-time client-side operation that adds metadata
+to the `Kptfile` specifying a `ResourceGroup` resource on the cluster associated
```
mortent, line 3: This sentence is a bit vague. Can we be a bit more concrete about the information in the Kptfile rather than just say metadata. And be a bit more specific about the `ResourceGroup` resource than just saying it is "associated" with the package.

site/book/06-deploying-packages/02-applying-a-package.md
```diff
@@ -0,0 +1,102 @@
...
+## Server-side vs Client-side apply
+
+By default, `live apply` command uses client-side apply. The updates are
+accomplished by calculating and sending a patch from the client. Server-side
+apply, which can be enabled with the `--server-side` flag, sends the entire
+resource to the server for the update.
+
+## Dry-run
+
+You can uses the `--dry-run` flag to get break down of operations that will be
```
mortent, line 48: I think we should also mention validation here. The main advantage of dry-run (in particular server-side dry-run) is that the resources pass through all the validation steps so any errors can be caught early and before any changes have been made to the live state.

site/reference/schema/crd-status-convention/README.md
```diff
@@ -0,0 +1,63 @@
...
+to start), or that something is taking longer than expected (similar to the
+`progressDeadlineSeconds` timeout on Deployments). If this condition is True, it
+should be interpreted that something might be wrong. It does not mean that the
+resource will never be reconciled. Most process in Kubernetes retry forever, so
+this should not be considered a terminal state.
+
+These conditions adhere to the [Kubernetes design principles] which include
+expressing conditions using abnormal-true polarity. There is currently a
+[proposal] to change to guidance for conditions. If this is accepted, the
+recommended conditions for kpt might also change, but we will continue to
```
mortent, line 24: This PR has merged, so we can probably remove the reference to abnormal-true polarity here. We should also mention that kpt does understand the `Ready` condition if it is found on a resource. It essentially works the same as the `Reconciling` condition, just opposite polarity.

site/book/06-deploying-packages/01-initializing-a-package-for-apply.md
mortent: Seeing now that this is explained in some more detail below.

site/book/06-deploying-packages/01-initializing-a-package-for-apply.md
mortent: Publishing a package with the inventory information would not be recommended. The inventory is specific to applying the packages in a specific cluster, so adding this information to the Kptfile should probably happen once the package has been customized. This will somewhat depend on how users want to handle things like creating variants of a package for different environments.

We have talked about adding a separate publish command. A first step might be to add a command that can provide the validation of the package prior to publishing. Some things that could be included are:
 * No inventory information.
 * Package is hydrated, i.e. running the `kpt fn render` command should not lead to a diff.
 * The `UpstreamLock` information matches the `Upstream` information in all subpackages.
 * ...

frankfarzan: I think we will need a publish command, otherwise, there's a growing laundry list of things the publisher needs to manually watch out for.

frankfarzan:
I'm going to submit this and address any follow up comments in a PR.

site/book/06-deploying-packages/02-applying-a-package.md
mortent: Yeah, validation errors will be caught during apply, but only when the resource in question is being applied to the cluster. Catching validation errors at this point will likely lead to the package being only partially applied.

PR #2214, opened 2021-06-10
include stderr in results when function succeeds

### LINKED ISSUE #2200: Kpt fn render swallows stderr output when exit code is 0
Current UX when exit code is 0:
Run `kpt fn render --results-dir=results`. There are no stderr in the console and no stderr in the structured results.

Proposed UX when exit code is 0:
Run `kpt fn render --results-dir=results`, there are no stderr in the console, but we should have stderr in the structured results.

Motivation:
Using stderr is a common way for debugging. (e.g. starlark).

Related code: https://github.com/GoogleContainerTools/kpt/blob/24e8929b1332b8185e6f439a6b4787c444305b76/internal/fnruntime/runner.go#L142-L147

Shell32-Natsu:
close #2200

### DIFF
--- a/e2e/testdata/fn-eval/fn-success-with-stderr/starlark.yaml
+++ b/e2e/testdata/fn-eval/fn-success-with-stderr/starlark.yaml
@@ -17,4 +17,4 @@ kind: StarlarkRun
 metadata:
   name: httpbin-gen
 source: |
-  print("error message")
+  print("function succeeded, reporting it on stderr")
--- a/e2e/testdata/fn-eval/fn-success-with-stderr/starlark.yaml
+++ b/e2e/testdata/fn-eval/fn-success-with-stderr/starlark.yaml
@@ -1,0 +1,20 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: fn.kpt.dev/v1alpha1
+kind: StarlarkRun
+metadata:
+  name: httpbin-gen
+source: |
+  print("error message")
--- a/internal/fnruntime/container.go
+++ b/internal/fnruntime/container.go
@@ -108,6 +108,12 @@ func (f *ContainerFn) Run(reader io.Reader, writer io.Writer) error {
 		return fmt.Errorf("unexpected function error: %w", err)
 	}
 
+	if errSink.Len() != 0 {
+		return &StderrNotEmpty{
+			Stderr: errSink.String(),
+		}
+	}
+
 	return nil
 }
 
--- a/internal/fnruntime/container.go
+++ b/internal/fnruntime/container.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/GoogleContainerTools/kpt/internal/printer"
 	"github.com/GoogleContainerTools/kpt/internal/types"
+	fnresult "github.com/GoogleContainerTools/kpt/pkg/api/fnresult/v1alpha2"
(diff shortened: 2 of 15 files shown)


internal/fnruntime/container.go
```diff
@@ -108,6 +108,12 @@ func (f *ContainerFn) Run(reader io.Reader, writer io.Writer) error {
 		return fmt.Errorf("unexpected function error: %w", err)
 	}
 
+	if errSink.Len() != 0 {
```
droot, line 111: If I understand correctly, `ContainerFn` will be wrapped by runtime.FunctionFilter in `kyaml`. I wonder if `returning error` will have some side-effect etc such as it will ignore some things considering it as a real function failure or something ?

internal/fnruntime/runner.go
```diff
@@ -137,12 +137,18 @@ func (fr *FunctionRunner) do(input []*yaml.RNode) (output []*yaml.RNode, err err
...
-			fnResult.Stderr = fnErr.Stderr
-			fnErr.FnResult = fnResult
+		var execErr *ExecError
+		var stderrNotEmpty *StderrNotEmpty
+		switch {
+		case goerrors.As(err, &execErr):
+			fnResult.ExitCode = execErr.ExitCode
+			fnResult.Stderr = execErr.Stderr
 			fr.fnResults.ExitCode = 1
+		case goerrors.As(err, &stderrNotEmpty):
```
droot, line 147: I am feeling a bit uncomfortable retrieving this info through an error mechanism, but not able to think a way around it.

e2e/testdata/fn-eval/fn-success-with-stderr/starlark.yaml
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: fn.kpt.dev/v1alpha1
+kind: StarlarkRun
+metadata:
+  name: httpbin-gen
+source: |
+  print("error message")
```
droot, line 20: error is so overloaded, can we put something other than that :)
"function succeeded, reporting it on stderr"

droot (changes requested):

e2e/testdata/fn-eval/fn-success-with-stderr/starlark.yaml
frankfarzan: yea, it would be anti-pattern to have an error and succeed

internal/fnruntime/container.go
Shell32-Natsu: I understand your concern, another possible solution is adding the pointer to `FnResult` to ContainerFn struct and update the stderr directly in ContainerFn. It shares a variable across multiple different layers. Do you think it's better?

internal/fnruntime/fnerrors.go
```diff
@@ -64,3 +59,15 @@ func (fe *ExecError) String() string {
 func (fe *ExecError) Error() string {
 	return fe.String()
 }
+
+// StderrNotEmpty is an error type which indicates that the function
```
frankfarzan, line 63: Conceptually, this is NOT an error type.

internal/fnruntime/container.go
droot: > Do you think it's better?

Yes, I think that is much better the current approach. Ideal approach will be to update the `Run(r, w) error` interface to include function results in the output, but we don't want to go there at this stage of the project.

internal/fnruntime/runner.go
droot: Your new proposal should address this concern.

internal/fnruntime/container.go
Shell32-Natsu: done

internal/fnruntime/runner.go
Shell32-Natsu: done

e2e/testdata/fn-eval/fn-success-with-stderr/starlark.yaml
Shell32-Natsu: done

internal/fnruntime/fnerrors.go
Shell32-Natsu: Removed this type

droot (approved):

PR #2225, opened 2021-06-11
Fix Go Action Workflow

runewake2:
Update to CI Workflows in order to update the jobs to target v0 and main branches instead of master and next.

### DIFF
--- a/.github/workflows/checkDocSync.yml
+++ b/.github/workflows/checkDocSync.yml
@@ -17,7 +17,7 @@ on:
   pull_request:
     branches:
       - master
-      - next
+      - main
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/checkDocSync.yml
```diff
@@ -17,7 +17,7 @@ on:
   pull_request:
     branches:
       - master
```
frankfarzan, line 19: master is now `v0`.

frankfarzan (approved):

.github/workflows/checkLicenses.yml
```diff
@@ -20,7 +20,7 @@ on:
       - main
   pull_request:
     branches:
-      - master
+      - v0
```
frankfarzan, line 23: Didn't update line 19

PR #2226, opened 2021-06-11
kustomize solution

mikebz:
There are several things in this PR:
1. It creates a new solution folder on the website.  There are several things that go beyond just using kpt that I believe people are asking for - kpt and kustomize is probably the first one.  Other guidance scenarios are around interoperability with ConfigSync, using kpt in Cloud Build, Docker in Docker etc.
2. The FAQ is expanded to answer the frequently asked question about the positioning of kpt vs kustomize
3. The solution utilizing the existing example to go through the end to end scenario, understand the reason why this solution is put together as it is and reference the various chapters in the book to learn more.

### DIFF
--- a/site/faq/README.md
+++ b/site/faq/README.md
@@ -52,37 +52,29 @@ resource configuration provides a number of desirable properties:
    YAML/JSON directly.
 
 
-### Does kpt and kustomize work together?
+### Do kpt and kustomize work together?
 
-Yes!  Because both tools subscribe to template/DSL free approach to configuraiton it's possible to use both tools together.  We have created a 
+Yes!  Because both tools subscribe to template/DSL free approach to 
+configuration it's possible to use both tools together.  We have created a 
 [kustomize solution] which shows how you can leverage both of these tools
 in your environment.
 
 #### Kustomize background
-[Kustomize] is a CNCF project that is a part of Kubernetes.  It's included in 
-the `kubectl` in order to allow users to customize their configurations without introducing templates.   Kustomize was started by Googlers and still has 
-large contributions by Google engineers.  A lot of the Google engineers working
-on kpt also work on kustomize and a lot of the underlying libraries are shared.
-
-While kpt and kustomize support customization of configuration yaml there are 
-important differences and scope to these projects.  Kpt supports end to end 
-scenarios of packaging, customization and application of resource to the 
-clusters.  For kustomize packaging is explicitly out of scope and application 
-is deferred to whatever kubectl provides.
+[Kustomize] is a CNCF project that is a part of Kubernetes.  While kpt and 
+kustomize support customization of KRM manifests without templates there 
+are important differences and scope to these projects. Kpt supports end to end scenarios of packaging, customization and actuation of resources to the  
+clusters.  For kustomize packaging is explicitly out of scope and application is deferred to whatever kubectl provides.
 
 #### kpt vs kustomize customization approach.
-While kpt is set to support both in place and out-of-place customization,
-kustomize is only focused on providing out-of-place hydration using the 
-overlay pattern.  Here are the strengths and investment areas for these 
-products:
+While kpt supports both in place and out-of-place customization, kustomize is 
+only focused on providing out-of-place hydration using the  [overlay pattern].  
+Here are the strengths and investment areas for these products:
 
 *kpt*
 - Allows you to edit the configuration in-place without creating complex patches
 - Focuses on rebase with resource merge strategy allowing for edited config to 
 be updated
-- Is more intuitive and has a continuous learning curve as you usually start 
-small with modifying several YAML files using an editor and then want to scale 
-with complexity of the application
+- Allows one to start customization by modifying a few YAML files using an editor.
 - Allows to mix programmatic changes (functions) with manual edits in the same 
 [workflow]
 
@@ -129,3 +121,4 @@ don't have to alias it. It is pronounced "kept".
 [kustomize solution]: /solutions/kustomize/
 [kustomize]: https://kustomize.io
 [workflow]: /book/02-concepts/02-workflows
+[overlay pattern]: https://github.com/kubernetes-sigs/kustomize/tree/master/examples/multibases
--- a/site/faq/README.md
+++ b/site/faq/README.md
@@ -51,6 +51,47 @@ resource configuration provides a number of desirable properties:
 7. it **supports display in UI and tools** which use either OpenAPI or the
    YAML/JSON directly.
 
+
+### Does kpt and kustomize work together?
+
+Yes!  Because both tools subscribe to template/DSL free approach to configuraiton it's possible to use both tools together.  We have created a 
+[kustomize solution] which shows how you can leverage both of these tools
+in your environment.
+
+#### Kustomize background
+[Kustomize] is a CNCF project that is a part of Kubernetes.  It's included in 
+the `kubectl` in order to allow users to customize their configurations without introducing templates.   Kustomize was started by Googlers and still has 
+large contributions by Google engineers.  A lot of the Google engineers working
+on kpt also work on kustomize and a lot of the underlying libraries are shared.
+
+While kpt and kustomize support customization of configuration yaml there are 
+important differences and scope to these projects.  Kpt supports end to end 
+scenarios of packaging, customization and application of resource to the 
+clusters.  For kustomize packaging is explicitly out of scope and application 
+is deferred to whatever kubectl provides.
+
+#### kpt vs kustomize customization approach.
+While kpt is set to support both in place and out-of-place customization,
+kustomize is only focused on providing out-of-place hydration using the 
+overlay pattern.  Here are the strengths and investment areas for these 
+products:
+
+*kpt*
+- Allows you to edit the configuration in-place without creating complex patches
+- Focuses on rebase with resource merge strategy allowing for edited config to 
+be updated
+- Is more intuitive and has a continuous learning curve as you usually start 
+small with modifying several YAML files using an editor and then want to scale 
+with complexity of the application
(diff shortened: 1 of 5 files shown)


site/solutions/kustomize/README.md
```diff
@@ -0,0 +1,183 @@
+---
+title: "kpt and kustomize"
+linkTitle: "kustomize"
+type: docs
+---
+
+## Overview
```
frankfarzan, line 7: We should just include this in the README of the example package.

I prefer we don't add additional top-level sections to the kpt.dev until there are multiple solutions. Otherwise, this will stand out and people may overfit on it.

mikebz: I will be adding a ConfigSync section shortly.  That's another critical journey we need to cover.

frankfarzan: This seems to duplicating the content here:

https://github.com/GoogleContainerTools/kpt/tree/main/package-examples/kustomize-pkg

Per offline discussion, I suggest try to make the solution packages self-contained, and not requiring touching `/site` every time we add an example. For now, we can keep it simple, and just point to the package README from the FAQ. Long term, we can provide an index page, and render the package README.me in kpt.dev as well, similar to what we do for http://catalog.kpt.dev/

droot: I think we need to do a bit of homework about onramp guides for `kustomize` users. For ex. I would like to show a kustomize users how to accomplish variants using `fork/edit` patterns and highlight the fact some of the common customizations such as `labels`, `namespace`, `name changes` etc, how they are done using `pipeline` construct. And second aspect of onramp is what is being captured here where both tools co-exist, but that requires users to deal with two tools (not sure if that scales in a big organization).

So I agree with @frankfarzan here, before, putting a prominent solution page, we should form an informed opinion about the approaches. So for now having a brief FAQ entry pointing to the example pkg containing all the details in the README of the example package seems like a good idea. Also, we also need to vet the example with a few users before me make it prominent on the site.

mikebz: Sounds good.  I have vetted this pattern with a couple of customers.  I think if you'd prefer to take this into a different direction I am cool with it.  I would like to get back to one of the customers (2nd customer who asked for guidance) next week if possible.

droot: > I would like to get back to one of the customers (2nd customer who asked for guidance) next week if possible.

Can we move the content from `site/solutions/kustomize/README.md` to the `examples/kustomize-pkg/README.md` to unblock the customer communication? And from `FAQ` adding a link to that README ?

site/faq/README.md
```diff
@@ -51,6 +51,47 @@ resource configuration provides a number of desirable properties:
...
+### Does kpt and kustomize work together?
+
+Yes!  Because both tools subscribe to template/DSL free approach to configuraiton it's possible to use both tools together.  We have created a 
+[kustomize solution] which shows how you can leverage both of these tools
+in your environment.
+
+#### Kustomize background
+[Kustomize] is a CNCF project that is a part of Kubernetes.  It's included in 
+the `kubectl` in order to allow users to customize their configurations without introducing templates.   Kustomize was started by Googlers and still has 
+large contributions by Google engineers.  A lot of the Google engineers working
```
mortent, line 64: This section feels a bit verbose. Can we make it a little more concise?

site/faq/README.md
```diff
@@ -51,6 +51,47 @@ resource configuration provides a number of desirable properties:
...
+on kpt also work on kustomize and a lot of the underlying libraries are shared.
+
+While kpt and kustomize support customization of configuration yaml there are
```
mortent, line 67: Say "KRM manifests" instead of "configuration yaml"

site/faq/README.md
```diff
@@ -51,6 +51,47 @@ resource configuration provides a number of desirable properties:
...
+
+#### Kustomize background
+[Kustomize] is a CNCF project that is a part of Kubernetes.  It's included in 
+the `kubectl` in order to allow users to customize their configurations without introducing templates.   Kustomize was started by Googlers and still has 
+large contributions by Google engineers.  A lot of the Google engineers working
+on kpt also work on kustomize and a lot of the underlying libraries are shared.
+
+While kpt and kustomize support customization of configuration yaml there are 
+important differences and scope to these projects.  Kpt supports end to end 
+scenarios of packaging, customization and application of resource to the
```
mortent, line 69: I think actuation would be a better word than application here.

site/faq/README.md
```diff
@@ -51,6 +51,47 @@ resource configuration provides a number of desirable properties:
...
+on kpt also work on kustomize and a lot of the underlying libraries are shared.
+
+While kpt and kustomize support customization of configuration yaml there are 
+important differences and scope to these projects.  Kpt supports end to end 
+scenarios of packaging, customization and application of resource to the 
+clusters.  For kustomize packaging is explicitly out of scope and application 
+is deferred to whatever kubectl provides.
+
+#### kpt vs kustomize customization approach.
+While kpt is set to support both in place and out-of-place customization,
```
mortent, line 74: "kpt supports both in-place and out-place customization"

site/solutions/kustomize/README.md
```diff
@@ -0,0 +1,183 @@
...
+kind: Kptfile
+metadata:
+  labels:
+    environ: dev
+  name: dev-dev
+```
+
+This rendered configuration can be deployed with `kubectl` but you can also take 
+advantage of `kpt live apply` if you extend your kustomization.yaml to include
+the necessary kpt metadata.  You can learn more about kpt `apply` command in the
```
mortent, line 131: What does it mean to extend `kustomization.yaml` to include kpt metadata?

site/solutions/kustomize/README.md
```diff
@@ -0,0 +1,183 @@
...
+      containers:
+      - image: nginx:1.14.1
+        name: nginx
+        ports:
+        - containerPort: 80
+---
+apiVersion: kpt.dev/v1alpha2
+info:
+  description: sample description
+kind: Kptfile
```
mortent, line 122: What is the reason for including the Kptfile in the list of kustomize resources?

site/faq/README.md
```diff
@@ -51,6 +51,47 @@ resource configuration provides a number of desirable properties:
 7. it **supports display in UI and tools** which use either OpenAPI or the
    YAML/JSON directly.
 
+
+### Does kpt and kustomize work together?
```
natasha41575, line 55: s/does/do

site/faq/README.md
```diff
@@ -51,6 +51,47 @@ resource configuration provides a number of desirable properties:
...
+
+Yes!  Because both tools subscribe to template/DSL free approach to configuraiton it's possible to use both tools together.  We have created a
```
natasha41575, line 57: s/configuraiton/configuration

site/faq/README.md
```diff
@@ -51,6 +51,47 @@ resource configuration provides a number of desirable properties:
...
+While kpt and kustomize support customization of configuration yaml there are 
+important differences and scope to these projects.  Kpt supports end to end 
+scenarios of packaging, customization and application of resource to the 
+clusters.  For kustomize packaging is explicitly out of scope and application 
+is deferred to whatever kubectl provides.
+
+#### kpt vs kustomize customization approach.
+While kpt is set to support both in place and out-of-place customization,
+kustomize is only focused on providing out-of-place hydration using the 
+overlay pattern.  Here are the strengths and investment areas for these
```
natasha41575, line 76: suggestion: maybe have "overlay pattern" link to https://github.com/kubernetes-sigs/kustomize/tree/master/examples/multibases ?

site/faq/README.md
```diff
@@ -51,6 +51,47 @@ resource configuration provides a number of desirable properties:
...
+While kpt is set to support both in place and out-of-place customization,
+kustomize is only focused on providing out-of-place hydration using the 
+overlay pattern.  Here are the strengths and investment areas for these 
+products:
+
+*kpt*
+- Allows you to edit the configuration in-place without creating complex patches
+- Focuses on rebase with resource merge strategy allowing for edited config to 
+be updated
+- Is more intuitive and has a continuous learning curve as you usually start
```
natasha41575, line 83: "is more intuitive" is a little subjective, is it necessary to state it here or would it make sense to let the product speak for itself?

site/faq/README.md
mikebz: sure, I have cut down a bit.  This language came from the internal doc we all reviewed earlier.

site/faq/README.md
mikebz: done

site/faq/README.md
mikebz: done

site/solutions/kustomize/README.md
mikebz: reworded

site/solutions/kustomize/README.md
mikebz: Good question the reason is that customers want to use kpt packaging and kpt live apply but want to leverage their existing investment into kustomize build.  `kustomize build kustomize-pkg/overlays/dev | kpt live apply - ` doesn't work unless you pass in the inventory which is a part of the Kptfile (as you know)

site/faq/README.md
mikebz: good idea, adding.

site/faq/README.md
mikebz: Good call.  Changing.  This could start an opinion war.

mikebz:
@frankfarzan feel free to take this branch over and merge in the place you feel appropriate.

frankfarzan:
> @frankfarzan feel free to take this branch over and merge in the place you feel appropriate.

Done: https://github.com/GoogleContainerTools/kpt/pull/2243

Thanks for the content.

PR #2275, opened 2021-06-17
Fix support for local-config annotation

mortent:
This adds back the support for the local config annotation that was removed in error in an earlier PR. This also adds support for using `config.kubernetes.io/local-config: "false"` to explicitly set a resource as not being local config. This means that even if the file with the resource is referenced from the hydration pipeline, it will be applied by kpt. 

Any other values than `"true"` or `"false"` are also allowed and they will cause the resource to be considered local config. This is a little strange, but I don't think we should let kpt error out on other values.

### DIFF
--- a/pkg/live/rgpath_test.go
+++ b/pkg/live/rgpath_test.go
@@ -97,6 +97,32 @@ func TestPathManifestReader_Read(t *testing.T) {
 				},
 			},
 		},
+		"Function config resources which are marked as not being local config remains": {
+			manifests: map[string]string{
+				"Kptfile":           kptFileWithPipeline,
+				"deployment-a.yaml": deploymentA,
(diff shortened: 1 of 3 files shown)


pkg/live/rgpath_test.go
```diff
@@ -127,6 +153,23 @@ func TestPathManifestReader_Read(t *testing.T) {
 			namespace:      "test-namespace",
 			expectedErrMsg: "unknown resource types: Custom.custom.io",
 		},
+		"local-config is filtered out": {
```
frankfarzan, line 156: orthogonal to this: I'm pretty much convince that declaring packages in Go like this is an anti-pattern. At some point, need to move away from this.

frankfarzan (approved):

PR #2298, opened 2021-06-22
Fix cloud build job

mortent:
Fix cloud build job on `v0` branch. This PR changes two things:
 * Check the code out into a proper go structure and set the correct GOPATH.
 * Keep using `master` for the images and the binaries generated to avoid breaking users.

### DIFF
--- a/release/branch/cloudbuild.yaml
+++ b/release/branch/cloudbuild.yaml
@@ -53,14 +53,14 @@ steps:
   # Clone the kpt repo. The code already set up by cloud build is a copy
   # and not actually a git checkout, so we need to do this manually.
   - name: 'gcr.io/cloud-builders/git'
-    args: ['clone', 'git@github.com:${_GITHUB_USER}/kpt.git']
+    args: ['clone', 'git@github.com:${_GITHUB_USER}/kpt.git', 'go/src/github.com/GoogleContainerTools/kpt']
     volumes:
       - name: home
         path: /root
 
   # Make sure we check out the version of the code that triggered the run.
   - name: 'gcr.io/cloud-builders/git'
-    dir: 'kpt'
+    dir: 'go/src/github.com/GoogleContainerTools/kpt'
     args: ['checkout', '${COMMIT_SHA}']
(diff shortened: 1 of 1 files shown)


release/branch/cloudbuild.yaml
```diff
@@ -86,60 +86,60 @@ steps:
...
   # push the container image
   - name: 'gcr.io/cloud-builders/docker'
-    args: ['push', 'gcr.io/$PROJECT_ID/kpt:${BRANCH_NAME}']
-    dir: 'kpt'
+    args: ['push', 'gcr.io/$PROJECT_ID/kpt:master']
+    dir: 'go/src/github.com/GoogleContainerTools/kpt'
 
   # push the binaries
   - name: 'gcr.io/cloud-builders/gsutil'
     args: ['cp', '-r', '-a', 'public-read', '/workspace/bin/', 'gs://${_GCS_BUCKET}/']
```
droot, line 136: Not much familiar with the setup, so take it with a grain of salt. If I understand correctly, the binaries for `v0` branch will be available under gsbucket/.../master/kpt_OS_ARCH, right ? if that is correct, then, it can get confusing for people because people think `master` has the latest. Not sure, how much of a big deal that is.

droot (commented):
drive by comment.

release/branch/cloudbuild.yaml
mortent: The files are pushed to GCS under `kpt-dev/bin/master/OS_ARCH/kpt(.exe)`. We also publish docker images as `gcr.io/kpt-dev/kpt:master`. Users might rely on these stable URIs for fetching the latest (pre 1.0) version of kpt. Changing from master to v0 would break this, so I think we should just keep using master. Relatively soon we will discontinue support for that branch and we can remove the old binaries.

droot: Got it. Thanks.

droot (approved):

etefera (approved):
Thanks!

PR #2334, opened 2021-06-24
fn eval and render should understand image refs via SHA

natasha41575:
fixes https://github.com/GoogleContainerTools/kpt/issues/2322, and makes validation consistent for `fn render` and `fn eval`

### DIFF
--- a/pkg/api/kptfile/v1/validation.go
+++ b/pkg/api/kptfile/v1/validation.go
@@ -60,7 +60,7 @@ func (p *Pipeline) validate(pkgPath types.UniquePath) error {
 }
 
 func (f *Function) validate(fnType string, idx int, pkgPath types.UniquePath) error {
-	err := validateFunctionName(f.Image)
+	err := ValidateFunctionName(f.Image)
 	if err != nil {
 		return &ValidateError{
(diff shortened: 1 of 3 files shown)


pkg/api/kptfile/v1/validation.go
```diff
@@ -106,14 +106,16 @@ func (f *Function) validate(fnType string, idx int, pkgPath types.UniquePath) er
 //	path-component                  := alpha-numeric [separator alpha-numeric]*
 // 	alpha-numeric                   := /[a-z0-9]+/
 //	separator                       := /[_.]|__|[-]*/
-func validateFunctionName(name string) error {
+func ValidateFunctionName(name string) error {
```
droot, line 109: Can we rename this to `ValidateFunctionImageURL` ?

droot (changes requested):

pkg/api/kptfile/v1/validation.go
natasha41575: done

droot:
@natasha41575 rebasing should fix the CI failures.

droot (approved):

PR #2339, opened 2021-06-25
fix renamed fn image names

mengqiy:


package-examples/wordpress/Kptfile
```diff
@@ -8,7 +8,7 @@ info:
   description: This is an example wordpress package with mysql subpackage.
 pipeline:
   mutators:
-    - image: gcr.io/kpt-fn/set-label:v0.1
+    - image: gcr.io/kpt-fn/set-labels:v0.1
```
frankfarzan, line 11: can you test the examples work as before?

frankfarzan: Also, is there any reference to `unstable` remaining in the kpt repo?

frankfarzan (approved):

pkg/api/kptfile/v1/validation_test.go
```diff
@@ -166,23 +166,23 @@ func TestValidateFunctionName(t *testing.T) {
 	}
 	inputs := []input{
 		{
-			"gcr.io/kpt-functions/generate-folders",
+			"gcr.io/kpt-fn/generate-folders",
```
mengqiy, line 169: Changes in this file doesn't affect the correctness. But I still want to change it to avoid using the old gcr name.

mengqiy:
This PR need @droot's approval as CODEOWNER

droot (approved):

PR #2353, opened 2021-06-30
Include git command (and directory) in git verbose output

justinsb:
Otherwise we can't see the actual command being run (e.g. fetch /
ls-remote etc)


mikebz:
is this a debugging request?  What happens if you have it?  Does this need to happen only in a situation where you have a problem?

mortent (approved):

PR #2359, opened 2021-07-02
Adds error message for missing directory on get.

etefera:
In response to confusing error messaging in https://github.com/GoogleContainerTools/kpt/issues/2323

```
$ kpt pkg get git@github.com:GoogleContainertools/kpt.git/bibbidi-bobbidi-boo
Package "bibbidi-bobbidi-boo":
Fetching git@github.com:GoogleContainertools/kpt@main
From github.com:GoogleContainertools/kpt
 * branch            main       -> FETCH_HEAD
Error: path "/bibbidi-bobbidi-boo" does not exist in repo "git@github.com:GoogleContainertools/kpt" 
```


mikebz:
I am not 100% sold that we should provide this message.  It seems to me that working seamlessly with existing manifest directories makes it feel like Kpt meets customers where they are.  I am sorry I didn't take a strong position on this bug and potentially led to wasted effort.  If it's not blocking anything for the customer I would not display an error by default.

etefera:
> I am not 100% sold that we should provide this message. It seems to me that working seamlessly with existing manifest directories makes it feel like Kpt meets customers where they are. I am sorry I didn't take a strong position on this bug and potentially led to wasted effort. If it's not blocking anything for the customer I would not display an error by default.

We already generate an error message for this case (as described in https://github.com/GoogleContainerTools/kpt/issues/2323), it's just more descriptive in the case of a missing file.

mortent:
So I don't think we should reference the Kptfile here, as that is not really the problem. The problem is when a user specifies a path to a package within a repo that doesn't exist. As an example:
```
kpt pkg get git@github.com:GoogleContainertools/kpt.git/bibbidi-bobbidi-boo
```
returns the error
```
Error: lstat /tmp/kpt-get-159126868/bibbidi-bobbidi-boo: no such file or directory 
```
This should be an error, but it should say that the directory `bibbidi-bobbidi-boo` doesn't exist in the repository at `github.com:GoogleContainertools/kpt`, and not reference a temporary directory on the local filesystem. So I think you are pretty close here, but a missing Kptfile in upstream isn't an error.

A corner case here is whether we should return an error on an empty directory (which would include an empty repo). I think for now we don't return an error in that situation.

internal/util/fetch/fetch.go
```diff
@@ -199,9 +199,14 @@ func ClonerUsingGitExec(ctx context.Context, repoSpec *git.RepoSpec) error {
 		return errors.E(op, errors.Internal, fmt.Errorf("error copying package: %w", err))
 	}
 
+	// Verify that the requested path exists in the repo.
+	pkgPath := filepath.Join(dir, repoSpec.Path)
+	if _, err = os.Stat(pkgPath); os.IsNotExist(err) {
+		return errors.E(op, errors.Internal, err, fmt.Errorf("path %q does not exist in repo %q", repoSpec.Path, repoSpec.OrgRepo))
```
droot, line 205: @etefera Can you share the final CLI output ? (may be update in the PR description).

etefera: Updated PR description with the following:
```
$ kpt pkg get git@github.com:GoogleContainertools/kpt.git/bibbidi-bobbidi-boo
Package "bibbidi-bobbidi-boo":
Fetching git@github.com:GoogleContainertools/kpt@main
From github.com:GoogleContainertools/kpt
 * branch            main       -> FETCH_HEAD
Error: path "/bibbidi-bobbidi-boo" does not exist in repo "git@github.com:GoogleContainertools/kpt" ```

droot (approved):
Change looks good to me. I will have @mortent or @phanimarupaka approve it.

droot:
@mortent since you had some comments on it, can you pl. take a final look on this one. Thanks

phanimarupaka (approved):

PR #2367, opened 2021-07-07
Update cli utils dependency

mortent:
This updates the cli-utils dependency to the latest version. This includes support for the `depends-on` annotation.

The goal here is to update the cli-utils dependency with minimal changes to the output format of the `kpt live` commands. We do want to update the format at some point to better show the steps of the apply/prune operation, but it needs some design work.
As part of this, the `pkg/print` package from cli-utils is copied into thirdparty.


e2e/live/end-to-end-test.sh
```diff
@@ -439,11 +439,11 @@ ${BIN_DIR}/kpt live apply e2e/live/testdata/rg-test-case-1a > $OUTPUT_DIR/status
 # The ResourceGroup inventory CRD is automatically installed on the initial apply.
 assertContains "installing inventory ResourceGroup CRD"
 assertContains "namespace/rg-test-namespace"
+assertContains "1 resource(s) applied. 0 created, 1 unchanged, 0 configured, 0 failed"
 assertContains "pod/pod-a created"
 assertContains "pod/pod-b created"
 assertContains "pod/pod-c created"
 assertContains "4 resource(s) applied. 3 created, 1 unchanged, 0 configured, 0 failed"
-assertContains "0 resource(s) pruned, 0 skipped, 0 failed"
```
droot, line 446: so `prune` summary wasn't useful ? It will be comforting to know if any resources were pruned.

internal/cmdapply/cmdapply.go
```diff
@@ -199,10 +199,19 @@ func runApply(r *Runner, invInfo inventory.InventoryInfo, objs []*unstructured.U
 
 	// Run the applier. It will return a channel where we can receive updates
 	// to keep track of progress and any issues.
-	if err := r.Applier.Initialize(); err != nil {
+	poller, err := factory.NewStatusPoller(r.factory)
```
droot, line 202: Not much familiar with the code, but `factory.NewSomething(r.factory)` feels a bit odd to read as code reader. May be `status.NewPoller(r.factory)` or `poller.NewStatusPoller(r.factory)` ?

e2e/live/end-to-end-test.sh
mortent: We are changing the output so it only shows up if anything is actually pruned. The current output has confusing for users since the context of the summary output is not clear.

internal/cmdapply/cmdapply.go
mortent: Good point. I have used an alias for the import so it is now `status.NewStatusPoller(r.factory)`

droot (approved):

PR #2370, opened 2021-07-08
Remove the `next` tag from kpt images

mortent:
We should no longer build images with the `next` tag. Releases should only be built using the semver tags. 

We might want to set up a CD pipeline for building images directly from `main` using the `unstable` tag, although I don't think we have actually had any requests for this yet.

### DIFF
--- a/release/tag/goreleaser.yaml
+++ b/release/tag/goreleaser.yaml
@@ -29,11 +29,9 @@ builds:
 dockers:
   - image_templates:
       - "gcr.io/kpt-dev/kpt:{{ .Tag }}"
-      - "gcr.io/kpt-dev/kpt:next"
     dockerfile: "release/images/Dockerfile"
   - image_templates:
       - "gcr.io/kpt-dev/kpt-gcloud:{{ .Tag }}"
-      - "gcr.io/kpt-dev/kpt-gcloud:next"
     dockerfile: "release/images/Dockerfile-gcloud"
 archives:
   - id: archived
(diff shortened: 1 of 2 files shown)


release/tag/goreleaser.yaml
```diff
@@ -29,11 +29,9 @@ builds:
 dockers:
   - image_templates:
       - "gcr.io/kpt-dev/kpt:{{ .Tag }}"
-      - "gcr.io/kpt-dev/kpt:next"
     dockerfile: "release/images/Dockerfile"
   - image_templates:
       - "gcr.io/kpt-dev/kpt-gcloud:{{ .Tag }}"
```
droot, line 34: I am trying to understand the impact of this. We will stop updating the `kpt-gcloud:next` and `kpt:next`. So unless we delete them from the registry, existing users (who use these two images) will continue to be using the old version of kpt. We can probably keep them around for sometime (say 2-4 months), and then delete ?

mortent: Yeah, this just means that new images will no longer get the `next` tag. I think this is aligned with how we handle functions, where we use semver tags for releases and the unstable tag for builds from main/master. We currently don't build kpt on every PR merge, but if we want to do that, we should use the `unstable` tag.

Any users currently relying on the `next` tag will no longer be automatically updated. But they probably shouldn't be doing that anyway, since it is pretty risky.

droot: Ack. sounds good.

droot (approved):

PR #2372, opened 2021-07-08
Fix issue where live resolver passes the wrong error to the template engine

mortent:
The live error resolver has a bug where the top-level error is passed to the template engine instead of the specific type needed by the template.


mikebz:
thanks for fixing this @mortent !!

droot (approved):

PR #2390, opened 2021-07-12
 New e2e tests for kpt live

mortent:
This is the start for setting up a new framework for running e2e tests for kpt live. It is using the same structure as the existing fn tests.

### DIFF
--- a/e2e/testdata/live-apply/crd-and-cr/resources/Kptfile
+++ b/e2e/testdata/live-apply/crd-and-cr/resources/Kptfile
@@ -1,0 +1,8 @@
+apiVersion: kpt.dev/v1
+kind: Kptfile
+metadata:
+  name: crd-and-cr
+inventory:
+  namespace: crd-and-cr
+  name: crd-and-cr
+  inventoryID: 24a8006de03ceca5ad9c551b1ee467414b6cb5a6-1625960472425460000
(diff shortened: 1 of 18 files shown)


e2e/testdata/live-apply/crd-and-cr/resources/Kptfile
```diff
@@ -0,0 +1,8 @@
+apiVersion: kpt.dev/v1
+kind: Kptfile
+metadata:
+  name: crd-and-cr
+inventory:
+  namespace: crd-and-cr
+  name: crd-and-cr
+  inventoryID: 24a8006de03ceca5ad9c551b1ee467414b6cb5a6-1625960472425460000
```
droot, line 8: how is this generated ?

droot: Will be good if we can add a README.md describing steps for adding/debugging a test.

droot (commented):
The change is looking good, have minor comments about the documenting.

e2e/testdata/live-apply/crd-and-cr/resources/Kptfile
mortent: Updated the tests to avoid the auto-generated inventoryIds. Those are not important for the test.

Also added a README.md with the basic information about the kpt live e2e tests.

droot (approved):

PR #2394, opened 2021-07-13
Add kptfile schema to make generate

etefera:
Addresses https://github.com/GoogleContainerTools/kpt/issues/2038 by introducing a method to generate the kptfile schema using the go struct as the source of truth.


go.mod
```diff
@@ -3,8 +3,10 @@ module github.com/GoogleContainerTools/kpt
 go 1.16
 
 require (
+	github.com/Azure/go-autorest/autorest/adal v0.9.5 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.0
 	github.com/go-errors/errors v1.4.0
+	github.com/go-swagger/go-swagger v0.27.0 // indirect
```
droot, line 9: Since we are using `go install` to install the swagger tool, these shouldn't show up in `go.mod`.

pkg/api/kptfile/v1/types.go
```diff
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package defines the schema for Kptfile version v1.
+// Package defines Kptfile schema.
+// Version: v1
+// swagger:meta
```
droot, line 17: Did we checkout using `controller-gen` tool from kubebuilder project ? 
One good thing about `controller-gen` is that it generates the schema as well as CRD definition as well (which probably is required by cloudcode ?) @phanimarupaka can confirm.
Can we confirm if we use this schema with cloudcode, we get authoring assistance in cloudcode ?

go.mod
etefera: Removed.

Makefile
```diff
@@ -45,7 +45,8 @@ fmt:
 	go fmt ./...
 
 generate:
-	go install ./mdtogo
+	go install ./mdtogo 
+	(which swagger || go install github.com/go-swagger/go-swagger/cmd/swagger@latest)
```
droot, line 49: avoid using `latest` tag, pin it to the latest stable version.

pkg/api/kptfile/v1/types.go
droot: Updating this as per offline discussion. So we know cloudcode supports normal openapi schema along with the CRDs, so we are good.
And you also found out that controller-gen requires embedding the metadata from apimachinery. 

I think it's fine to move ahead with go-swagger for now.

droot (approved):
Looks good. Pl. merge is after pinning to latest stable version of go-swagger.

Makefile
etefera: Done, pinned to 0.27.0

PR #2395, opened 2021-07-13
 Normalize kpt capitalization in docs.

etefera:
Reenables markdown linting for the site, addressing capitalization in https://github.com/GoogleContainerTools/kpt/issues/1652


site/markdownlint.json
```diff
@@ -1,10 +1,12 @@
 {
-  "default": true,
+  "default": false,
```
droot, line 2: what does default means in this case ?

droot (approved):

site/markdownlint.json
etefera: It toggles the enforcement of a set of rules according to https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md
We're going to be piecewise enabling rules as they fit our style guide, so we're setting it to false here.

frankfarzan (approved):

site/installation/migration.md
```diff
@@ -188,7 +188,7 @@ information on the new structure of function results.
 
 ### `live`
 
-Kpt live in `v1.0` no longer uses an inventory object to track the grouping of
+kpt live in `v1.0` no longer uses an inventory object to track the grouping of
```
frankfarzan, line 191: In this case, we want `kpt live`

PR #2398, opened 2021-07-15
re-enable gatekeeper test

mengqiy:
The test was disabled due to unstable output.
It has been fixed.


droot (commented):
We need to update the expected result part for test to pass, otherwise the change is looking good.

droot (approved):

PR #2417, opened 2021-07-27
Fixes problem reading from stdin for live status command

seans3:
* Fixes `kpt live status` to read properly from `stdin` (only removes code).
* Fixes https://github.com/GoogleContainerTools/kpt/issues/2412
* Adds end-to-end tests for `kpt live apply/status/destroy` reading from `stdin`

### DIFF
--- a/e2e/live/end-to-end-test.sh
+++ b/e2e/live/end-to-end-test.sh
@@ -636,11 +636,36 @@ assertPodNotExists "pod-c" "rg-test-namespace"
 assertPodNotExists "pod-d" "rg-test-namespace"
 printResult
 
+# Test: Basic kpt live apply/status/destroy from stdin
+# 
+echo "Testing apply/status/destroy from stdin"
+echo "cat e2e/live/testdata/stdin-test/pods.yaml | kpt live apply -"
+cat e2e/live/testdata/stdin-test/pods.yaml | ${BIN_DIR}/kpt live apply - > $OUTPUT_DIR/status 2>&1
+assertContains "pod/pod-a created"
+assertContains "pod/pod-b created"
+assertContains "pod/pod-c created"
+assertContains "3 resource(s) applied. 3 created, 0 unchanged, 0 configured, 0 failed"
+printResult
+echo "cat e2e/live/testdata/stdin-test/pods.yaml | kpt live status -"
+cat e2e/live/testdata/stdin-test/pods.yaml | ${BIN_DIR}/kpt live status - > $OUTPUT_DIR/status 2>&1
+assertContains "namespace/stdin-test-namespace is Current: Resource is current"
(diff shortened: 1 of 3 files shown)


e2e/live/end-to-end-test.sh
```diff
@@ -636,11 +636,36 @@ assertPodNotExists "pod-c" "rg-test-namespace"
 assertPodNotExists "pod-d" "rg-test-namespace"
 printResult
 
+# Test: Basic kpt live apply/status/destroy from stdin
+# 
+echo "Testing apply/status/destroy from stdin"
```
mortent, line 641: I have created a framework for running e2e tests for live that is aligned with the framework we are already using for functions: https://github.com/GoogleContainerTools/kpt/tree/main/e2e and https://github.com/GoogleContainerTools/kpt/tree/main/e2e/testdata/live-apply. The bash script approach is difficult to debug and maintain, so can we add these tests using the framework instead? It almost certainly will require some changes to the framework to support reading from stdin.

seans3: How about in a future PR. 

I actually l don't think the bash script is difficult to debug and maintain. It took me only a few minutes to have a working test. Actually, the other framework takes a lot more work to get anything to work.

mortent: The challenge with the bash script isn't adding new tests, is being able to understand why something isn't working. It has dependencies between test-cases, it copies files in a way that is difficult to follow, and it doesn't actually tell us what is wrong when tests fail. Often they just fail with a message that some text string wasn't found. I agree a framework requires a bit more up-front knowledge, but it separates the test-cases and makes it easy to run them individually and it actually prints the full diff when the output isn't as expected.

I'm ok with doing it in a follow-up, but we should get it updated soon. Adding more test-cases to the script just means it is harder and more work to replace it.

mortent (approved):

e2e/live/end-to-end-test.sh
seans3: Most (if not all) of the dependencies between test-cases have been fixed. And adding a bit more code to tell us what went wrong is going to be a LOT less code than the ginkgo/gomega route. The gingko/gomega is significantly more difficult to understand what is being setup and what is happening, including lots of code inspecting low-level events. The bash script is much simpler with simple command lines. On whole I do not agree that the ginkgo/gomega framework is better in any dimension.

If we want to deprecate the bash script, let's just take it out of the build so we can still run it manually.

droot: @seans3 the framework that @mortent referred is not ginkgo/gomega based. Its a simple framework that lets you manage your e2e tests declaratively. We have followed the same for `fn` tests and it has been very useful.
I would recommend to check out the test cases for live that @mortent added recently and then compare the two approaches.

PR #2455, opened 2021-08-15
Show timestamp for pipeline execution duration

howardjohn:
This is helpful for debugging and understanding pipeline execution speed. Example output:
```
Package "grafana":
[RUNNING] "gcr.io/kpt-fn/set-namespace:v0.1"
[PASS] "gcr.io/kpt-fn/set-namespace:v0.1" in 4.2s
[RUNNING] "gcr.io/kpt-fn/set-namespace:v0.1"
[PASS] "gcr.io/kpt-fn/set-namespace:v0.1" in 2.9s
[RUNNING] "gcr.io/kpt-fn/set-namespace:v0.1"
[PASS] "gcr.io/kpt-fn/set-namespace:v0.1" in 2.7s

Successfully executed 3 function(s) in 1 package(s).
```

### DIFF
--- a/internal/fnruntime/runner.go
+++ b/internal/fnruntime/runner.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/GoogleContainerTools/kpt/internal/errors"
 	"github.com/GoogleContainerTools/kpt/internal/pkg"
@@ -111,10 +112,11 @@ func (fr *FunctionRunner) Filter(input []*yaml.RNode) (output []*yaml.RNode, err
 	if !fr.disableCLIOutput {
 		pr.Printf("[RUNNING] %q\n", fr.name)
 	}
+	t0 := time.Now()
 	output, err = fr.do(input)
 	if err != nil {
 		printOpt := printer.NewOpt()
-		pr.OptPrintf(printOpt, "[FAIL] %q\n", fr.name)
+		pr.OptPrintf(printOpt, "[FAIL] %q in %v\n", fr.name, time.Since(t0).Truncate(time.Millisecond*100))
 		printFnResult(fr.ctx, fr.fnResult, printOpt)
 		var fnErr *ExecError
 		if goerrors.As(err, &fnErr) {
@@ -124,7 +126,7 @@ func (fr *FunctionRunner) Filter(input []*yaml.RNode) (output []*yaml.RNode, err
 		return nil, err
 	}
 	if !fr.disableCLIOutput {
-		pr.Printf("[PASS] %q\n", fr.name)
+		pr.Printf("[PASS] %q in %v\n", fr.name, time.Since(t0).Truncate(time.Millisecond*100))
 		printFnResult(fr.ctx, fr.fnResult, printer.NewOpt())
 	}
 	return output, err
--- a/pkg/test/runner/runner.go
+++ b/pkg/test/runner/runner.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"testing"
 )
@@ -227,7 +228,7 @@ func (r *Runner) runFnEval() error {
 		}
 
 		// compare results
-		err = r.compareResult(i, fnErr, stdout, stderr, pkgPath, resultsDir)
+		err = r.compareResult(i, fnErr, stdout, sanitizeTimestamps(stderr), pkgPath, resultsDir)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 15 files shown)


phanimarupaka:
@howardjohn This is a backwards breaking change and if it should be done, this has to be an opt-in feature. Do you have an issue filed for this with your use-case mentioned ?

howardjohn:
Why is this a breaking change? Its a user-facing unstructured command line output. Are you expecting users to be parsing the result of this?

frankfarzan:
Unstructured human-consumed output of `fn` commands is not considered part of the API, so this is not a backwards breaking change (In contrast to https://kpt.dev/book/04-using-functions/03-function-results)

phanimarupaka:
Never mind. This is not a backwards breaking change. But seems like we need to refactor test framework to test this behavior.

howardjohn:
I will update the tests soon. Thanks!

mikebz:
Really cool addition to the pipeline output.  Thanks for the PR!

howardjohn:
OK looks like I got the tests passing. Had to do some regex which didn't feel great but wasn't too complex

internal/fnruntime/runner.go
```diff
@@ -111,10 +112,11 @@ func (fr *FunctionRunner) Filter(input []*yaml.RNode) (output []*yaml.RNode, err
 	if !fr.disableCLIOutput {
 		pr.Printf("[RUNNING] %q\n", fr.name)
 	}
+	t0 := time.Now()
 	output, err = fr.do(input)
 	if err != nil {
 		printOpt := printer.NewOpt()
-		pr.OptPrintf(printOpt, "[FAIL] %q\n", fr.name)
+		pr.OptPrintf(printOpt, "[FAIL] %q in %v\n", fr.name, time.Since(t0).Truncate(time.Millisecond*100))
```
droot, line 119: curious the reasoning for rounding to 100ms here ? I was playing with the default formatter %v for time, works well for this case https://play.golang.org/p/0ZcLEAwmuGb

internal/fnruntime/runner.go
```diff
@@ -124,7 +126,7 @@ func (fr *FunctionRunner) Filter(input []*yaml.RNode) (output []*yaml.RNode, err
 		return nil, err
 	}
 	if !fr.disableCLIOutput {
-		pr.Printf("[PASS] %q\n", fr.name)
+		pr.Printf("[PASS] %q in %v\n", fr.name, time.Since(t0).Truncate(time.Millisecond*100))
```
droot, line 129: nit: realized as reader `100*time.Millisecond` reads more natural :)

pkg/test/runner/runner.go
```diff
@@ -246,6 +247,18 @@ func (r *Runner) runFnEval() error {
...
 }
 
+// Match strings starting with [PASS] or [FAIL] and ending with " in N...". We capture the duration portion
+var timestampRegex = regexp.MustCompile(`\[(?:PASS|FAIL)].* in ([0-9].*)`)
+
+func sanitizeTimestamps(stderr string) string {
+	// Output will have non-deterministic output timestamps. We will replace these to static message for
+	// stable comparison in tests.
+	for _, m := range timestampRegex.FindAllStringSubmatch(stderr, -1) {
+		stderr = strings.ReplaceAll(stderr, m[1], "0s")
```
droot, line 257: This introduces friction in adding tests. Today, users just copy paste the output in the expected `stderr` section, not they need to replace timing with `0s` after copy-paste. Not a huge deal given the benefits, but don't have better solution in mind also.

/cc @mengqiy this will have ripple effects on function-catalog tests when we update the test runner in that repo.

droot (commented):
Thank you for the submission. Change looks good overall. I have minor comments/questions.

internal/fnruntime/runner.go
howardjohn: Looks great for artificial inputs, but real world inputs will be much worse. For example, https://play.golang.org/p/asojWEV2kBc you end up with things like `12.783198789s`. If we are getting execution down to a lot less than a second it would make sense to drop this to 1/10ms probably, but right now everything is in the seconds range. So I thought `4.9s` is sufficient enough. I don't mind `4.92s` or `4.923s` though - but `12.783198789s` is a bit excessive :slightly_smiling_face:

pkg/test/runner/runner.go
howardjohn: we could make `KPT_E2E_UPDATE_EXPECTED` (or a new var) auto update it if desired?

howardjohn: Or also sanitize the expected string as well. That may mean that there is a lot of churn in the golden file though

internal/fnruntime/runner.go
droot: > but real world inputs will be much worse.

Aah. Thats what I was missing. I think 100ms truncation is reasonable.

pkg/test/runner/runner.go
droot: I think adding a new var (KPT_E2E_UPDATE_EXPECTED_STDERR_TIMING) for auto-updating the expected stderr with `0ms` can work well. So auto-update behavior is to replace any timing info with `0s` or add `0s` if timing info is missing in the expected stderr (this can help minimize the ripple effect in the function-catalog-repo).

mengqiy: > this will have ripple effects on function-catalog tests when we update the test runner in that repo.

We are NOT testing presentation layer in the catalog. Instead we test it through the structured result API. So this change is not a problem for the catalog repo.

IIRC we have a few test in the kpt repo that are testing the presentation layer. They will be affected.

droot: ok, that's a fair point as well.

droot (approved):

PR #2472, opened 2021-08-29
Bump go module dependencies

howardjohn:
This is not fixing any concrete issues, but I have some upstream changes to some dependencies I was hoping to make and figured it would be good to get everything up to date before making those.


droot:
We have a few other projects that depends on kpt and they also use upstream k8s deps, so need to confirm version upgrade for k8s is fine with them.

PR #2479, opened 2021-09-04
pkg tree output.

mikebz:
just a quick addition of providing output to the comamnd.


mikebz:
nevermind, seems like no other commands provide output.

phanimarupaka (approved):

PR #2520, opened 2021-09-27
Ensured that property names are lowercase in kptfile schema

droot:
This PR ensures that the property names are lowercase in kptfile schema (yaml and json).

Kptfile schema is generated from the Go structs. This PR adds `json` tags to all the exported fields of kptfile Go structs.

Note to the reviewers: Note that the definition name still starts with uppercase. Haven't figured a way to configure that.

/cc @phanimarupaka


site/reference/schema/kptfile/kptfile.json
```diff
@@ -1,105 +1,125 @@
 {
-  "$schema": "http://json-schema.org/draft/2019-09/schema#",
-  "type": "object",
-  "anyOf": ["#/definitions/Kptfile"],
```
grmoon, line 4: I think we will still need this "anyOf" block in order for language server support to work properly

phanimarupaka: Yeah. That should be manual step before transferring the schema to language server. In the meanwhile we will try to figure out if there is automated way to a add this as part of generation.

phanimarupaka:
@grmoon Can you confirm the uppercase for the structs shouldn't matter.

site/reference/schema/kptfile/kptfile.json
grmoon, line 4: i believe we will still need all 3 of these fields in order for language server support to work without making manual modifications to the schema

phanimarupaka: @droot We can add this as part of make generate step, add these static fields automatically after generating schema.

grmoon: this can be done using jq, something similar to the following:

Screen Shot 2021-09-28 at 2 55 54 PM

grmoon:
> @grmoon Can you confirm the uppercase for the structs shouldn't matter.

confirming! so long as the properties of the `Kptfile` object definition are lowercase, it should work as expected

PR #2565, opened 2021-11-01
Update kpt book with depends-on annotation

mortent:
This updates the kpt book with an additional chapter for kpt live covering the depends-on functionality. It also adds a new example package that uses the functionality.

### DIFF
--- a/package-examples/wordpress-with-dependencies/deployment/deployment.yaml
+++ b/package-examples/wordpress-with-dependencies/deployment/deployment.yaml
@@ -1,0 +1,50 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: wordpress
+  namespace: default
+  labels:
+    app: wordpress
+  annotations:
+    config.kubernetes.io/depends-on: apps/namespaces/default/StatefulSet/wordpress-mysql
+spec:
+  selector:
+    matchLabels:
+      app: wordpress
+      tier: frontend
+  template:
+    metadata:
+      labels:
+        app: wordpress
+        tier: frontend
+    spec:
+      containers:
+        - name: wordpress
+          image: wordpress:4.8-apache
+          ports:
+            - name: wordpress
+              protocol: TCP
+              containerPort: 80
+          env:
+            - name: WORDPRESS_DB_HOST
+              value: wordpress-mysql
+            - name: WORDPRESS_DB_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: mysql-pass
+                  key: password
+  strategy:
+    type: Recreate
--- a/site/book/06-deploying-packages/03-handling-dependencies.md
+++ b/site/book/06-deploying-packages/03-handling-dependencies.md
@@ -1,0 +1,74 @@
+Sometimes resources within a package have dependencies that require
+one resource to be applied and reconciled before the next resource
+is applied. For example, a package that includes both Wordpress and
+MySQL might require that MySQL is running before Wordpress is started.
+
+In kpt, this is supported by declaring dependencies with the 
+`config.kubernetes.io/depends-on` annotation.
+
+If you look at the `wordpress-with-dependencies` package, a modified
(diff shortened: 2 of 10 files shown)


mikebz:
thanks for making the addition!

site/book/06-deploying-packages/03-handling-dependencies.md
```diff
@@ -0,0 +1,74 @@
+Sometimes resources within a package have dependencies that require
+one resource to be applied and reconciled before the next resource
+is applied. For example, a package that includes both Wordpress and
```
frankfarzan, line 3: wordsmith: `before another resource.`

site/book/06-deploying-packages/03-handling-dependencies.md
```diff
@@ -0,0 +1,74 @@
...
+MySQL might require that MySQL is running before Wordpress is started.
```
frankfarzan, line 4: You may want quality you mean by `Wordpress` and `Mysql` (i.e. a deployment not a package)

site/book/06-deploying-packages/03-handling-dependencies.md
```diff
@@ -0,0 +1,74 @@
...
+
+In kpt, this is supported by declaring dependencies with the 
+`config.kubernetes.io/depends-on` annotation.
+
+If you look at the `wordpress-with-dependencies` package, a modified
```
frankfarzan, line 9: Provide the `pkg get` command, and walk the user (At no point should the user be left not knowing what to do).

site/book/06-deploying-packages/03-handling-dependencies.md
```diff
@@ -0,0 +1,74 @@
...
+metadata:
+  name: wordpress
+  namespace: default
+  labels:
+    app: wordpress
+  annotations:
+    config.kubernetes.io/depends-on: apps/namespaces/default/StatefulSet/wordpress-mysql
+```
+
+The syntax for the resource references are `<group>/namespaces/<namespace>/<kind>/<name>`
```
frankfarzan, line 27: User bulletpoints and break to make this more readable.

site/book/06-deploying-packages/03-handling-dependencies.md
```diff
@@ -0,0 +1,74 @@
...
+for namespaced resources and `<group>/<kind>/<name>` for cluster-scoped resources.
```
frankfarzan, line 28: Does version not matter?

frankfarzan (changes requested):

package-examples/wordpress-with-dependencies/deployment/deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+# limitations under the License.
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: wordpress
+  namespace: default
+  labels:
+    app: wordpress
+  annotations:
+    config.kubernetes.io/depends-on: apps/namespaces/default/StatefulSet/wordpress-mysql
```
droot, line 22: Should `set-namespace` and `ensure-name-substring` be aware of these annotations ? 
To clarify the use-case, so I have a kpt pkg that have resources annotated with `depends-on` annotation. The pkg also uses functions that manipulates the namespace or name, should those function also update the `depends-on` annotation ?

/cc @mengqiy

site/book/06-deploying-packages/03-handling-dependencies.md
```diff
@@ -0,0 +1,74 @@
+Sometimes resources within a package have dependencies that require
+one resource to be applied and reconciled before the next resource
```
droot, line 2: s/reconciled/ready ? (does ready convey the intent better here ?)

Reconciled but failed will not go through ? right ?

Two things are happening with `depends-on`:
1. Ordering of resources during `apply`
2.  Waiting for a condition to be met on a resource (This part is not very clear here)

frankfarzan:
1. "Reconcile" is an established term in the book.
2. Please add a snippet for this feature in the chapter overview where we
highlight the major enhancements over kubectl:

https://kpt.dev/book/06-deploying-packages/

site/book/06-deploying-packages/03-handling-dependencies.md
mortent: Different versions all point to the same resource. So version doesn't matter here.

frankfarzan (approved):

package-examples/wordpress-with-dependencies/deployment/deployment.yaml
mortent: Yeah, that is a great point. We should see if we can add support for the depends-on annotation in the functions. I have created an issue for it: https://github.com/GoogleContainerTools/kpt/issues/2579

site/book/06-deploying-packages/03-handling-dependencies.md
mortent: Reconciled is the term we are using for this process, both in the book and elsewhere. It means the live state has been updated to reflect the desired state, so a resource that have failed the reconciliation process is not considered reconciled. I also want to avoid the term `Ready`, as it is a commonly used condition for k8s resources and means something slightly different.

PR #2593, opened 2021-11-17
fn render supports function binary executables

droot:
This PR adds support for executing binaries (as function) in `fn render`.

An example of kptfile that uses binary executable as declarative function:

```yaml
apiVersion: kpt.dev/v1
kind: Kptfile
metadata:
  name: app
pipeline:
  mutators:
    - exec: "sed -e 's/foo/bar/'"
```

`render` needs to be executed with `allow-exec` flag for it to run binary executables as functions.

```sh
$ kpt fn render --allow-exec
```

### Overloading the `image` field to support other runtimes where prefix of the image value can be used to derive the runtime as shown below:

```
apiVersion: kpt.dev/v1
kind: Kptfile
metadata:
  name: app
pipeline:
  mutators:
    - image: "exec: sed -e 's/foo/bar/'"

```

This is very confusing from UX perspective where `image` field could mean many things, so ruled it out.

### Using an explicit discriminator field `type` to determine different runtimes

```
apiVersion: kpt.dev/v1
kind: Kptfile
metadata:
  name: app
pipeline:
  mutators:
    - type: container
      image: gcr.io/kpt-fn/set-namespace:v0.1.1
    - type: exec
      exec: "sed -e `s/foo/bar`"
    - type: remote
      url: https://my-fn
```

The discriminator `type` increases the verbosity of the configuration. As we were exploring patterns for more runtimes, we realized there is a primary required field for these runtimes such as `exec`, `image`, `url` that can serve the purpose of discriminator field and decided to go with that.

### DIFF
--- a/pkg/api/kptfile/v1/types.go
+++ b/pkg/api/kptfile/v1/types.go
@@ -268,6 +268,10 @@ type Function struct {
 	//	image: set-labels
 	Image string `yaml:"image,omitempty" json:"image,omitempty"`
 
+	// Exec specifies the function binary executable.
+	// The executable needs to be in the $PATH.
+	Exec string `yaml:"exec,omitempty" json:"exec,omitempty"`
+
 	// `ConfigPath` specifies a slash-delimited relative path to a file in the current directory
 	// containing a KRM resource used as the function config. This resource is
 	// excluded when resolving 'sources', and as a result cannot be operated on
--- a/pkg/api/kptfile/v1/types.go
+++ b/pkg/api/kptfile/v1/types.go
@@ -269,7 +269,10 @@ type Function struct {
 	Image string `yaml:"image,omitempty" json:"image,omitempty"`
 
 	// Exec specifies the function binary executable.
-	// The executable needs to be in the $PATH.
+	// The executable can be fully qualified or it must exists in the $PATH e.g:
+	//
+	// 	 exec: set-namespace
+	// 	 exec: /usr/local/bin/my-custom-fn
 	Exec string `yaml:"exec,omitempty" json:"exec,omitempty"`
 
 	// `ConfigPath` specifies a slash-delimited relative path to a file in the current directory
--- a/pkg/api/kptfile/v1/types.go
+++ b/pkg/api/kptfile/v1/types.go
@@ -275,8 +275,6 @@ type Function struct {
 	// 	 exec: /usr/local/bin/my-custom-fn
 	Exec string `yaml:"exec,omitempty" json:"exec,omitempty"`
 
-	ExecArgs []string `yaml:"-" json:"-"`
-
 	// `ConfigPath` specifies a slash-delimited relative path to a file in the current directory
 	// containing a KRM resource used as the function config. This resource is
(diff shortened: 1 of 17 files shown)


mikebz:
very cool.  Looking forward to the results!

droot:
> very cool. Looking forward to the results!

I cleaned up this PR to make it production ready. Code is complete, now adding tests.

pkg/api/kptfile/v1/validation.go
```diff
@@ -67,14 +67,29 @@ func (p *Pipeline) validate(pkgPath types.UniquePath) error {
...
+		err := ValidateFunctionImageURL(f.Image)
+		if err != nil {
+			return &ValidateError{
+				Field:  fmt.Sprintf("pipeline.%s[%d].image", fnType, idx),
+				Value:  f.Image,
+				Reason: err.Error(),
+			}
 		}
 	}
+	// TODO(droot): validate the exec
```
mengqiy, line 92: How?

pkg/api/kptfile/v1/types.go
```diff
@@ -268,6 +268,15 @@ type Function struct {
 	//	image: set-labels
 	Image string `yaml:"image,omitempty" json:"image,omitempty"`
 
+	// Exec specifies the function binary executable.
```
mengqiy, line 271: Why don't we make it executable + args like `kpt fn eval --exec`?
If so, we can remove field `ExecArgs`.

pkg/api/kptfile/v1/validation.go
droot: Good point. Probably we can't or even if we can, there isn't much value.

pkg/api/kptfile/v1/types.go
droot: Actually I need to clarify. `Exec` is `executable + args` (see the tests), I am using ExecArgs for internal purposes (note the `-` tags for json and yaml). I will take another stab at cleaning it up to remove the confusion.

mengqiy (approved):

mikebz:
Really excited about this change, we need to think through the journey of how we can make the function download easier but I've been a big believer that the docker overhead at the level of string replace is too much.

mengqiy:
We should update our Kptfile schema at some point. It can be in a followup PR.

pkg/api/kptfile/v1/types.go
droot: Closing the loop on this. Did another pass today and was able to remove `ExecArgs` from kptfile definition. `exec` field supports exec-with-args as we do with `fn eval`.

Thanks for pointing out.

droot:
@phanimarupaka @mengqiy Pl. take a final look. Also added more description in the PR about the alternatives considered to complete the PR.

phanimarupaka (approved):

mengqiy (approved):

PatrickChenSe:
Looking forward to this change, need this to upgrade our kpt from 0.39

droot:
> Looking forward to this change, need this to upgrade our kpt from 0.39

Just merged the PR and will be available in the next release (most likely this week).

PR #2600, opened 2021-11-29
changed default image pull policy to if-not-present

droot:
`kpt fn` commands support command flag called `image-pull-policy` that determines image fetch behavior for KRM functions. Current default value is `AlwaysPull` which means `kpt` will always try to pull the image using `docker pull` even if the function image is present locally. In addition, current code also calls `docker inspect` to check if the image is present locally for `ifnotpresent`.

This PR does the following:
 - Changes the default value for `image-pull-policy` to `if-not-present` i.e. pull image only if it doesn't exists locally.
 - Removes the `docker pull` and `docker image inspect` calls instead relays the `image-pull-policy` flag to `docker run` command as is. This will eliminate two docker calls and thus, make the `fn render` and `fn eval` faster.
 - To support the above implementation, we had to filter out docker CLI command  (especially progress messages). This was needed because output from executing KRM function and `docker run` will be emitted on same `stderr` output stream.

### DIFF
--- a/e2e/testdata/fn-eval/missing-fn-image/.expected/config.yaml
+++ b/e2e/testdata/fn-eval/missing-fn-image/.expected/config.yaml
@@ -17,4 +17,4 @@ exitCode: 1
 image: gcr.io/kpt-fn/dne # non-existing image
 args:
   namespace: staging
-stdErr: 'Function image "gcr.io/kpt-fn/dne" doesn''t exist'
+stdErr: 'gcr.io/kpt-fn/dne:latest not found'
--- a/e2e/testdata/fn-render/fn-success-with-stderr/.expected/results.yaml
+++ b/e2e/testdata/fn-render/fn-success-with-stderr/.expected/results.yaml
@@ -5,6 +5,5 @@ metadata:
 exitCode: 0
 items:
   - image: gcr.io/kpt-fn/starlark:v0.2
-    stderr: |
-      function succeeded, reporting it on stderr
+    stderr: function succeeded, reporting it on stderr
     exitCode: 0
--- a/internal/fnruntime/container.go
+++ b/internal/fnruntime/container.go
@@ -232,8 +232,8 @@ func isdockerCLIoutput(s string) bool {
 		strings.Contains(s, ": Pulling from") ||
 		strings.Contains(s, ": Waiting") ||
 		strings.Contains(s, ": Pull complete") ||
-		strings.Contains(s, ": Digest: sha256") ||
-		strings.Contains(s, ": Status: Downloaded newer image") ||
+		strings.Contains(s, "Digest: sha256") ||
+		strings.Contains(s, "Status: Downloaded newer image") ||
 		strings.Contains(s, "Unable to find image") {
 		return true
 	}
--- a/internal/fnruntime/container.go
+++ b/internal/fnruntime/container.go
@@ -209,7 +209,7 @@ func (e *ContainerImageError) Error() string {
 
 // filterDockerCLIOutput filters out docker CLI messages
 // from the given buffer.
-func filterDockerCLIOutput(in *bytes.Buffer) string {
+func filterDockerCLIOutput(in io.Reader) string {
 	out := strings.Builder{}
 
 	s := bufio.NewScanner(in)
--- a/internal/fnruntime/container.go
+++ b/internal/fnruntime/container.go
@@ -195,18 +195,6 @@ func (e *ContainerImageError) Error() string {
 		e.Image, e.Output)
 }
 
-/*
-	"Unable to find image 'gcr.io/kpt-fn/starlark:v0.3' locally"
-    "v0.3: Pulling from kpt-fn/starlark"
-    "4e9f2cdf4387: Already exists"
-    "aafbf7df3ddf: Pulling fs layer"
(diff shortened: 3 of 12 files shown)


e2e/testdata/fn-eval/missing-fn-image/.expected/config.yaml
```diff
@@ -17,4 +17,4 @@ exitCode: 1
 image: gcr.io/kpt-fn/dne # non-existing image
 args:
   namespace: staging
-stdErr: 'Function image "gcr.io/kpt-fn/dne" doesn''t exist'
+stdErr: 'gcr.io/kpt-fn/dne:latest not found'
```
mengqiy, line 20: Why we have this change? I didn't find any changes related to this.

e2e/testdata/fn-render/fn-success-with-stderr/.expected/results.yaml
```diff
@@ -5,6 +5,5 @@ metadata:
 exitCode: 0
 items:
   - image: gcr.io/kpt-fn/starlark:v0.2
-    stderr: |
-      function succeeded, reporting it on stderr
+    stderr: function succeeded, reporting it on stderr
```
mengqiy, line 8: Is this change necessary?

mengqiy (commented):
Code changes LGTM

e2e/testdata/fn-eval/missing-fn-image/.expected/config.yaml
droot: Earlier we were calling `docker pull` explicitly and capturing the not found error (look at the deleted code), so now that's gone, we relay the message from `docker run` as is. That error is little verbose because it's coming from docker, but I didn't find it much value in wrapping it up because the verbose message from docker is pretty clear.

e2e/testdata/fn-render/fn-success-with-stderr/.expected/results.yaml
droot: This is needed because we are stripping away `\n` from the last line returned by the function when sanitizing the output from the function. It's because we are using `bufio.Scanner` to process the output. I might be able to work around it, but didn't think it matters. Do you think of it being a problem ?

mengqiy: I don't think that's a problem.

internal/fnruntime/container.go
```diff
@@ -250,3 +194,48 @@ func (e *ContainerImageError) Error() string {
 		"Error: Function image %q doesn't exist remotely. If you are developing new functions locally, you can choose to set the image pull policy to ifNotPresent or never.\n%v",
 		e.Image, e.Output)
 }
+
+// filterDockerCLIOutput filters out docker CLI messages
+// from the given buffer.
+func filterDockerCLIOutput(in io.Reader) string {
```
sdowell, line 200: What is the motivation for filtering the docker output from the buffer?

I haven't encountered this use case before, but I expected there would be a more elegant way to do this (e.g docker cli flag, I/O redirection, etc)

internal/fnruntime/container.go
```diff
@@ -250,3 +194,48 @@ func (e *ContainerImageError) Error() string {
...
+//  "aafbf7df3ddf: Pulling fs layer"
+//  "aafbf7df3ddf: Verifying Checksum"
+//  "aafbf7df3ddf: Download complete"
+//  "6b759ab96cb2: Waiting"
+//  "aafbf7df3ddf: Pull complete"
+//  "Digest: sha256:c347e28606fa1a608e8e02e03541a5a46e4a0152005df4a11e44f6c4ab1edd9a"
+//  "Status: Downloaded newer image for gcr.io/kpt-fn/starlark:v0.3"
+//
+func isdockerCLIoutput(s string) bool {
+	if strings.Contains(s, "Already exists") ||
```
sdowell, line 228: Minor nit: these could be made more strict by including the preceding colon
e.g. ": Already exists", ": Pulling fs layer", etc

sdowell:
Code changes LGTM

internal/fnruntime/container.go
droot: so kpt invokes the functions using cmd.Exec() the tricky thing is the output stream (stdout/stderr) is shared between the docker command as well as the function itself. And at the kpt level, we want to process these outputs differently. I did investigate if docker provides a simple mechanism to control its progress output, but unfortunately couldn't find one, so had to settle for this hack.

I thought a lot about the maintenance aspect of this hack. docker is a very stable CLI, so I don't expect a huge churn in this code and also, this filtering is only relevant in case of error so even if the filtering is not perfect, it isn't going to effect the user experience much. And the upside of removing the two commands (`docker pull` and `docker image inspect`)  outweighs the downside of this hack, so went with this.

internal/fnruntime/container.go
droot: Good point. Done.

mengqiy (commented):
LGTM
We can merge it after the fn catalog site part is done

internal/fnruntime/container.go
sdowell: Seems reasonable to me 🙂

sdowell (approved):
LGTM

PR #2613, opened 2021-12-08
Simplify and speed up "kpt live" e2e tests

mortent:
Simplifies the test config and schedules tests to avoid having to recreate the kind cluster (which is very slow).

### DIFF
--- a/e2e/testdata/live-apply/json-output/config.yaml
+++ b/e2e/testdata/live-apply/json-output/config.yaml
@@ -1,0 +1,39 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+parallel: true
+kptArgs:
+  - "--output=json"
+stdOut: |
+  {"eventType":"resourceApplied","group":"","kind":"ConfigMap","name":"cm","namespace":"json-output","operation":"Created","timestamp":"<TIMESTAMP>","type":"apply"}
+  {"eventType":"resourceApplied","group":"apps","kind":"Deployment","name":"third-nginx","namespace":"json-output","operation":"Created","timestamp":"<TIMESTAMP>","type":"apply"}
+  {"configuredCount":0,"count":2,"createdCount":2,"eventType":"completed","failedCount":0,"serverSideCount":0,"timestamp":"<TIMESTAMP>","type":"apply","unchangedCount":0}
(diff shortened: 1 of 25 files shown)


e2e/testdata/live-apply/apply-depends-on/config.yaml
```diff
@@ -12,20 +12,14 @@
...
-      {"eventType":"resourceApplied","group":"apps","kind":"Deployment","name":"second-nginx","namespace":"apply-depends-on","operation":"Created","timestamp":"<TIMESTAMP>","type":"apply"}
-      {"configuredCount":0,"count":1,"createdCount":1,"eventType":"completed","failedCount":0,"serverSideCount":0,"timestamp":"<TIMESTAMP>","type":"apply","unchangedCount":0}
+parallel: true
+
+stdOut: |
+  deployment.apps/first-nginx created
+  1 resource(s) applied. 1 created, 0 unchanged, 0 configured, 0 failed
+  deployment.apps/second-nginx created
+  1 resource(s) applied. 1 created, 0 unchanged, 0 configured, 0 failed
+
```
droot, line 22: any reason for taking out the json output ?

e2e/testdata/live-apply/json-output/config.yaml
```diff
@@ -0,0 +1,39 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+parallel: true
+kptArgs:
+  - "--output=json"
+stdOut: |
```
droot, line 18: ignore my previous comment. It looks like you are adding an explicit test for testing json-output.

pkg/test/live/config.go
```diff
@@ -26,33 +26,28 @@ type TestCaseConfig struct {
...
 	Inventory []InventoryEntry `yaml:"inventory,omitempty"`
 
-	// RequiresCleanCluster tells the test framework that a new cluster must
-	// be created for running this test.
-	RequiresCleanCluster bool `yaml:"requiresCleanCluster,omitempty"`
-
 	// PreinstallResourceGroup causes the test framework to verify that the
 	// ResourceGroup CRD is available in the cluster before running the test.
-	PreinstallResourceGroup bool `yaml:"preinstallResourceGroup,omitempty"`
+	NoResourceGroup bool `yaml:"noResourceGroup,omitempty"`
```
droot, line 41: nit: The comment is outdated.

droot (approved):
minor comments.

pkg/test/live/config.go
mortent: Fixed.

e2e/testdata/live-apply/json-output/config.yaml
mortent: Yeah, having the same test looking at multiple output formats turned out to be difficult to manage. So removed that functionality and instead have separate tests for alternative output formats.

yuwenma:
/lgtm

Thanks, @droot !

yuwenma:
> Simplifies the test config and schedules tests to avoid having to recreate the kind cluster (which is very slow).

really like

PR #2618, opened 2021-12-10
Use reconcile-timeout for live apply tests

mortent:
Updates the test runner to remove status events from the output before we compare the output to the expected results. This is because the status events does not have consistent ordering.


sdowell (approved):
LGTM

PR #2619, opened 2021-12-11
Add structs for ResourceGroup objects

rquitales:
This PR introduces the required types to move inventory information out from the Kptfile and into a separate resourcegroup.yaml resource file. This is one of the required steps in implementing features defined within the [kpt live apply from STDIN design doc](https://github.com/GoogleContainerTools/kpt/blob/main/docs/design-docs/01-live_Inventory-to-ResourceGroup.md).

Since the ResourceGroup CRD was hard coded previously, and ResourceGroup objects were always handled as unstructured objects, these type definitions did not exist prior to this PR.


rquitales:
@yuwenma

droot (approved):

pkg/api/resourcegroup/v1alpha1/types.go
```diff
@@ -0,0 +1,44 @@
...
+const (
+	RGFileName       = "resourcegroup.yaml"
+	RGFileKind       = "ResourceGroup"
+	RGFileGroup      = "kpt.dev"
+	RGFileVersion    = "v1alpha1"
+	RGFileAPIVersion = RGFileGroup + "/" + RGFileVersion
+)
+
+// TypeMeta is the TypeMeta for ResourceGroup instances.
+var TypeMeta = yaml.ResourceMeta{
```
yuwenma, line 33: TypeMeta could be misleading considering it is actually a `ResourceMeta` rather than a `TypeMeta`.  I'd suggest renaming to`DefaultMeta`

pkg/api/resourcegroup/v1alpha1/types.go
```diff
@@ -0,0 +1,44 @@
...
+	TypeMeta: yaml.TypeMeta{
+		APIVersion: RGFileAPIVersion,
+		Kind:       RGFileKind,
+	},
+}
+
+// ResourceGroup contains the inventory information about a package managed with kpt.
+// swagger:model resourcegroup
+type ResourceGroup struct {
+	yaml.ResourceMeta `yaml:",inline" json:",inline"`
```
yuwenma, line 43: Would `omit empty` be better?

pkg/api/resourcegroup/v1alpha1/types.go
```diff
@@ -0,0 +1,44 @@
...
+// TypeMeta is the TypeMeta for ResourceGroup instances.
+var TypeMeta = yaml.ResourceMeta{
+	TypeMeta: yaml.TypeMeta{
+		APIVersion: RGFileAPIVersion,
+		Kind:       RGFileKind,
+	},
+}
+
+// ResourceGroup contains the inventory information about a package managed with kpt.
+// swagger:model resourcegroup
```
yuwenma, line 41: just curious, what is the swagger:model needed for?

yuwenma (changes requested):

pkg/api/resourcegroup/v1alpha1/types.go
rquitales: I was replicating the naming convention used for the `Kptfile` types file, but using `DefaultMeta` makes more sense to me as well.

pkg/api/resourcegroup/v1alpha1/types.go
rquitales: The swagger marker is used to generate the [swagger schema](https://github.com/GoogleContainerTools/kpt/blob/3d90e0ed19bcc5759dc22105aec46e53a0e15715/Makefile#L47). This is mainly for use/reference by [kpt.dev](https://kpt.dev) under the references/schema section.

yuwenma: Got it. thanks.

yuwenma: sg.

yuwenma (approved):

PR #2627, opened 2021-12-20
Remove printers from third_party and use cli-utils

mortent:
This change removes the printers from third_party and instead uses the implementation in cli-utils directly. Maintaining the fork doesn't really provide much value and has proven difficult to maintain. The fork of the status command still remains, but the goal is to remove that too.

This introduces some changes in the output format:
- The printers will now expose the `WaitEvents` in the output, either as separate lines for the `events` and `json` printer, or as a separate column for the `table` output.
- The lines printing the summary will now only happen once for apply operations and zero or once for prune operations, rather than once for every apply/prune group.
- Printing of the status events are now only printed for the `events` and `json` output formats if the `--status-events` flag is provided.

Blocked by https://github.com/GoogleContainerTools/kpt/pull/2616


karlkfi:
Do we want to backport these changes to cli-utils printer's? Or are the ones in cli-utils "better"?

karlkfi:
Ignore my question. I misunderstood the change.

LGTM, but will need a kpt team member approval.

droot (approved):

mortent:
@karlkfi This change does lead to some changes in the kpt output format, most importantly that we no longer print the summary for every apply/prune/delete task, but rather combine it at the end. I think this is the right direction, as we do want to rely less on the flattened graph.
This change also changes the output for dry-run to use `(preview)` instead of `(dry-run)`. I don't think we should merge this with that change, and I have a PR to simplify the output in cli-utils (https://github.com/kubernetes-sigs/cli-utils/pull/501). I think this change improves the output and it means we don't need to use the `preview` term in the kpt output.

PR #2636, opened 2022-01-04
Add annotation reference docs

karlkfi:
This adds docs for the following annotations:

- `config.kubernetes.io/depends-on`
- `config.kubernetes.io/apply-time-mutation`
- `config.kubernetes.io/local-config`

It also lists the following internal annotations as reserved:

- `internal.config.kubernetes.io/path`
- `internal.config.kubernetes.io/index`
- `config.kubernetes.io/merge-source`


droot (approved):
Thanks Karl. This is excellent. It's good to have a place for source of truths for annotations.

karlkfi:
Updated schema section to look more like the K8s API reference docs.

karlkfi:
Are same-page anchor links not allowed?

```
Starting check.
Checking http://localhost:3000/reference/annotations/depends-on/
Checking http://localhost:3000/reference/annotations/apply-time-mutation/
[samePage]	🚧	#ObjectReference [x1]
xargs: sh: exited with status 255; aborting
make: *** [Makefile:115: site-check] Error 124
Error: Process completed with exit code 2.
```

site/reference/annotations/README.md
```diff
@@ -0,0 +1,29 @@
+# Annotations Reference
+
+<!--mdtogo:Short
```
mortent, line 3: It is unclear to me how the strings produced by these directives will be used. For the actual cli commands, we reference them in the cobra commands, but it is not clear where we would reference this content.
I think we can just remove them.

site/reference/annotations/apply-time-mutation/README.md
```diff
@@ -0,0 +1,235 @@
...
+depends on the chosen implimentation. In this case, kpt uses the
+[ajson](https://github.com/spyzhov/ajson) library. For details about what
+language features are supported, see the
+[json-path-comparison table](https://cburgmer.github.io/json-path-comparison/).
+
+### Example
+
+<!--mdtogo:Examples-->
+
+In this example, pod-b depends on pod-a with two substitutions that replace
```
mortent, line 117: The examples here are somewhat more elaborate than the examples for the cli commands and feels almost more like the content of the kpt book. Should we consider making this more like the other examples and reference the book for a more complete description?

site/reference/annotations/local-config/README.md
```diff
@@ -0,0 +1,100 @@
...
+### Behavior
+
+Resources with the `local-config` annotation set to `true` will not be applied
+to the cluster when using `kpt live apply`.
+
+### Example
+
+<!--mdtogo:Examples-->
+
+In this example, the ConfigMap `cm-a` is local and not applied.
```
mortent, line 42: I think we usually use backticks around resource types, so this should be `ConfigMap`.

site/reference/annotations/apply-time-mutation/README.md
karlkfi: I wasn't sure how to add apply-time-mutation to the book. And it definitely needs a reference page to specify the full schema.

It sounds like you'd rather have the examples in the book. If so, what should the apply-time-mutation page be called?

FWIW, many of the CLI command examples are actually executable using mdrip for testing. I wasn't sure if we needed that for the annotations, but if we do, it needs a kind cluster to run against.

site/reference/annotations/README.md
karlkfi: I mostly just copied what was in the CLI reference. It's not clear to me what the mdtogo tags are for.

karlkfi:
Fixed the same page errors by making the links relative to the root of the site.

site/reference/annotations/README.md
karlkfi: removed

site/reference/annotations/apply-time-mutation/README.md
mortent: Agree we need a reference page for this and other features/concepts that are too detailed for the book and doesn't fit easily into the cli reference structure. I think this approach looks pretty good for that. I think there are other concepts that we should describe in some more detail outside of the book. Update strategies and function selectors comes to mind.

I think the question is just whether we should leave these things out of the book and then just keep the examples in the reference pages. Since both the depends-on and the apply-time mutations are somewhat advanced features, maybe we don't need to cover it in the book.

For now, I'm ok with keeping this as is and also leave the depends-on chapter in the book. We should probably also reference the apply-time mutations reference page from the book, so anyone following the book can find it.

What do you think @droot ?

I don't think we need to look at running mdrip for these examples in this PR. Automatic verification of the docs is useful, but we should separate look into how to best do it.

karlkfi: How about this: 
1. we merge this PR as-is
2. if a customer asks for the apply-time-mutation in the book, or expresses confusion not finding it there, we can add that later.

mortent: Sounds good

karlkfi:
This is ready to merge.

PR #2645, opened 2022-01-12
Update reference docs and book for kpt live

mortent:
Update the kpt book and reference docs:
* Add the new `status-events` flag to the reference docs
* Update the output examples to reflect the updated behavior when no reconcile or prune timeout is provided.
* Update the output examples to reflect the less verbose reconcile output when the `status-events` flag is not used.


site/book/06-deploying-packages/02-applying-a-package.md
```diff
@@ -58,14 +70,15 @@ would see that 6 resources would be created:
...
 $ kpt live apply wordpress --dry-run
-service/wordpress created (dry-run)
-service/wordpress-mysql created (dry-run)
-deployment.apps/wordpress created (dry-run)
-deployment.apps/wordpress-mysql created (dry-run)
-persistentvolumeclaim/mysql-pv-claim created (dry-run)
-persistentvolumeclaim/wp-pv-claim created (dry-run)
-6 resource(s) applied. 6 created, 0 unchanged, 0 configured, 0 failed (dry-run)
-0 resource(s) pruned, 0 skipped, 0 failed (dry-run)
+Dry-run strategy: client
```
droot, line 73: I am thinking if there is a good way to indicate the steps are running in dry-run mode. I could imagine myself mid-way freaking out if I *actually* applied or it's dry-run and I will have to scroll back to look at the command args.

The previous layout with `(dry-run` suffix wasn't terrible. What's the reasoning behind this change ?

site/reference/cli/live/apply/README.md
```diff
@@ -88,18 +88,24 @@ PKG_PATH | -:
...
-  giving up. If this flag is not set, kpt live apply will not wait for
-  resources to reconcile.
+  giving up. If this flag is not set, kpt live apply will wait until
+  interrupted.
 
 --server-side:
   Perform the apply operation server-side rather than client-side.
   Default value is false (client-side).
+
+--status-events:
```
droot, line 104: is it a boolean ? If yes, we should include the default value (true or false)

if it is boolean, we can probably rename it to `--enable-status-events` or `--{display|show}-status-events` ?

droot (commented):
Looks good to me. Few questions.

site/book/06-deploying-packages/02-applying-a-package.md
mortent: There have been some discussions around this. The previous output format is just very verbose, as the output on every line is identical, and in fact only prints the value of the flag provided through the flag. This also aligns the different output formats, as only the events output format had the dry-run suffix.

site/reference/cli/live/apply/README.md
mortent: I have a PR to change the name to `--show-status-events`: https://github.com/GoogleContainerTools/kpt/pull/2648.

I have updated the docs with the new name and included the default value.

droot (approved):

PR #2647, opened 2022-01-13
fix: check correct docker version is installed

bharathkkb:
Previously if docker client was less than [20.10.0](https://docs.docker.com/engine/release-notes/#client-7), kpt would error out with ` "unknown flag: --pull"`. This improves the error message by validating that docker version is >= `20.10.0` before proceeding with any fn executions.


droot (approved):
Thanks @bharathkkb for the PR.

PR #2649, opened 2022-01-13
Add 2-page rationale for kpt

bgrant0607:
Starting point for an explanation of kpt. Once merged, this can be moved around and modified as needed.


bgrant0607:
I am probably not going to have time to set up a dev environment to run make site-run-server any time soon, so if this can't work without regenerating the site, I may need someone else to pick this up.

droot:
> I am probably not going to have time to set up a dev environment to run make site-run-server any time soon, so if this can't work without regenerating the site, I may need someone else to pick this up.

It's safe to merge this since it is not linked from anywhere yet. I will do quick test in my dev setup and also link it up from the home-page.

bgrant0607:
Thanks! We may also need to figure out how to crop the svg downloaded from the Google doc, depending on how it renders on the site. It might be as easy as editing the bounding box numbers.

PR #2666, opened 2022-01-26
oci-support feature branch

loudej:
oci-support rebased on main and squash merged for pull request.

See design docs for usage patterns. Primarily:

* Accepts `oci://image:tag` and `oci://image@sha256:digest` arguments in commands where a git location would be expected.
* Sub-package path is delimited by double-slash, e.g. `oci://image//subdir:tag`
* Adds `oci` support to `upstream` and `upstreamLock` in Kptfile
* Two new commands, `kpt pkg pull [DIR] --origin oci://...` and `kpt pkg push [DIR]`


martinmaly:
Thank you!

PR #2667, opened 2022-01-26
Implement Direct Package Fetch

martinmaly:
Expose a simple package manipulation API.


pkg/api/kptlib/clone.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+	"strings"
+
+	internalpkg "github.com/GoogleContainerTools/kpt/internal/pkg"
+	kptfilev1 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// TODO: Accept a virtual filesystem or other package abstraction
+func UpdateUpstreamFromGit(kptfileContents string, name string, lock kptfilev1.GitLock) (string, error) {
```
droot, line 27: Looks like a better place for this method will be`pkg/kptfile/kptfileutil` pkg.

droot (changes requested):

pkg/api/kptlib/clone.go
martinmaly: Thanks! I moved the file inside Porch for now as we finalize the locations for interfaces.

martinmaly:
Updated the code. Please take a look.

droot (approved):

PR #2671, opened 2022-01-26
Enable Running Locally Without Docker

martinmaly:


justinsb:
Not entirely sure what happened here - I thought it was working on the mac?  `make` isn't working on linux either right now, if that's any consolation!

IMO: If we change configurations for Mac, we should do so for Linux also.  We should also figure out how this should be used (e.g. do I now need to open 3 terminal windows?)

PR #2672, opened 2022-01-27
Fix Porch Build

martinmaly:
Location of ClonerUsingGitExec changed in the OCI branch.

### DIFF
--- a/.github/workflows/porch.yml
+++ b/.github/workflows/porch.yml
@@ -1,0 +1,43 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


.github/workflows/porch.yml
```diff
@@ -0,0 +1,43 @@
+# Copyright 2022 Google LLC
```
justinsb, line 1: Does github run the workflows from the target branch (or do we need to restrict to the porch branch?)

justinsb:
One question, but lgtm

/approve
/lgtm

.github/workflows/porch.yml
martinmaly: From what I can tell it runs them in the target branch.

PR #2674, opened 2022-01-27
Upgrade go-containerregistry to v0.8.0

martinmaly:
Upgrade go-containerregistry (cherry pick from oci-support).
Tidy go modules.
Ignore go modules in .build directory.


internal/util/man/man_test.go
```diff
@@ -155,7 +155,7 @@ Java server ConfigMap
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: java\-config
```
justinsb, line 158: Is this change related?

justinsb:
I should show you my scripts that I use for separating out PRs, and one question, but lgtm

/approve
/lgtm

internal/util/man/man_test.go
martinmaly: This relates to the upgrade of go-md2man to v2.0.1.
I was not able to figure out how to prevent go from upgrading it so I went all in and fixed the test. @droot reviewed this in oci branch (#2669)

PR #2682, opened 2022-01-27
Ensure release license file exists

sdowell:
* ci: ensure existence of lib.zip for releases
ensure existence of the lib.zip file for downstream releases. This is
done by creating an empty zip file and updating it with mozilla_repos if
mozilla_repos is not empty.

* ci: use array for mozilla_repos
Update mozilla_repos to use an array rather than a string with space
delimited elements. This is intended to provide a more straightforward
type for storing a list of elements as well as provide more explicit
word splitting in line with SC2086.

I am open to feedback for alternatives to an empty zip file, e.g. the base zip file contains a README or a dummy go file

### DIFF
--- a/scripts/docs/create-licenses/README.md
+++ b/scripts/docs/create-licenses/README.md
@@ -1,0 +1,5 @@
+# create-licenses
+
+This file is intended to be included with the `libs.zip` file that is
+distributed with each `kpt` release. The `libs.zip` file contains a packaging of
+source files necessary to satisfy the Mozilla license.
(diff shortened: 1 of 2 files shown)


scripts/docs/create-licenses/README.md
```diff
@@ -0,0 +1,5 @@
+# create-licenses
```
droot, line 1: Does this need license :)

droot (approved):

scripts/docs/create-licenses/README.md
sdowell: I assumed the README file does not need a license - does it? If so, which license should it have?

PR #2684, opened 2022-01-28
Render and Eval via Simple Memory FS

martinmaly:
For now the filesystem is inserted at lower level; In future changes
the mutation Apply will accept the filesystem abstraction to avoid
back-and-forth between filesystem and PackageResources.


porch/engine/pkg/engine/render.go
```diff
@@ -16,48 +16,80 @@ package engine
...
 			Image:     "render",
 			ConfigMap: map[string]string{},
 		},
 	}, nil
 }
+
+// TODO: Implement filesystem abstraction directly rather than on top of PackageResources
+func writeResources(fs filesys.FileSystem, resources repository.PackageResources) error {
+	for k, v := range resources.Contents {
+		fs.MkdirAll(path.Dir(k))
```
justinsb, line 68: Probably should check the error retval

justinsb (approved):
One nit, but lgtm

droot (approved):

porch/engine/pkg/engine/render.go
martinmaly: done! thanks!

PR #2688, opened 2022-01-29
Kpt location parsing

loudej:
`location.ParseReference` returning `location.Reference` for CLI would replace internal git and oci argument parsing in a way that's callable by porch or other clients. 

In the case that the caller already has structured data, where the location type and individual properties are known, the `location.Git{...}` and `location.Oci{...}` should be created directly, rather than creating the string notation to parsing.

The common interface under createable struct types, with a parse to go from a string and options to the common interface, is based on the pattern from gcrane pkg's `name.ParseReference(name, opts...)`.

The ability to mutate/lock references abstractly is intended to be a step towards eliminating methods from the internal remote.Updater and remote.Origin utility interface. The pattern there is taken from the `io.fs` approach where package functions like `mutate.Lock(Reference, string) => (ReferenceLock, error)` make it easy to call optional methods that are not supported on every kind of Reference.

### DIFF
--- a/demos/demo-location/main.go
+++ b/demos/demo-location/main.go
@@ -1,3 +1,17 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package main
 
 import (
--- a/demos/demo-location/main.go
+++ b/demos/demo-location/main.go
@@ -156,9 +156,8 @@ type CustomLocation struct {
 }
 
 type CustomLocationLock struct {
-	WhereItIs            string
-	LabelOrVersionString string
-	UniqueIDString       string
+	CustomLocation
+	UniqueIDString string
 }
 
 // compile-time check that duck types are correct
@@ -170,6 +169,14 @@ var _ mutate.LockSetter = CustomLocation{}
 // string when reference appears in console and log messages
 func (ref CustomLocation) String() string {
 	return fmt.Sprint(" WhereItIs:", ref.WhereItIs, " LabelOrVersionString:", ref.LabelOrVersionString)
+}
+
+func (ref CustomLocation) Type() string {
+	return "custom"
+}
+
+func (ref CustomLocation) Validate() error {
+	return nil
 }
 
 // string when reference appears in console and log messages
@@ -190,8 +197,10 @@ func (ref CustomLocation) SetIdentifier(labelOrVersion string) (location.Referen
 // depending on location, the exact meaning may be commit-id, image-digest, url-query parameter, exact resource name, etc.
 func (ref CustomLocation) SetLock(uniqueID string) (location.ReferenceLock, error) {
 	return CustomLocationLock{
-		WhereItIs:            ref.WhereItIs,
-		LabelOrVersionString: ref.LabelOrVersionString,
-		UniqueIDString:       uniqueID,
+		CustomLocation: CustomLocation{
+			WhereItIs:            ref.WhereItIs,
+			LabelOrVersionString: ref.LabelOrVersionString,
+		},
+		UniqueIDString: uniqueID,
 	}, nil
 }
--- a/demos/demo-location/main.go
+++ b/demos/demo-location/main.go
@@ -28,30 +28,16 @@ import (
 func main() {
 	ctx := context.Background()
 
-	opts := []location.Option{
+	opts := location.Options(
 		location.WithContext(ctx),
-		location.WithParser(func(value string) (location.Reference, error) {
-			parts := strings.SplitN(value, ":", 3)
-			if len(parts) == 3 && parts[0] == "custom" {
-				return CustomLocation{
-					WhereItIs:            parts[1],
-					LabelOrVersionString: parts[2],
-				}, nil
-			}
-			return nil, nil
-		}),
-		location.WithGit(),
-		location.WithOci(),
-		location.WithDir(),
-	}
-
-	optsStdin := []location.Option{
-		location.WithStdin(os.Stdin),
-	}
-
-	optsStdout := []location.Option{
-		location.WithStdout(os.Stdout),
-	}
+		location.WithParsers(
+			location.StdioParser,
+			CustomParser,
+			location.GitParser,
+			location.OciParser,
+			location.DirParser,
+		),
+	)
 
 	fmt.Println("- parsing argument to location")
 	fmt.Println()
@@ -61,7 +47,7 @@ func main() {
 		"oci://us-docker.pkg.dev/my-project-id/my-repo-name/my-blueprint:draft",
 		"example",
 		"sha256:9f6ca9562c5e7bd8bb53d736a2869adc27529eb202996dfefb804ec2c95237ba",
-		opts...,
+		opts,
 	)
 
 	example(
@@ -69,7 +55,7 @@ func main() {
 		"https://github.com/GoogleCloudPlatform/blueprints.git/catalog/gke@gke-blueprint-v0.4.0",
 		"main",
 		"2b8afca2ef0662cf5ea39c797832ac9c5ea67c7e",
-		opts...,
+		opts,
 	)
 
 	example(
@@ -77,7 +63,7 @@ func main() {
 		"path/to/dir",
 		"qa",
 		"",
-		opts...,
+		opts,
 	)
 
 	example(
@@ -85,7 +71,7 @@ func main() {
 		"custom:where:which",
 		"another",
 		"98175",
-		opts...,
+		opts,
 	)
 
 	// stdin and stdout options are added on individual calls to parse, because
@@ -95,7 +81,8 @@ func main() {
 		"-",
 		"",
 		"",
-		append(optsStdin, opts...)...,
+		location.WithStdin(os.Stdin),
+		opts,
 	)
 
 	example(
@@ -103,7 +90,8 @@ func main() {
 		"-",
 		"",
 		"",
-		append(optsStdout, opts...)...,
+		location.WithStdout(os.Stdout),
+		opts,
 	)
 
 	fmt.Println("- creating locations directly")
@@ -146,7 +134,6 @@ func main() {
 	locked, _ := mutate.Lock(updated, "98510723450981325098375013")
 	fmt.Printf("locked: %v\n", locked)
 	fmt.Println()
-
 }
 
 func example(caption string, arg string, identifier string, lock string, opts ...location.Option) {
@@ -207,6 +194,21 @@ var _ location.Reference = CustomLocation{}
 var _ location.ReferenceLock = CustomLocationLock{}
 var _ mutate.IdentifierSetter = CustomLocation{}
 var _ mutate.LockSetter = CustomLocation{}
+
+var CustomParser = location.NewParser(
+	[]string{
+		"Custom locations format is 'custom:[WHERE]:[LABEL-OR-VERSION]'",
+	},
+	func(parse *location.Parse) {
+		parts := strings.SplitN(parse.Value, ":", 3)
+		if len(parts) == 3 && parts[0] == "custom" {
+			parse.Result(CustomLocation{
+				WhereItIs:            parts[1],
+				LabelOrVersionString: parts[2],
+			})
+		}
+	},
+)
 
 // string when reference appears in console and log messages
 func (ref CustomLocation) String() string {
--- a/demos/demo-location/main.go
+++ b/demos/demo-location/main.go
@@ -158,7 +158,7 @@ type CustomLocation struct {
 type CustomLocationLock struct {
 	WhereItIs            string
 	LabelOrVersionString string
-	UniqueIdString       string
+	UniqueIDString       string
 }
 
 // compile-time check that duck types are correct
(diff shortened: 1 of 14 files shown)


pkg/location/reference.go
```diff
@@ -0,0 +1,13 @@
+package location
+
+import "fmt"
+
+type Reference interface {
+	fmt.Stringer
+	Type() string
+	Validate() error
```
martinmaly, line 8: Is Validate needed given that the underlying type can ensure that invalid instance cannot exist?

Or are there cases where otherwise valid reference moves through invalid state(s)?

pkg/location/reference.go
```diff
@@ -0,0 +1,13 @@
...
+}
+
+type ReferenceLock interface {
+	Reference
```
martinmaly, line 12: Since we have SetLock interface, I wonder if a counterpart here would be:

`Lock() string`

it may be nice to glean some useful information from the interface

pkg/location/type_dir.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+type Dir struct {
+	Directory string
+}
+
+var _ Reference = Dir{}
+
+func parseDir(location string, opt options) (Reference, error) {
+	if fs.ValidPath(location) {
```
martinmaly, line 15: does this operate in the context of local filesystem, or is this purely stateless validation analyzing only the input string?

pkg/location/type_dir.go
```diff
@@ -0,0 +1,33 @@
+package location
+
+import (
+	"fmt"
+	"io/fs"
+)
+
+type Dir struct {
+	Directory string
```
martinmaly, line 9: with FileSystem abstraction, Directory alone is insufficient to identify a package.

We may need Dir that relies on local file system and one that is paired with FileSystem instance?

pkg/location/type_git.go
```diff
@@ -0,0 +1,123 @@
...
+func newGit(location string, opt options) (Git, error) {
+	// args[1] is "" for commands that do not require an output path
+	gitTarget, err := parse.GitParseArgs(opt.ctx, []string{location, ""})
+	var zero parse.GitTarget
+	if err != nil || gitTarget == zero {
+		return Git{}, err
+	}
+
+	dir := gitTarget.Directory
+	if strings.HasPrefix(dir, "/") {
```
martinmaly, line 51: There is probably some filepath method (or set thereof) that will make sure this works on non-Unix (if it is to be used as part of the CLI interface?

pkg/location/type_git.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func parseGit(location string, opt options) (Reference, error) {
+	git, gitErr := newGit(location, opt)
+	var zero Git
+	if gitErr == nil && git != zero {
+		return git, nil
+	}
+
+	// TODO - figure out which gitErr must be returned, and which simply mean "it's not a git path"
```
martinmaly, line 72: I had similar question reading the parse protocol (first to recognize path wins) and non-recognition is signaled by returning nil; What if there is parser later in the parser list that will succeed?

pkg/location/reference.go
```diff
@@ -0,0 +1,13 @@
+package location
+
+import "fmt"
+
+type Reference interface {
+	fmt.Stringer
```
martinmaly, line 6: Should each reference be expressible as a URI?

file://
oci://
https://github.com/...

Most certainly seem to be; if not all, maybe we need an interface that can go back and forth from single-string (URI) representation?

pkg/location/type_git.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return Git{
+		Repo:      gitTarget.Repo,
+		Directory: dir,
+		Ref:       gitTarget.Ref,
+	}, nil
+}
+
+func parseGit(location string, opt options) (Reference, error) {
```
martinmaly, line 65: this may be unneeded (since newGit parses?)

pkg/location/type_git.go
```diff
@@ -0,0 +1,123 @@
...
+		return errors.E(op, errors.MissingParam, fmt.Errorf("must specify directory"))
+	}
+	return nil
+}
+
+func (ref Git) GetDefaultDirectoryName() (string, error) {
+	repo := ref.Repo
+	repo = strings.TrimSuffix(repo, "/")
+	repo = strings.TrimSuffix(repo, ".git")
+	return path.Base(path.Join(path.Clean(repo), path.Clean(ref.Directory))), nil
```
martinmaly, line 107: this feels strange that default directory name would (IIUC) read:

`https://github.com/GoogleContainerTools/kpt/examples/foo`

pkg/location/defaultdir.go
```diff
@@ -0,0 +1,14 @@
...
+
+type DirectoryNameDefaulter interface {
+	GetDefaultDirectoryName() (string, error)
+}
+
+func DefaultDirectoryName(ref Reference) (string, error) {
+	if ref, ok := ref.(DirectoryNameDefaulter); ok {
+		return ref.GetDefaultDirectoryName()
+	}
+	return "", fmt.Errorf("default directory not supported for reference: %v", ref)
```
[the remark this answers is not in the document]
loudej: Good idea, especially for the places which support the method it's essentially the final path segment. If there was a possible way to fail - it should have already been returned by Parse or Validate.

demos/demo-location/main.go
```diff
@@ -0,0 +1,234 @@
+package main
```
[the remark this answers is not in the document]
loudej: Done, thanks!

pkg/location/mutate/mutations.go
```diff
@@ -0,0 +1,29 @@
+package mutate
+
+import (
+	"fmt"
+
+	"github.com/GoogleContainerTools/kpt/pkg/location"
+)
+
+type IdentifierSetter interface {
```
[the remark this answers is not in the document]
loudej: And possibly comments on public surface area in general :) I'll put more text around DefaultDirectory / Interface / Lock to start

pkg/location/type_git.go
loudej: In this case I believe it's safe to assume this value where it appears in git uri and kptfile fields is always canonically '/'

That said the `type_dir.go` may not be as lucky - looking at how [DIR] args are pre-processed in kpt - a call to `filepath.Clean` should probably be the first step of parsing

pkg/location/type_git.go
loudej: That is a problem. Right now it's a requirement that parsers are added in the right order. That is, put stdio "-" first because it's most specific and "dir" last because it'll accept most values.

But that is a good example, when either "-" and "dir" can be used as an arg, you've more or less lost the ability to read a dir named "-". Well, unless you use "./-" as the path. TBH, that's also similar to the "go cmd X" problem where X has to start with "./" so it's not parsed as a module package name.

An improvement could be to have a priority order, but that breaks down when multiple custom types (that are not aware of each other) that can be solved by having the caller provide the priority for each.. location.WithGit(1), location.WithOci(2) location.WithDir(0) etc... Which would probably seem awful to use...

pkg/location/type_git.go
loudej: It's a subtle difference - but the different return value means parseGit can be used as a parser type - but the newGit call can be returned directly from NewGit without assert the type back to the struct.

TBH, parseGit could also be moved to an inline func in the option. That could be cleaner....

pkg/location/type_git.go
loudej: Nearly - though the final `path.Base` around the whole thing means "just the last segment"

Altogether it's the last segment of the dir in the repo --- or the last segment of the repo if the package is in the repo's root.

The line is copy-pasted from the existing parse code --- though it could be broken into smaller steps where the local variable names provide better clues to the meaning.

pkg/location/mutate/mutations.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+func Identifier(ref location.Reference, identifier string) (location.Reference, error) {
+	if ref, ok := ref.(IdentifierSetter); ok {
+		return ref.SetIdentifier(identifier)
+	}
+	return nil, fmt.Errorf("changing identifier not supported for reference: %v", ref)
+}
+
+type LockSetter interface {
```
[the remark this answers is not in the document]
loudej: Great question! Looking ahead a bit (and based on how it would be used for cli-oci feature) a Reference and ReferenceLock would be used in a few places: 

There's the Reference that's a [DIR] argument or "-" stdio stream. That's location info for direct read/write of package content.

Once package is read in, there's the Upstream Reference/ReferenceLock representation of the upstream/upstreamLock. Those are always driven by the kptfile contents and will be the remote git/oci type.

Those are the simple cases. The more interesting one like you mentioned is like what happens in the oci cli commands that were added "kpt pkg pull/push". There an Origin ReferenceLock added to the kptfile as it's pulled (it's purely client-side data and is removed again when it's pushed).

So in that case the Reference of Dir{} or stream will get you package content, and the Upstream will be where it was cloned from originally, and the Origin ReferenceLock will be where it was pulled from. (Similar to the origin of git folder, really.)

Last case worth mentioning --- if you have a location Reference that's directly to git/oci (and not to a pulled dir copy) that's where you'd expect that remote Reference to source the content directly. In that case Upstream would come from the kptfile of course and be just that data, but if you ask for the Origin of _that_ result there you'll expect the ReferenceLock to be the commit/digest on top of the remote Reference that was pulled.

TBH, the idea of Origin was Brian's idea earlier. He mentioned that if pkg pull added origin data, it would give a default push location. And also that saving the digest of the origin at pull time could be used for optimistic concurrency purposes at push time as well.

pkg/location/type_git.go
loudej: Okay - I pushed a refactor that does a few things.

* there's now a single WithParsers(...) option that takes the ordered `...parsers`. So the order of options is no longer an issue.

* you can now add StdioParser upfront all of the time, so that list no longer needs to vary. Instead, it is only active when you also have the WithStdin and WithStdout option. e.g. You'll have WithStdin option only on the cli args which are input and accept "-".

* the return value is now a result struct {ref, err, helpText} with code comments for semantics. `ref` and `err` are always final, and all of the helptext is returned if no parser returns ref or err.

So `err` means I **know** this is my format, and it's broken. and `helpText` means this _might_ be my format, and here's text to add to the warning if a later parser doesn't lock onto it unambiguously

pkg/location/reference.go
loudej: Ah, that is true, isn't it? Though TBH the oci:// one isn't strictly canonical... It is the same format helm uses for oci, but really it's pulled out of thin air.

Come to think of it - the Stringer isn't also strictly round tripping with parse. Instead it's more like a "%v" property-fmt style dump. Maybe it should be treated more like the inverse of parse?

I suppose a question would be - you'd probably still want to have non-file:// dir for CLI args --- if we did add a URI is there a place you're imagining that where URI-valued-string would be the standard ref notation?

martinmaly (approved):
we can start here and iterate. I do worry about the canonical or roundtripping representations and would prefer if the interfaces were a bit more opaque, but we can learn how this work in practice and iterate from here.

pkg/location/reference.go
loudej: Oh! Yes, agreed. In the next PR where some of this is used under `kpt pkg get` there are methods that are aiming to be in the kptfileutil package. It's possible there could be a better home for those, but that's where methods like `kptfileutil.UpdateUpstreamLockFromGit` are today so seemed like a good starting point.

```go
ref, err := kptfileutils.NewReferenceFromUpstream(kf)
refLock, err := kptfileutils.NewReferenceLockFromUpstreamLock(kf)
kf.Upstream, err = kptfileutils.NewUpstreamFromReference(ref)
kf.UpstreamLock, err = kptfileutils.NewUpstreamLockFromReferenceLock(refLock)
```

pkg/location/defaultdir.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2021 Google LLC
```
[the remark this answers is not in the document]
loudej: Oh, yes. I'll send a 2nd PR for that rather than fold into this (I think it will hit a _lot_ more than the files already in this change)

pkg/location/options.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	return opt
+}
+
+type parser func(value string, opt options) (Reference, error)
+
+// Option is a functional option for location parsing.
+type Option func(*options)
+
+// WithDefaultTag sets the default tag that will be used if one is not provided.
```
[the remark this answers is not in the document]
loudej: Sorry - missed this question.

In the Git parser there is one place where (if there is no branch/ref name given in the arg) it will reach out to the git repository to find the default branch name.

Because the git parsing calls through to the existing implementation, passing the context as an option enabled that behavior to be unchanged.

pkg/location/parse_test.go
```diff
@@ -0,0 +1,185 @@
...
+			},
+			wantErr: false,
+		},
+		{
+			name: "StreamLocationNotExpected",
+			args: args{
+				location: "-",
+			},
+			want: Dir{
+				Directory: "-",
```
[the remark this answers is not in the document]
loudej: that is true. there is an ambiguity added by "-" as stdio which is a valid dir name. 

some points that kind of mitigate the concern, 
* It's the code calling ParseReference that can decide which formats are accepted for that particular arg. So it's not a blanket collision, instead it's only where a CLI makes a choice, for example, that a `[DIR]` destination arg is actually a `[DIR] | -` filesystem or stdio arg.
* For those cases, you can point to `kubectl apply -f -` to read from stdin instead of a folder name. It doesn't avoid the ambiguity, but at least it establishes precedent.
* Finally the person calling the CLI can also use an "/absolute" or "./relative" path to guarantee it matches the `[DIR]` format only. So that works for a hardcoded `"./-"` dir if needed, or `"./${RELPATH}"` if it's in a script and the path isn't known in advance...

pkg/location/reference.go
loudej: Well, I suppose a short answer would be that `Validate()` was added to be called from places in kpt commands that call validate on arguments...

Beyond that though - the types like `location.Dir{}` being simple struct with public fields for convenience does mean the calling code can initialize or modify the values as needed.

So in the end that would be the case where invalid an ref can be created pretty easily, and calling Validate at the point of consumption would be the guarantee.

PR #2694, opened 2022-01-31
Create dockerfile / makefile / manifest for pushing controllers to cluster

justinsb:
This lets us start to run these on the server, instead of just locally.


martinmaly (changes requested):
For now holding off while this is marked as WIP. Please let me know when ready for review.

martinmaly (approved):

PR #2697, opened 2022-01-31
Introduce applyset abstraction

justinsb:
Replacing our very primitive apply logic with something better using
server-side apply.


porch/controllers/remoterootsync/pkg/applyset/comparison.go
```diff
@@ -0,0 +1,98 @@
...
+
+// objectsAreEquals returns true if the two objects are equal.
+// If there is ambiguity / failure, we return false.
+// This method is used to determine whether it is safe to reuse the cached state when replacing objects.
+func objectsAreEqual(l, r *unstructured.Unstructured) bool {
+	if l == nil {
+		return r == nil
+	}
+	if r == nil {
+		return l == nil
```
martinmaly, line 30: nit: this statement will never return true (if it did, line 27 would have)

porch/controllers/remoterootsync/pkg/applyset/comparison.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog/v2"
+)
+
+// objectsAreEquals returns true if the two objects are equal.
+// If there is ambiguity / failure, we return false.
+// This method is used to determine whether it is safe to reuse the cached state when replacing objects.
+func objectsAreEqual(l, r *unstructured.Unstructured) bool {
```
martinmaly, line 25: would reflect.DeepEqual be suitable so we don't need to reimplement the logic?

porch/controllers/remoterootsync/pkg/applyset/comparison.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		return lv == rv
+	case int64:
+		rv, ok := r.(int64)
+		if !ok {
+			return false
+		}
+		return lv == rv
+	default:
+		// TODO: Drop to Warningf
```
martinmaly, line 94: Tricky. not sure if we have guarantee that we'll be seeing only int64 types for numbers, that seems like a common case to encounter depending on yaml parser(s). See another question re: reflect.DeepEqual which may help. Otherwise I'd probably downgrade to error so user input doesn't crash the controller.

porch/controllers/remoterootsync/pkg/applyset/tracker.go
```diff
@@ -0,0 +1,101 @@
...
+		tracker := &l.items[i]
+		key := computeKey(tracker.desired)
+		existingTrackers[key] = tracker
+	}
+
+	newList := &objectTrackerList{}
+
+	for _, obj := range objects {
+		key := computeKey(obj)
+		// TODO: Detect duplicate keys?
```
martinmaly, line 74: how do we detect a resource that should be deleted (is in the cluster now, was previously deployed but the new version of package no longer includes it)?

porch/controllers/remoterootsync/pkg/applyset/applyset.go
```diff
@@ -0,0 +1,216 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/klog/v2"
+)
+
+// ApplySet is a set of objects that we want to apply to the cluster.
```
martinmaly, line 31: how are deletions handled?

porch/controllers/remoterootsync/pkg/applyset/applyset.go
```diff
@@ -0,0 +1,216 @@
...
+	if r.total != (r.applySuccessCount + r.applyFailCount) {
+		klog.Warningf("consistency error (apply counts): %#v", r)
+	} else if r.total != (r.healthyCount + r.unhealthyCount) {
+		// This "invariant" only holds when all objects could be applied
+		klog.Warningf("consistency error (healthy counts): %#v", r)
+	}
+}
+
+// applySkipped records that the apply of an object was skipped because it was already applied.
+func (r *ApplyResults) applySkipped(gvk schema.GroupVersionKind, nn types.NamespacedName) {
```
martinmaly, line 123: will future PRs utilize the gvk and nn parameters?

martinmaly (commented):
Couple of questions re: possible code reuse and plans for delete.

porch/controllers/remoterootsync/pkg/controllers/remoterootsyncset/remoterootsync_controller.go
```diff
@@ -100,17 +101,23 @@ func (r *RemoteRootSyncSetReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, nil
 	}
 
+	var result ctrl.Result
+
 	var patchErrs []error
 	for _, clusterRef := range subject.Spec.ClusterRefs {
-		err := r.applyToClusterRef(ctx, &subject, clusterRef)
+		results, err := r.applyToClusterRef(ctx, &subject, clusterRef)
```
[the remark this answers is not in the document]
justinsb: The referenced type could be a KCC ContainerCluster, or a KCC ConfigControllerInstance, or anything "cluster-like"

porch/controllers/remoterootsync/pkg/applyset/health.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"encoding/json"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog/v2"
+)
+
+// isHealthy reports whether the object should be considered "healthy"
+// TODO: Replace with kstatus library
```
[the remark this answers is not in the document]
justinsb: I _hope_ there's nothing blocking it, I'm just trying to keep the commits small.

porch/controllers/remoterootsync/pkg/applyset/comparison.go
justinsb: Good call - it's probably good enough - and we end up likely pulling this out into an interface a few commits down anyway!  Removed & replaced with reflect.DeepEqual

porch/controllers/remoterootsync/pkg/applyset/tracker.go
justinsb: Pruning is coming later.  I'll add a TODO!

porch/controllers/remoterootsync/pkg/applyset/applyset.go
justinsb: They aren't yet, I added a TODO.  The applyset has state but the state is in-memory and doesn't survive restarts.  For pruning, we (likely) need to track in the cluster, and there are a few approaches out there for doing so.

porch/controllers/remoterootsync/pkg/applyset/applyset.go
justinsb: That's the intent, yes.  We're able to expose per-object status, and often that will be exposed to users, particularly on failure.  We don't do so yet, so right now we're just keeping the count.

I started passing this because pruning gets tricky - we don't always know the version for example, depending on how we are tracking ownership - so I wanted to track what information we actually had.

martinmaly (approved):

PR #2699, opened 2022-01-31
Tidy Go Modules

martinmaly:
- Fix Go Modules
- Run Tidy in Porch Workflow


justinsb (approved):

justinsb:
lgtm, though I don't _think_ it suffices to run `go mod tidy` to verify that no tidying is needed in CI.  But we can do that in another PR!

martinmaly:
agreed. At this point I suspect the build will fail if our references are off. Will iterate ...

PR #2700, opened 2022-01-31
Define a FunctionEvaluator gRPC Interface

martinmaly:
- Add install-protoc.sh script
- Define a FunctionEvaluator interface


porch/func/evaluator/evaluator.proto
```diff
@@ -0,0 +1,127 @@
...
+}
+
+message EvaluateFunctionRequest {
+  // Serialized ResourceList (https://kpt.dev/reference/schema/resource-list/)
+  bytes resource_list = 1;
+
+  // kpt image identifying the function to evaluate
+  string image = 2;
+
+  oneof config {
```
loudej, line 37: IIRC in the `ResourceList.config.kubernetes.io/v1` - the `functionConfig` in `results` out are already part of the resource_list?

Wondering because, if the existing code on either end of the conversation will already have that marshaling, if it stays in the `bytes resource_list` you wouldn't need to have code to map between the proto and ResourceList types.

martinmaly: You are right! I got carried away. This will simplify it considerably. Thank you!

martinmaly: Fixed. Thank you! ptal

porch/func/server/server.go
```diff
@@ -0,0 +1,34 @@
...
+
+	pb "github.com/GoogleContainerTools/kpt/porch/func/evaluator"
+	"google.golang.org/grpc"
+)
+
+func main() {
+	lis, _ := net.Listen("tcp", fmt.Sprint(":9445"))
+	server := grpc.NewServer()
+	pb.RegisterFunctionEvaluatorServer(server, &evaluator{})
+	server.Serve(lis)
```
justinsb, line 29: We probably should check/log errors, but we can do that later...

justinsb (approved):

porch/hack/install-protoc.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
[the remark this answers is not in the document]
martinmaly: Will do. Since the results of protoc generation are checked in we shouldn't need to do this often and hopefully can remove it when containerized.

porch/func/server/server.go
martinmaly: ack, will do.

PR #2701, opened 2022-01-31
WIP: check file headers

justinsb:
- Makefile: Add CI for porch controllers
- Check headers in CI


martinmaly (approved):
Not sure if this is WIP or ready but simple enough change to gladly re-review if it is WIP after all.

justinsb:
Thanks - I think we did this via the `fix-all` approach, so I'm going to close.

PR #2702, opened 2022-01-31
Create fix-all and verify-fix-all

justinsb:
We run all the formatters / header makers etc.

In CI, we run them and make sure that the git checkout is still clean.

Ideally we would also run generate, but it is very slow, and it also
requires GOPATH to be set.  We can do that in CI later.

### DIFF
--- a/porch/hack/verify-fix-all.sh
+++ b/porch/hack/verify-fix-all.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 The Kubernetes Authors.
+#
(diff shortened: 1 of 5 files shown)


porch/hack/verify-fix-all.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 The Kubernetes Authors.
```
martinmaly, line 3: what governs whether we use K8s authors vs. Google LLC?

martinmaly (changes requested):

martinmaly (approved):

PR #2705, opened 2022-02-01
Enable migration to store inventory in separate ResourceGroup file

rquitales:
This is a PR that implements the kpt live migrate functionality defined within the kpt live apply from STDIN design doc.

Additions:
* Migrate from ConfigMap to resourcegroup.yaml
* Migrate from Kptfile inventory to resourcegroup.yaml
* Test cases
Note: This is a stacked PR on top of #2615, so some code changes may appear duplicated...

The files changed in this PR are:
```
internal/cmdmigrate/migratecmd.go
internal/cmdmigrate/migratecmd_test.go
```

### DIFF
--- a/internal/cmdmigrate/migratecmd_test.go
+++ b/internal/cmdmigrate/migratecmd_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/GoogleContainerTools/kpt/internal/pkg"
 	"github.com/GoogleContainerTools/kpt/internal/printer/fake"
+	rgfilev1alpha1 "github.com/GoogleContainerTools/kpt/pkg/api/resourcegroup/v1alpha1"
 	"github.com/stretchr/testify/assert"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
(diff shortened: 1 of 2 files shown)


mortent:
It is not clear to me how this addition will work with the current migration. Is it a separate migration from the existing one, or do a migration from now on simply put the inventory into the separate file rather than update the Kptfile? If so, why do we need the two different implementations?

It also seems like a lot of duplicated code between the `Run` and `runLiveMigrateWithRGFile` function. Is there a way to avoid some of this?

internal/cmdmigrate/migratecmd_test.go
```diff
@@ -177,6 +178,118 @@ func TestKptMigrate_updateKptfile(t *testing.T) {
...
+		dryRun        bool
+		isError       bool
+	}{
+		"Missing Kptfile is an error": {
+			kptfile:    "",
+			rgFilename: "resourcegroup.yaml",
+			dryRun:     false,
+			isError:    true,
+		},
+		"Kptfile with existing inventory will create ResourceGroup": {
```
mortent, line 195: So in this situation, do we only move the inventory information from the Kptfile into the rg file without attempting any update of the live state? At first it seems like we maybe wouldn't care about the live state, but it might be that a previous run failed midway through, so we should verify the live state and complete any in-process migrations.

rquitales: Synced up on this offline. For existing Kptfiles that contain an inventory, the cluster's inventory should be tracking all resources as a ResourceGroup CR on the cluster. If there is drift between the cluster state and the local config, the differences will be reconciled when the user next runs `kpt live apply`.
For a ConfigMap migration, this new migrate CUJ still follows the existing migration flow as from CM to Kptfile. The only difference is that we write the inventory to `resourcegroup.yaml` instead of the local Kptfile.

rquitales:
> It is not clear to me how this addition will work with the current migration. Is it a separate migration from the existing one, or do a migration from now on simply put the inventory into the separate file rather than update the Kptfile? If so, why do we need the two different implementations?
> 
> It also seems like a lot of duplicated code between the `Run` and `runLiveMigrateWithRGFile` function. Is there a way to avoid some of this?

The logic in `runLiveMigrateWithRGFile` will completely replace the logic in `Run` when we expose this functionality externally. We essentially have "dead code" currently so we don't break existing functionality for kpt releases, but can also introduce the new migrate CUJ logic for PR review and integration testing.

mortent (approved):

PR #2709, opened 2022-02-01
Enable apply/destroy to use inventory in separate ResourceGroup file 

rquitales:
This is a PR that implements the `kpt live apply` and `destroy` functionality defined within the kpt live apply from STDIN design doc.

Additions:
* Apply and destroy using inventory stored in resourcegroup.yaml
* Test cases
Note: This is a stacked PR on top of #2705, so some code changes may appear duplicated...

#### Files changed:
```
internal/testutil/pkgbuilder/builder.go
pkg/live/load.go
pkg/live/load_test.go
pkg/live/rgpath.go
pkg/live/rgstream.go
```
Refactoring/splitting code:
```
internal/pkg/pkg.go
```

Small changes (2 lines):
```
internal/cmdapply/cmdapply.go
internal/cmddestroy/cmddestroy.go
internal/cmdmigrate/migratecmd.go
thirdparty/cli-utils/status/cmdstatus.go
```

### DIFF
--- a/pkg/live/load.go
+++ b/pkg/live/load.go
@@ -24,6 +24,7 @@ import (
 	"github.com/GoogleContainerTools/kpt/internal/pkg"
 	"github.com/GoogleContainerTools/kpt/internal/util/strings"
 	kptfilev1 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
+	rgfilev1alpha1 "github.com/GoogleContainerTools/kpt/pkg/api/resourcegroup/v1alpha1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/kubectl/pkg/cmd/util"
 	"sigs.k8s.io/cli-utils/pkg/common"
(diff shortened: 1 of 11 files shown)


pkg/live/load.go
```diff
@@ -58,26 +67,52 @@ func (e *MultipleInventoryInfoError) Error() string {
...
+		if err != nil {
+			return nil, kptfilev1.Inventory{}, err
+		}
+
+		diskInv, err := readInvInfoFromDisk(cwd, rgfile)
+		if err != nil {
+			return nil, kptfilev1.Inventory{}, err
+		}
+
+		if diskInv.Name != "" && invInfo.Name != "" {
```
mortent, line 103: Can we check this against empty structs instead of only checking the `Name` property?

pkg/live/rgpath.go
```diff
@@ -64,6 +65,11 @@ func (r *ResourceGroupPathManifestReader) Read() ([]*unstructured.Unstructured,
 		if err != nil {
 			return objs, err
 		}
+
+		// Skip if current file is a resourcegroup resource.
```
mortent, line 69: This comment would be much more useful if it explained why.

mortent (commented):
Mostly looks good. Just a few comments.

pkg/live/load.go
rquitales: Have updated the code to check if the inventory is valid or not, rather than only using the `Name` property. Currently this check enforces `Name` and `Namespace` to not be empty. Also added new test cases to check for this!

pkg/live/rgpath.go
rquitales: Done. Replaced comment with:

> Skip if current file is a ResourceGroup resource. We do not want to apply/delete any ResourceGroup CRs when we run any `kpt live` commands on a package. Instead, we have specific logic in place for handling ResourceGroups in the live cluster.

mortent (approved):

PR #2710, opened 2022-02-01
Integrate porch with renderer interface

phanimarupaka:
This PR integrates porch renderer with kpt renderer.

### DIFF
--- a/pkg/fn/eval.go
+++ b/pkg/fn/eval.go
@@ -17,7 +17,6 @@ package fn
 import (
 	"context"
 
-	"github.com/GoogleContainerTools/kpt/internal/types"
 	fnresult "github.com/GoogleContainerTools/kpt/pkg/api/fnresult/v1"
 	v1 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
@@ -26,10 +25,10 @@ import (
 
 type EvalOptions struct {
 	// FnResultList stores the result of the function evaluation
-	FnResultList fnresult.ResultList
+	FnResultList *fnresult.ResultList
 }
 
 type Evaluator interface {
 	Eval(ctx context.Context, pkg filesys.FileSystem, fn v1.Function, opts EvalOptions) error
-	NewRunner(ctx context.Context, fn *v1.Function, pkgPath types.UniquePath, opts EvalOptions) (kio.Filter, error)
+	NewRunner(ctx context.Context, fn *v1.Function, opts EvalOptions) (kio.Filter, error)
 }
--- a/pkg/fn/eval.go
+++ b/pkg/fn/eval.go
@@ -24,8 +24,8 @@ import (
 )
 
 type EvalOptions struct {
-	// FnResultList stores the result of the function evaluation
-	FnResultList *fnresult.ResultList
+	// ResultList stores the result of the function evaluation
+	ResultList *fnresult.ResultList
 }
 
 type Evaluator interface {
--- a/pkg/fn/eval.go
+++ b/pkg/fn/eval.go
@@ -17,13 +17,19 @@ package fn
 import (
 	"context"
 
+	"github.com/GoogleContainerTools/kpt/internal/types"
+	fnresult "github.com/GoogleContainerTools/kpt/pkg/api/fnresult/v1"
 	v1 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
+	"sigs.k8s.io/kustomize/kyaml/kio"
 )
 
 type EvalOptions struct {
+	// FnResultList stores the result of the function evaluation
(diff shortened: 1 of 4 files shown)


pkg/fn/eval.go
```diff
@@ -17,13 +17,19 @@ package fn
...
+	"sigs.k8s.io/kustomize/kyaml/kio"
 )
 
 type EvalOptions struct {
+	// FnResultList stores the result of the function evaluation
+	FnResultList fnresult.ResultList
 }
 
 type Evaluator interface {
 	Eval(ctx context.Context, pkg filesys.FileSystem, fn v1.Function, opts EvalOptions) error
```
phanimarupaka, line 33: @martinmaly Do we need this ? Can't we just leverage NewRunner function and kio.Filter interface ? Refer to fnruntime package.

martinmaly: it may work. What does pkgPath do? If we use kio.Filter there should be no need for paths.

pkg/fn/eval.go
```diff
@@ -17,13 +17,19 @@ package fn
...
+	"github.com/GoogleContainerTools/kpt/internal/types"
+	fnresult "github.com/GoogleContainerTools/kpt/pkg/api/fnresult/v1"
 	v1 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
+	"sigs.k8s.io/kustomize/kyaml/kio"
 )
 
 type EvalOptions struct {
+	// FnResultList stores the result of the function evaluation
+	FnResultList fnresult.ResultList
```
martinmaly, line 29: this would need to be pointer (or opts below need to be pointer)

pkg/fn/eval.go
```diff
@@ -17,13 +17,19 @@ package fn
...
 }
 
 type Evaluator interface {
 	Eval(ctx context.Context, pkg filesys.FileSystem, fn v1.Function, opts EvalOptions) error
+	NewRunner(ctx context.Context, fn *v1.Function, pkgPath types.UniquePath, opts EvalOptions) (kio.Filter, error)
```
martinmaly, line 34: what about something like:

```
type Evaluator /* maybe we should rename this */ interface {
  NewProcessor(ctx context.Context, fn v1.Function, opts) (ResourceListProcessor, error)
}
```

https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/fn/framework/framework.go#L67

so we don't need to split up the pieces of data (ResultList in opts, RNodes separately too)

pkg/fn/eval.go
phanimarupaka: updated.

pkg/fn/eval.go
phanimarupaka: pkgPath can be used for printing status messages. Removing it for now, you can reintroduce it based on the implementation.

pkg/fn/eval.go
```diff
@@ -17,13 +17,18 @@ package fn
...
 )
 
 type EvalOptions struct {
+	// FnResultList stores the result of the function evaluation
+	FnResultList *fnresult.ResultList
 }
 
 type Evaluator interface {
 	Eval(ctx context.Context, pkg filesys.FileSystem, fn v1.Function, opts EvalOptions) error
+	NewRunner(ctx context.Context, fn *v1.Function, opts EvalOptions) (kio.Filter, error)
```
martinmaly, line 33: Couple suggestions:

NewRunner --> NewFilter
EvalOptions --> FilterOptions

pkg/fn/eval.go
```diff
@@ -17,13 +17,18 @@ package fn
...
 
+	fnresult "github.com/GoogleContainerTools/kpt/pkg/api/fnresult/v1"
 	v1 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
+	"sigs.k8s.io/kustomize/kyaml/kio"
 )
 
 type EvalOptions struct {
+	// FnResultList stores the result of the function evaluation
+	FnResultList *fnresult.ResultList
```
martinmaly, line 28: I'd propose FnResultList --> ResultList

martinmaly (changes requested):
couple nitpicks

pkg/fn/eval.go
phanimarupaka: FilterOptions seems generic. NewRunner is consistent with fnruntime.

pkg/fn/eval.go
phanimarupaka: updated.

pkg/fn/eval.go
martinmaly: If we are going to stick with the fnruntime nomenclature, let's switch over completely:

Evaluator --> FunctionRunner
eval.go --> runner.go
EvalOptions --> RunnerOptions

wdyt?

I still wish NewRunner returned Runner and not Filter, but oh well :)

martinmaly (approved):

PR #2715, opened 2022-02-02
create gcloudconfig.yaml to store local default value

yuwenma:


internal/util/get/local.go
```diff
@@ -0,0 +1,118 @@
...
+		fmt.Println("`Organization` or `Folder` not found")
+	}
+
+	return map[string]string{
+		"name":      pkgName,
+		"namespace": projectID,
+		"projectID": projectID,
+		"zone":      zone,
+		"region":    region,
+		"domain":    account,
```
bgrant0607, line 65: This is the whole account, right, not just the domain? We need to split out the domain. Better here than in every function.

morgante: I'm actually not sure. Domain isn't used frequently while some functions will need access to the user.

bgrant0607: I wouldn't expect the user's account to be used to provision resources in production.

I have a specific purpose in mind for domain, which is to use it for org-wide groups like gke-security-groups@
https://github.com/GoogleCloudPlatform/blueprints/blob/main/catalog/gke/cluster/cluster.yaml#L39
and the foundational groups in https://github.com/GoogleCloudPlatform/blueprints/tree/main/catalog/iam-foundation

We can add additional fields to the context in the future as needed.

liamfallon:
GCE PR

PR #2716, opened 2022-02-02
Oci pkg get

loudej:
Draft PR to preview shifting the `kpt pkg get` from for oci-support onto new virtual fs types

Adds "content" layer above "location" layer to make it simple to work with location as FileSystem and Reader abstractly.

The change from remote.Upstream interface to location.Reference in get command object also updates some tests - the difference is small but looks like an improvement.

### DIFF
--- a/demos/demo-content/main.go
+++ b/demos/demo-content/main.go
@@ -15,23 +15,42 @@
 package main
 
 import (
+	"context"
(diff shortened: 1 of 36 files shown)


loudej:
The PR branch is moving to different fork

PR #2719, opened 2022-02-03
Use apply-setters directly

martinmaly:
For now use the apply-setters in the engine. Next we'll update function runner
to run some functions in-process rather than execing them.

### DIFF
--- a/porch/engine/pkg/kpt/render_test.go
+++ b/porch/engine/pkg/kpt/render_test.go
@@ -1,0 +1,192 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kpt
+
+import (
+	"testing"
+
+	"github.com/GoogleContainerTools/kpt/internal/printer/fake"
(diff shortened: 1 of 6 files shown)


porch/engine/pkg/kpt/render_test.go
```diff
@@ -0,0 +1,192 @@
...
+data:
+  name: updated-bucket-name
+  namespace: updated-namespace
+  project-id: updated-project-id
+  storage-class: updated-storage-class
+`
+)
+
+func TestRenderWithFunctionConfigFile(t *testing.T) {
+	t.Skip("kpt renderer does not correctly construct function config")
```
martinmaly, line 71: @phanimarupaka for now I am skipping this test; when we fix the renderer handling of configPath we can remove this line.

porch/engine/pkg/kpt/render_test.go
```diff
@@ -0,0 +1,192 @@
...
+    enabled: false
+`
+)
+
+func TestRenderWithFunctionConfigInline(t *testing.T) {
+	fs := &memfs{}
+	if err := fs.MkdirAll("/inline-bucket"); err != nil {
+		t.Errorf("Failed MkdirAll: %v", err)
+	}
+	for k, v := range map[string]string{
```
justinsb, line 165: Not necessarily for this PR, but a pattern I like is to have a directory with inputs under testdata for each test, and simply read the directory and in this case load it into `fs`

porch/engine/pkg/kpt/render_test.go
```diff
@@ -0,0 +1,192 @@
...
+		PkgPath:    "/inline-bucket",
+		Runner:     &runner{},
+		FileSystem: fs,
+	}
+
+	if err := r.Execute(fake.CtxWithDefaultPrinter()); err != nil {
+		t.Errorf("Render failed: %v", err)
+	}
+
+	got, err := fs.ReadFile("/inline-bucket/bucket.yaml")
```
justinsb, line 184: And then I would probably put the expected data into a different directory :-)

justinsb (approved):

PR #2721, opened 2022-02-03
Create push-images target to push all our images

justinsb:
Also a few tweaks to make the porch image build faster.

### DIFF
--- a/porch/hack/Dockerfile
+++ b/porch/hack/Dockerfile
@@ -40,11 +40,26 @@ RUN echo "Downloading engine modules ..." \
 RUN echo "Downloading repository modules ..." \
  && cd porch/repository && go mod download
 
+# Prebuild some library dependencies to warm the cache
+RUN cd porch/apiserver; go build -v \
+  google.golang.org/grpc \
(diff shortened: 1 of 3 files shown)


porch/hack/Dockerfile
```diff
@@ -40,11 +40,26 @@ RUN echo "Downloading engine modules ..." \
...
+COPY porch/apiserver porch/apiserver
+COPY porch/controllers porch/controllers
+COPY porch/engine porch/engine
+COPY porch/repository porch/repository
+
 
-RUN cd porch/apiserver; go build -o /porch ./cmd/porch
+RUN cd porch/apiserver; go build -v -o /porch ./cmd/porch
 
 FROM debian:bullseye
```
justinsb, line 64: @martinmaly would you be OK if I switched this to gcr.io/distroless/static ?   I don't think we depend on e.g. any binaries in the image

martinmaly: yes, let's do that.

martinmaly (approved):

PR #2727, opened 2022-02-03
Upgrade kpt dependencies

martinmaly:


droot (approved):
We will do this on main also soon.

PR #2728, opened 2022-02-03
Add capability to override function runner in fnrender

droot:
This PR just shows how to provide a hook for override function executor when using `fn render` as library. Don't worry about the naming... they are just for the placeholder.

This approach wraps the provided function executor so that all the goodness of fn orchestration such as comment/indentation preservation is preserved.

Raising this PR for a discussion with @phanimarupaka @martinmaly


martinmaly:
Merged with #2730

PR #2729, opened 2022-02-03
Build Function Runner Docker Image

martinmaly:


porch/func/Makefile
```diff
@@ -23,3 +27,11 @@ $(COMPILED_PROTO): evaluator/evaluator.proto
 	  --go_out=./evaluator --go_opt=paths=source_relative \
 	  --go-grpc_out=./evaluator --go-grpc_opt=paths=source_relative \
 	  ./evaluator/evaluator.proto 
+
+.PHONY: build-image
+build-image:
+	docker buildx build --tag $(IMAGE_NAME):$(IMAGE_TAG) --tag $(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_TAG) .
```
justinsb, line 33: The double tag is for running locally I guess?

justinsb: I also don't think this actually does anything externally visible without `--load` or `--push`

justinsb (approved):

porch/func/Makefile
martinmaly: yes, I am trying to figure out best way to build for local and remote running. May change things as I get further along.

I'm working on push in subsequent PRs

PR #2730, opened 2022-02-04
Use Lower-Level Function Runner Interface

martinmaly:
- Add capability to override function runner during fnrender
- Use Lower-Level Function Runner Interface
- Fix tests to use packages on disk


droot (approved):
Fantastic, looks good to me.

PR #2742, opened 2022-02-04
tests: Fix test flake from timestamp sanitization (#2713)

martinmaly:
* Fix test flake from timestamp sanitization

When timestamps "overlap", sanitization did not work correctly.

For example:

in 1s ... in 2.1s

The first substitution would replace 1s => 0s

in 0s ... in 2.0s

The second substitution is predetermined, and replaces 2.1s => 0s, but
2.1s is no longer in the string.

* Use new sanitization function in tests, remove old function


justinsb (approved):
Not sure if I can approve,but lgtm :-)

PR #2744, opened 2022-02-04
Introduce mapTaskToMutation function

justinsb:
This is a small refactor so that we can reuse this logic in future.


justinsb:
A small refactor that makes my experimental branch much easier to carry :-)

(On that branch, I'm experimenting with re-executing these functions)

martinmaly:
I will be going through this code soon since I need to change the mutation interface.... just heads up.

martinmaly (approved):

PR #2746, opened 2022-02-05
Fix kpt file system interactions

martinmaly:
* missing file.Close() call (filesys' memory filesystem doesn't support
  files opened multiple times)
* pass filesystem into newFnConfig (it reads fn config from file)


phanimarupaka (approved):
Please replace api with kyaml in imports.

martinmaly:
done. unfortunately vscode tends to pick the wrong default :)

PR #2749, opened 2022-02-05
Fix clone directory rewriting

martinmaly:
After changes to git cloning we return PackageResources with configs
at the "root directory". Therefore we don't need to rewrite config paths
anymore.

### DIFF
--- a/porch/engine/pkg/engine/clone.go
+++ b/porch/engine/pkg/engine/clone.go
@@ -20,9 +20,8 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
-	"strings"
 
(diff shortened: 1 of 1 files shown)


justinsb (approved):

justinsb:
I am thinking we do need to document the conventions here ... i.e. do we clone into the root or packageName, what are the branch names in git, what is the pattern for OCI etc.  But this lgtm!

martinmaly:
Totally agree re: documenting this. My plan is that as I build the tests I can add the documentation in parallel. I am bound to find inconsistencies in the code and fixing them while testing & documenting seems most efficient.

PR #2752, opened 2022-02-05
Streamline Docker Image Building

martinmaly:
* Use the same naming pattern
* Enable nested Makefiles to inherit values from parent
* Enable explicit image tagging (i.e. using
  `IMAGE_TAG=$(git rev-parse --short HEAD) make build-images`
* Enable `make build-images` and `make push-images`


justinsb (approved):
Thanks - looks great.

And TIL about Makefile exports :-)

PR #2760, opened 2022-02-08
Content layer of Reference/Content pattern

loudej:
Incremental PR of oci rebase.

- Implements the content layer of reference/content pattern.
- demos/demo-content/main.go methods openContent and readFromFileSystem are example usage

Exported types need comments.


martinmaly (commented):
only initial review

pkg/location/extensions/extensions.go
```diff
@@ -0,0 +1,59 @@
...
+type DefaultIdentifierGetter interface {
+	GetDefaultIdentifier(ctx context.Context) (string, error)
+}
+
+// RelPather is present on Reference types that
+// will return a relative path if one reference is a sub-package
+// location in another. The comparison is strict, meaning all criteria
+// other than the directory component (like repo, ref, image, tag, etc.) must be equal.
+type RelPather interface {
+	Rel(targref Reference) (string, error)
```
droot, line 58: shouldn't it be returning `location.Reference` instead of string ?

pkg/location/extensions/extensions.go
```diff
@@ -0,0 +1,59 @@
...
+type ReferenceLock interface {
+	Reference
+}
+
+type IdentifierGetter interface {
+	GetIdentifier() (string, bool)
+}
+
+type LockGetter interface {
+	GetLock() (string, bool)
```
droot, line 37: should this return `location.ReferenceLock` ?

pkg/location/extensions/extensions.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"context"
+	"fmt"
+)
+
+type Reference interface {
+	fmt.Stringer
+	Type() string
+	Validate() error
+}
```
droot, line 26: I am wondering if this should have `ID()` method that returns string ?

pkg/location/location.go
```diff
@@ -0,0 +1,36 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package location
+
+import (
+	"fmt"
+)
+
+type Location struct {
```
droot, line 21: Thinking from an end-user API's perspective:

```

loc := content.Location{}

src, err := content.Open(loc)

```

pkg/location/location.go
```diff
@@ -0,0 +1,36 @@
...
+	Reference     Reference
+	ReferenceLock ReferenceLock
+}
```
droot, line 24: Having fun with naming here :) ..alternatives for `Reference` and `ReferenceLock` might be `Addr` and `ResolvedAddr` 

```

loc.Addr

loc.ResolvedAddr

```

demos/demo-content/main.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	if err := readFromFS(ctx, ref); err != nil {
+		return fmt.Errorf("open content as FS: %v", err)
+	}
+	return nil
+}
+
+func openContent(ctx context.Context, ref location.Reference) error {
+	// open it as content
+	src, err := open.Content(ref, open.WithContext(ctx))
```
droot, line 131: having `open` as the package name is little awkward. I am almost thinking if we can do this `content.Open(...)`

pkg/content/provider/oci/oci.go
```diff
@@ -0,0 +1,142 @@
...
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+type ociProvider struct {
+	image v1.Image
+	dir   string
+	fsys  filesys.FileSystem
```
droot, line 37: How is the auth setup done ? Is there a way to plug `authProvider` or the library is assuming something ?

pkg/content/provider/oci/oci.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if err := pullAndExtract(p.image, p.dir, fsys, "/"); err != nil {
+		return nil, "", err
+	}
+
+	p.fsys = fsys
+	return fsys, "/", nil
+}
+
+// pullAndExtract uses current credentials (gcloud auth) to pull and
```
droot, line 93: ok, this addresses the question I had earlier.

pkg/content/provider/oci/oci.go
```diff
@@ -0,0 +1,142 @@
...
+func (p *ociProvider) Close() error {
+	return nil
+}
+
+func (p *ociProvider) ProvideFileSystem() (filesys.FileSystem, string, error) {
+	if p.fsys != nil {
+		return p.fsys, "/", nil
+	}
+
+	fsys := filesys.MakeFsInMemory()
```
droot, line 83: sweet :)

pkg/content/provider/oci/oci.go
```diff
@@ -0,0 +1,142 @@
...
+				return err
+			}
+		default:
+			file, err := fsys.Create(path)
+			if err != nil {
+				return err
+			}
+			defer file.Close()
+
+			_, err = io.Copy(file, tarReader)
```
droot, line 134: This project https://github.com/mholt/archiver is an interesting one that provides `fs.FS` from an archive. Probably works better with streams.

pkg/content/extensions/extensions.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+type ChangeCommitter interface {
+	Content
+	CommitChanges() (location.Location, error)
+}
+
+type FileSystemProvider interface {
+	Content
+	ProvideFileSystem() (filesys.FileSystem, string, error)
```
droot, line 37: we can drop `Provide` from the name here.

pkg/content/extensions/extensions.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+type FSProvider interface {
+	Content
+	ProvideFS() (fs.FS, error)
```
droot, line 42: we can drop `Provide` from the name here.

pkg/content/extensions/extensions.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+type ReaderProvider interface {
+	Content
+	ProvideReader() (kio.Reader, error)
```
droot, line 47: we can drop `Provide` from the name here.

pkg/content/extensions/extensions.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+type RealPathProvider interface {
+	Content
+	ProvideRealPath() (string, error)
```
droot, line 52: we can drop `Provide` from the name here.

droot (commented):
I mostly have comments about some naming. But pl. take with grain of salt as this was a quick scan.

pkg/content/content.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2021 Google LLC
```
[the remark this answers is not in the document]
loudej: I _think_ the source of truth for that is in this file?

https://github.com/GoogleContainerTools/kpt/blob/5169954be82fabcfb6f7c0a01f4803c81cb65c67/scripts/update-license.sh#L23

I would change it, but the magnitude of that diff would probably need to be best in a PR that's entirely by itself. Maybe as a main branch PR that's brought back into porch?

pkg/content/open/open.go
```diff
@@ -0,0 +1,165 @@
...
+			return open{}, err
+		}
+		return open{
+			Content: provider,
+			Location: location.Location{
+				Reference: ref,
+			},
+		}, nil
+	case location.Git:
+		provider, lock, err := git.Open(opt.ctx, ref)
```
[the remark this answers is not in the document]
loudej: Yes, 100% agreed this is where there should be an interception. Switching reference.(type) cases onto type-specific Open funcs is very much just a stepping stone.

Along the lines of your first suggestion - similar to the `location.ParseReference(arg, location.WithParsers(...))` an opener option `open.Content(ref, open.WithContentOpeners(...))` would give the caller the first chance at taking control of how well-known or custom reference types are brought to life.

Testing the reference itself for a `ContentOpener` extension interface would also be a way for custom Reference types to have a convenient default case built in. But that alone is insufficient - comes with the problems like you've pointed out. It also threatens to make the reference smarter than it should be - anything beyond standardizing access to some of the string data is suspicious. 

(Tangent - probably worth filing an issue to consider if the "default branch name" getter should move off of the reference.)

pkg/content/extensions/extensions.go
```diff
@@ -0,0 +1,53 @@
...
+	Content
+	CommitChanges() (location.Location, error)
+}
+
+type FileSystemProvider interface {
+	Content
+	ProvideFileSystem() (filesys.FileSystem, string, error)
+}
+
+type FSProvider interface {
```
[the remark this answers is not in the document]
loudej: This does stand out as unused for CLI purposes, doesn't it? The reason I assumed it was important enough to include in public surface area --- filesys.FileSystem is a useful kyaml fs abstraction and has mutability by default --- but other code that will be using kpt would probably consider `io/fs.FS` a much more natural abstract file system to call through.

Esp. if they're using other go pkgs to process the content --- probably safe to assume filesys.FileSystem won't be what's needed.

pkg/content/extensions/extensions.go
```diff
@@ -0,0 +1,53 @@
...
+	Content
+	ProvideFS() (fs.FS, error)
+}
+
+type ReaderProvider interface {
+	Content
+	ProvideReader() (kio.Reader, error)
+}
+
+type RealPathProvider interface {
```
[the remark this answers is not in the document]
loudej: TBH - this can probably be removed. It was for the `kpt pkg diff` case --- the idea being when you _must_ have a real OS file system path you could use the existing one for content that was already on disk. (like dir and git that is based on exec git clone)

In practice --- the diff command actually must copy from src to staging location.Dir for other reasons (because prepping for diff is destructive). So I'll just remove this extension.

pkg/content/commit.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/GoogleContainerTools/kpt/pkg/content/extensions"
+	"github.com/GoogleContainerTools/kpt/pkg/location"
+)
+
+func Commit(content Content) (location.Location, error) {
+	switch content := content.(type) {
+	case extensions.ChangeCommitter:
+		return content.CommitChanges()
+	default:
+		return location.Location{}, nil
```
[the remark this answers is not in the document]
loudej: That's an interesting one. The idea there is that if the content is a remote location type (like OCI) this would flush any changes made via filesys.FileSystem or kio.Writer. Returning nil was speculating what's the expectation if you flush changes to a location type like Dir (or a purely in-memory which is not implemented yet).

That is to say --- does nil a success because the changes have already been made to the dir destination? Or should a "not supported" err type be detected?

Returning nil was assuming if a destination arg (like in kpt fn sink) could be either oci:// directly or dir path, that every place that ended with a call to commit on when dest was a dir the "not supported" was "don't care - it's optional" then needing to test for not supported err is only extra work.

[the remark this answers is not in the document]
loudej: Thoughts on that? It seems like a case where maybe a 3-valued return (location, bool, err) could be used --- with `{}, false, nil` being the not supported answer?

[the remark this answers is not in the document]
martinmaly: some go packages use error value singular to signal special cases (io.EOF); that could work here too.

pkg/content/filesystem.go
```diff
@@ -0,0 +1,35 @@
...
+func FileSystem(content Content) (paths.FileSystemPath, error) {
+	switch content := content.(type) {
+	case extensions.FileSystemProvider:
+		fsys, path, err := content.ProvideFileSystem()
+		return paths.FileSystemPath{
+			FileSystem: fsys,
+			Path:       path,
+		}, err
+	default:
+		return paths.FileSystemPath{}, fmt.Errorf("not implemented")
```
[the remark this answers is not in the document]
loudej: exactly so - well spotted! only the cases that were needed so far have been implemented.

I files #2764 to capture this, and referenced it with TODO(issue#) comments in the three places where there are switches like this.

pkg/content/open/open.go
```diff
@@ -0,0 +1,165 @@
...
+			Content: provider,
+			Location: location.Location{
+				Reference:     ref,
+				ReferenceLock: lock,
+			},
+		}, nil
+	case location.Oci:
+		provider, lock, err := oci.Open(
+			ref,
+			remote.WithAuthFromKeychain(gcrane.Keychain),
```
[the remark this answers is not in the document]
loudej: Yes - that will need to be done with additional open.WithXxxOption variations that pass through for the built-in types. Filed the remaining work for that as https://github.com/GoogleContainerTools/kpt/issues/2765.

pkg/content/provider/git/git.go
```diff
@@ -0,0 +1,116 @@
...
+		Path:    filepath.Join("/", ref.Directory),
+		Ref:     ref.Commit,
+	}
+
+	return open(ctx, ref, repoSpec)
+}
+
+func open(ctx context.Context, ref location.Reference, repoSpec *git.RepoSpec) (_result content.Content, _ location.ReferenceLock, _err error) {
+
+	if err := remote.ClonerUsingGitExec(ctx, repoSpec); err != nil {
```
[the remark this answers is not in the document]
loudej: Yep, exactly like you mentioned earlier, in the `c, err := open.Content(ref, ...options)` an option will need to enable getting in front of the built in switch cases - and for new public api that transitively and recursively opens upstream packages I assume it would also take the opts ...open.Option at the calling point? e.g. to carry WithContext(ctx), any credentials for built-in openers, any custom openers added by the caller, etc.

pkg/location/extensions/extensions.go
loudej: In this case the difference between two Reference would be a relative path - which is the string return value. The place it's used is in `kpt pkg update` in the "should update sub package" where the logic is asking if the nested kptfile.upstream is pointing to exactly the same remote location - but in a folder that's nested.

See also pkg/location/reference.go `func Rel(baseref Reference, targref Reference) (string, error)` which is the api that's actually called (not the RelPather directly)

Actually a `type RelativePath string` could be nice to have, but that's also a huge change. Maybe a simpler update would be to have named return arguments like this?

func Rel(baseref Reference, targref Reference) (relPath string, err error)` 

The argument names baseref and targref are not great - they came straight from filepath.Rel argument names. Suggestions welcomed there as well.

pkg/location/extensions/extensions.go
loudej: That's a good question, but I think the problem is that the name "GetLock" is very bad.

The value being returned here is the Digest string, or Commit string, or for a custom the appropriate string that uniquely identifies the immutable revision of remote content.

So at the moment to turn a Reference to a ReferenceLock - you call:
`refLock ReferenceLock, err error := mutate.Lock(ref Reference, lock string)`

And to get the lock string back out of a ReferenceLock - you call:
`lock string, err error := location.Lock(refLock)`

_**So**_ I think the `lock string` needs a better name for digest-or-commit value. If that was called a `Foo` then the above pair would change to:

`refLock ReferenceLock, err error := mutate.Lock(ref Reference, foo string)`
and
`foo string, err error := location.GetFoo(refLock)`

pkg/location/extensions/extensions.go
loudej: Sure, what would be the semantics? Is the idea that `ID()` would return a nearly total string value that would round-trip through parsing if possible - and the `fmt.Stringer.String()` would return a string that's potentially more optimized for display/log/error purposes?

@martinmaly had earlier questions related to round-tripping that could be related. Also this would help with "-" locations... As a Content.ID() it works great for round-tripping, but as a Stringer.String() a single dash is a bit of miserable display name.

pkg/location/location.go
loudej: TBH - the struct Location was only added as a convenience when a method _returned_ both values - in places like content.Open when a ref or refLock is an _argument_ you'd be interested in one or the other, but both at once never really happened. Closest would be like upgrade - but even then the "@version" mutates the Reference and two calls to content.Open was more natural...

In fact - even when Location is used to return both values - there's always other values being returned as well. So even there it's not necessary.

In hindsight I think it'd be best to remove the `content.Location{}` 2-value struct because like you say it _looks_ like something you should use to open content, but isn't

pkg/location/location.go
loudej: Interesting --- do you mean to change the field names to 
```
Addr Reference
ResolvedAddr ReferenceLock
```
or rename the two 
```
type Reference interface {...}
type ReferenceLock interface {...}
```
as
```
type Addr interface {...}
type ResolvedAddr interface {...}
```
?

I think (see above) the `location.Location{}` struct should probably be deleted because the purpose it serves doesn't deserve such a significant sounding name... So those field names would go away with it...

That said if there's a better name than "Reference" for the core interface, now would be the time to rename it. :) Though I tend to suggest avoiding abbreviations in type names...
```
type Address interface {...}
type ResolvedAddress interface {...}

addr, err := location.ParseAddress(arg, location.WithZzz(), location.WithYyy())
src, err := content.Open(addr)
```
or
```
type Location interface {...}
type ResolvedLocation interface {...}

loc, err := location.ParseLocation(arg, location.WithZzz(), location.WithYyy())
src, err := content.Open(loc)
```

I _kind_ of like "address" thinking just of git and oci... But thinking about `location.Dir{}` and "-" streams... A custom location could even have a `map[string]string` as its only field if you wrote a content-opener and filesystem-provider for it... Point being I do prefer Address initially but I think it breaks down thinking about the non-remote kinds of location/reference.

pkg/content/provider/oci/oci.go
loudej: Yes, there's an issue filed that underlying library options will need to pass in as additional open options... 

https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/options.go

Because they're a different type, it'll be slightly nested: 

```
src, err := open.Content(ref, c.OpenOptions...) // when calculated earlier, in PreRunE
// or inline, for example
src, err := open.Content(
    ref,
    open.WithContext(ctx),
    open.WithContainerRegistryOptions(
        remote.WithAuthFromKeychain(gcrane.Keychain)),
        remote.WithTransport(etc),
        remote.WithRetryBackoff(etc),
    ),
)
```

pkg/content/provider/oci/oci.go
loudej: Indeed - and the `/pkg/content/provider/*/*.go` are exactly where those kind of choices and changes can be made in a place that's dedicated only to the task of exposing one type of location as content.

Doesn't need to understand argument parsing, package structures, or file formats... Also being mainly lowercase doesn't need to worry about becoming public surface area that will cause breaking changes if updated.

And also to @martinmaly question on related files, can be replaced if someone gives you a custom ContentOpener as an open.Option... e.g. in places where the kptlib is used that shouldn't exec git to clone to tmp folder --- all you need is to intercept the moment where the default &gitProvider{} would have been returned with your own --- and everything else in the system runs as they would because they don't care that's a difference

pkg/content/extensions/extensions.go
loudej: This extension was removed entirely, but otherwise, 👍

demos/demo-content/main.go
loudej: That convention does have an odd first impression, doesn't it? With only one "verbish" go sub package to start it's hard to tell if it's paying for itself.

The place that pattern comes from is in the oci lib --- for example instead of `images.Validate(img, opts ...images.ValidateOption)` there's a `validate.Image(img, opts ...validate.Option)`

One way it really shines is that the option factories for a verbish group are very focused to what apply to methods in that group. e.g. `validate.Image(img, validate.Fast)` where Fast is of type `validate.Option`... Other examples remote.Image(name, ...remote.Option)`, daemon.Image, daemon.Write, tarball.Image, etc...

---

but looping back to Kpt apis specifically, where you see that is `src := open.Content(reference)` to open as content,`src := open.FileSystem(reference)` to open but with filesystem as access choice, `src := open.Reader(reference)` to open but with reader as access choice...

in the base package there is `fs := content.FileSystem(src)` and `r := content.Reader(src)` also - but everything you'll expect to find there is "I've got a content as the argument, and the methods often relate to a return value or property. Frequently something which would be optional and duck typed but default behavior is provided"

Say! That's the other example. `io/fs` kind of acts this way where `x := fs.ReadFile(fsys fs.FS, path string)` where the package name "fsXxx." kind of means "these are methods you may find on a FS, passed as the first argument" like "content.Xxx" is a method that you may find on a Content, which you pass as the first argument.

But the way you get your hands on a "FS" is somewhere else, like `os.DirFS(path)`

pkg/content/content.go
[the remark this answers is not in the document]
loudej: 👍 filed as https://github.com/GoogleContainerTools/kpt/issues/2766

pkg/content/commit.go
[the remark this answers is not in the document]
loudej: Yeah, I think you're right. Similar to `return nil, ErrNotFound` instead of `return nil, nil` with the mental note "nil means not found"... Err matching "normal" conditions is always irritating, but there are places where `if err != nil { return err }` becomes `if err!=nil && !errors.Is(ErrNotFound) {return err}`

Maybe it's a lingering "dotnet-ism" --- there's a mantra "don't throw exceptions for non-exceptional circumstances" --- though that could just as much relate to the burden of forcing callers to add try/catch blocks.

pkg/content/open/open.go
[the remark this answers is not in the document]
martinmaly: TODO may be sufficient to mark that we need to enable pluggable reference resolvers.

pkg/content/commit.go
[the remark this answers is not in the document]
loudej: added content.ErrNotCommittable

pkg/content/open/open.go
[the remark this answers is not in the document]
loudej: It's not really a large change, I've pushed an update that adds a `open.WithProviders(...)` option where a provider is a first-chance function call.

pkg/content/paths/paths.go
```diff
@@ -0,0 +1,41 @@
...
+	Path       string
+}
+
+// String returns the path in string format.
+func (fsp FileSystemPath) String() string {
+	return fsp.Path
+}
+
+type FSPath struct {
+	FileSystem filesys.FileSystem
```
martinmaly, line 35: Should this be of type fs.FS?

(consider removing FS support for now)

pkg/location/reference_test.go
```diff
@@ -0,0 +1,139 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := SetRevision(tt.args.ref, tt.args.revision)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("SetRevision() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
```
martinmaly, line 93: consider using https://github.com/google/go-cmp

pkg/content/open/open.go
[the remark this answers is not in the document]
loudej: @martin, this is updated if you'd like to re-review --- I'll resolve in the meantime, and feel free to reopen

pkg/location/extensions/extensions.go
loudej: There was a another thread which @droot also mentioned Address and ResolvedAddress as names. I think that was for some Ref RefLock properties, not the Ref/RefLock interface names themselves --- but if there is a better name now could be a good time to change

pkg/location/type_oci.go
```diff
@@ -92,12 +115,28 @@ func parseOci(value string) (Reference, error) {
...
+
+func ociString(image name.Reference, directory string) string {
+	if directory != "" && directory != "." && directory != "/" {
+		if image, ok := image.(name.Tag); ok {
+			return fmt.Sprintf("oci://%s//%s:%s", image.Context().Name(), directory, image.TagStr())
+		}
+		if image, ok := image.(name.Digest); ok {
+			return fmt.Sprintf("oci://%s//%s@%s", image.Context().Name(), directory, image.DigestStr())
+		}
+		return fmt.Sprintf("oci://%s//%s", image.Context().Name(), directory)
```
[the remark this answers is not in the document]
loudej: It's taken from Terraform's convention for paths inside git repositories:

https://www.terraform.io/language/modules/sources#modules-in-package-sub-directories

Kpt does have a convention in git of assuming if there is a ".git" in the string that's the repo::subpackage delimiter --- There is also an error message from the CLI if parsing the repo name is ambiguous that ".git" should be added. But there's not a similar suffix that goes with OCI repos...

TBH --- the `//` convention could be added to the kpt git repo parsing as well in a non-breaking way, e.g.

`https://github.com/org/repo//path@main`

pkg/location/type_oci.go
```diff
@@ -114,19 +153,43 @@ func (ref Oci) Type() string {
 	return "oci"
 }
 
+func (ref Oci) Rel(target Reference) (string, error) {
```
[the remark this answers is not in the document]
loudej: TBH the "Rel" is much less abtract than the previous internal "shouldupdatesubpackage"

The case it's used is in upgrade - when you 1) update a package to a different upstream version and, 2) the local you're updating has subpackages, and 3) those subpackages are related by be relative-subfolder from the upstream --- then _only_ in that case are those subpackages simultaneously updated with the base package. Essentially a recursive update in the one special case where the subpkg upstream is relative subpath in the same remote.

So even in the case of a git 3-way merge - where all 3 copies are on os disk - you still need to know "given these two fully qualified locations, what's the relative path relationship?"

pkg/location/type_git.go
```diff
@@ -127,15 +170,34 @@ func (ref Git) GetDefaultDirectoryName() (string, bool) {
...
 }
 
-// SetIdentifier is called from mutate.Identifier
-func (ref Git) SetIdentifier(name string) (Reference, error) {
+func (ref Git) DefaultRevision(ctx context.Context) (string, error) {
+	gur, err := gitutil.NewGitUpstreamRepo(ctx, ref.Repo)
+	if err != nil {
+		return "", err
+	}
+	b, err := gur.GetDefaultBranch(ctx)
```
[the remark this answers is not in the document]
loudej: That could be a question for @droot --- it's true it's the only place that the reference/parsing layer and it's only because in the existing CLI will fully qualify the git repo (adding default branch name) if it's not provided.

That said --- **Open** on a reference does return _both_ the ref and the refLock for that reason. The CLI could be changed to allow `location.Git{Ref string}` at parse-time remain empty-string to imply the default branch --- and the Ref returned could be treated as the correct value to persist in Kptfile.upstream

pkg/content/reader.go
```diff
@@ -0,0 +1,44 @@
...
+func Reader(content Content) (kio.Reader, error) {
+	switch content := content.(type) {
+	case extensions.ReaderProvider:
+		return content.Reader()
+	case extensions.FileSystemProvider:
+		fsys, path, err := content.FileSystem()
+		if err != nil {
+			return nil, err
+		}
+		return &kio.LocalPackageReader{
```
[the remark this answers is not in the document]
loudej: In that case it is possible to use the `content.FileSystem` as access layer, as you say, if you need to set exact flags on the local package reader.

The times that you would bind to `content.Reader` directly as the access layer is where code is doing linear transforms (like sink/source/etc) --- operations that work equally well with stdio or ResourceList.yaml as the content.

In those cases the kio reader doesn't have many options (just a few like include metadata which can be added to the Reader binding call) --- but to convert to filesystem and back to reader in those cases isn't necessary. (extra work, and a way for unwanted deltas to appear in the data)

pkg/content/commit.go
```diff
@@ -0,0 +1,36 @@
...
+var (
+	ErrNotCommittable = errors.New("not committable")
+)
+
+func Commit(content Content) (location.Reference, location.ReferenceLock, error) {
+	switch content := content.(type) {
+	case extensions.ChangeCommitter:
+		return content.CommitChanges()
+	default:
+		return nil, nil, fmt.Errorf("invalid %T operation: %w", content, ErrNotCommittable)
```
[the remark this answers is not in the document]
loudej: errors.Is() detects the error when wrapped (there should be a unit test to assert that IIRC)

pkg/content/content.go
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package content
+
+import (
+	"github.com/GoogleContainerTools/kpt/pkg/content/extensions"
+)
+
+type Content = extensions.Content
```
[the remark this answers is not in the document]
loudej: Because the `pkg/content` code needs to see a few go pkg which need to accept `content.Content` as argument.

The alias enables `content.Content` and `content.EverythingElse()` to be in one place from the consumer's PoV, but in the places where that caused a cycle the transitive location could use `extensions.Content` leaf location for type definitions.

TBH, it originally contained only the duck-typing interfaces but `content.Content` was eventually was closed back over as well.

PR #2770, opened 2022-02-09
Refactor Server Startup

martinmaly:
* Start background processing in Run function
* Shut down when stopCh is closed

### DIFF
--- a/porch/apiserver/pkg/registry/porch/background.go
+++ b/porch/apiserver/pkg/registry/porch/background.go
@@ -28,14 +28,23 @@ import (
 	"github.com/GoogleContainerTools/kpt/porch/repository/pkg/cache"
 )
 
+func RunBackground(coreClient client.WithWatch, cache *cache.Cache, stopCh <-chan struct{}) {
+	b := background{
+		coreClient: coreClient,
+		cache:      cache,
+	}
+	ctx := context.Background()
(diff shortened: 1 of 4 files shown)


porch/apiserver/pkg/registry/porch/background.go
```diff
@@ -28,14 +28,23 @@ import (
...
+
 // background manages background tasks
 type background struct {
 	coreClient client.WithWatch
 	cache      *cache.Cache
 }
 
-// RunForever will run until ctx is done
-func (b *background) RunForever(ctx context.Context) {
+// run will run until ctx is done or stopCh is closed
```
justinsb, line 46: Personally, I prefer just putting all my eggs into the context basket.  So at the point where someone passes us a stopCh,I would  immediately translate it to a context.  But that's just me :-)

justinsb (approved):

porch/apiserver/pkg/registry/porch/background.go
martinmaly: I agree. the stopCh is k8s apiserver concept though that they use to stop the main server so we are caught in the middle there. Looking at the apiserver code, it may be possible to refactor further. I'll see if we can drive the whole apiserver shutdown from context only.

PR #2771, opened 2022-02-09
Create Engine Using Options

martinmaly:
Create options for the common engine configuration options
(more to come).

PR #2772, opened 2022-02-09
Support lazy credential resolution

martinmaly:
* Add CredentialResolver interface
* Add WithCredentialResolver engine option
* Use lazy credential resolution when interacting with Git


martinmaly:
This PR has dependency on #2771 and #2770 so it is rebased on top of them

justinsb (approved):
Nice - this code feels like cleaner, as well as a functionality improvement.

PR #2774, opened 2022-02-10
Use Context to Drive Server Shutdown

martinmaly:
Turns out k8s apiserver supports both and we can get the core context to
listen on instead of just getting the close channel.


justinsb (approved):
Wow - nice find & thanks for doing this!

PR #2776, opened 2022-02-10
Easier deployment onto GKE

justinsb:
Create doc to describe how to deploy onto a GKE cluster, and some supporting manifest / script changes.


martinmaly (approved):
Let's go with your change first; let me know what you think about mine as a follow-up, in particular kustomize (I felt we needed to uphold the CaD principles :)

PR #2781, opened 2022-02-12
`kpt fn render` includes local and meta resources by default

droot:
This PR adds support for including `local` and `meta` resources by default in `fn render`.

To exclude meta resource (just the `Kptfile`), users can run `kpt fn render --exclude-meta-resources` .


internal/cmdrender/executor.go
```diff
@@ -92,12 +92,18 @@ func (e *Executor) Execute(ctx context.Context) error {
 	at := attribution.Attributor{Resources: hctx.root.resources, CmdGroup: "fn"}
 	at.Process()
 
+	includeMetaResources := true
```
yuwenma, line 95: will this eventually be changed as a configurable flag? Or kpt always include meta resources in `fn render`

droot:
ref: https://github.com/GoogleContainerTools/kpt/issues/2590

internal/cmdrender/executor.go
droot: Actually, thought some more and realized exposing this as `configurable` flag will make it consistent with other `fn` commands and also provide a way to address the breaking change. Users can run `fn render --include-meta-resources false` to get the previous behavior. I will update the PR.

droot:
Quick update: Now this PR is against the target feature branch `variant-constructor` where we plan to accumulate all changes related to `variant constructor pattern` /cc @yuwenma @bgrant0607

internal/cmdrender/executor.go
```diff
@@ -168,6 +176,10 @@ type hydrationContext struct {
 	// imagePullPolicy controls the image pulling behavior.
 	imagePullPolicy fnruntime.ImagePullPolicy
 
+	// indicate if package meta resources such as Kptfile
+	// to be excluded from the function in put during render.
+	encludeMetaResources bool
```
yuwenma, line 181: ditto

PR #2784, opened 2022-02-12
Fix missing error handling

justinsb:
We weren't checking errors when building a CaDEngine.


martinmaly (approved):
thanks! I noticed those too and was going to fix it; thanks for beating me to it!

PR #2786, opened 2022-02-12
Create simple kpt-porch CLI

justinsb:
This will let us iterate while developing new commands for kpt.

First command uploads resources into porch.


martinmaly (approved):
actually, this can give us more velocity and we can easily refactor in a bit, and incorporate additional UX learning

PR #2791, opened 2022-02-12
Makefile: split push-and-deploy into push and deploy

justinsb:
This is helpful when iterating on the manifests!


martinmaly (approved):
although I just merged change that contains this very split :( sorry about the dupe.

PR #2792, opened 2022-02-12
remoterootsyncset: don't require gcloud

justinsb:
This was legacy code from development.


martinmaly (approved):
does this rely on workload identity now?

PR #2794, opened 2022-02-14
update resourcegroup CRD

Liujingfang1:
- Add the new fields for capturing the apply and reconcile status
- Update the pre-steps of apply. If the ResourceGroup CRD is
  installed but doesn't match the latest declaration, the ResourceGroup
  CRD should be re-applied.

### DIFF
--- a/e2e/live/end-to-end-test.sh
+++ b/e2e/live/end-to-end-test.sh
@@ -129,7 +129,30 @@ function downloadPreviousKpt {
   fi
   tar -xvf kpt.tar.gz > $OUTPUT_DIR/kptdownload 2>&1
   mv kpt $BIN_DIR/previouskpt
-  echo "Downloading latest kpt binary...SUCCESS"
+  echo "Downloading previous kpt binary...SUCCESS"
+  rm kpt.tar.gz LICENSES.txt lib.zip
+  set +e
+}
+
+function downloadKpt1.0 {
+  set -e
+  echo "Downloading v1.0.0-beta.13 kpt binary..."
+  uname="$(uname -s)"
+  if [[ "$uname" == "Linux" ]]
+  then
+    echo "Running on Linux"
+    curl -LJ -o kpt.tar.gz https://github.com/GoogleContainerTools/kpt/releases/download/v1.0.0-beta.13/kpt_linux_amd64-1.0.0-beta.13.tar.gz > $OUTPUT_DIR/kptdownload 2>&1
+  elif [[ "$uname" == "Darwin" ]]
+  then
+    echo "Running on Darwin"
+      curl -LJ -o kpt.tar.gz https://github.com/GoogleContainerTools/kpt/releases/download/v1.0.0-beta.13/kpt_darwin_amd64-1.0.0-beta.13.tar.gz > $OUTPUT_DIR/kptdownload 2>&1
+    else
+      echo "ERROR: Unknown OS $uname"
+      exit 1
+  fi
+  tar -xvf kpt.tar.gz > $OUTPUT_DIR/kptdownload 2>&1
+  mv kpt $BIN_DIR/kpt1.0.0
+  echo "Downloading 1.0.0 kpt binary...SUCCESS"
   rm kpt.tar.gz LICENSES.txt lib.zip
   set +e
 }
@@ -409,6 +432,7 @@ echo
 buildKpt
 
 downloadPreviousKpt
+downloadKpt1.0
 
 echo
 set +e                          # Do not stop the test for errors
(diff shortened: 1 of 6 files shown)


internal/cmdutil/util.go
```diff
@@ -73,3 +77,12 @@ type NoResourceGroupCRDError struct{}
...
 	return "type ResourceGroup not found"
 }
+
+// ResourceGroupCRDNotLatestError is an error type that will be used when a
+// cluster has a ResourceGroup CRD that doesn't match the
+// latest declaration.
+type ResourceGroupCRDNotLatestError struct{}
+
+func (*ResourceGroupCRDNotLatestError) Error() string {
+	return "type ResourceGroup CRD needs update"
```
mortent, line 87: What is the UX here? Once users update to the latest version of kpt, they will get this message on the next attempt to apply. But what happens then? Do we provide an easy was to update? We should make the error message here more actionable.

pkg/live/inventoryrg.go
```diff
@@ -228,20 +230,71 @@ var crdGroupKind = schema.GroupKind{
...
 	_, err = mapper.RESTMapping(ResourceGroupGVK.GroupKind())
 	if err != nil {
 		klog.V(7).Infof("error retrieving ResourceGroup RESTMapping: %s\n", err)
+		return false, false
+	}
+	return true, resourceGroupCRDMatchesLatest(mapper, factory)
+}
+
+func resourceGroupCRDMatchesLatest(mapper meta.RESTMapper, factory cmdutil.Factory) bool {
+	rgCRDStr, ok := resourceGroupCRDs["v1"]
```
mortent, line 253: I think we can probably ignore the error check here. We know it is there and the code will panic if for some reason it is not.

pkg/live/inventoryrg.go
```diff
@@ -228,20 +230,71 @@ var crdGroupKind = schema.GroupKind{
...
+
+	mapping, err := mapper.RESTMapping(crdGroupKind)
+	if err != nil {
+		klog.V(7).Infof("Failed to get mapping of CRD type: %s", err)
+		return false
+	}
+
+	liveCRD, err := dc.Resource(mapping.Resource).Get(context.TODO(), "resourcegroups.kpt.dev", metav1.GetOptions{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "apiextensions.k8s.io/v1",
```
mortent, line 278: Do we need to handle clusters that doesn't have the v1 version of the CRD resource type?

mortent (commented):
Could we expand a little bit on the end-to-end UX here? For example:
* When will users see the error message about needing to update the CRD and what should they do about it?
* Can we just update it instead of returning an error?
* Could we continue even if there is an old version of the RG CRD? It would just mean that the additional fields will not be set?

internal/cmdutil/util.go
Liujingfang1: They could run `kpt live install-resource-group`

pkg/live/inventoryrg.go
Liujingfang1: I don't think we need to continue providing support for v1beta1 CRD. The v1 version of the CRD is available since 1.16, which is quite a few versions before the current one 1.23.

pkg/live/inventoryrg.go
Liujingfang1: Do you mean that we don't check if the ResourceGroup CRD in the cluster is the same as the one in the current kpt release? If we do that, users may see some error from the apply library, but it doesn't provide the meaning action to fix the error.

internal/cmdutil/util.go
Liujingfang1: The change updates the pre-steps of apply. If the ResourceGroup CR is installed, but it doesn't match the latest one in the kpt release, then it updates the ResourceGroup CRD in the cluster.
This update fails in the case that users don't have the permission to GET/UPDATE a CRD resource.
So we can change the error message here to be "The ResourceGroup CRD needs update, please run `kpt live install-resource-group`" to update it.

Liujingfang1:
> Could we continue even if there is an old version of the RG CRD? It would just mean that the additional fields will not be set?

I think we could. However, it will leave the ResourceGroup CRD in the cluster with a previous schema. Then the ResourceGroup CR is not able to
catch the apply/reconcile status.

pkg/live/inventoryrg.go
Liujingfang1: Please ignore my previous comments. I see what you mean. Let me update it.

Liujingfang1: done

pkg/live/inventoryrg.go
mortent: I'm not sure we are ready to assume all cluster have the v1 version yet. In any case, we should do that in a separate PR and clean up any remaining code for the v1beta1 version. Is it a lot of work to handle this for v1beta1 too?

internal/cmdutil/util.go
mortent: Yeah, with the updated error message, this looks good.

pkg/live/inventoryrg.go
Liujingfang1: Not much work. Let me add the code for handling v1beta1 version as well.

Liujingfang1: Added this.

Liujingfang1:
Here is the end-to-end UX:
- When the CRD needs install or update
    - when --install-resource-group=false, it doesn't install or update the ResourceGroup CRD
    - when --install-resource-group=true, it installs or updates the ResourceGroup CRD

Also added two new testcases to cover this UX.

Liujingfang1:
@mortent The PR is ready for another review.

mortent (approved):
Looks good, just a comment on the test.

e2e/live/end-to-end-test.sh
```diff
@@ -836,6 +860,65 @@ assertPodExists "pod-c" "test-rg-namespace"
...
+assertContains "pod/pod-b unchanged"
+assertContains "pod/pod-c unchanged"
+printResult
+echo "cat e2e/live/testdata/stdin-test/pods.yaml | kpt live destroy -"
+cat e2e/live/testdata/stdin-test/pods.yaml | ${BIN_DIR}/kpt live destroy - > $OUTPUT_DIR/status 2>&1
+assertContains "pod/pod-a deleted"
+assertContains "pod/pod-b deleted"
+assertContains "pod/pod-c deleted"
+assertContains "4 resource(s) deleted, 0 skipped"
+printResult
```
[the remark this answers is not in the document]
Liujingfang1: Added comments to explain this test. The package was applied first by a previous released kpt, where the ResourceGroup CRD is different. Then the package is applied with the built kpt. It updates the CRD and applies or prunes the package successfully.

e2e/live/end-to-end-test.sh
```diff
@@ -836,6 +860,65 @@ assertPodExists "pod-c" "test-rg-namespace"
 assertPodExists "pod-d" "test-rg-namespace"
 printResult
 
+###########################################################
+#  Test Update ResourceGroup CRD during apply
```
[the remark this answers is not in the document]
Liujingfang1: Ack

PR #2797, opened 2022-02-15
Split git_test To Use Git Server

martinmaly:
Enable e2e tests to use git server


justinsb (approved):

porch/repository/pkg/git/testing.go
```diff
@@ -0,0 +1,635 @@
...
+		return p.walkObjectTree(obj.Target)
+	default:
+		// Error out on unhandled object types.
+		return fmt.Errorf("Unknown object %s %s %T\n", obj.ID(), obj.Type(), obj)
+	}
+	return nil
+}
+
+// initRepo is a helper that creates a first commit, ensuring the repo is not empty.
+func initRepo(repo *git.Repository) error {
```
justinsb, line 471: TBD, but we might want this in a different package in the end - depends on whether we further split "git server" from "pure test code"

martinmaly: great point. I noticed this after I'd made the initial split. I'll refactor this in subsequent PRs.

We'll also want different ways to initialize the repositories so I'll have couple of initializers (empty repo, initialized from contents in testdata, ...)

PR #2798, opened 2022-02-15
Make Git Server hostable in cluster

martinmaly:
And add makefile targets to build and push its images.


martinmaly:
Based off of #2797

porch/test/git/main.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	addressChannel := make(chan net.Addr)
+
+	go func() {
+		if err := server.ListenAndServe(ctx, fmt.Sprintf(":%d", *port), addressChannel); err != nil && err != http.ErrServerClosed {
+			fmt.Fprintf(os.Stderr, "Listen failed: %v", err)
```
justinsb, line 75: We may want to klog.Fatalf or panic here, otherwise we don't exit and the pod won't restart (I think!)

(Though if we have a health/readiness check, that would get it in the end, I guess)

martinmaly: good point; fixed.

PR #2801, opened 2022-02-15
Return Git PackageRevisionResources Correctly

martinmaly:
Newly created Git package returned package resources at its old commit
SHA where no package contents existed, thus returning empty resources.
As package changes, advance the reference to return up-to-date package
revision contents.


justinsb (approved):
I think we made a similar change in OCI, so this makes sense!

PR #2802, opened 2022-02-16
Use external resourcegroup object file to store inventory

rquitales:
This PR modifies all kpt live commands to use an external resourcegroup object file for storing inventory information.

Changes:
- `kpt live` commands introduce a `--rg-file` flag
- `kpt live init` and `kpt live migrate` commands' `Run` logic updated to remove feature gating
- Updated integration tests to pass
- Updated e2e tests to pass
- Added new e2e test cases to test the new feature (init, apply, destroy and migrate)


e2e/live/testdata/migrate-case-2a/namespace.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright 2019 The Kubernetes Authors.
```
yuwenma, line 1: Suggest modifying and adding the "Copyright **2022** The Kubernetes Authors..." here and in other files

internal/cmdapply/cmdapply.go
```diff
@@ -83,6 +84,7 @@ func NewRunner(ctx context.Context, factory util.Factory,
 		"dry-run apply for the resources in the package.")
 	c.Flags().BoolVar(&r.printStatusEvents, "show-status-events", false,
 		"Print status events (always enabled for table output)")
+	c.Flags().StringVar(&r.rgFile, "rg-file", rgfilev1alpha1.RGFileName, "The name of the ResourceGroup file for storing inventory information")
```
yuwenma, line 87: nit: "the file path to the ResourceGroup object". qq: Shall we mention inventory here or only ResourceGroup?

mortent: Agree we should rephrase this a bit.

e2e/live/testdata/migrate-case-2a/namespace.yaml
mortent: Are these files copied from a Kubernetes repo? If not, why are we using the Kubernetes Authors copyright?

e2e/live/end-to-end-test.sh
```diff
@@ -836,6 +850,163 @@ assertPodExists "pod-c" "test-rg-namespace"
 assertPodExists "pod-d" "test-rg-namespace"
 printResult
 
+###########################################################
+#  Test Migrate from Kptfile inventory to ResourceGroup
+###########################################################
```
mortent, line 855: I'm not crazy about adding more of this type of tests. It has proven very difficult to debug for others. How much effort would it be to instead implement this with the other e2e framework?

mortent (changes requested):
This seems to pass empty inventoryID to the actuation library if it is not provided by the user. I think the idea was that it would be generated based on the name and namespace if not provided. Leaving it empty can lead to strange behavior.

e2e/live/testdata/migrate-case-2a/namespace.yaml
rquitales: The files in this directory were copied over from `e2e/live/testdata/migrate-case-1a` as is. A new directory is used since the e2e shell scripts doesn't clean up the testdata directory after being run, hence it seemed like a cleaner approach for me to have a separate testdata folder for the new migrate e2e test case. Let me know how the license header should be handled if we do stick with having the new test case in the shell script, rather than the Go test harness.

e2e/live/end-to-end-test.sh
rquitales: It will require a quite bit more effort, since we'd have to come up with a solution for building the older kpt binaries from our Go test suite and the logic for testing migrations. But also open to implementing that since it will be a cleaner solution. Alternatively, creating the migrate e2e Go tests in a follow-up PR might be better?

internal/cmdapply/cmdapply.go
rquitales: Since we still mention inventory elsewhere, I'm leaning more towards maintaining that vocabulary. This is especially since we still support the inventory being provided in the Kptfile - we just throw a warning message encouraging users to migrate to a resourcegroup file instead. But yes, I'll need to update the help message regardless. Lmk what your preferences are in how this should be rephrased.

rquitales:
> This seems to pass empty inventoryID to the actuation library if it is not provided by the user. I think the idea was that it would be generated based on the name and namespace if not provided. Leaving it empty can lead to strange behavior.

The design doc mentions that the default annotation provided would be "-" if not specified by the user, but this doesn't conform to the K8s annotation requirements. I opted for using an empty value `""` since I observed some strange behaviour when the annotation was completely removed on the live cluster, but not when the annotation was set but with an empty value.

I know @yuwenma wanted to simplify the inventoryID logic which is why we looked into not using a name+namespace generated inventoryID. However, given that generating a name+namespace based inventoryID is idempotent/reproducible, I don't see any issues with reverting back to this logic since recalculating this value on every `kpt live` run is also a cheap computation.

rquitales:
@yuwenma @mortent Friendly ping to see if we can get a general consensus before I update this PR? Thanks

e2e/live/testdata/migrate-case-2a/namespace.yaml
yuwenma: so as Morten mentioned, shall we update this to kpt Authors?

internal/cmdliveinit/cmdliveinit.go
```diff
@@ -153,78 +151,11 @@ type ConfigureInventoryInfo struct {
...
-	return nil
-}
-
-// func runLiveInitWithRGFile is a modified version of ConfigureInventoryInfo.Run that stores the
-// package inventory information in a separate resourcegroup file. The logic for this is branched into
-// a separate function to enable feature gating.
-func (c *ConfigureInventoryInfo) runLiveInitWithRGFile(ctx context.Context) error {
-	const op errors.Op = "cmdliveinit.runLiveInitWithRGFile"
-	pr := printer.FromContextOrDie(ctx)
-
```
yuwenma, line 227: In [ln 165](https://github.com/GoogleContainerTools/kpt/pull/2802/files#diff-a4edb85561333f6d7f4b2003a19f26017970bfcd20cdf92a42398024a399b6f8L234),  I just realized the user message should not be "ResourceGroup" (it used to be Kptfile which refers to the file name). Can you update the message as ("initializing %v  data (namespace: %v)", RGFileName, namespace)

yuwenma:
One problem: `kpt live apply` shall not rely on the default resourcegroup.yaml to read the inventory info when you enable the rg-file flag. the `kpt live apply` should read resource-group from the input by looking for the kind `ResourceGroup` and apiVersion.

e2e/live/testdata/migrate-case-2a/namespace.yaml
rquitales: Given that these files were copied directly from e2e/live/testdata/migrate-case-1a, and they contained the Kubernetes authors as copyright, it seems like we should maintain the Kubernetes authors still?

internal/cmdliveinit/cmdliveinit.go
rquitales: Ack.

rquitales:
> One problem: `kpt live apply` shall not rely on the default resourcegroup.yaml to read the inventory info when you enable the rg-file flag. the `kpt live apply` should read resource-group from the input by looking for the kind `ResourceGroup` and apiVersion.

@yuwenma I think this should already be handled if I'm understanding this user journey correctly. Here are the following combinations of local and STDIN resourcegroup:

1) User pipes K8s resources along with resourcegroup to `kpt live apply -`, we will use the resourcegroup in stream, and Kptfile in current working directory.
2) User pipes K8s resources along with resourcegroup and Kptfile to `kpt live apply -`, we use both resourcegroup and Kptfile in stream.

3) User pipes K8s resources along with resourcegroup to `kpt live apply -`, but the current working directory already has a resourcegroup, we throw a multiple inventory error. Perhaps we should always default to using the resourcegroup in the stream for this case and ignore the local directory's resourcegroup?

4) User pipes K8s resources WITHOUT resourcegroup, but Kptfile that contains an inventory to `kpt live apply -`, we successfully apply the resources to maintain backwards compatibility with existing `kpt live apply -` functionality.

yuwenma:
> > One problem: `kpt live apply` shall not rely on the default resourcegroup.yaml to read the inventory info when you enable the rg-file flag. the `kpt live apply` should read resource-group from the input by looking for the kind `ResourceGroup` and apiVersion.
> 
> Yuwen Ma I think this should already be handled if I'm understanding this user journey correctly. Here are the following combinations of local and STDIN resourcegroup:
> 
> 1. User pipes K8s resources along with resourcegroup to `kpt live apply -`, we will use the resourcegroup in stream, and Kptfile in current working directory.
> 2. User pipes K8s resources along with resourcegroup and Kptfile to `kpt live apply -`, we use both resourcegroup and Kptfile in stream.
> 3. User pipes K8s resources along with resourcegroup to `kpt live apply -`, but the current working directory already has a resourcegroup, we throw a multiple inventory error. Perhaps we should always default to using the resourcegroup in the stream for this case and ignore the local directory's resourcegroup?
> 4. User pipes K8s resources WITHOUT resourcegroup, but Kptfile that contains an inventory to `kpt live apply -`, we successfully apply the resources to maintain backwards compatibility with existing `kpt live apply -` functionality.

The CUJs you described is correct. But i'm not talking about the `kpt live apply -`. We should make kpt live be backward compatible with current behavior: `kpt live apply` (*without `-`*) should read the resourceGroup from input resources by looking for kind "ResourceGroup". This guarantees whether user migrate or not, the `kpt live apply` can also work. ''When I'm testing your code, I tried `--rg-file` flag to store the resourceGroup in a custom file and `kpt live apply` raises error about not finding the inventory.

internal/cmdapply/cmdapply.go
yuwenma: I actually has a strong preference to not surface "inventory" in the resourcegroup case. One of the motivations for the `kpt live apply` change is that we got many feedback the "inventory" "resourceGroup" confuse users. It's harder to let them realize the value of `kpt live apply`.

As for your concern about  "we still support the inventory being provided in the Kptfile", thinking from the users' perspective, they use either inventory or resourcegroup, but not using them at the same time (except migration).

mortent:
I agree with @yuwenma here. We should detect the ResourceGroup resource type regardless of the filename. A clarification of the usage of the `--rg-file` flag would be useful. I think I understand the use-case for allowing it to be read from a file while the other resources are read from STDIN, but not clear what the use-case is when we read the other resources from disk.

And regarding the inventory ID, we can't leave it empty. Although instead of setting it to `<NAME>-<NAMESPACE>`, we should leverage the existing resource reference format we are already using for the [`depends-on` annotation](https://kpt.dev/reference/annotations/depends-on/). So we should set it to `kpt.dev/namespaces/<NAMESPACE>/ResourceGroup/<NAME>`

yuwenma:
@rquitales Gentle ping

internal/cmdapply/cmdapply.go
rquitales: Sure - there needs to be a time to make the hard switch anyway, might as well make it now. Have updated the `--rg-file` help message, and other log output to not refer resourcegroup as inventory.

internal/cmdliveinit/cmdliveinit.go
rquitales: Done

rquitales:
> I agree with @yuwenma here. We should detect the ResourceGroup resource type regardless of the filename. A clarification of the usage of the `--rg-file` flag would be useful. I think I understand the use-case for allowing it to be read from a file while the other resources are read from STDIN, but not clear what the use-case is when we read the other resources from disk.
> 
> And regarding the inventory ID, we can't leave it empty. Although instead of setting it to `<NAME>-<NAMESPACE>`, we should leverage the existing resource reference format we are already using for the [`depends-on` annotation](https://kpt.dev/reference/annotations/depends-on/). So we should set it to `kpt.dev/namespaces/<NAMESPACE>/ResourceGroup/<NAME>`

Regarding inventoryID: the latest commit uses the `namespace-name` as the label value for inventoryID - happy to change this to a format we can all agree on. We can't use the depends-on format (ie: `kpt.dev/namespaces/<NAMESPACE>/ResourceGroup/<NAME>`) for the following 2 reasons: `/` is not an allowed character in the value, and value can only contain up to 63 characters. This character limit means that the resource's name and namespace can't be too long... Also, the current logic, as per the initial design doc is to not store the inventoryID in the local resourcegroup object if not specified explicitly when running `kpt live init`. This means that changing the format that we use in future kpt versions might cause breaking issues.

For the `kpt live apply` with custom resourcegroup, we support the `--rg-file` flag as well. So if a customer does not want to use the default `resourcegroup.yaml` filename when running `kpt live init` or `kpt live migrate`, they will also need to explicitly use the `--rg-file` flag with `kpt live apply`. I agree that finding the ResourceGroup inventory by parsing the yaml files would be a nice to have feature, and we can follow up with another PR that introduces this. I think this would need a refactor of the current `live.Load()` function since we don't actually scan any yaml files currently.

rquitales:
@yuwenma Had a conversation with @mortent and we prefer punting off the simplification of the inventoryID handling in the ResourceGroup object. Currently, the inventoryID is stored as a Kubernetes label. There are current restrictions on label formats, specifically, it can be no more than 63 characters long. Using a format that is open ended like `<namespace>-<name>` could cause issues with the 63 char limit. We can hash the `name+namespace` and store it in the cluster's ResourceGroup. However, if we don't also save it in the local resourcegroup.yaml file, as what the design doc describes, this will cause breaking changes in the future if we decide to change the format of the inventoryID.
I propose that we continue storing the inventoryID in the local resourcegroup.yaml until we can solidify on an inventoryID format.

Regarding the `kpt live apply` with custom resource group filename, it seems like the default use case for most users would be to use the default `resourcegroup.yaml` file name. If they specify a custom file name instead, they can still use the `--rg-file` flag when they run `kpt live apply`, for example: `kpt live apply --rg-file=custom-rg.yaml /dir/to/pkg`. If you have a strong preference to include inventory scanning by finding the ResourceGroup GVK, I can refactor the current `live.load` function to support this.

mortent:
@rquitales I synced with @yuwenma on this.
So we should roll back the change for the inventoryID and handle that in a separate effort.

For the `--rg-file` flag, lets take it away and lets rely solely on identifying inventory based on the resource type among the resources provided either in package or through stdin.

rquitales:
> @rquitales I synced with @yuwenma on this. So we should roll back the change for the inventoryID and handle that in a separate effort.
> 
> For the `--rg-file` flag, lets take it away and lets rely solely on identifying inventory based on the resource type among the resources provided either in package or through stdin.

@yuwenma @mortent To clarify, when a user inits or migrates to the newer ResourceGroup object, should we now only support outputting the inventory to a file named `resourcegroup.yaml`? And if they want a custom name for that file, they can choose to manually rename it? Or is the `-rg-file` flag removal only for `kpt live apply/destroy/status` since we'll scan for the inventory information either from the package on disk, or in stream.

mortent:
We can allow users to specify a custom name for the `kpt live init` command. We just remove the `-rg-file` flag from the `apply/destroy/status` commands.

pkg/live/load.go
```diff
@@ -369,6 +296,14 @@ func validateInventory(inventory kptfilev1.Inventory) error {
 
 func generateInventoryObj(inv kptfilev1.Inventory) *unstructured.Unstructured {
 	// Create and return ResourceGroup custom resource as inventory object.
+
+	// When inventoryID is not specified by the local resourcegroup, we generate one using
+	// depends-on annotation format that we store on the live cluster. This implementation detail
+	// is hidden from the local resourcegroup unless the one was explicitly generated by the client.
+	if inv.InventoryID == "" {
```
mortent, line 303: I think the idea was that we would not implement any changes to the inventoryID in this PR? We know that generating an inventoryID this way is not safe.

yuwenma:
@rquitales  Kpt is preparing for a new launch, so we shall complete this feature or put it in a reasonable status.
 Can you resolve Morten's comment and finish the part described in  https://github.com/GoogleContainerTools/kpt/pull/2802#issuecomment-1069554960?

PR #2743, opened 2022-02-04
Update kpt dependencies

martinmaly:
- Upgrade dependencies
- Some more updates

PR #2804, opened 2022-02-16
Update to go1.17

mortent:
Update kpt to require go 1.17. This is in preparation to update the cli-utils library that have this requirement due to dependencies on k8s libraries.


martinmaly:
Indeed. #2743 is the PR where porch branch got upgraded to 1.17 and newer versions of dependencies

PR #2807, opened 2022-02-16
Add Clone Package Test

martinmaly:
Fix issues with git server's reporting of symbolic references


martinmaly:
This PR is also based on #2797

PR #2814, opened 2022-02-18
Revert OCI and Location

martinmaly:
Im preparation to merge porch branch into main:

- Revert "Kpt location parsing (#2688)"
- Revert "oci-support feature branch (#2666)"


droot (approved):
@loudej Can you also take a look at it. Thanks

internal/util/fetch/fetch.go
```diff
@@ -62,34 +76,183 @@ func (c Command) validate(kf *kptfilev1.KptFile) error {
...
+	if len(g.Ref) == 0 {
+		return errors.E(op, errors.MissingParam, fmt.Errorf("must specify ref"))
+	}
+	if len(g.Directory) == 0 {
+		return errors.E(op, errors.MissingParam, fmt.Errorf("must specify directory"))
+	}
+	return nil
+}
+
+// cloneAndCopy fetches the provided repo and copies the content into the
```
loudej, line 97: Just a note - this would be a good file to bookmark - that in a PR to main if there are no diffs or if there are that the main is authoritative.

Unless the text of the code came from the current main (rather than a "commit revert") - then having no diffs should be almost certain.

internal/util/parse/parse.go
```diff
@@ -22,113 +22,18 @@ import (
...
-
-type GitTarget struct {
+type Target struct {
 	kptfilev1.Git
 	Destination string
 }
 
-func GitParseArgs(ctx context.Context, args []string) (GitTarget, error) {
-	g := GitTarget{}
+func GitParseArgs(ctx context.Context, args []string) (Target, error) {
```
loudej, line 35: Same note - this file is good to spot check. Actually - rather than wait for PR to main - you could spot-check now with a copy-paste of the entire raw file from main in your working copy to verify that it remains "unchanged".

pkg/kptfile/kptfileutil/util.go
```diff
@@ -212,104 +196,33 @@ func UpdateKptfile(localPath, updatedPath, originPath string, updateUpstream boo
...
-	const op errors.Op = "kptfileutil.UpdateUpstreamLock"
+// UpdateUpstreamLockFromGit updates the upstreamLock of the package specified
+// by path by using the values from spec. It will also populate the commit
+// field in upstreamLock using the latest commit of the git repo given
+// by spec.
+func UpdateUpstreamLockFromGit(path string, spec *git.RepoSpec) error {
+	const op errors.Op = "kptfileutil.UpdateUpstreamLockFromGit"
 	// read KptFile cloned with the package if it exists
-	kptfile, err := pkg.ReadKptfile(filesys.FileSystemOrOnDisk{}, path)
+	kpgfile, err := pkg.ReadKptfile(filesys.FileSystemOrOnDisk{}, path)
```
loudej, line 206: spin off --- keeping this PR a pure revert is a good idea, but there's a minor local variable typo here that would be nice to fix in main

loudej (approved):
LGTM - the oci-support was intended to be a "light touch" change and these diffs all look like the places where existing code was adjusted when it was git-specific (e.g. where it was coded directly against kptfile.upsteam.git structs)

internal/util/fetch/fetch.go
martinmaly: good point. I will still need to be very careful with the merge into main.

pkg/kptfile/kptfileutil/util.go
martinmaly: Yeah :) I actually fixed this typo and much of my battling with rebases last night was trying to keep this typo fixed; ultimately I gave up; will fix in main.

internal/util/parse/parse.go
martinmaly: ack. Will double check this file too

martinmaly: confirmed. they are near identical; porch contains the filesystem changes which makes them not exactly identical.

internal/util/fetch/fetch.go
martinmaly: Confirmed. this file is near identical; only diff relative to main are filesystem changes.

PR #2816, opened 2022-02-18
Pre-merge Cleanup

martinmaly:
* Tidy go modules
* Fix year in license header


martinmaly:
@droot If your preference is to merge this cleanup into porch branch _before_ merging porch into main, let's merge this PR first and I'll re-do the merge porch --> main PR.
Otherwise I can also merge the big PR first and then do this after. LMK

site/sidebar.md
```diff
@@ -64,10 +64,6 @@
     - [ResourceList](reference/schema/resource-list/)
     - [CRD Status Convention](reference/schema/crd-status-convention/)
     - [Config Connector Status Convention](reference/schema/config-connector-status-convention/)
-  - [Annotations Reference](reference/annotations/)
```
droot, line 67: This shouldn't be removed. This is a generated file. Looks like `porch` branch doesn't have the sidebar template updated.

droot:
> @droot If your preference is to merge this cleanup into porch branch _before_ merging porch into main, let's merge this PR first and I'll re-do the merge porch --> main PR. Otherwise I can also merge the big PR first and then do this after. LMK

Since it's a minor cleanup, I would prefer doing it after our big PR merge. Will be reviewing the big PR soon.

site/sidebar.md
droot: Oh, I see it now. In that case, may be merging this PR first sounds reasonable.

droot (approved):

site/sidebar.md
martinmaly: Thanks

PR #2819, opened 2022-02-22
store the actuation/reconcile status in the inventory object

Liujingfang1:
In the `Store()` function, it saves the actuation and reconcile status in
the InventoryResourceGroup object.

In the `GetObject()` function, it constructs the unstructured object of
the ResourceGroup CR with .status.resourceStatuses assigned.

Updated the unit test to cover this change.


Liujingfang1:
@mortent @karlkfi

pkg/live/inventoryrg.go
```diff
@@ -193,20 +199,52 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+			"status", "resourceStatuses")
 	} else {
 		klog.V(4).Infof("storing inventory (%d) resources", len(objs))
 		err := unstructured.SetNestedSlice(invCopy.UnstructuredContent(),
 			objs, "spec", "resources")
 		if err != nil {
 			return nil, err
 		}
+		err = unstructured.SetNestedSlice(invCopy.UnstructuredContent(),
+			objStatus, "status", "resourceStatuses")
```
karlkfi, line 238: why did we call this `resourceStatuses` and not just `resources`? it's already under the `status` parent map.

pkg/live/inventoryrg.go
```diff
@@ -193,20 +199,52 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+		if err != nil {
+			return nil, err
+		}
+		generation := invCopy.GetGeneration()
+		err = unstructured.SetNestedField(invCopy.UnstructuredContent(),
+			generation, "status", "observedGeneration")
```
karlkfi, line 244: nice catch here. Does nomos cli use the observedGeneration?

karlkfi (approved):
/lgtm

I don't know if i have permission to approve here, but this looks solid.

@Liujingfang1 says that new fields will be ignored by the apiserver if the CRD hasn't been updated, and these are all new fields. So this should be a safe addition.

pkg/live/inventoryrg.go
Liujingfang1: Nomos cli doesn't use the observedGeneration field. It reads the information in `.status.resourceStatuses` field.

pkg/live/inventoryrg.go
Liujingfang1: To make it clear that this field contains the status information for the resources.

karlkfi: just seems redundant... but it's not new to this PR, so whatever.

pkg/live/inventoryrg.go
```diff
@@ -193,20 +199,52 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+	for _, objMeta := range icm.objMetas {
+		status, found := objStatusMap[objMeta]
+		if found {
+			klog.V(4).Infof("storing inventory obj refercence and its status: %s/%s", objMeta.Namespace, objMeta.Name)
+			objStatus = append(objStatus, map[string]interface{}{
+				"group":     objMeta.GroupKind.Group,
+				"kind":      objMeta.GroupKind.Kind,
+				"namespace": objMeta.Namespace,
+				"name":      objMeta.Name,
+				"status":    "UnKnown",
```
mortent, line 213: Why the capitalized `K` here?

Liujingfang1: Fixed it.

mortent (approved):

PR #2822, opened 2022-02-23
added support for hooks

droot:
This PR adds support for `hooks` as a generic mechanism.

This is WIP mainly to test a few ideas.


internal/cmdhook/cmdhook.go
```diff
@@ -0,0 +1,119 @@
+// Copyright 2021 Google LLC
```
martinmaly, line 1: we should update the year. here and in the scripts.

PR #2823, opened 2022-02-23
Start In-Cluster Git Server For e2e Tests

martinmaly:
* Create git-server Deployment resource
* Enable serving empty repository by default

### DIFF
--- a/porch/apiserver/pkg/e2e/suite.go
+++ b/porch/apiserver/pkg/e2e/suite.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


porch/apiserver/pkg/e2e/suite.go
```diff
@@ -356,3 +359,187 @@ func createInitialCommit(t *testing.T, repo *gogit.Repository) {
...
+	ctx := context.TODO()
+
+	// Determine git-server image name. Use the same container registry and tag as the Porch server,
+	// replacing base image name with `git-server`. TODO: Make configurable?
+
+	var porch appsv1.Deployment
+	t.GetF(ctx, client.ObjectKey{
+		Namespace: "porch-system",
+		Name:      "porch-server",
+	}, &porch)
```
justinsb, line 383: This is proving to be a nice pattern!

porch/apiserver/pkg/e2e/suite.go
```diff
@@ -356,3 +359,187 @@ func createInitialCommit(t *testing.T, repo *gogit.Repository) {
...
+							ImagePullPolicy: coreapi.PullIfNotPresent,
+						},
+					},
+				},
+			},
+		},
+	})
+
+	t.Cleanup(func() {
+		t.DeleteE(ctx, &appsv1.Deployment{
```
justinsb, line 432: No need to change, but it might suffice to just delete the namespace

justinsb (approved):

porch/apiserver/pkg/e2e/suite.go
martinmaly: When we run against dev instance, namespace controller is not running so deleting namespace wouldn't quite work, but long-term I'd like to get to the point where deleting the namespace would suffice. For now I'll leave it to possibly catch more issues.

porch/apiserver/pkg/e2e/suite.go
martinmaly: thanks!

PR #2829, opened 2022-02-23
porch: raise memory limit, also remove CPU limit

justinsb:
The memory limit was too tight and we were hitting it (and I don't
think the limit was high enough that we should prioritize the usage -
yet!)

For CPU, the general advice is to set a request but not to set a
limit.


droot (approved):

martinmaly (approved):
Thanks! I was curious what reasonable limit are. ack this was too tight. Thanks for fixing

PR #2834, opened 2022-02-24
Reconnect Watch with Backoff

martinmaly:
Implement simple exponential backoff to reconnect failed watch.

### DIFF
--- a/porch/apiserver/pkg/registry/porch/background.go
+++ b/porch/apiserver/pkg/registry/porch/background.go
@@ -42,6 +42,11 @@ type background struct {
 	cache      *cache.Cache
 }
 
+const (
+	minReconnectDelay = 100 * time.Millisecond
+	maxReconnectDelay = 10 * time.Second
+)
+
 // run will run until ctx is done
 func (b *background) run(ctx context.Context) {
 	klog.Infof("Background routine starting ...")
@@ -56,13 +61,17 @@ func (b *background) run(ctx context.Context) {
 		}
 	}()
 
+	reconnect := newBackoffTimer(minReconnectDelay, maxReconnectDelay)
+	defer reconnect.Stop()
+
 	// Start ticker
 	ticker := time.NewTicker(1 * time.Minute)
 	defer ticker.Stop()
 
 loop:
 	for {
-		if watcher == nil {
+		select {
+		case <-reconnect.channel():
 			var err error
 			klog.Infof("Starting watch ... ")
 			var obj configapi.RepositoryList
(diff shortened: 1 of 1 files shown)


porch/apiserver/pkg/registry/porch/background.go
```diff
@@ -56,15 +61,19 @@ func (b *background) run(ctx context.Context) {
...
 	defer ticker.Stop()
 
 loop:
 	for {
-		if watcher == nil {
+		select {
+		case <-reconnect.channel():
 			var err error
-			klog.Infof("Starting watch ... ")
+			klog.Info("Starting watch ... ")
```
justinsb, line 76: FYI I personally always use Infof, because the much more common failure is forgetting to use `f`, and I don't think the cost of Info vs Infof should be noticeable (though I haven't benchmarked)

porch/apiserver/pkg/registry/porch/background.go
```diff
@@ -73,19 +82,22 @@ loop:
...
-		select {
 		case event, eventOk := <-events:
 			if !eventOk {
 				klog.Errorf("Watch event stream closed. Will restart watch from bookmark %q", bookmark)
 				watcher.Stop()
 				events = nil
 				watcher = nil
+
+				// Initiate reconnect
+				reconnect.reset()
```
justinsb, line 100: I think that because we haven't implemented watch we immediately go into this path

porch/apiserver/pkg/registry/porch/background.go
```diff
@@ -151,3 +163,38 @@ func (b *background) cacheRepository(ctx context.Context, repo *configapi.Reposi
 	}
 	return nil
 }
+
+type backoffTimer struct {
+	min, max, curr time.Duration
+	timer          *time.Timer
+}
+
+func newBackoffTimer(min, max time.Duration) *backoffTimer {
```
justinsb, line 172: There is a backoff somewhere in kubernetes apimachinery, but IIRC it's a bit convoluted

justinsb (approved):

porch/apiserver/pkg/registry/porch/background.go
martinmaly: that has not been my experience. I see the watch working, actually; since we watch CRDs only, the main apiserver handles it. It is when the main apiserver goes down or disconnects (or if porch is running standalone without main apiserver) that I see this code path execute.

porch/apiserver/pkg/registry/porch/background.go
martinmaly: oh, I think I found something I can use ... stay tuned.

martinmaly: decided to leave this as-is for now. the apimachiner backoff could be used but would require additional code to support reset and all that is not worth it.

porch/apiserver/pkg/registry/porch/background.go
martinmaly: :) sg. putting it back.

PR #2837, opened 2022-02-26
Force Push

martinmaly:
For the time being, use force when pushing to remote(s). Eventually
we'll implement conflict resolution but until deletion and conflict
resolution are supported, Force is a good way to avoid hard to recover
scenarios


justinsb (approved):

justinsb:
lgtm.

This is a bigger todo, we might want to file a tacking bug/issue on it.  We might also need to improve the OCI standard to offer similar guarantees.

martinmaly:
filed issue.

PR #2838, opened 2022-02-26
Sort Git Tree Entries Correctly

martinmaly:
The Git tree entry sort oder is very particular. Tree entries are sorted
as if tree entries (directories) had '/' appended to them. Git seems to
rely on this order, causing errors if it is not followed.


justinsb (approved):

justinsb:
lgtm - this rings a bell!  We should also check if there's a helper in go-git, maybe?

martinmaly:
I checked but the code is internal utility.

PR #2841, opened 2022-02-28
feat: Write function to Kptfile

yuwenma:
### description
- Add a boolean flag `--save` to `kpt fn eval`. If set to true, the kpt function will be added to the Kptfile after `kpt fn eval` successfully executes.
- The functionConfig to be written to Kptfile can be FunctionConfigMap, functionConfigPath, no functionConfig, deferring to the arguments. 
- If writing to the Kptfile fails, it won't fail the `kpt fn eval`.
- e2e test coverage:
   
     - `./save-fn/image` runs `kpt fn eval -s -i .. -- [args]`. It writes the fn image and configMap as fnConfig to Kptfile
     - `./save-fn/exec` runs `kpt fn eval -s --exec .. --fn-config=..`. It writes the executable and configPath as fnConfig to Kptfile
     - `./save-fn/match-selector` runs `kpt fn eval -s  --match-kind -i ..`. It writes the function with selectors to Kptfile.
     -  `./save-fn/no-save-when-fail` does not have `Kptfile`, so it only updates the resources, but not write the fn to the Kptfile.

Context: This flag is an easy approach for package authoring.  Another option could be to add a new set of sub commands like `kpt editor` to do CLI-side authoring. But we do not have many feature requirements on these new subcommands.

### DIFF
--- a/thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
+++ b/thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
@@ -13,15 +13,18 @@ import (
 
 	docs "github.com/GoogleContainerTools/kpt/internal/docs/generated/fndocs"
 	"github.com/GoogleContainerTools/kpt/internal/fnruntime"
+	"github.com/GoogleContainerTools/kpt/internal/pkg"
 	"github.com/GoogleContainerTools/kpt/internal/printer"
 	"github.com/GoogleContainerTools/kpt/internal/util/argutil"
 	"github.com/GoogleContainerTools/kpt/internal/util/cmdutil"
 	kptfile "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
+	"github.com/GoogleContainerTools/kpt/pkg/kptfile/kptfileutil"
 	"github.com/GoogleContainerTools/kpt/thirdparty/cmdconfig/commands/runner"
 	"github.com/GoogleContainerTools/kpt/thirdparty/kyaml/runfn"
 	"github.com/google/shlex"
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/kustomize/kyaml/errors"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
 	"sigs.k8s.io/kustomize/kyaml/fn/runtime/runtimeutil"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
@@ -46,6 +49,9 @@ func GetEvalFnRunner(ctx context.Context, parent string) *EvalFnRunner {
 	_ = r.Command.RegisterFlagCompletionFunc("image", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return cmdutil.FetchFunctionImages(), cobra.ShellCompDirectiveDefault
 	})
+	r.Command.Flags().BoolVarP(
+		&r.SafeImage, "save", "s", false,
+		"save the function image and fn-config to Kptfile. Require `--image`.")
 	r.Command.Flags().StringVar(
 		&r.Exec, "exec", "", "run an executable as a function")
 	r.Command.Flags().StringVar(
@@ -90,6 +96,7 @@ type EvalFnRunner struct {
 	OutContent           bytes.Buffer
 	FromStdin            bool
 	Image                string
+	SafeImage            bool
 	Exec                 string
 	FnConfigPath         string
 	RunFns               runfn.RunFns
@@ -102,6 +109,7 @@ type EvalFnRunner struct {
 	IncludeMetaResources bool
 	Ctx                  context.Context
 	Selector             kptfile.Selector
+	dataItems            []string
 }
 
 func (r *EvalFnRunner) runE(c *cobra.Command, _ []string) error {
@@ -109,7 +117,63 @@ func (r *EvalFnRunner) runE(c *cobra.Command, _ []string) error {
 	if err != nil {
 		return err
 	}
-	return cmdutil.WriteFnOutput(r.Dest, r.OutContent.String(), r.FromStdin, printer.FromContextOrDie(r.Ctx).OutStream())
+	if err = cmdutil.WriteFnOutput(r.Dest, r.OutContent.String(), r.FromStdin,
+		printer.FromContextOrDie(r.Ctx).OutStream()); err != nil {
+		return err
+	}
+	if r.SafeImage {
+		r.SaveFnToKptfile()
+	}
+	return nil
+}
+
+// AddImageToKptfile adds the function image and config to Kptfile
+func (r *EvalFnRunner) SaveFnToKptfile() {
+	pr := printer.FromContextOrDie(r.Ctx)
+	kptFile, err := pkg.ReadKptfile(filesys.FileSystemOrOnDisk{}, r.Dest)
+	if err != nil {
+		pr.Printf("image not added: Kptfile not exists\n")
+		return
+	}
+	// TODO(yuwenma): Right now we cannot tell a function is a mutator or validator. Once kpt supports
+	// OCI images, we can add annotations to image and find out the function type from these annotations.
+	if kptFile.Pipeline == nil {
+		kptFile.Pipeline = &kptfile.Pipeline{}
+	}
+	if kptFile.Pipeline.Mutators == nil {
+		kptFile.Pipeline.Mutators = []kptfile.Function{}
+	} else {
+		for _, m := range kptFile.Pipeline.Mutators {
+			if m.Name == r.Image || m.Image == r.Image {
+				pr.Printf("skip adding image: already exists in Kptfile\n")
+				return
+			}
+		}
+	}
+	var newMutator kptfile.Function
+	if r.FnConfigPath != "" {
+		newMutator = kptfile.Function{Name: r.Image, Image: r.Image, ConfigPath: r.FnConfigPath}
+	} else {
+		data := map[string]string{}
+		for i, s := range r.dataItems {
+			kv := strings.SplitN(s, "=", 2)
+			if i == 0 && len(kv) == 1 {
+				continue
+			}
+			data[kv[0]] = kv[1]
+		}
+		if len(data) == 0 {
+			newMutator = kptfile.Function{Name: r.Image, Image: r.Image}
+		} else {
+			newMutator = kptfile.Function{Name: r.Image, Image: r.Image, ConfigMap: data}
+		}
+	}
+	kptFile.Pipeline.Mutators = append(kptFile.Pipeline.Mutators, newMutator)
+	if err = kptfileutil.WriteFile(r.Dest, kptFile); err != nil {
+		pr.Printf("image is not added to Kptfile: %v\n", err)
+		return
+	}
+	pr.Printf("image is added to Kptfile\n")
 }
 
 // getCLIFunctionConfig parses the commandline flags and arguments into explicit
@@ -232,6 +296,8 @@ func (r *EvalFnRunner) preRunE(c *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
+	} else if r.AddImage {
+		return fmt.Errorf("--add requires --image")
 	}
 	if err := cmdutil.ValidateImagePullPolicyValue(r.ImagePullPolicy); err != nil {
 		return err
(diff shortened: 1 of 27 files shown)


yuwenma:
### Update
I pretty like the name `--save` as proposed in https://github.com/GoogleContainerTools/kpt/issues/2249. Updated

site/reference/cli/fn/eval/README.md
```diff
@@ -130,6 +130,10 @@ fn-args:
   it doesn't exist. Structured results emitted by the functions are aggregated and saved
   to `results.yaml` file in the specified directory.
   If not specified, no result files are written to the local filesystem.
+
+--safe:
+  Safe the function image and fn-config to Kptfile. Require `--image`.
```
droot, line 135: Few comments:

- display the short flag as well. I am guessing `-s` is the short flag
- Instead of saving `Require --image`, may be, we should specify the behavior that declarative form is saved only if the function is successfully evaluated. That should cover all the validity requirements for the command.
- I think we can improve the text. I think on a high level, this flag turns the imperative invocation of the function into declarative form and adds to `pipeline` of the package.

site/reference/cli/fn/eval/README.md
```diff
@@ -155,6 +159,11 @@ $ kpt fn eval DIR -i gcr.io/example.com/my-fn --fn-config my-fn-config
 $ kpt fn eval DIR -i gcr.io/example.com/my-fn:v1.0.0 -- foo=bar
 ```
 
+```shell
+# execute container my-fn and add it to Kptfile after execution.
+$ kpt fn eval DIR -s -i gcr.io/example.com/my-fn:v1.0.0 -- foo=bar
+```
```
droot, line 165: So one thing we will have to do is exclude the docs from this PR because as soon as the PR is merged, docs changes become live on the `kpt.dev` and so this reference docs should actually be merged after this change is released to the public.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -90,6 +96,7 @@ type EvalFnRunner struct {
 	OutContent           bytes.Buffer
 	FromStdin            bool
 	Image                string
+	AddImage             bool
```
droot, line 99: Can we rename it to something like `SaveToKptfile` or `AddToPipeline` or `SaveCommand` to reflect the usage better.

Because if user use `--exec`, we will save that as well to the `Kptfile`, right ?

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -102,14 +109,71 @@ type EvalFnRunner struct {
...
+		return err
+	}
+	if r.AddImage {
+		r.AddImageToKptfile()
+	}
+	return nil
+}
+
+// AddImageToKptfile adds the function image and config to Kptfile
+func (r *EvalFnRunner) AddImageToKptfile() {
```
droot, line 131: pl. refer to renaming suggestion above.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -102,14 +109,71 @@ type EvalFnRunner struct {
...
+		for _, m := range kptFile.Pipeline.Mutators {
+			if m.Name == r.Image || m.Image == r.Image {
+				pr.Printf("skip adding image: already exists in Kptfile\n")
+				return
+			}
+		}
+	}
+	var newMutator kptfile.Function
+	if r.FnConfigPath != "" {
+		newMutator = kptfile.Function{Name: r.Image, Image: r.Image, ConfigPath: r.FnConfigPath}
```
droot, line 155: Can we change `kptFile` variable to `kf` or something to differentiate between `kptfile` pkg import from the local variable. Will help with the readability.

droot, line 155: One thing we will have to do is: if `fnConfigPath` is absolute, that will have to be converted to relative otherwise it will result in a invalid Pipeline.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -102,14 +109,71 @@ type EvalFnRunner struct {
...
+			newMutator = kptfile.Function{Name: r.Image, Image: r.Image}
+		} else {
+			newMutator = kptfile.Function{Name: r.Image, Image: r.Image, ConfigMap: data}
+		}
+	}
+	kptFile.Pipeline.Mutators = append(kptFile.Pipeline.Mutators, newMutator)
+	if err = kptfileutil.WriteFile(r.Dest, kptFile); err != nil {
+		pr.Printf("image is not added to Kptfile: %v\n", err)
+		return
+	}
```
droot, line 175: I would suggest splitting this function in two parts (two separate smaller functions):

1. First part converts a given `fn eval` invocation into `kptfile.Function`
2. Second part takes a `kptfile.Function` pointer and saves it to a given `kptFile` struct

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -102,14 +109,71 @@ type EvalFnRunner struct {
...
+	// TODO(yuwenma): Right now we cannot tell a function is a mutator or validator. Once kpt supports
+	// OCI images, we can add annotations to image and find out the function type from these annotations.
+	if kptFile.Pipeline == nil {
+		kptFile.Pipeline = &kptfile.Pipeline{}
+	}
+	if kptFile.Pipeline.Mutators == nil {
+		kptFile.Pipeline.Mutators = []kptfile.Function{}
+	} else {
+		for _, m := range kptFile.Pipeline.Mutators {
+			if m.Name == r.Image || m.Image == r.Image {
```
droot, line 147: We should add support for `exec` and `selectors`.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -232,6 +296,8 @@ func (r *EvalFnRunner) preRunE(c *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
+	} else if r.AddImage {
+		return fmt.Errorf("--add requires --image")
```
droot, line 300: We shouldn't need this check. Only pre-requisite for `--save` is that function should evaluate successfully.

thirdparty/cmdconfig/commands/cmdeval/cmdeval_test.go
```diff
@@ -462,5 +466,97 @@ func TestCmd_flagAndArgParsing_Symlink(t *testing.T) {
 	assert.Equal(t, filepath.Join("path", "to", "pkg", "dir"), r.RunFns.Path)
 }
 
+func TestAddImageToKptfile(t *testing.T) {
```
droot, line 469: Pl. take a look at `e2e` tests for `fn eval`. They are under `e2e/testdata/fn-eval`. Add tests for at least the following:

1. On successful `fn eval`, function is added to the pipeline
2. On failure, function is not added
3. --image and --exec 
4. ConfigMap and configPath option
5. selectors if we add support for them

thirdparty/cmdconfig/commands/cmdeval/cmdeval_test.go
```diff
@@ -462,5 +466,97 @@ func TestCmd_flagAndArgParsing_Symlink(t *testing.T) {
...
+			r.AddImageToKptfile()
+			assert.Equal(t, strings.TrimSpace(tc.ExpectedStdout), strings.TrimSpace(outBuf.String()))
+			actualKf, err := pkg.ReadKptfile(filesys.FileSystemOrOnDisk{}, r.Dest)
+			if tc.ExpectedPipeline != nil {
+				assert.NoError(t, err)
+				assert.Equal(t, actualKf.Pipeline.String(), tc.ExpectedPipeline.String())
+			}
+		})
+	}
+}
```
droot, line 559: We are trying to move away from these style tests and favoring `e2e` tests. `fn eval` has legacy code (thats the reason you see it in `thirdparty`). Would recommend removing this test :(

All our function-catalog tests are written e2e style.

droot (commented):
Thanks for picking this up @yuwenma . I have some comments. Happy to discuss offline if you need to discuss or clarify something.

site/reference/cli/fn/eval/README.md
```diff
@@ -130,6 +130,10 @@ fn-args:
   it doesn't exist. Structured results emitted by the functions are aggregated and saved
   to `results.yaml` file in the specified directory.
   If not specified, no result files are written to the local filesystem.
+
+--safe:
```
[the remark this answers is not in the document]
yuwenma: Ah, good catch. thanks

site/reference/cli/fn/eval/README.md
yuwenma: done.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
yuwenma: It won't be saved if users use `--exec`. I added a check in `preRunE` to make sure this flag only works with `--image` flag.

Renamed.

site/reference/cli/fn/eval/README.md
yuwenma: I see. I'll comment them out and add a TODO

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
yuwenma: I don't see examples that `kpt fn eval` support selectors or `kpt fn render` support exec function. Do you mind giving some pointers how these two works in this cmds?

thirdparty/cmdconfig/commands/cmdeval/cmdeval_test.go
yuwenma: I see. thanks

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
yuwenma: ah, found the exec case in the https://kpt.dev/book/04-using-functions/01-declarative-function-execution

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
yuwenma: Found the exec example. but still not see selectors in `kpt fn eval`

site/reference/cli/fn/eval/README.md
yuwenma: Here's the other PR https://github.com/GoogleContainerTools/kpt/pull/2848

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
yuwenma: I see. good catch! Thanks

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
yuwenma: done.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
yuwenma: oh nevermind, it's the --match-XXX.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
yuwenma: done

droot (approved):

droot:
@yuwenma Thanks for addressing the comments. Feel free to merge it in.

morgante:
@yuwenma It looks like this refuses to update an existing function in the `Kptfile`:
```
$ kpt fn eval -i gcr.io/kpt-fn/set-namespace:v0.1.0 --save -- namespace=config-controld
[RUNNING] "gcr.io/kpt-fn/set-namespace:v0.1.0"
[PASS] "gcr.io/kpt-fn/set-namespace:v0.1.0" in 1.5s
skip adding image: already exists in Kptfile
```

As discussed in #2249, this should *update* an existing declaration if one is found. This makes it easier for consumer guides to provide imperative docs.

PR #2854, opened 2022-03-02
Load Packages From Git Tags

martinmaly:
Simple tags without slashes are assumed to be version tags. All packages
in the tree associated with the tag will be loaded.
Tags containing slashes are of the form <package path>/<version>.
Specific package will be loaded at that version, if it exists.

### DIFF
--- a/porch/repository/pkg/git/git.go
+++ b/porch/repository/pkg/git/git.go
@@ -43,6 +43,8 @@ const (
 	refMain plumbing.ReferenceName = "refs/heads/main"
 	// TODO: support customizable pattern of draft branches.
 	refDraftPrefix = "refs/heads/drafts/"
+	refTagsPrefix  = "refs/tags/"
+	refHeadsPrefix = "refs/heads/"
 )
 
 type GitRepository interface {
@@ -154,6 +156,7 @@ func (r *gitRepository) ListPackageRevisions(ctx context.Context) ([]repository.
 
 	var main *plumbing.Reference
 	var drafts []repository.PackageRevision
+	var result []repository.PackageRevision
 
 	for {
 		ref, err := refs.Next()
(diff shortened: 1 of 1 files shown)


porch/repository/pkg/git/git.go
```diff
@@ -496,6 +515,58 @@ func parseDraftName(draft *plumbing.Reference) (name, revision string, err error
...
+
+	dirTree, err := tree.Tree(path)
+	if err != nil {
+		klog.Warningf("Skipping %q; cannot find %q (corrupted repository?): %w", name, path, err)
+		return nil, nil
+	}
+
+	if kptfileEntry, err := dirTree.FindEntry("Kptfile"); err == nil {
+		if !kptfileEntry.Mode.IsRegular() {
+			klog.Warningf("Skippping %q: cannot find Kptfile: %w", name, err)
```
justinsb, line 552: Should this message be "is not a file"?

porch/repository/pkg/git/git.go
```diff
@@ -396,16 +404,27 @@ func (r *gitRepository) discoverFinalizedPackages(main *plumbing.Reference) ([]r
 		return nil, err
 	}
 
+	var revision string
+	rev := ref.Name().String()
+	switch {
+	case strings.HasPrefix(rev, refTagsPrefix):
+		revision = rev[len(refTagsPrefix):]
```
justinsb, line 411: FYI, I like strings.TrimPrefix here.  It would be nice if there was a way to avoid having to duplicate HasPrefix TrimPrefix (maybe if TrimPrefix returned a bool as well), but I haven't found one.

porch/repository/pkg/git/git.go
```diff
@@ -496,6 +515,58 @@ func parseDraftName(draft *plumbing.Reference) (name, revision string, err error
 	return name, revision, nil
 }
 
+func (r *gitRepository) loadTaggedPackages(tag *plumbing.Reference) ([]repository.PackageRevision, error) {
+	name := tag.Name().String()
+	if !strings.HasPrefix(name, refTagsPrefix) {
+		return nil, fmt.Errorf("invalid tag ref name: %q", name)
```
justinsb, line 521: I'm surprised that `go vet` isn't complaining about this.  Typically it gets annoyed when you have an `else` if you're returning from the `if-true` block

justinsb (approved):

porch/repository/pkg/git/git.go
martinmaly: sg. I'll switch to TrimPrefix. easy way to avoid accidental +-1 error :)

porch/repository/pkg/git/git.go
martinmaly: odd indeed. I fixed it and using TrimPrefix as per above also

porch/repository/pkg/git/git.go
martinmaly: good catch! fixed.

PR #2858, opened 2022-03-02
Add Deployment Repository Field

martinmaly:
Use first-class field to mark repository as 'deployment repository'.

Packages in a deployment repository are considered ready to be deployed
by the deployment mechanism.

### DIFF
--- a/porch/controllers/pkg/apis/porch/v1alpha1/types.go
+++ b/porch/controllers/pkg/apis/porch/v1alpha1/types.go
@@ -53,6 +53,8 @@ type RepositorySpec struct {
 	Title string `json:"title,omitempty"`
 	// User-friendly description of the repository
 	Description string `json:"description,omitempty"`
+	// The repository is a deployment repository; final packages in this repository are deployment ready.
+	Deployment bool `json:"deployment"`
 	// Type of the repository (i.e. git, OCI)
 	Type RepositoryType `json:"type,omitempty"`
 	// Content stored in the repository (i.e. Function, PackageRevision - the literal values correspond to the API resource names).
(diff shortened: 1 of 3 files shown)


porch/controllers/pkg/apis/porch/v1alpha1/types.go
```diff
@@ -53,6 +53,8 @@ type RepositorySpec struct {
 	Title string `json:"title,omitempty"`
 	// User-friendly description of the repository
 	Description string `json:"description,omitempty"`
+	// The repository is a deployment repository; final packages in this repository are deployment ready.
+	Deployment bool `json:"deployment"`
```
droot, line 57: I noticed that we are missing `yaml` tags here. Would be good to add them if this type definition is eventually used in CLIs etc where we accept YAML as input for this.

droot (commented):
A drive by comment.

porch/controllers/pkg/apis/porch/v1alpha1/types.go
martinmaly: Interestingly, I've been using these types with kubectl for some time and even without yaml annotations they appear to work all correctly. Likewise, k8s types lack the yaml tags, too.

justinsb: `yaml` tags shouldn't be necessary.  `sigs.k8s.io/yaml` infers them from json tags.  If you accidentally use the other yaml library then things break without the yaml tags, but my understanding is that other things break too so this is a good break, and it's a sign I should fix my import!

justinsb: Now if someone could just teach `goimports` which yaml import I want :-)

justinsb (approved):

porch/controllers/pkg/apis/porch/v1alpha1/types.go
droot: hmm..we had to add `yaml` tags to kptfile schema, I can't remember the reason (it was probably something to do with generating openapi schema).

PR #2861, opened 2022-03-02
Package Initialization

martinmaly:


porch/engine/pkg/engine/init.go
```diff
@@ -0,0 +1,72 @@
...
+	b, err := yaml.MarshalWithOptions(kptfile, &yaml.EncoderOptions{SeqIndent: yaml.WideSequenceStyle})
+	if err != nil {
+		return repository.PackageResources{}, nil, fmt.Errorf("failed to serialize Kptfile: %w", err)
+	}
+
+	if resources.Contents == nil {
+		resources.Contents = map[string]string{}
+	}
+
+	path := path.Join(m.spec.Subpackage, kptfilev1.KptFileName)
```
justinsb, line 64: Nit: Shadowing `path` will give confusing results.  I try to do something like `kptfilePath := path.Join` or just `p := path.Join`

porch/engine/pkg/engine/engine.go
```diff
@@ -102,6 +115,14 @@ func (cad *cadEngine) CreatePackageRevision(ctx context.Context, repositoryObj *
 
 func (cad *cadEngine) mapTaskToMutation(ctx context.Context, obj *api.PackageRevision, task *api.Task) (mutation, error) {
 	switch task.Type {
+	case api.TaskTypeInit:
```
justinsb, line 118: I do wonder if an init could be a clone with no upstream, but you're right that this is likely clearer!

justinsb (approved):
LGTM!

porch/engine/pkg/engine/engine.go
martinmaly: We could also do clone from a 'singleton' (similar to what Docker does with `FROM scratch`.

The growing problem is representing actions as tasks, especially as Porch does more things automatically, things that there are no tasks for.

PR #2871, opened 2022-03-05
Return TypeMeta from Porch API

martinmaly:
Kubectl has trouble printing resources otherwise.


justinsb (approved):
Thanks for tracking this down!

PR #2883, opened 2022-03-08
kpt pkg init generates dummy package context

droot:
With this change:
`kpt pkg init` will generate `package-context.yaml` file with following content:

```
# package-context.yaml

apiVersion: v1
kind: ConfigMap
metadata:
  name: kptfile.kpt.dev
  annotations:
    config.kubernetes.io/local-config: "true"
data:
  name: example
```

### DIFF
--- a/internal/builtins/pkg_context.go
+++ b/internal/builtins/pkg_context.go
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	pkgContextFile = "package-context.yaml"
+	PkgContextFile = "package-context.yaml"
 	pkgContextName = "kptfile.kpt.dev"
 )
 
@@ -118,7 +118,7 @@ data: {}
 		return nil, err
 	}
 	annotations := map[string]string{
-		kioutil.PathAnnotation: path.Join(path.Dir(kptfilePath), pkgContextFile),
+		kioutil.PathAnnotation: path.Join(path.Dir(kptfilePath), PkgContextFile),
 	}
 
 	for k, v := range annotations {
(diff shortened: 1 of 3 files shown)


internal/builtins/pkg_context.go
```diff
@@ -131,3 +131,18 @@ data: {}
 	})
 	return cm, nil
 }
+
+// DummyPkgContext returns content for package context that contains
+// placeholder value for package name. This will be used to create
+// abstract blueprints.
+func DummyPkgContext() string {
```
natasha41575, line 138: WDYT about changing the name to `AbstractPkgContext` so the name is more clear about what it is?

internal/builtins/pkg_context.go
```diff
@@ -131,3 +131,18 @@ data: {}
...
+	return fmt.Sprintf(`apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: %s
+  annotations:
+    config.kubernetes.io/local-config: "true"
+data:
+  name: example
+`, pkgContextName)
```
natasha41575, line 147: Can we use this `DummyPkgContext` to replace lines 106-114 above, since the name is being set in lines 129-130 anyways? Then we will not have to write out this yaml twice.

internal/builtins/pkg_context.go
droot: Yes, I like your suggestion. Done.

internal/builtins/pkg_context.go
droot: That's good point.  Done!

droot:
Thanks @natasha41575 Addressed both the comments.

internal/builtins/pkg_context.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	pkgContextFile = "package-context.yaml"
+	PkgContextFile = "package-context.yaml"
 	pkgContextName = "kptfile.kpt.dev"
```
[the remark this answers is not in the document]
droot: Yes, eventually.

internal/builtins/pkg_context.go
droot: This is resolved in another discussion.

PR #2884, opened 2022-03-09
Create Porch e2e Test Workflow

martinmaly:


.github/workflows/porch-e2e.yml
```diff
@@ -0,0 +1,78 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Porch End-to-End Tests
+on:
+  push:
+    paths-ignore:
+      - "docs/**"
+      - "site/**"
```
droot, line 20: May be we can trigger this on changes in porch codebase ?

martinmaly: I'll look to restrict more; we shouldn't do 'just porch' because of dependencies on kpt libraries (i.e. renderer) that can cause Porch to fail if bugs are introduced.

martinmaly: done; ptal. I'd be uncomfortable restricting more of the kpt code at this point. We can evaluate if we see that this test is causing a bottleneck

.github/workflows/porch-e2e.yml
```diff
@@ -0,0 +1,98 @@
...
+        working-directory: ./porch
+      - name: Wait For Server
+        run: |
+          for i in {1..10}; do
+            sleep $i
+            if kubectl api-resources; then
+              echo "Server is up"
+              break
+            fi
+          done
```
droot, line 95: Any specific reason for doing the above steps (install kind, install porch, wait) inline here. I was thinking if we can have these steps in a setup script or go-setup-function then it will be easier to run these locally as well right ?

.github/workflows/porch-e2e.yml
droot: hmm..you are right, we should use `paths` instead of `paths-ignore` ([gh workflow docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) to trigger on selected paths (porch/, kpt/internal/render...) ?

martinmaly: I don't think we should use the inverse logic because it will make it harder to update. I'd like to (at least while we have high velocity development) err on the side of automatically running tests when we add directories etc. than missing adding paths to the list :(

droot: That's a fair point.

.github/workflows/porch-e2e.yml
martinmaly: Some of the steps seem not applicable to run locally (such as running the engineerd/setup-kind workflow step).

While we can encode some of this in a script (and I'm fine doing it) deep scripts make testing much harder. So for Porch we follow philosophy - environment is set up externally to the test, and then we run go test within the environment.

The e2e test has enough smarts to detect appropriate way to start Git server and will spin up one in-memory or via k8s depending on needs.

droot (approved):

PR #2886, opened 2022-03-10
Generate Context When Creating Deployment Package

martinmaly:
When creating (cloning or creating new) package revision in a deployment
repository, call `builtins/gen-pkg-context` function to generate package
context.

### DIFF
--- a/porch/engine/pkg/engine/engine.go
+++ b/porch/engine/pkg/engine/engine.go
@@ -24,6 +24,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/GoogleContainerTools/kpt/internal/fnruntime"
 	"github.com/GoogleContainerTools/kpt/pkg/fn"
 	api "github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
 	configapi "github.com/GoogleContainerTools/kpt/porch/controllers/pkg/apis/porch/v1alpha1"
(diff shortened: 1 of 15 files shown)


porch/engine/pkg/engine/engine.go
```diff
@@ -102,6 +103,15 @@ func (cad *cadEngine) CreatePackageRevision(ctx context.Context, repositoryObj *
 		mutations = append(mutations, mutation)
 	}
 
+	// If creating a package in a deployment repository, generate context
+	if repositoryObj.Spec.Deployment {
```
droot, line 107: One observation: `kpt CLI`'s [implementation](https://github.com/GoogleContainerTools/kpt/blob/main/internal/util/get/get.go#L125) is deviating significantly from this server side logic.

May be not a huge pain at the moment, but can become an issue as we evolve the deployment hook more.

porch/engine/pkg/engine/testdata/context/expected/bucket.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: storage.cnrm.cloud.google.com/v1beta1
+kind: StorageBucket
+metadata:
+  name: bucket-name
+  namespace: bucket-namespace
+  annotations:
+    cnrm.cloud.google.com/project-id: bucket-project
+    config.kubernetes.io/path: 'bucket.yaml'
```
droot, line 8: Do we expect `path` annotations when a package is written out ?

droot (approved):
Drive by comments.

porch/engine/pkg/engine/engine.go
martinmaly: Can we change kpt to create library of the 'customize for deployment' functionality, similarly to what we did with render (it will need to accept function runner for anything other than builtins but this could be a better solution. How hard would it be to refactor kpt to do that?

porch/engine/pkg/engine/testdata/context/expected/bucket.yaml
martinmaly: Yes for this output. The packages arrive from the server with annotations and are compared to golden files without modification for verbatim validation.

porch/engine/pkg/engine/engine.go
droot: we have a package `hook` under [internal/hook](https://github.com/GoogleContainerTools/kpt/blob/main/internal/hook/executor.go#L60) which exposes the this as a library (It also get's invoked in `kpt pkg get` command as library only`). 

So might be able to use it with little changes.

PR #2894, opened 2022-03-15
WIP: include functionConfigs by defaults

droot:
This PR intends (WIP) to the following:

- Do not treat `functionConfigs` as meta-resources. It means the folllowing:
       - Include `functionConfigs` by default in the function inputs in `kpt fn`
       - Apply `functionConfigs` to the cluster in `kpt live` commands unless `local-config` annotation is specified
- `--include-meta-resources` flag in `kpt fn commands` will treat only `Kptfile` as the meta-resource
- `kpt fn render` now supports `--include-meta-resources` flag like other `kpt fn` commands  

Note to reviewers: I am still working on fixing the tests. Some tests will become invalid (the one that tests `functionConfigs` behavior) and will be removed.


e2e/testdata/fn-eval/include-meta-resources-v1alpha1/.expected/config.yaml
```diff
@@ -16,5 +16,6 @@ runCount: 2
 testType: eval
 image: gcr.io/kpt-fn/set-namespace:v0.1.3
 includeMetaResources: true
+exitCode: 1
```
mengqiy, line 19: Why?

e2e/testdata/fn-render/fn-success-with-stderr/.expected/diff.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/starlark.yaml b/starlark.yaml
+index edbb132..1b85ac9 100644
+--- a/starlark.yaml
++++ b/starlark.yaml
+@@ -11,7 +11,6 @@
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-
```
mengqiy, line 9: This diff is not useful. We should to tweak it to get rid of this diff.

e2e/testdata/fn-render/generator-subpkgs/.expected/diff.patch
```diff
@@ -1,25 +1,44 @@
...
-+  name: parent-map
-+  namespace: def
-+data:
-+  parent-key: parent-value
+diff --git a/db/fn-config.yaml b/db/fn-config.yaml
+index 66b0d31..7a0f3af 100644
+--- a/db/fn-config.yaml
++++ b/db/fn-config.yaml
+@@ -1,17 +1,3 @@
+-# Copyright 2021 Google LLC
```
mengqiy, line 6: It seems the license header become part of the diff. It's verbose and not useful.
I guess we don't want the license header to be part of the diff.

e2e/testdata/fn-render/modify-legacy-path-annotation/.expected/diff.patch
```diff
@@ -2,3 +2,15 @@ diff --git a/deployment.yaml b/newfilename.yaml
...
 rename to newfilename.yaml
+diff --git a/starlark-fn.yaml b/starlark-fn.yaml
+index f4393c2..a56ce8a 100644
+--- a/starlark-fn.yaml
++++ b/starlark-fn.yaml
+@@ -11,7 +11,6 @@
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-
```
mengqiy, line 13: Same comment here and below

mengqiy:
> kpt should support a way to exclude KRM resources (especially Kptfile and local-configs) so that users who are impacted by this breaking change can use the exclude directive to migrate to this change.

This is important step. Otherwise, users may not have a clear way to upgrade.

> Horizontal transforms (set-namespace, set-labels etc. should handle unrecognized resources such as Kptfile gracefully).

It will impact not only horizontal transformers but also any functions uses OpenAPI (e.g. kubeval).

We also need to verify if the comments in Kptfile can be preserved after going through KRM functions.

e2e/testdata/fn-eval/include-meta-resources-v1alpha1/.expected/config.yaml
droot: Actually this PR removes `--includeMetaResources` commandline flag, so instead of deleting this test, I re-purposed to test the error if someone specifies this flag. I should probably add comment explaining that.

e2e/testdata/fn-render/generator-subpkgs/.expected/diff.patch
droot: Removed the license header.

e2e/testdata/fn-render/modify-legacy-path-annotation/.expected/diff.patch
droot: Fixed.

bharathkkb:
>  Apply functionConfigs to the cluster in kpt live commands unless local-config annotation is specified

@droot I am wondering what is the usecase for this? Afaict in a cluster functionConfig would not mean anything and only seems useful in the context of a function.

droot:
> > Apply functionConfigs to the cluster in kpt live commands unless local-config annotation is specified
> 
> @droot I am wondering what is the usecase for this? Afaict in a cluster functionConfig would not mean anything and only seems useful in the context of a function.

I should have worded it better (updated it in the PR description). Basically from `kpt live` standpoint, the model is very simple: A KRM resource is either `local` or not. If it is `local` mark it as `local-config` else it is assumed to be applied to the cluster. From function's perspective, input can come from the items in the resource list, or functionConfig field in resourceList. It shouldn't be be treated in any special manner if it is specified as `configPath` in the pipeline. The current model around `meta-resources` leads to lot of confusion and restriction in `kpt fn` commands (and edge cases as well), and so far we haven't seen any benefits.

mengqiy (approved):

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -57,8 +57,6 @@ func GetEvalFnRunner(ctx context.Context, parent string) *EvalFnRunner {
 		&r.Exec, "exec", "", "run an executable as a function")
 	r.Command.Flags().StringVar(
 		&r.FnConfigPath, "fn-config", "", "path to the function config file")
-	r.Command.Flags().BoolVarP(
-		&r.IncludeMetaResources, "include-meta-resources", "m", false, "include package meta resources in function input")
```
natasha41575, line 61: If we want to print a helpful message when they are using this flag, we can't remove these lines. As you asked, I created a draft PR for how to gracefully let users know this is no longer available.

See https://github.com/GoogleContainerTools/kpt/pull/2939

droot: Yes, re-introduced `include-meta-resources` as hidden flag and throwing an error as suggested in your PR. Thanks.

droot:
@mortent Can you take a look at the `kpt live` specific changes ? Thanks.

pkg/live/rgpath.go
```diff
@@ -35,34 +33,26 @@ func (r *ResourceGroupPathManifestReader) Read() ([]*unstructured.Unstructured,
...
 
 	for _, n := range nodes {
-		relPath, _, err := kioutil.GetFileAnnotations(n)
-		if err != nil {
-			return objs, err
-		}
-		if fcPaths.Has(relPath) && !isExplicitNotLocalConfig(n) {
-			continue
-		}
+		// Note(droot): Since we stopped givign special treatment to functionConfigs
```
yuwenma, line 47: clean up the code?

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -304,6 +307,12 @@ func checkFnConfigPathExistence(path string) error {
 }
 
 func (r *EvalFnRunner) preRunE(c *cobra.Command, args []string) error {
+	// Let users know that --include-meta-resources is not longer necessary
```
yuwenma, line 310: no longer

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -304,6 +307,12 @@ func checkFnConfigPathExistence(path string) error {
...
+	// since meta resources are included by default.
+	if r.IncludeMetaResources {
+		return fmt.Errorf("--include-meta-resources is no longer necessary because meta resources are now included by default")
```
yuwenma, line 313: I like this user friendly message!

thirdparty/cmdconfig/commands/cmdsource/cmdsource.go
```diff
@@ -37,15 +35,21 @@ func GetSourceRunner(ctx context.Context, name string) *SourceRunner {
...
+		PreRunE: r.preRunE,
 	}
 	c.Flags().StringVarP(&r.Output, "output", "o", cmdutil.Stdout,
 		fmt.Sprintf("output resources are written to stdout in provided format. Allowed values: %s|%s", cmdutil.Stdout, cmdutil.Unwrap))
 	c.Flags().StringVar(&r.FunctionConfig, "fn-config", "",
 		"path to function config file.")
 	c.Flags().BoolVar(&r.IncludeMetaResources,
 		"include-meta-resources", false, "include package meta resources in the command output")
+	if err := c.Flags().MarkHidden("include-meta-resources"); err != nil {
+		panic(err)
```
yuwenma, line 47: once this is triggered, users will see the tracing log with kpt source code? Do we want to expose that?

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
yuwenma: Shall we return errors or moving forward? 

Returning errors could break some user auto-process, suggest giving a warning and moving forward instead.

yuwenma (approved):
some minor comments about backward compatible handling

pkg/live/rgpath.go
droot: Done. Moved the comment to an appropriate location.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
droot: Done.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
droot: I thought about just giving warning first but given that the default behavior for `--include-meta-resources` ("false") is no longer supported so keeping the flag around didn't make sense, so had to take it out.

droot:
> Overall LGTM. Some minor comments about backward compatible handling

Thanks for the review Yuwen. Addressed the comments.

mortent (approved):

droot:
Thanks everyone for reviewing!

nadworny:
hi @droot
I'm a bit confused if this part has been implemented?
`kpt should support a way to exclude KRM resources (especially Kptfile and local-configs) so that users who are impacted by this breaking change can use the exclude directive to migrate to this change.`

If so, how can I make the `kpt fn source` to exclude certain resources, in my case the `local-config`?

droot:
> hi @droot I'm a bit confused if this part has been implemented? `kpt should support a way to exclude KRM resources (especially Kptfile and local-configs) so that users who are impacted by this breaking change can use the exclude directive to migrate to this change.`
> 
> If so, how can I make the `kpt fn source` to exclude certain resources, in my case the `local-config` ones?

We added mechanism to exclude resources on the basis of labels, annotations, group/version/kind/namespace/name in `kpt fn eval` and `kpt fn render`. For example,  https://kpt.dev/book/04-using-functions/02-imperative-function-execution?id=specifying-exclusions, https://kpt.dev/book/04-using-functions/01-declarative-function-execution?id=specifying-exclusions

I don't think we added to kpt fn source.

If you are combining `kpt fn source` with `kpt fn eval`, then you can use the exclude support in `kpt fn eval`. Can you share how how you are using `kpt fn source` or your use-case, I might be able to suggest a solution ?

nadworny:
Thanks for the quick reply.

The use case is following:
- We are using terraform to apply k8s resources generated by the kpt
- First we get the package and render it
- After we use `kpt fn source` to generate the yaml files which are then read by the terraform
- Since resources of kind Kptfile or SetLabels aren't valid CRDs in our cluster, we want to exclude them

droot:
> Thanks for the quick reply.
> 
> The use case is following:
> 
> * We are using terraform to apply k8s resources generated by the kpt
> * First we get the package and render it
> * After we use `kpt fn source` to generate the yaml files which are then read by the terraform
> * Since resources of kind Kptfile or SetLabels aren't valid CRDs in our cluster, we want to exclude them completely
> 
> If I understand it correctly, the function exclude capability mentioned above will simply not apply transformations to the excluded resources but will still pass the whole pkg to the next pipe.

Thanks for the sharing the use-case. The best solution will be for `kpt fn` commands to provide more flexbility around filtering resources especially for `-o unwrap` option where people are piping to other tools. Sharing a workaround here:

It requires using another tool called [kubectl-slice](https://github.com/patrickdappollonio/kubectl-slice). `kubectl-slice` can filter out resources on the basis of comma separates names.

```
kpt fn source -o unwrap|kubectl-slice --exclude-name kptfile.kpt.dev,mariadb --stdout
```

Hope this helps.

nadworny:
I wasn't aware of kubectl-slice - nice tool!

I have found another workaround using yq:
`kpt fn source . | yq '.items[] |= del(select(.kind == "Kptfile"))'`

PR #2895, opened 2022-03-15
Porch Command e2e Tests

martinmaly:
Initial set of e2e tests of Porch integration commands


droot (approved):
This is neat.

martinmaly:
I'm glad you think so :) I suspect that the test may need a bit more power for some more advanced scenarios but we can refactor when it is needed.

PR #2903, opened 2022-03-16
support compiled-in fns in func-runner

mengqiy:
I added starlark and apply-replacements for now.


porch/func/go.mod
```diff
@@ -3,33 +3,60 @@ module github.com/GoogleContainerTools/kpt/porch/func
...
+	github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/starlark v0.3.0
+	github.com/GoogleContainerTools/kpt-functions-sdk/go/fn v0.0.0-20220316202203-f9115a993ebd
 	google.golang.org/grpc v1.44.0
 	google.golang.org/protobuf v1.27.1
 	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/klog/v2 v2.40.1
 	sigs.k8s.io/kustomize/kyaml v0.13.3
 )
 
+replace github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/starlark v0.3.0 => github.com/mengqiy/kpt-functions-catalog/functions/go/starlark v0.0.0-20220316204754-e2611d41524d
```
mengqiy, line 16: Will remove this after https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/778 merges

porch/func/go.mod
```diff
@@ -3,33 +3,60 @@ module github.com/GoogleContainerTools/kpt/porch/func
 go 1.17
 
 require (
+	github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/apply-replacements v0.0.0-20220316042125-d3eb289008bc
```
mengqiy, line 6: We can use a released version after we release `apply-replacements`

mengqiy:
@martinmaly This is a draft PR. Early feedbacks are welcome.

porch/func/go.mod
martinmaly: sounds good.

porch/func/go.mod
martinmaly: ack, sounds good.

porch/func/internal/builtinevaluator.go
```diff
@@ -0,0 +1,65 @@
...
+type builtinEvaluator struct {
+	pb.UnimplementedFunctionEvaluatorServer
+
+	fnMapping map[string]fn.ResourceListProcessor
+}
+
+func newBuiltInEvaluator() (*builtinEvaluator, error) {
+	return &builtinEvaluator{
+		fnMapping: map[string]fn.ResourceListProcessor{
+			"gcr.io/kpt-fn/starlark:v0.3.0": &starlark.Processor{},
```
martinmaly, line 36: Something to think about (I don't have a good answer for it yet) is to what extent, especially for built-ins (where we can keep fairly high back compatibility bar) we can use the same compiled-in function for whole range of versions...

This is great for now though.

porch/func/internal/evaluator.go
```diff
@@ -32,6 +33,43 @@ import (
...
+	}
+	resp, err = e.exec.EvaluateFunction(ctx, req)
+	if errors.As(err, &ufe) {
+		return nil, status.Errorf(codes.NotFound, ufe.Error())
+	}
+	return resp, err
+}
+
+type executableEvaluator struct {
+	pb.UnimplementedFunctionEvaluatorServer
```
martinmaly, line 71: We should not need this line anymore since the `evaluator` struct above implements the gRPC interface and we don't expose the `executableEvaluator` as gRPC (it is delegated to by the `evaluator`

porch/func/internal/builtinevaluator.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/apply-replacements/replacements"
+	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/starlark/starlark"
+	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
+	pb "github.com/GoogleContainerTools/kpt/porch/func/evaluator"
+	"k8s.io/klog/v2"
+)
+
+type builtinEvaluator struct {
+	pb.UnimplementedFunctionEvaluatorServer
```
martinmaly, line 28: ditto. thhis line should be unnecessary; we don't need to use the builtinEvalutor as the direct gRPC server endpoint.

martinmaly (commented):
This looks great!
only a couple of nitpicks. I found this which may be helpful when adding a test (sorry, I notice that I didn't write any tests; my bad).

https://pkg.go.dev/google.golang.org/grpc/test
in particular, the buffconn may be helpful.

martinmaly (commented):
For the presubmit errors. Please run `make tidy` in `porch` that should update all go.mod and go.sum files; that appears to have been the cause of the faiure.

mengqiy:
Closing this in favor of  #2934

PR #2904, opened 2022-03-17
Package Type API

martinmaly:
Add PackageRevision `type` field to capture whether the packge is draft,
proposed, of final.

### DIFF
--- a/porch/api/porch/types_packagerevisions.go
+++ b/porch/api/porch/types_packagerevisions.go
@@ -40,13 +40,21 @@ type PackageRevisionList struct {
 	Items []PackageRevision
 }
 
+type PackageRevisionType string
+
+const (
+	PackageRevisionTypeDraft    PackageRevisionType = "draft"
+	PackageRevisionTypeProposed PackageRevisionType = "proposed"
+	PackageRevisionTypeFinal    PackageRevisionType = "final"
+)
+
 // PackageRevisionSpec defines the desired state of PackageRevision
 type PackageRevisionSpec struct {
-	PackageName string `json:"packageName,omitempty"`
-
-	Revision string `json:"revision,omitempty"`
-
+	PackageName    string `json:"packageName,omitempty"`
+	Revision       string `json:"revision,omitempty"`
 	RepositoryName string `json:"repository,omitempty"`
+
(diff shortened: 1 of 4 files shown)


porch/api/porch/types_packagerevisions.go
```diff
@@ -40,14 +40,22 @@ type PackageRevisionList struct {
...
+	PackageRevisionTypeFinal    PackageRevisionType = "final"
+)
 
+// PackageRevisionSpec defines the desired state of PackageRevision
+type PackageRevisionSpec struct {
+	PackageName    string `json:"packageName,omitempty"`
+	Revision       string `json:"revision,omitempty"`
 	RepositoryName string `json:"repository,omitempty"`
 
+	Type PackageRevisionType `json:"type,omitempty"`
```
justinsb, line 57: Let's not change it in this PR (I assume it'll break the whole chain), but this reads more like `state` or `lifecycle` than `type` to me.

porch/api/porch/v1alpha1/types_packagerevisions.go
```diff
@@ -41,14 +41,22 @@ type PackageRevisionList struct {
...
 }
 
-// PackageRevisionSpec defines the desired state of PackageRevision
-type PackageRevisionSpec struct {
-	PackageName string `json:"packageName,omitempty"`
+type PackageRevisionType string
 
-	Revision string `json:"revision,omitempty"`
+const (
+	PackageRevisionTypeDraft    PackageRevisionType = "draft"
```
justinsb, line 47: I _think_ we generally capitalize user-visible "enums" (e.g. Draft, Proposed, Final)

justinsb (approved):

porch/api/porch/types_packagerevisions.go
martinmaly: I decided to do this in this PR ... there will be potential for less long-term fallout. Thanks!

porch/api/porch/v1alpha1/types_packagerevisions.go
martinmaly: Looking at prior art, this is correct. I fixed these constants but there are more to fix too that we currently lower-case. I'll add tracking issue for that.

PR #2907, opened 2022-03-17
Repo Registration Test

martinmaly:
This is an experimental unit test.

* Add RESTMapper to Porch client to avoid using discovery
* Create first test for Repository registration command

### DIFF
--- a/internal/cmdreporeg/command_test.go
+++ b/internal/cmdreporeg/command_test.go
@@ -13,3 +13,215 @@
 // limitations under the License.
 
 package cmdreporeg
+
+import (
+	"encoding/json"
+	"flag"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/GoogleContainerTools/kpt/internal/printer/fake"
+	"github.com/google/go-cmp/cmp"
+	"gopkg.in/yaml.v3"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+var (
+	update = flag.Bool("update", false, "update golden files")
+)
+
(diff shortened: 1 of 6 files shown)


internal/cmdreporeg/command_test.go
```diff
@@ -13,3 +13,215 @@
...
+	"testing"
+
+	"github.com/GoogleContainerTools/kpt/internal/printer/fake"
+	"github.com/google/go-cmp/cmp"
+	"gopkg.in/yaml.v3"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+var (
+	update = flag.Bool("update", false, "update golden files")
```
justinsb, line 35: Nit: I personally have used env vars for this ... I've created a library that I'm pretty happy with, but unfortunately it is off on an experimental branch: https://github.com/justinsb/kpt/blob/main_integration/porch/functions/pkg/testhelpers/golden.go#L53  Not a blocker because we can move across fairly easily later!

internal/util/porch/client.go
```diff
@@ -34,7 +36,10 @@ func CreateClient(flags *genericclioptions.ConfigFlags) (client.Client, error) {
 		return nil, err
 	}
 
-	c, err := client.New(config, client.Options{Scheme: scheme})
+	c, err := client.New(config, client.Options{
+		Scheme: scheme,
+		Mapper: createRESTMapper(),
```
justinsb, line 41: I didn't see how this is used (and it doesn't appear to be test code).  Am I missing it?

martinmaly: When I was building the test against the fake server the go-client would by default interrogate the server openapi specs to construct the rest mapping. So to avoid having to fake all of the openapi methods I provided own rest mapper for the types that the CLI uses.

So far I haven't found a better way :(

PR #2909, opened 2022-03-17
Add ci step to deploy firebase functions

sdowell:
The existing deployment step deploys the static hosted content on merge. This change is intended to additionally add a step which deploys the recently added firebase functions upon merge.


liamfallon:
Firebase is not used anymore

PR #2911, opened 2022-03-17
Git Test Repository Scripts

martinmaly:
The scripts can be used to create an initial git repository to
subsequently populate with commits and archive it for use with
tests.

### DIFF
--- a/porch/hack/create-test-repo.sh
+++ b/porch/hack/create-test-repo.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


justinsb:
To be honest I don't love these, because they're pretty opaque.  But it is a nice simple approach for complex git repos.  I'll think for a few minutes ... the alternative in my mind is having a directory under testdata and constructing the git repo from it, but that doesn't do well with history.

justinsb:
I do think we should use it sparingly (I'd much rather we built up the git repo in the tests by creating packages or otherwise), but I can see that these are valuable!

justinsb (approved):

martinmaly:
Thanks. I expect as we implement more of the package lifecycle we will be able to build more complex repositories directly, and use this technique for edge cases only. I have the same reservations but couldn't find a better way that would preserve history (I considered series of directories to be checked into Git, but that seemed strange too.

PR #2912, opened 2022-03-17
Git Repo Loading and Tests

martinmaly:
- Improve Git repository loading
- Recognize package revision types (draft, proposed)
- Add more tests


porch/repository/pkg/git/git.go
```diff
@@ -42,9 +42,12 @@ import (
 const (
 	refMain plumbing.ReferenceName = "refs/heads/main"
 	// TODO: support customizable pattern of draft branches.
-	refDraftPrefix = "refs/heads/drafts/"
-	refTagsPrefix  = "refs/tags/"
-	refHeadsPrefix = "refs/heads/"
+	refDraftPrefix        = "refs/heads/drafts/"
+	refProposedPrefix     = "refs/heads/proposed/"
```
justinsb, line 46: If we do need more metadata in future, we could attach git notes (or do what gerrit does, with a "metadata" branch)

porch/repository/pkg/git/git.go
```diff
@@ -211,12 +174,12 @@ func (r *gitRepository) ListPackageRevisions(ctx context.Context) ([]repository.
 }
 
 func (r *gitRepository) CreatePackageRevision(ctx context.Context, obj *v1alpha1.PackageRevision) (repository.PackageDraft, error) {
-	main, err := r.repo.Reference(refMain, true)
-	if err != nil {
-		return nil, err
+	var base plumbing.Hash
+	if main, err := r.repo.Reference(refMain, true); err == nil {
```
justinsb, line 178: Nit: I feel like checking the error is a good idea here, both because it _might_ be something serious (I'm guessing we're anticipating a "not found" type error).  But also checking for not-found also makes the code more self-documenting.

porch/repository/pkg/git/git.go
```diff
@@ -643,3 +630,106 @@ func (r *gitRepository) getRepo() (string, error) {
...
+	}
+
+	// Fetch
+	switch err := r.repo.Fetch(&gogit.FetchOptions{
+		RemoteName: originName,
+		Auth:       auth,
+		Tags:       gogit.AllTags,
+	}); err {
+	case nil: // OK
+	case gogit.NoErrAlreadyUpToDate:
```
justinsb, line 647: Well that's a frustrating API :-)

porch/repository/pkg/git/git.go
```diff
@@ -643,3 +630,106 @@ func (r *gitRepository) getRepo() (string, error) {
...
+	// There may be risks involved such as automated rebase which may lead to incorrect results on the package.
+	// One possibility is to replay package history ... something to consider in the future.
+	for name, remoteRef := range remoteBranches {
+		localRef, ok := localBranches[name]
+
+		if !ok {
+			localRef = plumbing.NewHashReference(plumbing.ReferenceName(name), remoteRef.Hash())
+			// local branch doesn't exist. create it
+			if err := r.repo.Storer.SetReference(localRef); err != nil {
+				klog.Warningf("Skipping failed local ref %q: %v", name, err)
```
justinsb, line 696: Do we expect this?  Maybe just return an error and fail the whole update?

porch/repository/pkg/git/git.go
```diff
@@ -643,3 +630,106 @@ func (r *gitRepository) getRepo() (string, error) {
...
+			}
+		} else if remoteRef.Hash() != localRef.Hash() {
+			remoteCommit, err := r.repo.CommitObject(remoteRef.Hash())
+			if err != nil {
+				klog.Warningf("Skipping unresolvable remote reference %s: %v", remoteRef, err)
```
justinsb, line 701: This also seems like we've got ourselves into an "impossible" situation => return error?

porch/repository/pkg/git/git.go
```diff
@@ -643,3 +630,106 @@ func (r *gitRepository) getRepo() (string, error) {
...
+				continue
+			}
+			localCommit, err := r.repo.CommitObject(localRef.Hash())
+			if err != nil {
+				klog.Warningf("Overwriting unresolvable local reference %s: %v", localRef, err)
+				new := plumbing.NewHashReference(localRef.Name(), remoteCommit.Hash)
+				if err := r.repo.Storer.SetReference(new); err != nil {
+					klog.Errorf("Failed to set local reference: %s to %s", new.Name(), new.Hash())
```
justinsb, line 709: return error here as well (?)

porch/repository/pkg/git/git.go
```diff
@@ -643,3 +630,106 @@ func (r *gitRepository) getRepo() (string, error) {
...
+				continue
+			}
+
+			// If local commit is ancestor of remote, fast-forward local commit to match
+			ancestor, err := localCommit.IsAncestor(remoteCommit)
+			if err != nil {
+				klog.Warningf("Failed to determine whether %s is ancestor of %s: %v", localRef, remoteRef, err)
+			}
+
+			// TODO: Better conflict resolution policy
```
justinsb, line 720: So in my mind, the remote git store is the source of truth.  I guess the question is ... what happened here to cause the conflict?  I would imagine we would _always_ take the remote git store value, but there could well be a case I'm missing...

porch/repository/pkg/git/git_test.go
```diff
@@ -284,3 +285,207 @@ func initRepo(repo *gogit.Repository) error {
...
+
+	revisions, err := git.ListPackageRevisions(ctx)
+	if err != nil {
+		t.Fatalf("Failed to list packages from %q: %v", tarfile, err)
+	}
+	if got, want := len(revisions), 0; got != want {
+		t.Errorf("Number of packges in empty repository: got %d, want %d", got, want)
+	}
+
+	// TODO: Enable the rest of the test when gogit can push into an empty repository.
```
justinsb, line 329: I do think it can, it just needs us to do a little dance IIRC ... we can try to figure it out together!

porch/repository/pkg/git/package.go
```diff
@@ -139,7 +141,24 @@ func (p *gitPackageRevision) GetUpstreamLock() (kptfile.Upstream, kptfile.Upstre
...
-				Commit:    p.sha.String(),
+				Commit:    p.commit.String(),
 			},
 		}, nil
 }
+
+func (p *gitPackageRevision) getPackageRevisionType() v1alpha1.PackageRevisionType {
+	switch {
+	case p.ref == nil || p.ref.Name().IsTag():
+		return v1alpha1.PackageRevisionTypeFinal
```
justinsb, line 152: Does `ref == nil`  mean Final?

justinsb (approved):

porch/repository/pkg/git/git.go
martinmaly: notes do seem like a useful technique!

porch/repository/pkg/git/git.go
martinmaly: ack, done.

porch/repository/pkg/git/git.go
martinmaly: yes :) Errors used quite a bit to signal success, and some of the errors are are named as errors (`ErrEmptyRemoteRepository`)

porch/repository/pkg/git/git.go
martinmaly: sg.

porch/repository/pkg/git/git.go
martinmaly: sg

porch/repository/pkg/git/git.go
martinmaly: done

porch/repository/pkg/git/git.go
martinmaly: race conditions can happen. Porch is processing package update; in parallel the remote repo gets new commit.

So long as Porch is the exclusive user of the Git repo all should be fine, but with other actors involved, we may see conflicts where draft branch is updated to 2 different, divergent, commits. May be mostly theoretical scenario.

porch/repository/pkg/git/git_test.go
martinmaly: sure! I debugged into the go-git internals and it _seemed_ to me that the code is overly conservative. I'll find some time to dive deeper into this.

porch/repository/pkg/git/package.go
martinmaly: at this point the field should never be nil. I didn't want to panic the server ... that seemed cruel :)

I'll just turn this into an (internal) error since that's what it would be if it were to occur.

PR #2920, opened 2022-03-20
Support calling into executor pods directly

justinsb:
If suitable function pods are already running, we can call into them
directly.

This is essentially a proof-of-concept; in practice we will want to
launch these pods ourselves / manage them etc.

### DIFF
--- a/porch/apiserver/pkg/apiserver/apiserver.go
+++ b/porch/apiserver/pkg/apiserver/apiserver.go
@@ -168,13 +168,21 @@ func (c completedConfig) New() (*PorchServer, error) {
 	renderer := kpt.NewRenderer()
 
 	cache := cache.NewCache(c.ExtraConfig.CacheDirectory, credentialResolver)
-	cad, err := engine.NewCaDEngine(
+	options := []engine.EngineOption{
 		engine.WithCache(cache),
-		engine.WithGRPCFunctionRuntime(c.ExtraConfig.FunctionRunnerAddress),
 		engine.WithCredentialResolver(credentialResolver),
 		engine.WithRenderer(renderer),
 		engine.WithReferenceResolver(referenceResolver),
-	)
+	}
+
+	if c.ExtraConfig.FunctionRunnerAddress != "" {
+		options = append(options, engine.WithGRPCFunctionRuntime(c.ExtraConfig.FunctionRunnerAddress))
+	} else {
+		ns := "porch-functions-system"
+		options = append(options, engine.WithKubeFunctionRuntime(coreClient, ns))
+	}
+
+	cad, err := engine.NewCaDEngine(options...)
 	if err != nil {
 		return nil, err
 	}
--- a/porch/config/deploy/5-rbac.yaml
+++ b/porch/config/deploy/5-rbac.yaml
@@ -30,3 +30,15 @@ rules:
 - apiGroups: ["flowcontrol.apiserver.k8s.io"]
   resources: ["flowschemas", "prioritylevelconfigurations"]
   verbs: ["get", "watch", "list"]
+
+---
+
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: porch-function-executor
+rules:
+# Needed to launch / read function executor pods
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "watch", "list"]
--- a/porch/engine/pkg/engine/k8sruntime.go
+++ b/porch/engine/pkg/engine/k8sruntime.go
@@ -1,0 +1,170 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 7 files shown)


justinsb:
WDYT @mengqiy - do you want to take this and build on it, or should we go another way?

porch/apiserver/pkg/apiserver/apiserver.go
```diff
@@ -168,13 +168,21 @@ func (c completedConfig) New() (*PorchServer, error) {
...
 		engine.WithRenderer(renderer),
 		engine.WithReferenceResolver(referenceResolver),
-	)
+	}
+
+	if c.ExtraConfig.FunctionRunnerAddress != "" {
+		options = append(options, engine.WithGRPCFunctionRuntime(c.ExtraConfig.FunctionRunnerAddress))
+	} else {
+		ns := "porch-functions-system"
+		options = append(options, engine.WithKubeFunctionRuntime(coreClient, ns))
```
martinmaly, line 182: this seems to read as though the runtimes are mutually exclusive. We still need high performance runtime for cached functions.

(since there appears to be no example of what the Kube function pod looks like, it is hard to tell what expected performance they will have)

porch/config/deploy/5-rbac.yaml
```diff
@@ -30,3 +30,7 @@ rules:
 - apiGroups: ["flowcontrol.apiserver.k8s.io"]
   resources: ["flowschemas", "prioritylevelconfigurations"]
   verbs: ["get", "watch", "list"]
+# Needed to launch / read executor pods
```
martinmaly, line 33: I'm not sure if it is possible to restrict to the `porch-functions-system` namespace. Do you happen to know?

porch/engine/pkg/engine/k8sruntime.go
```diff
@@ -0,0 +1,170 @@
...
+	address := k.pod.Status.PodIP
+	if address == "" {
+		return fmt.Errorf("pod did not have podIP")
+	}
+	address += ":8888"
+
+	klog.Infof("dialing grpc function runner %q", address)
+
+	// TODO: pool connections
+	cc, err := grpc.Dial(address, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
martinmaly, line 142: is the intention for this to work for arbitrary functions?
I understand we can use image rewriting (which will have issues with image signatures later on), including an example of a concrete function would help.

porch/apiserver/pkg/apiserver/apiserver.go
justinsb: They are currently mutually exclusive, and I changed this PR so that they aren't enabled by default (I think of it as a first step to start investigating this path).

I can add a "try multiple engines" runtime, but we should try to figure out what we need.

There are some example servers on my experimental branch, and they use [something similar to your code](https://github.com/justinsb/kpt/blob/main_integration/porch/functions/pkg/yo/krmfunction/server.go), except the functions [implement the server directly](https://github.com/justinsb/kpt/blob/main_integration/porch/config/samples/functions/main.go) rather than being invoked as binaries.

Still lots of experiments to be done here, e.g. should we multiplex functions into an image?  How long should we run a pod for?  Is the performance good enough?  My impression is that the performance is good enough, but I'm also (manually) pre-launching all the pods!

porch/config/deploy/5-rbac.yaml
justinsb: Good call - it is.  We would just create a RoleBinding rather than a ClusterRoleBinding.  A RoleBinding can bind either to a ClusterRole or a Role, but then only binds in that Namespace.

porch/engine/pkg/engine/k8sruntime.go
justinsb: Yes, this works for arbitrary functions - if they implement the GRPC protocol.  We _could_ take a non-enlightened function and just overlay your existing function-runner onto it.  But technically it's possible that a function might not be safe to run in parallel / multiple times in the same pod.  I don't know if we want to require that functions are safe to run in parallel / multiple times (maybe supporting an opt-out mechanism).  If we're going to do that, maybe we should just require them to build in GRPC support!

justinsb:
Split RBAC roles so that we only grant pod permissions on the porch-functions-system namespace

porch/apiserver/pkg/apiserver/apiserver.go
justinsb: Experimental "multiple-engines" runtime, if we need it: https://github.com/GoogleContainerTools/kpt/pull/2937

porch/engine/pkg/engine/k8sruntime.go
martinmaly: let's track this as a follow-up.

martinmaly (approved):

porch/engine/pkg/engine/k8sruntime.go
mengqiy: I'm working a PoC that has a different flavor of pod evaluator. It doesn't force every KRM function to support gRPC.
It 1) creates pod 2) evaluates function 3) brings it down. More concretely, it does the following:
- Create a pod with 1 regular container (containing the real KRM function) and 1 initContainer (containing a gRPC wrapper server). These 2 containers share the emptyDir volume.
- In initContainer, it copies a wrapper gRPC server binary to the emptyDir volume.
- In the regular container, we run the gPRC server which will invoke the KRM function binary.

I should be able to create a draft PR today.

PR #2928, opened 2022-03-22
Include api and controllers (APIs) in kpt module

martinmaly:
Include the `api` and `controllers` modules directly inside `kpt`.
BY doing so the use of pseudo-modules is no longer necessary and
`go get` on the `kpt` module is possible.


martinmaly:
Hmm ... for some reason kpt build scripts don't seem to like this change. We may have to go with #2929 at least short-term until we can figure out what is tripping the build up. It doesn't seem quite clear form the errors, and they don't happen when running locally.

martinmaly:
Will need to investigate why this breaks kpt build :)

PR #2929, opened 2022-03-22
Kpt depends on specific version of Porch

martinmaly:
By having kpt module depend on specific version of `api` and `controllers` there
is no need to use pseudo-modules. The redirects in sub-modules continue to work
but it will require update to kpt to take dependency on newer versions of `api`
and `controllers`.

This is an alternative to #2928

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,8 +3,8 @@ module github.com/GoogleContainerTools/kpt
 go 1.17
 
 require (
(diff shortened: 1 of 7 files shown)


mengqiy (approved):

mengqiy:
Thank you for creating this PR!
I'm going to merge it since we don't have a clue about how to make https://github.com/GoogleContainerTools/kpt/pull/2928 CI passing.

PR #2931, opened 2022-03-23
Create Deterministic Yaml Output

martinmaly:
The go-yaml serializer orders nodes deterministically. Use it to create output
that can be compared reliably.

### DIFF
--- a/e2e/porch_test.go
+++ b/e2e/porch_test.go
@@ -30,7 +30,7 @@ import (
 
 	"github.com/GoogleContainerTools/kpt/pkg/test/porch"
 	"github.com/google/go-cmp/cmp"
-	"gopkg.in/yaml.v3"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
 const (
--- a/e2e/porch_test.go
+++ b/e2e/porch_test.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/GoogleContainerTools/kpt/pkg/test/porch"
 	"github.com/google/go-cmp/cmp"
+	"gopkg.in/yaml.v3"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


e2e/porch_test.go
```diff
@@ -88,6 +89,10 @@ func runTestCase(t *testing.T, git string, tc porch.TestCaseConfig) {
 
 		err := cmd.Run()
 
+		if command.Yaml {
+			reorderYamlStdout(t, &stdout)
```
droot, line 93: Do we want the command to render a stable output ?

e2e/porch_test.go
```diff
@@ -108,6 +113,25 @@ func runTestCase(t *testing.T, git string, tc porch.TestCaseConfig) {
...
 
+func reorderYamlStdout(t *testing.T, buf *bytes.Buffer) {
+	var data interface{}
+	if err := yaml.Unmarshal(buf.Bytes(), &data); err != nil {
+		// not yaml.
+		return
+	}
+
+	var stable bytes.Buffer
+	encoder := yaml.NewEncoder(&stable)
```
droot, line 124: I am little nervous using another yaml serializer. Should we stick to "sigs.k8s.io/kustomize/kyaml/yaml" ?

e2e/porch_test.go
martinmaly: This would likely require changes to kyaml writers because they are currently in use and they generate the unstable output.
Unless the contract on ResourceList defines ordering (which I don't think it does) I would rather leave the primary implementation as-is to not make it adhere to a different contract, and have the test compensate for order issues.

That said I assume that dealing with ordering is something kpt tests must do already (for example test for `kpt fn source` ... I can look how the ordering issues are handled there.

e2e/porch_test.go
martinmaly: actually, sigs.k8s.io/kustomize/kyaml/yaml may do since it uses (a clone of an earlier version) of the same module. The deterministic ordering was introduced in this yaml encoder in v2 so with some luck kyaml inherited it.

martinmaly: kyaml fork of go-yaml works.

martinmaly:
Updated; please take a look.

droot (approved):

PR #2933, opened 2022-03-23
add local-config annotation to kptfile

droot:
With this change, `Kptfile` created on `kpt pkg init` will have local-config annotation set to "true".

### DIFF
--- a/internal/cmdinit/cmdinit.go
+++ b/internal/cmdinit/cmdinit.go
@@ -35,6 +35,7 @@ import (
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
+	"sigs.k8s.io/kustomize/kyaml/kio/filters"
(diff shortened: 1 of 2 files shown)


internal/cmdinit/cmdinit.go
```diff
@@ -104,6 +105,10 @@ func (r *Runner) runE(c *cobra.Command, args []string) error {
 					NameMeta: yaml.NameMeta{
 						Name: r.Name,
 					},
+					// mark Kptfile as local-config
+					Annotations: map[string]string{
+						filters.LocalConfigAnnotation: "true",
```
yuwenma, line 110: we also have a local-config const var in our new sdk
`fn.KptLocalConfig`

yuwenma (approved):
One minor comment.

internal/cmdinit/cmdinit.go
droot: There is other code in kpt (for ex. `kpt live`) that also uses this annotation. We can switch to using new const in all places as separate effort.

PR #2934, opened 2022-03-23
support compiled-in functions in porch apiserver

mengqiy:
This PR supersedes https://github.com/GoogleContainerTools/kpt/pull/2903.
Comparing to #2903, the advantages of implementing it this way:
- Faster since we remove one hop when evaluating builtin functions
- It implements `kpt.FunctionRuntime` interface. And it lives in the `engine` module which is independent from `apiserver` module. It will be easier to integrate with kpt CLI in the future if needed.

The disadvantage:
- there will be more dependencies to manage


porch/engine/pkg/engine/builtinruntime.go
```diff
@@ -0,0 +1,92 @@
...
+	fnMapping map[string]fnsdk.ResourceListProcessor
+}
+
+func newBuiltinRuntime() *builtinRuntime {
+	return &builtinRuntime{
+		fnMapping: map[string]fnsdk.ResourceListProcessor{
+			"gcr.io/kpt-fn/apply-replacements:unstable":                                                                fnsdk.ResourceListProcessorFunc(replacements.ApplyReplacements),
+			"gcr.io/kpt-fn/apply-replacements@sha256:c7a1b2d96e7f02e8aca6b1d15c15542c73f5a9f072c7fc6a3569160f8662e056": fnsdk.ResourceListProcessorFunc(replacements.ApplyReplacements),
+			"gcr.io/kpt-fn/set-namespace:unstable":                                                                     fnsdk.ResourceListProcessorFunc(transformer.SetNamespace),
+			"gcr.io/kpt-fn/set-namespace@sha256:36bf1a791436a75c0108c071bdb34c2cfe97329d548ac47329119dad87d6c6e2":      fnsdk.ResourceListProcessorFunc(transformer.SetNamespace),
```
droot, line 41: Just a note for us: so this mapping needs to be updated whenever we updated the vendor function's version. And for the short term, we will only maintain one version of a function.

droot: I was wondering what are those `SHA1` are for ? probably testing ?

droot (commented):
Sweet! I have some drive my questions. @martinmaly will be the one approving it.

porch/engine/pkg/engine/builtinruntime.go
```diff
@@ -0,0 +1,92 @@
...
+		return fmt.Errorf("func eval failed: %w", err)
+	}
+	y, err := rl.ToYAML()
+	if err != nil {
+		return fmt.Errorf("failed to marshal resourceList as yaml: %w", err)
+	}
+	if _, err = w.Write(y); err != nil {
+		return fmt.Errorf("failed to write function runner output: %w", err)
+	}
+	return nil
```
martinmaly, line 91: In my experiments with in-process function execution I came up with different implementation for this (though I am not sure whether one approach has benefits over the other). Would you mind taking a look and seeing if there is functionality that is available with one approach vs. the other?

https://github.com/GoogleContainerTools/kpt/blob/main/porch/engine/pkg/kpt/eval.go#L63-L69

porch/engine/pkg/engine/delegatingruntime.go
```diff
@@ -0,0 +1,76 @@
...
+	"io"
+
+	v1 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
+	"github.com/GoogleContainerTools/kpt/pkg/fn"
+	"github.com/GoogleContainerTools/kpt/porch/engine/pkg/kpt"
+)
+
+type delegatingRuntime struct {
+	builtin *builtinRuntime
+	grpc    *grpcRuntime
```
martinmaly, line 29: for more composability, consider what if this were `FunctionRunner` interface

porch/engine/pkg/engine/error.go
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package engine
+
+import "fmt"
+
+type UnsupportedFunctionError struct {
```
martinmaly, line 19: should this be part of the FunctionRuntime contract?
(if so we'd want to declare this error in kpt: https://github.com/GoogleContainerTools/kpt/tree/main/pkg/fn)

please check with @droot if he'd be ok with moving this there

porch/engine/pkg/engine/options.go
```diff
@@ -46,9 +42,28 @@ func WithCache(cache *cache.Cache) EngineOption {
...
+		runtime, err := newGRPCFunctionRuntime(address)
+		if err != nil {
+			return fmt.Errorf("failed to create function runtime: %w", err)
+		}
+		engine.runtime = runtime
+		return nil
+	})
+}
+
+func WithDelegatingFunctionRuntime(address string) EngineOption {
```
martinmaly, line 64: we have only few options for function runtimes but consider whether different pattern for constructing them would work better, particularly given the comment above - what if delegating runtime just delegated to another FunctionRuntime...

(I don't have specific desired outcome in mind, just something to consider)

martinmaly (changes requested):
Couple things for consideration in other comments;

The main thing blocking my approval is that I'd like to see is test validation (see dedicated comment). Rest is desired but optional.

And preferably some more test scenarios (i.e. unit test) so e2e test is not the only line of defense.

In a follow-up, we may be able to clean up the experiment I'd put in place over here: 

https://github.com/GoogleContainerTools/kpt/blob/main/porch/engine/pkg/kpt/eval.go
https://github.com/GoogleContainerTools/kpt/tree/main/porch/engine/pkg/internal

porch/engine/pkg/engine/builtinruntime.go
mengqiy: > I was wondering what are those SHA1 are for ? probably testing ?

Some users prefer to use SHA instead of tag. Add the SHA here ensures we can serve these users.

We will be using released version for these functions. They are not changing as rapidly as `unstable` tag.
e.g. We will use `gcr.io/kpt-fn/set-namespace:v0.3.0` and it SHA when we import set-namespace:v0.3.0 in `go.mod`.

porch/apiserver/go.mod
```diff
@@ -19,8 +19,8 @@ require (
 	go.opentelemetry.io/otel/sdk v0.20.0
 	go.opentelemetry.io/otel/sdk/metric v0.20.0
 	google.golang.org/grpc v1.44.0
-	k8s.io/api v0.23.2
-	k8s.io/apimachinery v0.23.2
+	k8s.io/api v0.23.4
+	k8s.io/apimachinery v0.23.4
```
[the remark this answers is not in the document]
mengqiy: `go mod tidy` updated these dependencies. I guess we can update all k8s dependencies to `v0.23.4 to be safer.

porch/apiserver/pkg/e2e/e2e_test.go
```diff
@@ -108,12 +108,21 @@ func (t *PorchSuite) TestGitRepository(ctx context.Context) {
 				{
 					Type: "eval",
 					Eval: &porchapi.FunctionEvalTaskSpec{
-						Image: "gcr.io/kpt-fn/set-namespace:v0.2.0",
+						Image: "gcr.io/kpt-fn/set-namespace:unstable",
```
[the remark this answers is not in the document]
mengqiy: Changing this to make the test use builtin functions for `set-namespace`. `gcr.io/kpt-fn/set-namespace:v0.2.0` maps to the executable extraction approach.

porch/apiserver/go.mod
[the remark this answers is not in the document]
droot: Yes, we should use same version of k8s deps across kpt/porch. @mengqiy Will it make sense to update the deps in a separate PR for both kpt and porch? and rebase this on top of that ?

porch/engine/go.mod
```diff
@@ -13,7 +16,7 @@ require (
 	github.com/go-git/go-git/v5 v5.4.3-0.20220119145113-935af59cf64f
 	github.com/google/go-cmp v0.5.7
 	google.golang.org/grpc v1.44.0
-	k8s.io/apimachinery v0.23.2
+	k8s.io/apimachinery v0.23.4
```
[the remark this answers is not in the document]
droot: +1 to what Martin said.

porch/engine/pkg/engine/builtinruntime.go
droot: Ack.

porch/engine/pkg/engine/error.go
droot: Moving under `pkg/fn` sounds good to me.

porch/apiserver/go.mod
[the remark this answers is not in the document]
mengqiy: Sure 
https://github.com/GoogleContainerTools/kpt/pull/2940

mengqiy:
Rebased on top of https://github.com/GoogleContainerTools/kpt/pull/2940.
Cherry-picked Justin's commit from https://github.com/GoogleContainerTools/kpt/pull/2937 and fix the CI failure.
Added unit tests for builtin runtime.

porch/engine/pkg/engine/builtinruntime.go
mengqiy: I simplified this since we add something in our SDK: https://github.com/GoogleContainerTools/kpt-functions-sdk/pull/532

porch/engine/pkg/engine/delegatingruntime.go
mengqiy: `MultiRuntime` from Justin's PR is a better way to do it. I updated it.

porch/engine/pkg/engine/error.go
mengqiy: I cherry-picked the commit from Justin's PR.

porch/engine/pkg/engine/options.go
mengqiy: Using `MultiRuntime` now

porch/apiserver/pkg/e2e/e2e_test.go
```diff
@@ -108,12 +108,21 @@ func (t *PorchSuite) TestGitRepository(ctx context.Context) {
...
 							"namespace": "bucket-namespace",
 						},
 					},
 				},
+				{
+					Type: "eval",
+					Eval: &porchapi.FunctionEvalTaskSpec{
+						Image: "gcr.io/kpt-fn/set-annotations:v0.1.4",
+						ConfigMap: map[string]string{
+							"foo": "bar",
```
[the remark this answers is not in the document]
mengqiy: I added some unit tests for the builtin runtime. We exercised the logic of the function in the unit test. So I guess we don't need to do it again in the e2e test. Please LMK if you disagree.

pkg/fn/eval_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	fn := &v1.Function{Image: "foo"}
+	err = &NotFoundError{Function: *fn}
+
+	var notFoundErr *NotFoundError
+	if !errors.As(err, &notFoundErr) {
+		t.Fatalf("expected NotFoundError to satisfy errors.As")
+	}
+
+	if got, want := notFoundErr.Function, *fn; !reflect.DeepEqual(got, want) {
```
martinmaly, line 36: for such comparisons, I'd recommend using https://github.com/google/go-cmp
It comes with good Diff algorithm too which can be used to generate more helpful error messages.

Example: https://github.com/GoogleContainerTools/kpt/blob/main/porch/engine/pkg/kpt/render_test.go#L76

pkg/fn/eval_test.go
```diff
@@ -0,0 +1,43 @@
...
+		t.Fatalf("function in NotFoundError did not match expected; got %#v, want %#v", got, want)
+	}
+
+	if got, want := notFoundErr.Error(), "function \"foo\" not found"; !reflect.DeepEqual(got, want) {
```
martinmaly, line 40: DeepEqual on strings seems unnecessary

porch/engine/pkg/engine/builtinruntime_test.go
```diff
@@ -0,0 +1,74 @@
...
+`))
+	var buf bytes.Buffer
+	err = fr.Run(reader, &buf)
+	if err != nil {
+		t.Fatalf("unexpected error when running the function runner: %v", err)
+	}
+	output := buf.String()
+	if !strings.Contains(output, "namespace: test-ns") {
+		t.Fatalf("expect namespace to be set, but it didn't. We got:\n%v\n", output)
+	}
```
martinmaly, line 61: I'd recommend parsing the result and checking structurally, rather than via substring.
Is the functionConfig removed from the result? (there may be false positive if it is not).
I'd also recommend checking for the desired target namespace, which would be ... `test-ns`?

porch/engine/pkg/engine/builtinruntime_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+func TestBuiltinRuntimeNotFound(t *testing.T) {
+	br := newBuiltinRuntime()
+	funct := &v1.Function{
+		Image: "gcr.io/kpt-fn/not-exist:unstable",
+	}
+	_, err := br.GetRunner(context.Background(), funct)
+	var fnNotFoundErr *fn.NotFoundError
+	if !errors.As(err, &fnNotFoundErr) {
+		t.Fatalf("expect error to be %T, but got %v", fnNotFoundErr, err)
```
martinmaly, line 72: " ... but got %T, %v", err, err) ... ? to print the type too?

martinmaly (approved):
Couple additional comments. Feel free to merge after addressing them

porch/engine/pkg/engine/options.go
```diff
@@ -46,13 +42,29 @@ func WithCache(cache *cache.Cache) EngineOption {
...
 
+func WithBuiltinFunctionRuntime() EngineOption {
+	return EngineOptionFunc(func(engine *cadEngine) error {
+		runtime := newBuiltinRuntime()
+		if engine.runtime == nil {
+			engine.runtime = runtime
+		} else {
+			engine.runtime = fn.NewMultiRuntime([]fn.FunctionRuntime{engine.runtime, runtime})
+		}
+		return nil
```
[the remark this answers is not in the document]
mengqiy: It will be 
```
MultiRuntime(
  MultiRuntime(
    first-runtime,
    second-runtime,
  ),
  third-runtime,
)
```
instead of
```
MultiRuntime(
  first,
  second,
  third,
)
```

The correctness is NOT a problem here, since `first-runtime` will be tried first even though it's nested.
I can tweak it if you don't like the nested appraoch.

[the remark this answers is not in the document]
martinmaly: I understand that the code in the PR will look for functions in correct order. my observation is that it defeats the purpose of MultiRuntime; we only use it as TwoRuntime, that's all.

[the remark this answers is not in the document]
mengqiy: Updated

PR #2935, opened 2022-03-24
Implement delete-refs capability in test server

martinmaly:
Git server must implement delete-refs so that client can issue reference
deletions.


justinsb (approved):
lgtm

PR #2938, opened 2022-03-25
proposal for changing the function build process to support multi platform images

mikebz:
proposal for changing the function build process to support multi platform images.

### DIFF
--- a/docs/design-docs/04-arm-build.md
+++ b/docs/design-docs/04-arm-build.md
@@ -1,0 +1,66 @@
+# Title
+
+* Author(s): Mike Borozdin, @mikebz
+* Approver: Mengqi, Sunil
+
+## Why
+
+More people are getting Apple M1 machines.  The current docker images for
+functions do not work.  Here is an example:
+https://github.com/GoogleContainerTools/kpt/issues/2874
+While it's not a problem for CI/CD pipelines where architecture is mostly amd64
+for client development purposes the users are stuck.
+
+## Design
+
+Building arm64 and amd64 multi platform images is possible:
+https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/
+but it has it's caveats.  The build is done in a builder which needs to be 
+created:
+https://docs.docker.com/engine/reference/commandline/buildx_create/
+The caveat is that the multi platform images do not load into docker images:
+https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#-load-the-single-platform-build-result-to-docker-images---load
+
+Right now the build pipeline has several steps:
+1) build
+2) tag
+3) push
+
+In different scenarios only the first step is executed.  If we switch to 
+`buildx` we will need to build/tag/push in one step.  It is not clear if
+we need to build things locally (maybe the OWNERs can give guidance on that).
+
+However if we make the changes needed to the shell scripts that produce the
+images all that the users of docker images will need to do is just update
+to the next version of the function image.  Docker promises to use the right 
+architecture automatically.
+
+## User Guide
+
+The user guide for using the new images should not change.  People using
+amd64 systems (Linux or Mac) can continue using the same systems.  People using
+arm64 systems can already get the arm64 kpt binary and will update their
+pipelines to the next version of functions.  The functions should select the 
+right architecture to run.
+
+## Open Issues/Questions
+
+Please list any open questions here in the following format:
+
+## Alternatives Considered
+
+An alternative to using a multi architecture images is to build a special image
+for each platform and then tag them differently.  The benefits of this are:
+1) (possibly) the images are smaller
+2) users can select the right architecture in their Kptfile or imperability 
+invoking the right function image.
+
+The problem is that most of the time CI/CD systems are amd64 linux and the arm64
+is primarily for the convenience of client systems.  The users most likely
+do not want to change their hydration pipeline definition from client to CI/CD
+systems.
+
+### \<Approach\>
+
+Links and description of the approach, the pros and cons identified during the 
+design.
--- a/docs/design-docs/04-arm-build.md
+++ b/docs/design-docs/04-arm-build.md
@@ -26,9 +26,9 @@ Right now the build pipeline has several steps:
 2) tag
 3) push
 
-In different scenarios only the first step is executed.  If we switch to 
-`buildx` we will need to build/tag/push in one step.  It is not clear if
-we need to build things locally (maybe the OWNERs can give guidance on that).
+In different scenarios only the two steps are executed.  If we switch to 
+`buildx` we will need to build/tag/push in one step.  Locally developers 
+might want to build and tag images locally.
 
 However if we make the changes needed to the shell scripts that produce the
 images all that the users of docker images will need to do is just update
--- a/docs/design-docs/04-arm-build.md
+++ b/docs/design-docs/04-arm-build.md
@@ -26,7 +26,7 @@ Right now the build pipeline has several steps:
 2) tag
 3) push
 
-In different scenarios only the two steps are executed.  If we switch to 
+In different scenarios only the first two steps are executed.  If we switch to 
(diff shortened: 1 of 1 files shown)


docs/design-docs/04-arm-build.md
```diff
@@ -0,0 +1,66 @@
...
+https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#-load-the-single-platform-build-result-to-docker-images---load
+
+Right now the build pipeline has several steps:
+1) build
+2) tag
+3) push
+
+In different scenarios only the first step is executed.  If we switch to 
+`buildx` we will need to build/tag/push in one step.  It is not clear if
+we need to build things locally (maybe the OWNERs can give guidance on that).
```
droot, line 31: Isn't that needed when I am developing the function locally and testing out ?

docs/design-docs/04-arm-build.md
```diff
@@ -0,0 +1,66 @@
...
+
+However if we make the changes needed to the shell scripts that produce the
+images all that the users of docker images will need to do is just update
+to the next version of the function image.  Docker promises to use the right 
+architecture automatically.
```
droot, line 36: Is there a min docker version that supports that ?

One thing we should do is to reliably detect this failure (their docker version not supporting an arch.) and suggesting them to use the right version once we publish the function images supporting all platforms.

droot (commented):
Thanks @mikebz for the proposal. I have some minor comments, but will defer to @mengqiy for the final approval who is our resident expert of function build/release pipeline.

@mengqiy Can you pl. take a look.

docs/design-docs/04-arm-build.md
mikebz: yes, didn't think that you might want to do that without pushing things to your own container registry.  Good point.

docs/design-docs/04-arm-build.md
mikebz: Just found the original post when Docker announced this: https://www.docker.com/blog/docker-official-images-now-multi-platform/ - it's 2017.  I think someone needs to be going back 5 years to get a version that doesn't support it.

docs/design-docs/04-arm-build.md
```diff
@@ -0,0 +1,66 @@
...
+https://docs.docker.com/engine/reference/commandline/buildx_create/
+The caveat is that the multi platform images do not load into docker images:
+https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#-load-the-single-platform-build-result-to-docker-images---load
+
+Right now the build pipeline has several steps:
+1) build
+2) tag
+3) push
+
+In different scenarios only the first step is executed.  If we switch to
```
mengqiy, line 29: `build` and `tag` are usually done together. No matter if the users are going to push the images, they almost always tags images.

docs/design-docs/04-arm-build.md
mengqiy: +1 on Sunil said

docs/design-docs/04-arm-build.md
```diff
@@ -0,0 +1,66 @@
...
+The caveat is that the multi platform images do not load into docker images:
+https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#-load-the-single-platform-build-result-to-docker-images---load
+
+Right now the build pipeline has several steps:
+1) build
+2) tag
+3) push
+
+In different scenarios only the first step is executed.  If we switch to 
+`buildx` we will need to build/tag/push in one step.  It is not clear if
```
mengqiy, line 30: Quoted from docker buildx doc:
>--output,-o Output destination (format: type=local,dest=path)
--push Shorthand for --output=type=registry

If seems buildx support output to registry or local.

If we can't use buildx, we can still use `docker manifests` to achieve the goal.

docs/design-docs/04-arm-build.md
```diff
@@ -0,0 +1,66 @@
...
+we need to build things locally (maybe the OWNERs can give guidance on that).
+
+However if we make the changes needed to the shell scripts that produce the
+images all that the users of docker images will need to do is just update
```
mengqiy, line 34: Does it hurt the performance? When pulling a multi-arch image, is docker smart enough to only pull the matching part or everything for different arch? If the later, the image size to download is doubled or tripled (if we support arm32/v7 and arm64/v8)

docs/design-docs/04-arm-build.md
droot: ha ha.... 5 years, so we are good then :)

docs/design-docs/04-arm-build.md
Pitta: My understanding is that multi-arch images are all individual images that are all valid under a single tag. Your docker runtime knows what the local arch is, and it should only fetch the matching container. I can't think of a reason why this would cause a performance issue in practice.

mikebz: this one is a hard one to predict.  I am going to need to do a test probably to verify what this looks like.  We have a number of customers who are relying on basic images to be fairly speedy and their primary concern is the CI/CD pipeline.  One of the other things to do is to possibly add the emulation flag to the kpt binary...  this will basically allow people to emulate amd64 on a mac.

mikebz: Here are the findings:  the image is not 2x the size.  It looks like separate images are created for arm64 and amd64 and they are not bigger in size.  the top tag becomes a pointer.  Regular docker build and push results:
```
docker buildx imagetools inspect us-central1-docker.pkg.dev/mikebz-ex1/mbzcr/check-labels:v0.2
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
   "config": {
      "mediaType": "application/vnd.docker.container.image.v1+json",
      "size": 1865,
      "digest": "sha256:0013ea02a33011596b9ec69f8e2bf01de1d25c72dd790f7a4e6b34328c8dc38a"
   },
   "layers": [
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 2814512,
         "digest": "sha256:40e059520d199e1a1a259089077f2a0c879951c9a4540490bad3a0d7714c6ae7"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 3117030,
         "digest": "sha256:31e6c0326e6befcafbcc24f20fdf48cd199230f11ecc3f8cf9009f926f645163"
      }
   ]
}
```

Individual image looks like this:
image



Then buildx result:
```
docker buildx imagetools inspect us-central1-docker.pkg.dev/mikebz-ex1/mbzcr/check-labels:v0.3                                  
Name:      us-central1-docker.pkg.dev/mikebz-ex1/mbzcr/check-labels:v0.3
MediaType: application/vnd.docker.distribution.manifest.list.v2+json
Digest:    sha256:15a41dced1a488bd2e0690fa047873931b673fe1b3a5a792e23b4984d1a71f4d
           
Manifests: 
  Name:      us-central1-docker.pkg.dev/mikebz-ex1/mbzcr/check-labels:v0.3@sha256:edea73063326e28693555bfd909c47c394af6719601235663c3203dfde43ae21
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/amd64
             
  Name:      us-central1-docker.pkg.dev/mikebz-ex1/mbzcr/check-labels:v0.3@sha256:56808f70266b9ba8a4c80df7059932313a1d41860828733de5e831ea3a35f562
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/arm64
  ```
The images are pushed simultaneously:
image

and the individual images look like this:

image

docs/design-docs/04-arm-build.md
mengqiy: Then it sounds like we have to use `docker manifests`?

mengqiy (approved):

PR #2942, opened 2022-03-26
fix git url parsing

natasha41575:
fix https://github.com/GoogleContainerTools/kpt/issues/2936

Instead of blindingly looking for ".git" anywhere in the URL, use `url.Parse` to identify the path, and check for ".git" in the path. 

I noticed the URL parsing code doesn't have any unit tests, so I added a couple.


internal/util/parse/parse.go
```diff
@@ -84,11 +90,32 @@ func GitParseArgs(ctx context.Context, args []string) (Target, error) {
...
+	return g, nil
+}
+
+// parseURL parses a pkg url and returns the repo, directory, and version
+func parseURL(ctx context.Context, pkgURL string) (repo string, dir string, version string, err error) {
+	parsedUrl, err := url.Parse(pkgURL)
+	if err != nil {
+		return "", "", "", err
+	}
+	parts := strings.Split(parsedUrl.Path, ".git")
```
martinmaly, line 116: this may yield incorrect results with inputs such as:

https://my-site.com/root.gitops/foo

Imo we need to check for .git($|/)

internal/util/parse/parse_test.go
```diff
@@ -182,3 +183,33 @@ func Test_pkgURLFromGHURL(t *testing.T) {
...
+		dir     string
+		version string
+	}
+	tests := map[string]struct {
+		ghURL string
+		expected
+	}{
+		"simple": {
+			ghURL:    "https://my.git-server.com/repository.git/directory@main",
+			expected: expected{repo: "https://my.git-server.com/repository", dir: "/directory", version: "main"},
```
martinmaly, line 199: we should actually keep the .git part of the repository, shouldn't we?

with GitHub it happens to work that removing .git suffix points at the same repo but there is no such guarantee in the general case and the `.git` is part of the input URL, seems incorrect to remove it.

I'd expect output:
repo: https://my.git-server.com/repository.git
dir: /directory
version: main

internal/util/parse/parse.go
```diff
@@ -38,8 +39,13 @@ func GitParseArgs(ctx context.Context, args []string) (Target, error) {
 		return g, nil
 	}
 
+	parsedUrl, err := url.Parse(args[0])
+	if err != nil {
+		return g, err
+	}
```
droot, line 45: I think this will break if user specifies url in the form of `git@github.com:GoogleContainerTools/kpt.git`

I looked at the code a little bit and realized looking for `.git` is more about whether the url uses `git@github.com:org/repo.git` vs http based url `https://github.com/GoogleContainerTools/kpt.git`

droot: To feel more confident about the change, It will be good to have test that tests GitParseArgs function and test it for three forms of pkg URLs `git@....`, `http://`,`github.com:` etc.

internal/util/parse/parse_test.go
natasha41575: I'm not that familiar with this code, but I find that keeping `.git` here causes some pkg update tests to fail. I suspect that kpt is appending the .git suffix elsewhere (or doing some other magic to make it work). 

I'll look into a bit more to make sure, but I think the test for porch that uses the URL `http://git-server.test-git-namespace.svc.cluster.local:8080.git/@main` should catch any problems with nongithub URLs.

martinmaly: The URL used in Porch is an unfortunate hack I had to put in place due to the parsing bug so we can update it.

the URL I'd like to be using is:

`http://git-server.test-git-namespace.svc.cluster.local:8080/@main`

There is a separate problem where kpt will actually try to clone repository as part of parsing the address to glean where the separation between repository address and directory path within the repository is.

natasha41575: Sorry for the delay, just getting back to this now.

I see the porch test URLs have been updated in main, I'm assuming they're now the URLs that you want so there's no need to update them in this PR anymore.

martinmaly: correct.

internal/util/parse/parse.go
natasha41575: I think you're right that it'll break. Since we're checking for `.git($|/)` now I don't see much use for url.Parse anymore so I took this out. 

Added tests for `GitParseArgs` to `parse_test.go` to check the different url types

internal/util/parse/parse.go
natasha41575: done, thank you!

internal/util/parse/parse_test.go
natasha41575: > we should actually keep the .git part of the repository, shouldn't we?
> 
> with GitHub it happens to work that removing .git suffix points at the same repo but there is no such guarantee in the general case and the .git is part of the input URL, seems incorrect to remove it.
> 
> I'd expect output:
> repo: https://my.git-server.com/repository.git
> dir: /directory
> version: main

I checked this code with a local filepath `kpt pkg get /path/to/repo.git` and it was successful. Is that enough to give us confidence that this will work with nongithub URLs?

natasha41575:
@martinmaly I've updated the PR based on your and Sunil's feedback, would you mind taking another look while Sunil is OOO?

internal/util/parse/parse_test.go
```diff
@@ -182,3 +184,87 @@ func Test_pkgURLFromGHURL(t *testing.T) {
...
+	}{
+		"git@ url with .git suffix": {
+			ghURL: "git@github.com:GoogleContainerTools/kpt.git",
+			expected: Target{Git: v1.Git{
+				Repo:      "git@github.com:GoogleContainerTools/kpt",
+				Directory: "/",
+				Ref:       "main",
+			},
+				Destination: "kpt"},
+			// in CI we get the error "git@github.com: Permission denied (publickey)" because the ssh key is not set up.
```
martinmaly, line 236: Why does URL parsing require SSH keys and why would it get permission denied?

martinmaly (approved):
The main issue appears fixed. I do have a question about removing the `.git` suffix and the situation where parser gets `Access Denied` error

internal/util/parse/parse_test.go
natasha41575: `GitParseArgs` doesn't just parse the URL, it parses the URL and makes a call to [NewGitUpstreamRepo](https://github.com/GoogleContainerTools/kpt/blob/0e62e256d68c5daec50dd28e523566b84a31d4f2/internal/gitutil/gitutil.go#L141), somewhere in which it fetches the upstream repo.

martinmaly: Thank you!

internal/util/parse/parse_test.go
```diff
@@ -182,3 +184,87 @@ func Test_pkgURLFromGHURL(t *testing.T) {
...
+			skip: true,
+		},
+		"http url with .git suffix": {
+			ghURL: "https://github.com/GoogleContainerTools/kpt.git",
+			expected: Target{Git: v1.Git{
+				Repo:      "https://github.com/GoogleContainerTools/kpt",
```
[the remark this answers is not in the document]
martinmaly: Thanks for the follow-up; let's not block the PR on this; I'll chat with @droot about the history.

internal/util/parse/parse_test.go
martinmaly: This is great, thank you!

martinmaly (approved):

PR #2945, opened 2022-03-29
porch uses `kpt pkg init` library

droot:
With this change: porch and `kpt pkg init` both workflow use the same package initialization code.


porch/engine/pkg/engine/init_test.go
```diff
@@ -0,0 +1,66 @@
...
+			Site:        "http://kpt.dev/testpkg",
+		},
+	}
+
+	testdata, err := filepath.Abs(filepath.Join(".", "testdata", "init", "testpkg"))
+	if err != nil {
+		t.Fatalf("Failed to find testdata: %v", err)
+	}
+
+	rendered, _, err := init.Apply(context.Background(), repository.PackageResources{})
```
martinmaly, line 43: super nit: the package is not `rendered` so much as `initialized` :)

martinmaly (approved):
Looks great, just one nit, likely result of copy'n'paste

porch/engine/pkg/engine/init_test.go
droot: copy paste :) Fixed.

PR #2946, opened 2022-03-29
Create Lifecycle Strategies

martinmaly:
`packageCommon` now carries update strategy similar to what k8s
apiserver uses (though simpler) to validate package revision lifecycle
values and enforce approval flow via the subresource.

### DIFF
--- a/porch/apiserver/pkg/registry/porch/packagerevision.go
+++ b/porch/apiserver/pkg/registry/porch/packagerevision.go
@@ -17,6 +17,7 @@ package porch
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	api "github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
 	configapi "github.com/GoogleContainerTools/kpt/porch/controllers/pkg/apis/porch/v1alpha1"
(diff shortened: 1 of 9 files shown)


porch/apiserver/pkg/registry/porch/packagerevision.go
```diff
@@ -269,3 +202,55 @@ func (r *packageRevisions) Delete(ctx context.Context, name string, deleteValida
...
+	oldRevision := old.(*api.PackageRevision)
+	newRevision := obj.(*api.PackageRevision)
+
+	switch lifecycle := oldRevision.Spec.Lifecycle; lifecycle {
+	case "", api.PackageRevisionLifecycleDraft, api.PackageRevisionLifecycleProposed:
+		// valid
+
+	default:
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "lifecycle"), lifecycle, fmt.Sprintf("can only update package with lifecycle value one of %s",
+			strings.Join([]string{
```
justinsb, line 226: We do have generics now ;-)

justinsb (approved):

porch/apiserver/pkg/registry/porch/packagerevision.go
martinmaly: not yet :( we are still building with 1.17 :)
https://github.com/GoogleContainerTools/kpt/blob/main/.github/workflows/porch.yml#L35

but we can now upgrade :)

PR #2947, opened 2022-03-29
feat: change `--save` behavior from skip to Override

yuwenma:
- `--save` flag skips the Kptfile editing if it finds the fn in the mutators list. This PR changes the behavior to override the Kptfile with the new function arguments.
- bug fix: modify the Kptfile read/write path.

related:
- #2249 
- https://github.com/GoogleContainerTools/kpt/pull/2841#issuecomment-1081295669

### DIFF
--- a/thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
+++ b/thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
@@ -164,7 +164,7 @@ func (r *EvalFnRunner) NewFunction() *kptfile.Function {
 // SaveFnToKptfile adds the evaluated function and its arguments to Kptfile `pipeline.mutators` list.
 func (r *EvalFnRunner) SaveFnToKptfile() {
 	pr := printer.FromContextOrDie(r.Ctx)
-	kf, err := pkg.ReadKptfile(filesys.FileSystemOrOnDisk{}, r.Dest)
+	kf, err := pkg.ReadKptfile(filesys.FileSystemOrOnDisk{}, r.RunFns.Path)
 	if err != nil {
 		pr.Printf("function not added: Kptfile not exists\n")
 		return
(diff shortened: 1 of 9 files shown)


yuwenma:
+CC @morgante

morgante (approved):

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -176,24 +176,33 @@ func (r *EvalFnRunner) SaveFnToKptfile() {
...
 		}
 	}
-	kf.Pipeline.Mutators = append(kf.Pipeline.Mutators, *r.NewFunction())
-	if err = kptfileutil.WriteFile(r.Dest, kf); err != nil {
+	if !found {
+		newMutators = append(newMutators, *r.NewFunction())
+		pr.Printf("adding function to Kptfile\n")
+	}
+	kf.Pipeline.Mutators = newMutators
+	if err = kptfileutil.WriteFile(r.RunFns.Path, kf); err != nil {
```
droot, line 201: One thing I am not able to recall is if comments in kptfile will be preserved across these writes or not ? something to verify.

droot (approved):

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
yuwenma: We do not have mechanism to preserve comments in Kptfile read/write. Once the kptfile updates, comments will be removed. 
@morgante How serious it is to preserve the comments in Kptfile (I think we want to keep the license header at least)?

morgante: Yeah it's somewhat critical that we not overwrite the Kptfile's comments.

yuwenma: I see. We need to develop a new way to retrieve and update Kptfile then (e.g. using kustomize/kyaml RNode). Would blueprint KRM functions edit Kptfile as well? If so, I will put the new kptfile read/write in kpt-functions-sdk

yuwenma: @droot  @mengqiy @natasha41575 I'm thinking of importing the kustomize/kyaml comments handlers to kpt to preserve the Kptfile comments. But this would introduce a non-trivial amount of code and in a RNode-like style wich is very different from kpt source code. What're your thoughts on this approach?

natasha41575: @yuwenma what extra code would it introduce? Would you be able to do something here like 

```
kfNodeOriginal, _ := yaml.Parse(kf.String()) // save original kptfile content as an RNode
.... // operations to mutate kptfile
kfNode, _ := yaml.Parse(kf.String()) // save mutated kptfile content as an RNode
kyaml.CopyComments(kfNodeOriginal, kfNode) // copy the saved comments  from kfNodeOriginal to kfNode
yaml.Unmarshal(kf, kfNode.MustString()) // convert the kfNode back to kpt Kptfile struct
```

Does that make sense? Wondering if this approach would work, and if it would be relatively simple to do.

natasha41575: Maybe we can introduce a wrapper around `kyaml.CopyComments` to do the conversions from kf to RNode and back?

yuwenma: Another option is preserve the license header only. I'm going to merge this PR and updates the solution later on.

morgante: > Another option is preserve the license header only. I'm going to merge this PR and updates the solution later on.

We shouldn't do that.

Would it be possible to reuse the kpt logic which automatically reattaches comments on resources after functions modify them? Ideally function authors should be able to modify the Kptfile without worrying about comment management.

yuwenma: "auto reattaches comments on resources" -> this is the kustomize/kyaml RNode we discussed above.

yuwenma: Tracked in https://github.com/GoogleContainerTools/kpt/issues/2952

PR #2948, opened 2022-03-29
Delete PackageRevisions in Git

martinmaly:
If package has its own Git ref (own branch or package-specific tag),
the ref is deleted. This is true for drafts, proposals, and final
packages tagged with their own tag (refs/tags/<package patch>/revision).

If the package is tagged with general tag of a different format,
we can't delete it and deletion fails.

If the package is in a branch, we delete it by creating a commit
that removes it from that branch.

This change also includes few bug fixes that ensure that we track
package lifecycle correctly through package updates.

### DIFF
--- a/porch/apiserver/pkg/e2e/e2e_test.go
+++ b/porch/apiserver/pkg/e2e/e2e_test.go
@@ -361,7 +361,8 @@ func (t *PorchSuite) TestInitTaskPackage(ctx context.Context) {
 func (t *PorchSuite) TestCloneIntoDeploymentRepository(ctx context.Context) {
 	const downstreamRepository = "deployment"
 	const downstreamPackage = "istions"
-	const downstreamName = downstreamRepository + ":" + downstreamPackage + ":v1"
+	const downstreamRevision = "v1"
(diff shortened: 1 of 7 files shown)


porch/apiserver/pkg/e2e/e2e_test.go
```diff
@@ -706,3 +808,50 @@ func withContent(content configapi.RepositoryContent) repositoryOption {
...
+			},
+		},
+	}
+	t.CreateF(ctx, pr)
+	return pr
+}
+
+func (t *PorchSuite) mustExist(ctx context.Context, key client.ObjectKey, obj client.Object) {
+	t.GetE(ctx, key, obj)
+	if got, want := obj.GetName(), key.Name; got != want {
```
justinsb, line 842: Not sure that this is possible ... presumably obj would be untouched if it was not found?

justinsb (approved):

porch/apiserver/pkg/e2e/e2e_test.go
martinmaly: ah yes, you are right. I'll just use GetF

PR #2950, opened 2022-03-30
Make upstreamRef a Pointer

martinmaly:
This allows upstreamRef to be actually optional rather than always
present in the API request/response.


martinmaly:
I had to disable one test which I can only re-enable once this change is merged and the version of porch API that kpt depends on is updated (see #2929 for context)

justinsb (approved):

PR #2954, opened 2022-03-30
Set Commit Author to Authenticated User

martinmaly:
- Create UserInfoProvider which apiserver provides to engine to determine
  authenticated user information.
- Create GitRepositoryOptions to make passing of options easier
- Pass UserInfoProvider to git repository layer
- Set Commit Author from Authenticated User

### DIFF
--- a/porch/repository/pkg/git/commit.go
+++ b/porch/repository/pkg/git/commit.go
@@ -353,18 +353,22 @@ func entrySortKey(e *object.TreeEntry) string {
 
 func storeCommit(store storer.EncodedObjectStorer, parent plumbing.Hash, tree plumbing.Hash, author, message string) (plumbing.Hash, error) {
 	now := time.Now()
-	authorSignature := object.Signature{
-		Name:  porchSignatureName,
-		Email: porchSignatureEmail,
-		When:  time.Time{},
-	}
+	var authorName, authorEmail string
 	if author != "" {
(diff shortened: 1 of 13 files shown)


porch/repository/pkg/git/commit.go
```diff
@@ -333,17 +351,27 @@ func entrySortKey(e *object.TreeEntry) string {
 	return e.Name
 }
 
-func storeCommit(store storer.EncodedObjectStorer, parent plumbing.Hash, tree plumbing.Hash, message string) (plumbing.Hash, error) {
+func storeCommit(store storer.EncodedObjectStorer, parent plumbing.Hash, tree plumbing.Hash, author, message string) (plumbing.Hash, error) {
 	now := time.Now()
+	var authorName, authorEmail string
+	if author != "" {
+		// Authenticated user info only provides one value...
+		authorName = author
```
justinsb, line 359: It might be an idea to have a richer user info type, rather than a string.  `nil` is also a bit safer than `""` (misusing a nil is a panic)

justinsb (approved):

justinsb:
lgtm!

porch/repository/pkg/git/commit.go
martinmaly: Good point. I'll turn it into a pointer.

PR #2956, opened 2022-03-31
 Discover catalog function in `kpt fn eval`

yuwenma:
### Description

This PR allows users to discover and filter Catalog functions via `kpt fn eval --type --keywords --image`.

Functions have two sources.
1. catalog function, which provides general solutions like `set-namespace`, `set-annotation` . its metadata is stored in ttps://catalog.kpt.dev/catalog-v2.json
2. porch functions, which is custom KRM functions whose meta config is stored on the cluster api resource.

In previous work:
- https://github.com/GoogleContainerTools/kpt/pull/2980 adds "--type" "--keyword" flags with autocompletion in `kpt fn eval` to discover and filter porch function.
- https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/824 will add the function types and keywords to catalog function metadata. 
 
This PR is the last change, which contains:
- Read the new Catalog metadata from "https://catalog.kpt.dev/catalog-v2.json"
- Map Catalog function to porch `v1alpha1.Function`, to merge the logic on function matching.
- Change the format of displayed function names
   - Porch function name is `<PackageRepository>:<ImageName>:<Version>`. When displayed in kpt, it has the namespace as prefix e.g. `default:kpt-functions:set-annotation:v0.1`
   - Catalog v2 function name is trimmed to only contain `<ImageName>:<Version>`, and exclude `gcr.io/kpt-fn`. e.g. `set-annotation:v0.1`. 
- Retrieve the Porch function "Image" by function name in function `evaluation`. 
- To record porch function namespace (used to get porch function image later on), we add the namespace as prefix in the displayed function name. We use a fake namespace  "__catalog_v2" to mark catalog functions, which does not live in porch but used the fixed url `gcr.io/kpt-fn/<FUNCTION NAME>:<TAG>` to get the image.

Fix bugs:
- a multiple matcher should choose a function if the function satisfies all matcher conditions, not XNOR.

### DIFF
--- a/internal/fnruntime/container.go
+++ b/internal/fnruntime/container.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/GoogleContainerTools/kpt/internal/printer"
 	"github.com/GoogleContainerTools/kpt/internal/types"
+	"github.com/GoogleContainerTools/kpt/internal/util/porch"
 	fnresult "github.com/GoogleContainerTools/kpt/pkg/api/fnresult/v1"
 	"sigs.k8s.io/kustomize/kyaml/fn/runtime/runtimeutil"
 )
@@ -175,11 +176,20 @@ func NewContainerEnvFromStringSlice(envStr []string) *runtimeutil.ContainerEnv {
 }
 
 // AddDefaultImagePathPrefix adds default gcr.io/kpt-fn/ path prefix to image if only image name is specified
-func AddDefaultImagePathPrefix(image string) string {
+func AddDefaultImagePathPrefix(ctx context.Context, image string) string {
+	segments := strings.Split(image, ":")
+	if len(segments) == 4 {
+		// Porch function
+		functionName := strings.Join(segments[1:], ":")
+		function, err := porch.FunctionGetter{Ctx: ctx}.Get(functionName, segments[0])
+		if err != nil {
+			return image
+		}
+		return function.Spec.Image
+	}
 	if !strings.Contains(image, "/") {
 		return fmt.Sprintf("gcr.io/kpt-fn/%s", image)
 	}
-	// TODO(yuwenma): append porch repo prefix if the image comes from porch function.
 	return image
 }
 
--- a/internal/fnruntime/container.go
+++ b/internal/fnruntime/container.go
@@ -175,13 +175,16 @@ func NewContainerEnvFromStringSlice(envStr []string) *runtimeutil.ContainerEnv {
 	return ce
 }
 
-// AddDefaultImagePathPrefix adds default gcr.io/kpt-fn/ path prefix to image if only image name is specified
+// AddDefaultImagePathPrefix converts the function short path to the full image url.
+// If the function is Catalog function, it adds "gcr.io/kpt-fn/".e.g. set-namespace:v0.1 --> gcr.io/kpt-fn/set-namespace:v0.1
+// If the function is porch function, it queries porch to get the function image by name and namespace.
+// e.g. default:set-namespace:v0.1 --> us-west1-docker.pkg.dev/cpa-kit-dev/packages/set-namespace:v0.1
 func AddDefaultImagePathPrefix(ctx context.Context, image string) string {
 	segments := strings.Split(image, ":")
 	if len(segments) == 4 {
 		// Porch function
 		functionName := strings.Join(segments[1:], ":")
-		function, err := porch.FunctionGetter{Ctx: ctx}.Get(functionName, segments[0])
+		function, err := porch.FunctionGetter{}.Get(ctx, functionName, segments[0])
 		if err != nil {
 			return image
 		}
--- a/internal/util/cmdutil/cmdutil.go
+++ b/internal/util/cmdutil/cmdutil.go
@@ -32,7 +32,6 @@ import (
 	"github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
 	"github.com/spf13/cobra"
 	"golang.org/x/mod/semver"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 )
@@ -215,7 +214,7 @@ func SuggestKeywords(cmd *cobra.Command) []string {
 }
 
 func DiscoverFunctions(cmd *cobra.Command) []v1alpha1.Function {
-	porchFns := porch.FunctionListGetter{Ctx: cmd.Context()}.Get()
+	porchFns := porch.FunctionListGetter{}.Get(cmd.Context())
 	catalogV2Fns := fetchCatalogFunctions()
 	return append(porchFns, catalogV2Fns...)
 }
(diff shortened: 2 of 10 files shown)


internal/util/porch/function.go
```diff
@@ -50,81 +49,48 @@ func (f FunctionGetter) Get() []v1alpha1.Function {
...
+	if err := v1alpha1.SchemeBuilder.AddToScheme(scheme); err != nil {
+		return function, err
 	}
-	return suggestedFunctions
+	codec := runtime.NewParameterCodec(scheme)
+	err = client.Get().
+		Timeout(Expiration).
+		Resource("functions").
+		Name(name).
+		Namespace("default").
```
yuwenma, line 83: Get a resource by name requires providing the namespace. but "default" is not a reasonable assumption. 
Another option is to list all functions, which may not be efficient?
Do you have any insights what the right behavior should be if the functions have same repo, same function name but under different namespace? @martinmaly

internal/util/cmdutil/cmdutil.go
```diff
@@ -194,67 +197,87 @@ func GetKeywordsFromFlag(cmd *cobra.Command) []string {
...
-	functions = append(functions, FetchFunctionImages()...)
-	var dedup []string
-	fnMap := map[string]bool{}
-	for _, function := range functions {
-		if _, ok := fnMap[function]; !ok {
-			dedup = append(dedup, function)
-			fnMap[function] = true
-		}
+	matchers := []function.Matcher{
+		function.TypeMatcher{FnType: cmd.Flag("type").Value.String()},
```
mengqiy, line 201: I wonder do you validate the --type flag value somewhere?

internal/util/cmdutil/cmdutil.go
```diff
@@ -194,67 +197,87 @@ func GetKeywordsFromFlag(cmd *cobra.Command) []string {
...
 			}
 		}
-		result = append(result, fmt.Sprintf("%s:%s", fnName, latestVersion))
+		fnName := fmt.Sprintf("%s:%s", fnName, latestVersion)
+		functions = append(functions,
+			v1alpha1.Function{
+				TypeMeta: metav1.TypeMeta{},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      fnName,
+					Namespace: function.CatalogV2,
```
mengqiy, line 273: What's this? Using `__catalog_v2` as namespace?

yuwenma: This is an invalid namespace used to distinguish catalog function from porch functions. This will be trimmed before showing to users.
What users will see is something like: "default" is the porch function namespace. `set-namespace:v0.3.3` comes from catalog
```bash
 kpt fn eval yu --keywords namespace -i [tab tab]
default:kpt-functions:set-annotation:v0.1-yuwenma  default:kpt-functions:set-namespace:v0.3-yuwenma
default:kpt-functions:set-annotation:v0.2-yuwenma  set-namespace:v0.3.3
default:kpt-functions:set-annotation:v0.3-yuwenma
```

internal/util/cmdutil/cmdutil.go
yuwenma: Ah, you catch it. Short answer: no.  
The behavior of default value is something that I haven't come up with a good answer yet (well, and maybe not have put enough thoughts on it):

Problem 1. 
`kpt fn eval` already has too many flags that sort of correlate to each other. `--keywords`, `--type`, `--exec`, `--image`. The usage of `--type` is determined and determines other behavior. I don't want to give users too much burden to understand all of these. So it may not be a good idea to use --type flag.

Problem 2. 
`--type` only really matters when used with "--save". At which point "mutator" is a valid default. But when used as a filter to functions, I'm not quite confident if we shall use it as a filter. 

Maybe the right approach should be:
-  `--type` is default to "" (rather than "mutator") with autocompletion enabled to suggest "validator" and "mutator".
- Matching functions by type accepts the "--type" to be empty string "", in such case suggest all functions. 
- Only when "--save" is true, will kpt validate "--type". It accepts value `["", "mutator", "validator"]`, and reassign `""` to `"mutator"`.

The advantage is that users are guided to only care the `--type` when they want to write function to Kptfile `pipeline.validators`. All other cases the `--type` is a nice to have thing.  

My main concern here is that these assumptions could further confuse users. WDYT? @mengqiy

internal/fnruntime/container.go
```diff
@@ -175,11 +176,20 @@ func NewContainerEnvFromStringSlice(envStr []string) *runtimeutil.ContainerEnv {
 }
 
 // AddDefaultImagePathPrefix adds default gcr.io/kpt-fn/ path prefix to image if only image name is specified
-func AddDefaultImagePathPrefix(image string) string {
+func AddDefaultImagePathPrefix(ctx context.Context, image string) string {
+	segments := strings.Split(image, ":")
+	if len(segments) == 4 {
+		// Porch function
+		functionName := strings.Join(segments[1:], ":")
+		function, err := porch.FunctionGetter{Ctx: ctx}.Get(functionName, segments[0])
```
martinmaly, line 184: I think golang style frowns upon putting context into structs but I also acknowledge we do it in kpt somewhat routinely because cobra seems to force that pattern.

consider whether in this case it is avoidable.

internal/util/cmdutil/cmdutil.go
martinmaly: we could consider using empty value for this placeholder, or (worst case) using some wrapper object for the Function resource(s).

that said, I think if documented for posterity, this approach seems fine to me.

internal/util/porch/function.go
```diff
@@ -30,14 +29,14 @@ import (
 
 const expirationNanoSecond = 10
 
-var PorchExpiration = time.Duration(expirationNanoSecond) * time.Second
+var Expiration = time.Duration(expirationNanoSecond) * time.Second
```
martinmaly, line 32: this feels a bit strange. A simpler way may be:

const Expiration time.Duration = 10 * time.Second

(I think Duration can be const)
the line 30 is actually not a duration, just an int so I would drop it as it doesn't seem required to the task.

martinmaly (approved):
Couple nits. The important thing IMO is to start getting some experience with this functionality so we can iterate on it. Therefore I am OK if some comments turn into TODOs for follow-up.

internal/util/porch/function.go
yuwenma: Thanks!

internal/util/cmdutil/cmdutil.go
yuwenma: Sg. I used wrapper and added doc string.

internal/fnruntime/container.go
yuwenma: Good to know! moved the context to the function parameter.

internal/fnruntime/container.go
```diff
@@ -175,11 +176,19 @@ func NewContainerEnvFromStringSlice(envStr []string) *runtimeutil.ContainerEnv {
 }
 
 // AddDefaultImagePathPrefix adds default gcr.io/kpt-fn/ path prefix to image if only image name is specified
-func AddDefaultImagePathPrefix(image string) string {
-	if !strings.Contains(image, "/") {
+func AddDefaultImagePathPrefix(ctx context.Context, image string) string {
+	segments := strings.Split(image, ":")
```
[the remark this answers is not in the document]
yuwenma: Thanks for the notes. Though not supported in golang, I'll add the IFCHANGE THENCHANGE syntax to the porch function.

internal/util/cmdutil/cmdutil.go
yuwenma: I filed https://github.com/GoogleContainerTools/kpt/issues/3014 for further discussion about this flag. I'm going to merge this PR.

PR #2957, opened 2022-03-31
`exclude` option for kpt fn eval and kpt fn render

natasha41575:
First commit: add `exclude` functionality as described in https://github.com/GoogleContainerTools/kpt/issues/2930
Second commit: add label and annotation selection to selectors/exclusions
Third commit: e2e tests

Design proposal is here: https://github.com/GoogleContainerTools/kpt/pull/2971. This PR should not be merged until the design proposal is in.

TODO: Docs

### DIFF
--- a/internal/fnruntime/utils.go
+++ b/internal/fnruntime/utils.go
@@ -140,16 +140,32 @@ type SelectionContext struct {
 }
 
 // SelectInput returns the selected resources based on criteria in selectors
-func SelectInput(input []*yaml.RNode, selectors []kptfilev1.Selector, _ *SelectionContext) ([]*yaml.RNode, error) {
+func SelectInput(input []*yaml.RNode, selectors, exclusions []kptfilev1.Selector, _ *SelectionContext) ([]*yaml.RNode, error) {
+	var selectedInput []*yaml.RNode
 	if len(selectors) == 0 {
-		return input, nil
+		selectedInput = input
+	} else {
+		for _, node := range input {
+			for _, selector := range selectors {
+				if isMatch(node, selector) {
+					selectedInput = append(selectedInput, node)
+				}
+			}
+		}
+	}
+	if len(exclusions) == 0 {
+		return selectedInput, nil
 	}
 	var filteredInput []*yaml.RNode
-	for _, node := range input {
-		for _, selector := range selectors {
-			if isMatch(node, selector) {
-				filteredInput = append(filteredInput, node)
+	for _, node := range selectedInput {
+		matchesExclusion := false
+		for _, exclusion := range exclusions {
+			if !exclusion.IsEmpty() && isMatch(node, exclusion) {
+				matchesExclusion = true
 			}
+		}
+		if !matchesExclusion {
+			filteredInput = append(filteredInput, node)
 		}
 	}
 	return filteredInput, nil
--- a/internal/fnruntime/utils.go
+++ b/internal/fnruntime/utils.go
@@ -175,7 +175,8 @@ func SelectInput(input []*yaml.RNode, selectors, exclusions []kptfilev1.Selector
 func isMatch(node *yaml.RNode, selector kptfilev1.Selector) bool {
 	// keep expanding with new selectors
 	return nameMatch(node, selector) && namespaceMatch(node, selector) &&
-		kindMatch(node, selector) && apiVersionMatch(node, selector)
+		kindMatch(node, selector) && apiVersionMatch(node, selector) &&
+		labelMatch(node, selector) && annoMatch(node, selector)
 }
 
 // nameMatch returns true if the resource name matches input selection criteria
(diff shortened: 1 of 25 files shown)


natasha41575:
@droot @yuwenma this PR is ready for review

pkg/api/kptfile/v1/types.go
```diff
@@ -291,6 +291,10 @@ type Function struct {
 	// `Selectors` are used to specify resources on which the function should be executed
 	// if not specified, all resources are selected
 	Selectors []Selector `yaml:"selectors,omitempty" json:"selectors,omitempty"`
+
+	// From `Selectors`, `Exclude` are used to specify resources on which the function should NOT be executed.
```
droot, line 295: `From Selectors` at the beginning feels misplaced.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -113,7 +133,14 @@ type EvalFnRunner struct {
...
 	Ctx                  context.Context
 	Selector             kptfile.Selector
+	Exclusion            kptfile.Selector
 	dataItems            []string
+
+	// we will need to parse these values into Selector and Exclusion
+	selectorLabel []string
+	selectorAnno  []string
+	excludeLabel  []string
+	excludeAnno   []string
```
droot, line 143: nit: Can we use full name here `excludeAnnotations` ? I have found `Anno` to be very untuitive :)

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -431,7 +462,32 @@ func (r *EvalFnRunner) preRunE(c *cobra.Command, args []string) error {
...
 
 	return nil
 }
+
+// parses annotation and label based selectors and exclusion from the command line input
+func (r *EvalFnRunner) parseSelectors() {
+	for _, s := range r.selectorAnno {
+		parts := strings.Split(s, "=")
+		key, value := parts[0], parts[1]
+		r.Selector.Annotations = map[string]string{key: value}
```
droot, line 476: As per the implementation, `r.Selector.Annotations` will end up with last `key=val`. Is that what we want here ?

I see some code duplication in this function, would be good if we can dedupe it.

internal/fnruntime/utils.go
```diff
@@ -140,17 +140,33 @@ type SelectionContext struct {
...
 	var filteredInput []*yaml.RNode
-	for _, node := range input {
-		for _, selector := range selectors {
-			if isMatch(node, selector) {
-				filteredInput = append(filteredInput, node)
+	for _, node := range selectedInput {
+		matchesExclusion := false
+		for _, exclusion := range exclusions {
+			if !exclusion.IsEmpty() && isMatch(node, exclusion) {
+				matchesExclusion = true
```
droot, line 164: probably safe to `break` here ?

internal/fnruntime/utils.go
```diff
@@ -140,17 +140,33 @@ type SelectionContext struct {
...
 			}
 		}
+		if !matchesExclusion {
+			filteredInput = append(filteredInput, node)
+		}
```
droot, line 169: this can probably be moved in the block where we mark `matchesExclusion = true` and break  and get rid of `matchesExclusion` ?

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
natasha41575, line 476: just realized this is buggy (should be r.Selector.Annotations[key] = value). Updating now

internal/fnruntime/utils.go
```diff
@@ -181,3 +198,25 @@ func kindMatch(node *yaml.RNode, selector kptfilev1.Selector) bool {
 func apiVersionMatch(node *yaml.RNode, selector kptfilev1.Selector) bool {
 	return selector.APIVersion == "" || selector.APIVersion == node.GetApiVersion()
 }
+
+// labelMatch returns true if the resource labels match input selection criteria
+func labelMatch(node *yaml.RNode, selector kptfilev1.Selector) bool {
+	nodeLabels := node.GetLabels()
+	for sk, sv := range selector.Labels {
+		if nv, ok := nodeLabels[sk]; !ok || sv != nv {
```
droot, line 206: ok --> found reflects the usage better (found label or not?)

droot (commented):
Looks pretty neat. I have very minor comments.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
natasha41575: I noticed this at the same time you did, just fixed it

internal/fnruntime/utils.go
natasha41575: I don't think we can move it in, because we need to do this if it doesn't match any of the exclusions (we are checking for `!matchesExclusion` here. So we need to check all exclusions before we can decide whether or not we can include this input.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
natasha41575: done

pkg/api/kptfile/v1/types.go
natasha41575: removed

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
natasha41575: done

internal/fnruntime/utils.go
natasha41575: done

internal/fnruntime/utils.go
natasha41575: done

internal/fnruntime/utils.go
droot: Aah...yep!

droot (approved):

PR #2961, opened 2022-04-01
Reduce number of go modules

justinsb:
We need porch/api from kpt, but kpt should not depend on porch other than api.
    
Other than porch/api, treat porch as one big go module.

Move the config API (the CRD) under porch/api as well.


porch/apiserver/Dockerfile
```diff
@@ -21,30 +21,18 @@ WORKDIR /go/src/github.com/GoogleContainerTools/kpt
...
 COPY porch/api/go.mod porch/api/go.sum porch/api/
-COPY porch/apiserver/go.mod porch/apiserver/go.sum porch/apiserver/
-COPY porch/controllers/go.mod porch/controllers/go.sum porch/controllers/
-COPY porch/engine/go.mod porch/engine/go.mod porch/engine/
-COPY porch/repository/go.mod porch/repository/go.sum porch/repository/
-COPY porch/func/go.mod porch/func/go.sum porch/func/
 
 RUN echo "Downloading root modules ..." \
  && go mod download
+RUN echo "Downloading porch modules ..." \
```
martinmaly, line 29: I'm surprised this one happens before the API module ...

porch/apiserver/pkg/e2e/suite.go
```diff
@@ -515,7 +515,7 @@ func (t *TestSuite) createInClusterGitServer() GitConfig {
 		}
 	}
 
-	t.Logf("Waiting for git-serever-service to be ready ...")
+	t.Logf("Waiting for git-server-service to be ready ...")
```
martinmaly, line 518: thanks!

martinmaly (approved):
Awesome

porch/apiserver/Dockerfile
justinsb: I don't think the order matters when we're doing go mod download.  Any preference?  The current order is alphabetical (sort of)

go.mod
```diff
@@ -2,18 +2,19 @@ module github.com/GoogleContainerTools/kpt
 
 go 1.17
 
+replace github.com/GoogleContainerTools/kpt/porch/api => ./porch/api
```
mengqiy, line 5: This breaks `go install`: https://kpt.dev/installation/?id=source

```
$ go install -v github.com/GoogleContainerTools/kpt@main
go install: github.com/GoogleContainerTools/kpt@main (in github.com/GoogleContainerTools/kpt@v1.0.0-beta.14.0.20220404215315-97c704191b82):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```

mengqiy: We have a couple of places in the catalog repo relies on this.

droot: too much fun with `go.mod` :) should we revert this PR to unblock PRs on `kpt-function-catalog` repo ?

martinmaly: I'd just go back to using hardcoded version for this dependency. should be easy fix. we did it once before.

martinmaly: #2967 should resolve the issue.

PR #2962, opened 2022-04-03
Improve Branch Tracking

martinmaly:
* Initialize fetch specs even on repo reopen to enable healing
  of cached repositories.
* Do not hardcode refMain; use branch registered with the repository
* Take more care with reference names
* Do not use force push.

### DIFF
--- a/porch/repository/pkg/git/draft.go
+++ b/porch/repository/pkg/git/draft.go
@@ -25,7 +25,6 @@ import (
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
-	"github.com/go-git/go-git/v5/storage"
 )
 
 type gitPackageDraft struct {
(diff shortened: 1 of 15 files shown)


porch/repository/pkg/git/draft.go
```diff
@@ -75,84 +69,73 @@ func (d *gitPackageDraft) UpdateLifecycle(ctx context.Context, new v1alpha1.Pack
...
-		newRef, err := setReference(repo.Storer, refNames.draft, d.commit)
-		if err != nil {
-			return nil, fmt.Errorf("failed to update package draft: %v", err)
-		}
 		// Push the package revision into a draft branch.
-		// TODO: implement config resolution rather than forcing the push (+)
-		pushRefSpecs = append(pushRefSpecs, config.RefSpec(fmt.Sprintf("+%s:%s", d.commit, refNames.draft)))
-		cleanup[refNames.proposed] = true // In case client downgraded packaget to draft from proposed
+		pushHelper.Push(d.commit, draftBranch.RefInLocal())
+		// Delete base branch?
```
justinsb, line 124: Maybe instead of "Delete base branch?" (which looks a bit like a TODO), we should say "Delete base branch (if exists)"

porch/repository/pkg/git/package.go
```diff
@@ -147,12 +146,12 @@ func (p *gitPackageRevision) GetUpstreamLock() (kptfile.Upstream, kptfile.Upstre
 }
 
 func (p *gitPackageRevision) getPackageRevisionLifecycle() v1alpha1.PackageRevisionLifecycle {
-	switch {
-	case p.ref == nil || p.ref.Name().IsTag():
+	switch ref := p.ref; {
+	case ref == nil:
 		return v1alpha1.PackageRevisionLifecycleFinal
-	case strings.HasPrefix(p.ref.Name().String(), refDraftPrefix):
+	case isDraftBranchNameInLocal(ref.Name()):
```
justinsb, line 152: Idea for the future - we could have a "parse ref" function which returns a struct including the kind (draft / proposed etc), the name etc.  I think we could also have the inverse function (struct -> ref name).

martinmaly: Yeah ... I've tried several experiments on this one and the current incarnation uses custom string types, but I was quite close creating struct already.
I am already moving all of the mapping logic into one location and I expect we'll turn it into a more generic component. At that point we'll need to up-level the abstraction too, most likely.

porch/repository/pkg/git/draft.go
martinmaly: done. thanks!

porch/repository/pkg/git/push.go
```diff
@@ -0,0 +1,81 @@
...
+
+func newPushHelper(repo *git.Repository) *pushHelper {
+	return &pushHelper{
+		repo:     repo,
+		pushRefs: map[plumbing.ReferenceName]plumbing.Hash{},
+		cond:     map[plumbing.ReferenceName]plumbing.Hash{},
+	}
+}
+
+func (h *pushHelper) Push(hash plumbing.Hash, to plumbing.ReferenceName) {
```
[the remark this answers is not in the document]
martinmaly: good idea. renamed

porch/repository/pkg/git/push.go
```diff
@@ -0,0 +1,81 @@
...
+	h.pushRefs[to] = hash
+}
+
+func (h *pushHelper) Delete(ref *plumbing.Reference) {
+	if ref != nil {
```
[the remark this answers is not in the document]
martinmaly: :) good catch!

PR #2963, opened 2022-04-03
Implement Git Repository Refresh

martinmaly:
Git repository is periodically refreshed.


porch/repository/pkg/git/git.go
```diff
@@ -124,6 +124,10 @@ type gitRepository struct {
 }
 
 func (r *gitRepository) ListPackageRevisions(ctx context.Context) ([]repository.PackageRevision, error) {
+	if err := r.update(ctx); err != nil {
```
justinsb, line 127: Should we rename `update`?  It's not self-evident what it does... (though admittedly I can't think what _else_ it could do!)

justinsb (approved):
Reviewed 2nd commit only as 1st commit is the prior PR

porch/repository/pkg/git/git.go
martinmaly: done. I went with `fetchRemoteRepository`
(sorry, I'm apparently quite bad with naming)

PR #2968, opened 2022-04-05
Lifecycle Commands Accept Multiple Packages

martinmaly:
Increase consistency across get, delete, propose, approve, reject

### DIFF
--- a/internal/cmdrpkgapprove/command.go
+++ b/internal/cmdrpkgapprove/command.go
@@ -16,6 +16,8 @@ package cmdrpkgapprove
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"github.com/GoogleContainerTools/kpt/internal/errors"
 	"github.com/GoogleContainerTools/kpt/internal/util/porch"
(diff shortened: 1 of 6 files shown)


internal/cmdrpkgapprove/command.go
```diff
@@ -86,15 +89,25 @@ func (r *runner) preRunE(cmd *cobra.Command, args []string) error {
...
 	namespace := *r.cfg.Namespace
-	name := args[0]
 
-	if err := porch.UpdatePackageRevisionApproval(r.ctx, r.client, client.ObjectKey{
-		Namespace: namespace,
-		Name:      name,
-	}, v1alpha1.PackageRevisionLifecycleFinal); err != nil {
-		return errors.E(op, err)
+	for _, name := range args {
+		switch err := porch.UpdatePackageRevisionApproval(r.ctx, r.client, client.ObjectKey{
```
yuwenma, line 97: qq: just curious about the switch/case usage here. Does it infer `UpdatePackageRevisionApproval` requires  other error handlers in the future?  If not, if/else seems to be more readable.

yuwenma (approved):

PR #2971, opened 2022-04-05
kpt fn exclude design proposal

natasha41575:
Design proposal for `kpt fn exclude`.

Issue: https://github.com/GoogleContainerTools/kpt/issues/2930
Implementation PR: https://github.com/GoogleContainerTools/kpt/pull/2957

### DIFF
--- a/docs/design-docs/05-kpt-fn-exclude.md
+++ b/docs/design-docs/05-kpt-fn-exclude.md
@@ -138,7 +138,7 @@ pipeline:
 We will support selection and exclusion with annotation and label selectors.
 
 ```shell
-$ kpt fn eval [PKG_DIR] -i set-labels:v0.1.5 --match-annotation foo=bar --exclude-annotation foo=baz
+$ kpt fn eval [PKG_DIR] -i set-labels:v0.1.5 --match-annotation foo=bar --exclude-annotation config.kubernetes.io/local-config=true
 ```
 
 ```yaml
@@ -151,7 +151,7 @@ pipeline:
         foo: bar
     exclude:
     - annotations:
-        foo: baz
+        config.kubernetes.io/local-config: "true"
 ```
 
 ## Open Issues/Questions
@@ -184,7 +184,6 @@ this annotation to every resource that they don't want functions to modify.
 
 Pros:
 - The user would not need to write an exclusion field for each function.
-- Simple, easy to understand UX.
 
 Cons:
 - It will not be possible to control at a function level. If users want some functions to modify the Kptfile/meta resources, and others not to touch it, 
@@ -222,7 +221,8 @@ Some problems with this approach:
 
 We can exclude all local-config resources (i.e. resources with the annotation `config.kubernetes.io/local-config: true`) from functions. This annotation is 
 currently used by kpt live apply to determine whether to apply the resource to the cluster. However, users may have some resources that they want kpt to modify or be 
-available to generator functions, but that they don't want to be applied to the cluster. In these cases, overloading the annotation in this way will be problematic.
+available as function inputs, but that they don't want to be applied to the cluster. In these cases, overloading the annotation in this way will
+break some functions.
 
 ### Mark meta resources with an annotation
 Kpt can add the annotation config.kubernetes.io/meta-resource: true to all meta resources, to allow users to easily target the Kptfile and functionConfigs. 
--- a/docs/design-docs/05-kpt-fn-exclude.md
+++ b/docs/design-docs/05-kpt-fn-exclude.md
@@ -1,0 +1,233 @@
+# Title
+
+* Author(s): Natasha Sarkar, natasha41575
+* Approver: Sunil Arora, droot
+
+## Why
+
+`kpt fn render` and `kpt fn eval` now include meta resources (functionConfigs and the Kptfile), 
+a breaking change from previous behavior. Users who use functions such as set-labels and set-namespace
+will now see their KRM functions acting on resources that were previously excluded by default. 
+For these users, we need to provide a way to preserve old behavior by introducing a mechanism to 
+exclude certain resources, such as resources of a certain GVK or local configuration resources, from 
+being acted on by kpt functions during `kpt fn render` and `kpt fn eval`.
+
+### Background
+
+Open source issue: https://github.com/GoogleContainerTools/kpt/issues/2930
+
+#### Redefining and inclusion of meta resources
+
+Previously, `kpt fn render` and `kpt fn eval` excluded functionConfigs and the Kptfile. A user could include t
+hese "meta resources'' by setting the `--include-meta-resources` flag. There is a separate discussion 
+and PR (https://github.com/GoogleContainerTools/kpt/pull/2894) with the following changes:
+
+- The definition of meta-resources. Kpt no longer considers functionConfigs as meta resources; they are KRM 
+resources like any other, and the fact that they are used to configure a KRM function does not mean they should be 
+treated specially. With this change, "meta-resources" only includes the Kptfile.
+- The `--include-meta-resources` flag is going away, and meta resources are processed by all kpt functions by default. 
+With this change, there is no mechanism for users to exclude the Kptfile.
+
+#### `kpt fn` selectors
+
+kpt fn render and kpt fn eval currently support selector-based mechanisms to target certain resources.
+
+Imperatively:
+
+```shell
+$ kpt fn eval [PKG_DIR] -i set-labels:v0.1.5 --match-kind Deployment
+```
+
+Declaratively:
+
+```yaml
+# pipeline of Kptfile
+pipeline:
+  mutators:
+  - image: set-labels:v0.1.5
+    selectors:
+    - kind: Deployment
+```
+
+These selectors only allow you to list out which resources you want to include. There is no current mechanism to exclude a certain kind of resource. 
+For example, the user may want to exclude the Kptfile.
+
+#### Prior art in kustomize
+The kustomize `replacements` feature (used in our new apply-replacements function), allows a mechanism to exclude certain resources from being modified. 
+`replacements` is one of the most popular among kustomize users.
+
+Kustomize has also had a few requests (such as this one: https://github.com/kubernetes/enhancements/pull/1232) to allow transformers to skip resources of a 
+certain GVK, which gives us confidence that a similar feature in kpt would be welcome.
+
+## Design
+
+### A new `exclude` option inline with selectors
+
+We can extend the current selector mechanism to allow exclusions. This can be achieved by new exclude flags for imperative workflow and a new exclude field in the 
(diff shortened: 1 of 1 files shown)


yuwenma (approved):

droot (approved):
@natasha41575 the proposal is looking good to me. We can also get a quick review from @bgrant0607 about it.

docs/design-docs/05-kpt-fn-exclude.md
```diff
@@ -0,0 +1,233 @@
...
+    exclude:
+    - kind: Deployment
+``` 
+
+### Selecting by labels or annotations
+
+We will support selection and exclusion with annotation and label selectors.
+
+```shell
+$ kpt fn eval [PKG_DIR] -i set-labels:v0.1.5 --match-annotation foo=bar --exclude-annotation foo=baz
```
bgrant0607, line 141: Excluding local config objects would probably be a better example.

docs/design-docs/05-kpt-fn-exclude.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Open Issues/Questions
+
+### How should kpt functions handle meta resources?
+
+One concern about this approach is that in order to exclude the Kptfile from being modified by horizontal
+transformations like set-namespace and set-labels, the user will have to write an exclusion field for each function.
+Arguably, the user shouldn't have to provide any special syntax for the Kptfile at all, because kpt and 
+kpt functions should be able to handle it correctly. A mitigation for this is that our own horizontal kpt functions
+can exclude the Kptfile within the function logic itself, and there will be no special handling logic in kpt.
```
bgrant0607, line 165: The SDK should make filtering out these local config objects easy.

docs/design-docs/05-kpt-fn-exclude.md
```diff
@@ -0,0 +1,233 @@
...
+  annotations:
+    config.kubernetes.io/hydration-override: false
+```
+
+If kpt sees `config.kubernetes.io/hydration-override: false`, it won't allow the functions to modify the resource. The user can add 
+this annotation to every resource that they don't want functions to modify.
+
+Pros:
+- The user would not need to write an exclusion field for each function.
+- Simple, easy to understand UX.
```
bgrant0607, line 187: I disagree this is easy to understand. It doesn't make sense to me. The right behavior entirely depends on what the function is doing. The logic really belongs in the function SDK.

docs/design-docs/05-kpt-fn-exclude.md
```diff
@@ -0,0 +1,233 @@
...
+
+Some problems with this approach:
+- It is verbose.
+- Verbosity increases if the user has more resources in the package.
+- The user will have to keep track of an exhaustive list of all their resources in their package just to exclude the Kptfile.
+- kpt and kpt functions should already know about the Kptfile, so this overhead on the user's side is unnecessary.
+
+### Rely on the existing annotation config.kubernetes.io/local-config
+
+We can exclude all local-config resources (i.e. resources with the annotation `config.kubernetes.io/local-config: true`) from functions. This annotation is
```
bgrant0607, line 223: This would break all functions that use local config pseudo-resources as implicit inputs. I had to workaround this behavior when we built the porch prototype.

docs/design-docs/05-kpt-fn-exclude.md
yuwenma: Yes, `set-namespace` excludes local-config resources inside its logic.

To have SDK filter out resources, we can check the type for Kptfile, but how can we tell which resources are function configs if not using a new annotation nor reusing local-config?

yuwenma: To be more specific, in a hydration pipeline if function1 and function2 have different function config `funcitonConfig1` and `functionConfig2 ` correspondingly, the resourceslist for `function1` is.
```
kind: ResourceList
functionConfig: 
    kind: functionConfig1
    ...
items:
- kind: Kptfile
   ...
- kind: functionConfig2
   ...
```
the resourceslist for `function2` is.
```
kind: ResourceList
functionConfig: 
    kind: functionConfig2
    ...
items:
- kind: Kptfile
   ...
- kind: functionConfig1
   ...
```
function1 does not know which input resource is function2's config, and vice versa.

natasha41575: In what cases will the user need to skip functionConfig? If they are our own horizontal functions like set-namespace, then setting the namespace of a functionConfig doesn't seem to really change or break any use cases. And if the user really wants to skip these functionConfigs, then they could use `kind`-based exclusion since every function has its own functionConfig kind, like `SetNamespace`. WDYT?

docs/design-docs/05-kpt-fn-exclude.md
natasha41575: I see your point, and totally agree that the logic belongs in the function SDK and should be controlled on a per-function level. What I was trying to say here is that the annotation is descriptive and clearly tells the user what it means. But I agree that it isn't clear or easy to understand why we would go this route, so I've taken out this line from the proposal.

docs/design-docs/05-kpt-fn-exclude.md
natasha41575: updated the example!

natasha41575:
Thanks for the review @bgrant0607! I updated the proposal based on your feedback. If you have any further feedback, let me know. I plan to merge this design proposal after your approval.

docs/design-docs/05-kpt-fn-exclude.md
yuwenma: My bad. I proposed this to simplify the annotations users need to deal with.

yuwenma:
@bgrant0607  I think there's a misalignment, just want to call it out.  @natasha41575 Correct me if I myself misunderstand it.

This proposal is motived from excluding Kptfile from resources (due to recent change on including meta resource as default), but it is trying to solve a bigger problem which is described in the ["Prior art in kustomize"](https://github.com/GoogleContainerTools/kpt/pull/2971/files#diff-8b2d3c513c8efd9535c12910bfe44513d61bdd987ac5c09b2b6b4a5846cb8f84R55) that "exclude certain resources from being modified ... is one of the most popular among kustomize users". It is not limited to Kptfile, but other KRM resources users do not want the functions to touch. 

Thus, my understanding is that different customers have different resource skipping list, and it won't be a function level feature, (thus the design focus is on what's the best syntax to skip those resources easily and reliably). When you suggest this filtering to be a function level feature, do you mean the function should provide a general solution on accepting a custom FieldSpec list, iterate and skip resources from this list?  The general syntax for such a function config might be
```functionConfig.yaml
apiVersion: kpt.dev.fn/v1alpha1
kind: SetXXX
exclude:
- kind: ConfigMap
- apiVersion: kpt.dev.fn/v1alpha1
```

natasha41575:
@yuwenma My understanding of what it means for filtering to be a "function level" feature from my discussions with @droot  is that the user should be able to control which resources are included for each function individually, rather than have kpt make the same decision for all functions in the pipeline. The design proposal currently is in alignment with that, because the user can specify `selectors`/`exclude` for each function separately.

yuwenma:
> @yuwenma My understanding of what it means for filtering to be a "function level" feature from my discussions with @droot is that the user should be able to control which resources are included for each function individually, rather than have kpt make the same decision for all functions in the pipeline. The design proposal currently is in alignment with that, because the user can specify `selectors`/`exclude` for each function separately.

Ah, saw the other [comment](https://github.com/GoogleContainerTools/kpt/issues/2930#issuecomment-1093364459). 
I think we shall define two phases for this exclude design proposal. 
Since the current solution is more of a workaround and eventually the filtering logic should be in the SDK.  
- Phase 1. As the current proposal says, filtering the resources before passing into the function. 
- Phase 2. 
      - kpt shall annotate Kptfile and inline ConfigMap with an internal annotation, propose `internal.config.kubernetes.io/kpt-meta`, the SDK should provide functions to filter resources by annotation (and this should be called inside a function)
      - FunctionConfig defines a general interface to filter resources by GVK. SDK provides the actual filtering logic (maybe a default behavior under the SDK resourceList `Process` so it's applied to all functions automatically).  It would be like

```yaml
kind: Resourcelist
items:
- apiVersion: v1
  kind: ConfigMap
  metadata:
       name: current-function-config
       annotations:
         internal.config.kubernetes.io/kpt-meta: true # marked as meta resource, exclude by default
- apiVersion: v1
  kind: ConfigMap
  metadata:
       name: other-function-config
  annotations:
       internal.config.kubernetes.io/kpt-meta: true # marked as meta resource, exclude by default
- apiVersion: kpt.dev/v1
  kind: Kptfile   # Kptfile, exclude by default
  ...
functionConfig:
  apiVersion: v1
  kind: ConfigMap
  metadata:
      name: current-function-config
      annotations:
        internal.config.kubernetes.io/exclude-meta: true
        internal.config.kubernetes.io/exclude: |     # function level exclude list.
           - kind:
           - kind:
```

bgrant0607:
I'm fine with the exclude feature as described. It seems useful for horizontal functions like set-labels, where the function itself wouldn't necessarily know which resources should be targeted.

For the problem of changing behavior with respect to meta-resources and local resources, I see it as a temporary workaround.

Functions need to ignore resources and resource types they don't care about or would operate incorrectly on. The pipeline should always pass through everything by default.

The function SDKs need to make it straightforward to implement such filtering. 

If some resources or resource types need to be made easier to identify, I'd be fine with introducing new intra-pipeline annotations to identify them.

ConfigMaps that are generated to be used as functionConfig from inline configMaps in the Kptfile should be automatically annotated as local-config objects. ConfigMaps and pseudo-resources in their own files would need to be annotated by their creators if they didn't want them to be instantiated in the cluster.

natasha41575:
> I'm fine with the exclude feature as described. It seems useful for horizontal functions like set-labels, where the function itself wouldn't necessarily know which resources should be targeted.

Sounds like we are all on board with the design as described, so I will go ahead and merge this PR along with the implementation PR. 

> The function SDKs need to make it straightforward to implement such filtering.

SGTM, and sounds like this would be something we discuss in Yuwen's phase 2 in [her comment](https://github.com/GoogleContainerTools/kpt/pull/2971#issuecomment-1097323671).  @yuwenma should we file a separate tracking issue for that work?

yuwenma:
> > I'm fine with the exclude feature as described. It seems useful for horizontal functions like set-labels, where the function itself wouldn't necessarily know which resources should be targeted.
> 
> Sounds like we are all on board with the design as described, so I will go ahead and merge this PR along with the implementation PR.
> 
> > The function SDKs need to make it straightforward to implement such filtering.
> 
> SGTM, and sounds like this would be something we discuss in Yuwen's phase 2 in [her comment](https://github.com/GoogleContainerTools/kpt/pull/2971#issuecomment-1097323671). @yuwenma should we file a separate tracking issue for that work?

Filed https://github.com/GoogleContainerTools/kpt/issues/3016

PR #2972, opened 2022-04-06
Define function from OCI image annotation

yuwenma:
- Define new Function fields to store function meta resources
- Map OCI image annotation to Function api resources


porch/repository/pkg/oci/oci.go
```diff
@@ -165,6 +168,58 @@ func (r *ociRepository) buildPackageRevision(ctx context.Context, name ImageDige
...
+func GetFunctionMeta(reference string, ctx context.Context) (*functionMeta, error) {
+	ref, err := name.ParseReference(reference)
+	if err != nil {
+		return nil, fmt.Errorf("parse image reference %v: %v", reference, err)
+	}
+	image, err := remote.Image(ref, remote.WithAuthFromKeychain(gcrane.Keychain), remote.WithContext(ctx))
+	if err != nil {
+		return nil, fmt.Errorf("pull remote image %v: %v", reference, err)
+	}
+	manifest, err := image.Manifest()
```
martinmaly, line 180: we may want to explore caching of manifests (in a follow-up). the containerregistry library does some caching (or can be told to) but it doesn't cache some parts of the images. I forget the detail; something to follow-up on.

porch/repository/pkg/oci/oci.go
```diff
@@ -165,6 +168,58 @@ func (r *ociRepository) buildPackageRevision(ctx context.Context, name ImageDige
...
+
+func GetDefaultFunctionConfig(manifest *v1.Manifest) []functionConfig {
+	val, ok := manifest.Annotations[ConfigMapFnKey]
+	if !ok {
+		return nil
+	}
+	return []functionConfig{
+		{
+			TypeMeta: metav1.TypeMeta{
+				Kind: "Configmap",
```
martinmaly, line 201: `ConfigMap` (uppercase "M")?

martinmaly: do we need ApiVersion, too?

porch/repository/pkg/oci/oci.go
```diff
@@ -165,6 +168,58 @@ func (r *ociRepository) buildPackageRevision(ctx context.Context, name ImageDige
...
+		FunctionTypes:    GetSliceFromAnnotation(FunctionTypesKey, manifest),
+		Description:      GetSingleFromAnnotation(DescriptionKey, manifest),
+		DocumentationUrl: GetSingleFromAnnotation(DocumentationURLKey, manifest),
+		Keywords:         GetSliceFromAnnotation(keywordsKey, manifest),
+		FunctionConfigs:  GetDefaultFunctionConfig(manifest),
+	}, nil
+}
+
+func GetDefaultFunctionConfig(manifest *v1.Manifest) []functionConfig {
+	val, ok := manifest.Annotations[ConfigMapFnKey]
```
martinmaly, line 194: identifying function config types other than config map will be interesting, in particular given that the casing is lowercase in the annotation keys, and TypeMeta uses PascalCase.

no need to change the code, just something to think about.

martinmaly (approved):
we'll want to think about testing soon as well...
there is an e2e test already that deals with functions that we can extend but we'll likely want more, too

porch/repository/pkg/oci/oci.go
yuwenma: Ah, that's a good idea! I'll explore the caching.

porch/repository/pkg/oci/oci.go
yuwenma: Good catch.

porch/repository/pkg/oci/oci.go
yuwenma: That's a good point! If detecting the actual PascalCase kind cannot be 100% accurate, maybe we can store the structured function config in json or jsonpath? e.g.
`set-image` function uses a functionconfig `SetImage`
```yaml
apiVersion: fn.kpt.dev/v1alpha1
kind: SetImage
metadata:
  name: my-func-config
image:
  name: nginx
  newName: bitnami/nginx
  newTag: 1.21.4
additionalImageFields:
- kind: MyKind
  create: false
  group: dev.example.com
  path: spec/manifest/images[]/image
  version: v1
```
It can be represented as
`kpt.dev.fn.meta.fnconfig.setimage.kind: SetImage`

Other SetImage meta data shares the same prefix `kpt.dev.fn.meta.fnconfig.setimage`. 


To structure the required fields, maybe we can marshall the required fields in json format. Relatively easy to rebuilt the structure by unmarshalling the json serialized string, bonus point is to save the comments. But it may hit the size limitation.
`kpt.dev.fn.meta.fnconfig.setimage.json: "{image: \n. name: nginx\n  newName: ...}"`

Another option is that maybe we can require the function config to be defined as CRD and the annotation stores where to download the CRD? (This may go al ittl bit crazy though), porch can install the CRD and the client tool just creates a new CR.

martinmaly: Good ideas. We can keep them in mind as we explore the more complex cases. My intuition at this point is that I'd start with putting data into annotations and if we run out of space, we can put them into the image directly ... I'm sure we'll figure it out, and we'll also have more information to make good decisions in the future.

PR #2974, opened 2022-04-06
Render When Needed

martinmaly:
* Re-render on update resources
* Render when performing mutations (but not when no mutations performed)


justinsb (approved):
lgtm!

PR #2976, opened 2022-04-06
Add spec.parent to PackageRevision

justinsb:
spec.parent will allow some form of context for both OCI & git.


martinmaly (approved):
Is the parent relationship encoded in the underlying package somehow (i.e. in kptfile or package contents)? or is it just pure Porch-level concept?

justinsb:
Thanks @martinmaly - I sent this one so I could unblock further exploration.  I don't _think_ we need to put it into the Kptfile (but time will tell).  For OCI in particular, we don't have the hierarchy that we get from git, and so I needed some "hint" that a child package should be nested inside a parent package.

Following this thought experiment through though, I think there are some interesting conclusions.  For example, I think a PackageRevision could map 1:1 with a Kptfile (i.e. that a package should only have a _single_  Kptfile in it).  I think that's aligned with how we're thinking about things, but not sure if you agree / if we've written it down?

PR #2980, opened 2022-04-07
feat: Kpt CLI Function discovery

yuwenma:
### Description

This PR extends the `kpt fn eval` autocompletion to help user to discover functions by using the porch function fields, which store the OCI image annotations. the work on porch side has been merged  https://github.com/GoogleContainerTools/kpt/pull/2972 

- Define `NewPorchFunctionClient`, which provides the restclient to query porch orchestrator. Set the restulapi call timeout  to be 3 seconds. This timeout is used to determine whether porch is enabled on the cluster side.  
- `ListKeywords` iterates the porch functions' `keywords` field, and suggest users a unique keyword list.
- `ListFunctions` iterates the porch functions and filter functions with the matching function type (default to `mutator`) and the `keywords` (if not given, accept all) a well as  the existing [catalog.kpt.dev](https://catalog.kpt.dev/) functions  
- All suggested functions will be filtered based on the function type (`mutator` or `validator`)/
 
### New UX experience 
- Add `--type` flag to register the autocomplete with value `mutator`, `validator`, and store the evaluated functions to the right kptfile list: `.pipeline.mutators` if mutator type, `.pipeline.validators` if validator type. 
- Add `--keywords` flag which pull the porch functions to get the `function.spec.keywords` and provide users a unique keywords list
- Modify the `--image` autocompletion function to rely on NewPorchFunctionClient to
  -  decide  whether porch is used
  -  autocomplete showing the `keywords` from porch functions
  - autocomplete showing the `functions` from porch and catalog.kpt.dev, filter by function type and keywords. 

### Fixes
- Change the image annotation prefix from `kpt.dev.fn` to `dev.kpt.fn` to follow the reverse domain convention
- Fix the annotation mapping for `description` and `documentationURL`

### Update:
- `--type` related code is merged in https://github.com/GoogleContainerTools/kpt/pull/2975

### TODO:
- Update `AddDefaultImagePathPrefix`  to append porch function prefix.  
- Redesign the kpt-functions-catalog release workflow, add the function meta resource to function OCI image annotation before pushing to [gcr.io/kpt-fn](gcr.io/kpt-fn), including the meta data in [https://catalog.kpt.dev/catalog-v2.json](https://catalog.kpt.dev/catalog-v2.json)
- Modify the existing autocompletion to merge the logic of fetching functions from catalog.kpt.fn with the porch function filtering. Use the same rules (by function type, by keywords) to suggest functions and keywords in `kpt fn eval` autocompletion.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/GoogleContainerTools/kpt
 go 1.17
 
 require (
-	github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220404223850-4fe7ba260b65
+	github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220407000100-7a5545349d71
 	github.com/cpuguy83/go-md2man/v2 v2.0.1
 	github.com/go-errors/errors v1.4.2
 	github.com/google/go-cmp v0.5.7
--- a/go.sum
+++ b/go.sum
@@ -72,8 +72,8 @@ github.com/Azure/go-autorest/tracing v0.6.0/go.mod h1:+vhtPC754Xsa23ID7GlGsrdKBp
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
 github.com/DataDog/datadog-go v3.2.0+incompatible/go.mod h1:LButxg5PwREeZtORoXG3tL4fMGNddJ+vMq1mwgfaqoQ=
-github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220404223850-4fe7ba260b65 h1:HTV58N6lLjmq4XNhlxzU/u435HO8zzXIYQET6WJc8aI=
-github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220404223850-4fe7ba260b65/go.mod h1:cMUBP0fY1F/1noOO1jJoIHM3R8Cp1BYWIG9XAZET5gI=
+github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220407000100-7a5545349d71 h1:ciqpo48GR1x4CI46l2t0+J1WXTQj0wSKI8ckixStVcQ=
+github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220407000100-7a5545349d71/go.mod h1:51Vk7QZ+XUzHCvQBi7t9tiWqTXvy6T13cv/inUXJJ0s=
 github.com/MakeNowJust/heredoc v0.0.0-20170808103936-bb23615498cd h1:sjQovDkwrZp8u+gxLtPgKGjk5hCxuy2hrRejBTA9xFU=
 github.com/MakeNowJust/heredoc v0.0.0-20170808103936-bb23615498cd/go.mod h1:64YHyfSL2R96J44Nlwm39UHepQbyR5q10x7iYa1ks2E=
 github.com/NYTimes/gziphandler v0.0.0-20170623195520-56545f4a5d46/go.mod h1:3wb06e3pkSAbeQ52E9H9iFoQsEEwGN64994WTCIhntQ=
--- a/internal/util/cmdutil/cmdutil.go
+++ b/internal/util/cmdutil/cmdutil.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/GoogleContainerTools/kpt/internal/fnruntime"
 	"github.com/GoogleContainerTools/kpt/internal/util/httputil"
+	"github.com/GoogleContainerTools/kpt/internal/util/porch"
 	"github.com/spf13/cobra"
 	"golang.org/x/mod/semver"
 	"sigs.k8s.io/kustomize/kyaml/kio"
@@ -176,6 +177,49 @@ func CheckDirectoryNotPresent(outDir string) error {
 		return err
 	}
 	return nil
+}
+
+func GetKeywordsFromFlag(cmd *cobra.Command) []string {
+	flagVal := cmd.Flag("keywords").Value.String()
+	flagVal = strings.TrimPrefix(flagVal, "[")
+	flagVal = strings.TrimSuffix(flagVal, "]")
+	splitted := strings.Split(flagVal, ",")
+	var trimmed []string
+	for _, val := range splitted {
+		trimmed = append(trimmed, strings.TrimSpace(val))
+	}
+	return trimmed
+}
+
+// SuggestFunctions looks for functions from kpt curated catalog list as well as the Porch
+// orchestrator to suggest functions.
+func SuggestFunctions(cmd *cobra.Command) []string {
+	var functions []string
+	matchers := []porch.Matcher{
+		porch.TypeMatcher{FnType: cmd.Flag("type").Value.String()},
+		porch.KeywordsMatcher{Keywords: GetKeywordsFromFlag(cmd)},
+	}
+	porchFunctions := porch.MatchFunctions(cmd.Context(), matchers...)
+	functions = append(functions, porch.ToShortNames(porchFunctions)...)
+	// TODO(yuwenma): We should filter these catalog.kpt.fn functions by its annotation as well! This requires changing
+	// the kpt-functions-catalog release workflow to include new annotations.
+	functions = append(functions, FetchFunctionImages()...)
+	var dedup []string
+	fnMap := map[string]bool{}
+	for _, function := range functions {
+		if _, ok := fnMap[function]; !ok {
+			dedup = append(dedup, function)
+			fnMap[function] = true
+		}
+	}
+	return dedup
+}
+
+// SuggestKeywords looks for all the unique keywords from Porch functions. This keywords
+// can later help users to select functions.
+func SuggestKeywords(cmd *cobra.Command) []string {
+	functions := porch.MatchFunctions(cmd.Context(), porch.TypeMatcher{FnType: cmd.Flag("type").Value.String()})
+	return porch.UnifyKeywords(functions)
 }
 
 // FetchFunctionImages returns the list of latest function images from catalog.kpt.dev
--- a/internal/util/porch/function.go
+++ b/internal/util/porch/function.go
@@ -1,0 +1,144 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package porch
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+
+	// Enable the GCP Authentication plugin
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+)
+
+const expirationNanoSecond = 10
+
+var PorchExpiration = time.Duration(expirationNanoSecond) * time.Second
+
+// FunctionGetter gets the list of v1alpha1.Functions from the cluster.
+type FunctionGetter struct {
+	ctx context.Context
+}
+
+func (f FunctionGetter) Get() []v1alpha1.Function {
+	kubeflags := genericclioptions.NewConfigFlags(true)
+	client, err := CreateRESTClient(kubeflags)
+	if err != nil {
+		return nil
+	}
+	scheme := runtime.NewScheme()
+	if err := v1alpha1.SchemeBuilder.AddToScheme(scheme); err != nil {
+		return nil
+	}
+	codec := runtime.NewParameterCodec(scheme)
+	var functions v1alpha1.FunctionList
+	if err := client.Get().
+		Timeout(PorchExpiration).
+		Resource("functions").
+		VersionedParams(&metav1.GetOptions{}, codec).
+		Do(f.ctx).
+		Into(&functions); err != nil {
+		return nil
+	}
+	return functions.Items
+}
+
+type Matcher interface {
+	Match(v1alpha1.Function) bool
+}
+
+var _ Matcher = TypeMatcher{}
+var _ Matcher = KeywordsMatcher{}
+
+type TypeMatcher struct {
+	FnType string
+}
+
+// Match determines whether the `function` (which can be multi-typed), belongs
+// to the matcher's FnType. type value should only be `validator` or `mutator`.
+func (m TypeMatcher) Match(function v1alpha1.Function) bool {
+	for _, actualType := range function.Spec.FunctionTypes {
+		if string(actualType) == m.FnType {
+			return true
+		}
+	}
+	return false
+}
+
+type KeywordsMatcher struct {
+	Keywords []string
+}
+
+// Match determines whether the `function` has keywords which match the matcher's `Keywords`.
+// Experimental: This logic may change to only if all function keywords are found from  matcher's `Keywords`,
+// can it claims a match (return true).
+func (m KeywordsMatcher) Match(function v1alpha1.Function) bool {
+	if len(m.Keywords) == 0 {
+		// Accept all functions if keywords are not given.
+		return true
+	}
+	for _, actual := range function.Spec.Keywords {
+		for _, expected := range m.Keywords {
+			if actual != expected {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func MatchFunctions(ctx context.Context, matchers ...Matcher) []v1alpha1.Function {
+	var suggestedFunctions []v1alpha1.Function
+	functions := FunctionGetter{ctx: ctx}.Get()
+	for _, function := range functions {
+		for _, matcher := range matchers {
+			if matcher.Match(function) {
+				suggestedFunctions = append(suggestedFunctions, function)
+			}
+		}
+	}
+	return suggestedFunctions
+}
+
+// ToShortNames trims the function image to remove the OCI repo prefix and
+// only show the actual image and tag (or digest).
+// TODO(yuwenma): or may users prefer the Function `meta.name`?
+func ToShortNames(functions []v1alpha1.Function) []string {
+	var shortNameFunctions []string
+	for _, function := range functions {
+		slash := strings.LastIndex(function.Spec.Image, "/")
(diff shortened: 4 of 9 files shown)


thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -164,43 +175,58 @@ func (r *EvalFnRunner) NewFunction() *kptfile.Function {
...
+// SaveFnToKptfile adds the evaluated function and its arguments to Kptfile `pipeline.mutators` list.
+func (r *EvalFnRunner) SaveFnToKptfile() {
+	pr := printer.FromContextOrDie(r.Ctx)
+	kf, err := pkg.ReadKptfile(filesys.FileSystemOrOnDisk{}, r.RunFns.Path)
+	if err != nil {
+		pr.Printf("function not added: Kptfile not exists\n")
+		return
+	}
+	if kf.Pipeline == nil {
+		kf.Pipeline = &kptfile.Pipeline{}
```
martinmaly, line 214: this may have side effect on the Kptfile if no function is added; this will still add the Pipeline section, albeit empty one.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -164,43 +175,58 @@ func (r *EvalFnRunner) NewFunction() *kptfile.Function {
...
+	}
+	switch r.FnType {
+	case "mutator":
+		if kf.Pipeline.Mutators == nil {
+			kf.Pipeline.Mutators = []kptfile.Function{}
+		}
+		newFns := r.updateFnList(kf.Pipeline.Mutators)
+		kf.Pipeline.Mutators = newFns
```
martinmaly, line 222: append works on nil lists to all this can be simplified as:

kf.Pipeline.Mutators = r.updateFnList(kf.Pipeline.Mutators)

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -41,14 +41,23 @@ func GetEvalFnRunner(ctx context.Context, parent string) *EvalFnRunner {
...
 	r.Command.Flags().StringVarP(&r.Dest, "output", "o", "",
 		fmt.Sprintf("output resources are written to provided location. Allowed values: %s|%s|<OUT_DIR_PATH>", cmdutil.Stdout, cmdutil.Unwrap))
 	r.Command.Flags().StringVarP(
 		&r.Image, "image", "i", "", "run this image as a function")
 	_ = r.Command.RegisterFlagCompletionFunc("image", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-		return cmdutil.FetchFunctionImages(), cobra.ShellCompDirectiveDefault
+		return cmdutil.SuggestFunctions(cmd), cobra.ShellCompDirectiveDefault
+	})
+	r.Command.Flags().StringArrayVarP(
+		&r.Keywords, "keywords", "k", []string{}, "discover the keywords correlated functions")
```
martinmaly, line 53: I'm having hard time figuring out the help text. I am also having even harder time coming up with a better suggestion :)

my best try: "filter functions that match one or more keywords"

porch/repository/pkg/oci/oci.go
```diff
@@ -215,7 +215,7 @@ func GetSliceFromAnnotation(key string, manifest *v1.Manifest) []string {
 }
 
 func GetSingleFromAnnotation(key string, manifest *v1.Manifest) string {
-	if val, ok := manifest.Annotations[FunctionTypesKey]; ok {
+	if val, ok := manifest.Annotations[key]; ok {
```
martinmaly, line 218: good catch!

internal/porchutil/function.go
```diff
@@ -0,0 +1,155 @@
...
+		if !ContainsKeywords(function, keywords) {
+			continue
+		}
+		suggestedFunctions = append(suggestedFunctions, GetFunctionShortName(function))
+	}
+	return suggestedFunctions, nil
+}
+
+func visitFunctions() ([]v1alpha1.Function, error) {
+	kubeconfig, ok := os.LookupEnv("KUBECONFIG")
```
martinmaly, line 131: hmm ... I've never needed to do this; the k8s go-client has wrappers that take care of it.

This is what I am doing (not claiming that it is correct though :) but it _seems_ to work:

https://github.com/GoogleContainerTools/kpt/blob/main/internal/util/porch/client.go#L30

internal/porchutil/function.go
```diff
@@ -0,0 +1,155 @@
...
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	if err != nil {
+		log.Fatal(err)
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		log.Fatal(err)
+	}
+	var functions v1alpha1.FunctionList
+	raw, err := clientset.RESTClient().Get().Timeout(PorchExpiration).AbsPath(functionAPI).DoRaw(context.TODO())
```
martinmaly, line 146: interesting ... this will list all functions in all namespaces in the k8s cluster.
not something to figure out now but the namespace-deness of k8s is making things less straightforward since it is unclear which namespace we'd be querying if we had to pick.

So this is fine for now.

internal/porchutil/function.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("fail to call %v:%v", functionAPI, err)
+	}
+	err = json.Unmarshal(raw, &functions)
```
martinmaly, line 150: this shouldn't be needed either. I find the controller-runtime client does the unmarshaling too.

internal/porchutil/function.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	var keywords []string
+	for keyword := range keywordsMap {
+		keywords = append(keywords, keyword)
+	}
+	return keywords, nil
+}
+
+func getKeywordsFromFlag(cmd *cobra.Command) []string {
+	flagVal := cmd.Flag("keywords").Value.String()
```
martinmaly, line 93: Cobra doesn't have facility to accept lists as flag values? I'm just wondering if we can avoid the custom parser.

internal/porchutil/function.go
```diff
@@ -0,0 +1,155 @@
...
+			if actual != expected {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// ListKeywords returns all the unique keywords from Porch Functions.
+func (p *PorchFunctionClient) ListKeywords(cmd *cobra.Command) ([]string, error) {
```
martinmaly, line 74: another way to do this could be that this function accepts:

`func match(*Function) bool`

and we can compose the matching on type and on keywords using that.

To me the 'red' flag is the interdependence of Porch Client (which conceptually can/should be purely around Porch API concepts) and cobra Commands, resulting in leaky abstraction

internal/porchutil/function.go
```diff
@@ -0,0 +1,155 @@
...
+
+var (
+	functionAPI     = "/apis/" + v1alpha1.SchemeGroupVersion.String() + "/functions"
+	PorchExpiration = time.Duration(expirationNanoSecond) * time.Second
+)
+
+// NewPorchFunctionClient provides the PorchFunctionClient which connects to Porch server
+// and gets the functions.
+func NewPorchFunctionClient() *PorchFunctionClient {
+	functions, err := visitFunctions()
```
martinmaly, line 32: nit: visitFunctions signals (to me) that the function is iterating over an existing data, rather than fetching it. consider whether another name would convey the purpose more clearly

internal/porchutil/function.go
```diff
@@ -0,0 +1,155 @@
+package porchutil
```
martinmaly, line 1: missing header

also, we have this package: https://github.com/GoogleContainerTools/kpt/tree/main/internal/util/porch

would it be a good fit for this kind of utility? (please advise, I am not super familiar with kpt repo organization conventions)

internal/porchutil/function.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/client-go/kubernetes"
+
+	// Enable the GCP Authentication plugin
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+const expirationNanoSecond = 3
```
martinmaly, line 22: that's very short amount of time :)

internal/porchutil/function.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return &PorchFunctionClient{IsPorchAvailable: false, Functions: nil}
+	}
+	return &PorchFunctionClient{IsPorchAvailable: true, Functions: functions}
+}
+
+type PorchFunctionClient struct {
+	IsPorchAvailable bool
+	Functions        []v1alpha1.Function
+}
```
martinmaly, line 42: slight conceptual misalignment. this struct is actually not a client. I wonder if it would work as well to have a simpler interface, possibly avoiding the IsPorchAvailable bool and return list of functions directly?

(I am not feeling too strongly though)

martinmaly (approved):
some comments for your consideration.

e2e/testdata/fn-eval/save-fn/validator-type/.expected/diff.patch
```diff
@@ -0,0 +1,34 @@
+diff --git a/Kptfile b/Kptfile
+index d9e2f05..c6e80ef 100644
```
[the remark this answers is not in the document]
yuwenma: they are stable.

e2e/testdata/fn-eval/save-fn/validator-type/.expected/config.yaml
```diff
@@ -0,0 +1,23 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+testType: eval
+image: set-namespace:v0.1.3
+args:
+  namespace: staging
+stdErr: |
+  [RUNNING] "gcr.io/kpt-fn/set-namespace:v0.1.3"
+  [PASS] "gcr.io/kpt-fn/set-namespace:v0.1.3" in 0s
```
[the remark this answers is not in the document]
yuwenma: Exactly.

internal/porchutil/function.go
yuwenma: The IETF says [current request timeout implementation select times between 30-120s](https://tools.ietf.org/id/draft-thomson-hybi-http-timeout-00.html#:~:text=Knowing%20the%20time%20available%20to,empirically%20determined%20to%20be%20safe.)

Maybe we can set that to 30s?

martinmaly: this is for tab completion, correct? IMO 30s is too long.
let's try some value now (I'd recommend using the Duration type) and we'll adjust.

internal/porchutil/function.go
yuwenma: that's true. I defined the struct to make the caller code be more specific that only when porch available will we use  porch functions
```
if client.IsPorchAvailable():
   client.ListFunctions
```
I don't have a strong preference but agree the naming is misaligned. I'll update.

internal/porchutil/function.go
droot: I think we should use one `porch` client for all porch APIs. This seems to duplicate the code in `internal/util/porch` ?

internal/porchutil/function.go
droot: I wonder what the experience is like in `porch` disconnected mode, does it fail fast ? (you might need two timeout, one for connectionTimeout, another one timing out on waiting from response).

internal/porchutil/function.go
droot: @yuwenma would recommend using the existing porch client, it uses `controller-runtime` underneath and that does all of this for you for free.

internal/util/cmdutil/cmdutil.go
```diff
@@ -178,6 +179,44 @@ func CheckDirectoryNotPresent(outDir string) error {
...
+	functions = append(functions, FetchFunctionImages()...)
+	var dedup []string
+	fnMap := map[string]bool{}
+	for _, function := range functions {
+		if _, ok := fnMap[function]; !ok {
+			dedup = append(dedup, function)
+			fnMap[function] = true
+		}
+	}
+	return dedup
```
droot, line 204: IMO, we should fire off both lookups in parallel (porch and functionImages) so that autocomplete UX feels instantaneous (more important from `porch disconnected` mode)

go.sum
```diff
@@ -74,6 +74,10 @@ github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym
 github.com/DataDog/datadog-go v3.2.0+incompatible/go.mod h1:LButxg5PwREeZtORoXG3tL4fMGNddJ+vMq1mwgfaqoQ=
 github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220404223850-4fe7ba260b65 h1:HTV58N6lLjmq4XNhlxzU/u435HO8zzXIYQET6WJc8aI=
 github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220404223850-4fe7ba260b65/go.mod h1:cMUBP0fY1F/1noOO1jJoIHM3R8Cp1BYWIG9XAZET5gI=
+github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220406191334-921abb821a86 h1:CczJ3rznKSvyj9VDpLZdTcRy2ihGdeHR7ImvlufBwkQ=
+github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220406191334-921abb821a86/go.mod h1:51Vk7QZ+XUzHCvQBi7t9tiWqTXvy6T13cv/inUXJJ0s=
+github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220407000100-7a5545349d71 h1:ciqpo48GR1x4CI46l2t0+J1WXTQj0wSKI8ckixStVcQ=
+github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220407000100-7a5545349d71/go.mod h1:51Vk7QZ+XUzHCvQBi7t9tiWqTXvy6T13cv/inUXJJ0s=
```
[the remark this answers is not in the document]
yuwenma: wow, how did you catch that... TBH, I never really pay attention to go.mod

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -164,43 +175,58 @@ func (r *EvalFnRunner) NewFunction() *kptfile.Function {
...
 	if !found {
-		newMutators = append(newMutators, *r.NewFunction())
+		newFns = append(newFns, *r.NewFunction())
 		pr.Printf("adding function to Kptfile\n")
 	}
-	kf.Pipeline.Mutators = newMutators
+	return newFns
+}
+
+// SaveFnToKptfile adds the evaluated function and its arguments to Kptfile `pipeline.mutators` list.
```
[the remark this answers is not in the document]
yuwenma: good catch. thanks

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
yuwenma: This is called after the function run, so we always expect new functions to be added (unless it already exists in Kptfile).

internal/porchutil/function.go
yuwenma: Sounds good. Yeah, I actually did some code refactoring before uploading but didn't rethink about the function names.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
martinmaly: ack.  thanks

internal/porchutil/function.go
yuwenma: Yeah, I explored this a little bit. I used `StringArrayVarP` for "keywords", which users can provide an array of keywords in the form as `--keywords XX --keywords YY`. But when parsing the flag from commands, it does not provide such options.

internal/porchutil/function.go
yuwenma: really like this suggestion! thank you. Refactored the code.

internal/util/cmdutil/cmdutil.go
yuwenma: I'll do that in the follow up PR since I also plan to make some major changes to the `fetchFunctions` (including OCI annotations and filter functions by annotation)

internal/porchutil/function.go
yuwenma: Yeah, I didn't noticed that client code. Updated.

internal/porchutil/function.go
yuwenma: set timeout to 10.

internal/porchutil/function.go
martinmaly: agreed. let's keep this; for autocomplete being broader is good, and we can adjust in the future.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
martinmaly: quite possibly. IMO we need to (find a way) to communicate that the keywords here are used to filter functions, using the keywords stored on the functions.

(hopefully better try than my previous one): "Filter function; include functions whose keywords match one or more of the keywords provide in this flag"

(may be too wordy though; perhaps we can trim it down)

internal/porchutil/function.go
martinmaly: I see. re-reading the code it seems that for the completion, less arg parsing has been done. That _may_ mean that if customer enters --keywords XX --keywords YY ... the code may not see both options?

internal/util/porch/function.go
```diff
@@ -0,0 +1,160 @@
...
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+
+	// Enable the GCP Authentication plugin
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+)
+
+const expirationNanoSecond = 10
+
+var (
+	functionAPI     = "/apis/" + v1alpha1.SchemeGroupVersion.String() + "/functions"
```
martinmaly, line 35: GitHub observers this is unused

internal/util/porch/function.go
```diff
@@ -0,0 +1,160 @@
...
+	flagVal := cmd.Flag("keywords").Value.String()
+	flagVal = strings.TrimPrefix(flagVal, "[")
+	flagVal = strings.TrimSuffix(flagVal, "]")
+	splitted := strings.Split(flagVal, ",")
+	var trimmed []string
+	for _, val := range splitted {
+		trimmed = append(trimmed, strings.TrimSpace(val))
+	}
+	return trimmed
+}
```
martinmaly, line 160: [nit] I'd recommend moving this function out to the command-related code, i.e. into cmdutil.go to keep the porch code unaware of cobra.

martinmaly (approved):
Couple more nits but LGTM once tests pass.

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/GoogleContainerTools/kpt
 go 1.17
 
 require (
-	github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220404223850-4fe7ba260b65
+	github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220407000100-7a5545349d71
```
martinmaly, line 6: we now need `github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220404223850-4fe7ba260b65` which is already at HEAD so this file can be reverted.

yuwenma: thanks.

martinmaly: I'm sorry ... I was wrong. I thought my PR was already merged; alas, it wasn't. your code was correct.

yuwenma: I pulled the latest and got v0.0.0-20220411164219-e3555a1d90a9. Shall I use this instead?

PR #2986, opened 2022-04-08
Do not use kpt Git URL parsing

martinmaly:
The parser analyzes local environment and has other issues.
Rather, temporarily accept `--directory`, `--branch` and `--ref`
flags for `kpt alpha rpkg clone` and `kpt alpha repo reg`
respectively.


droot (approved):
Thank you so much for the fix (especially so last minute).

PR #2991, opened 2022-04-11
Use Updated Porch API in kpt

martinmaly:
Advance Porch API dependency in Porch and adjust code to match.


porch/apiserver/pkg/registry/porch/background.go
```diff
@@ -151,7 +151,17 @@ func (b *background) runOnce(ctx context.Context) error {
 	}
 
 	for i := range repositories.Items {
-		b.cacheRepository(ctx, &repositories.Items[i])
+		repo := &repositories.Items[i]
+
+		if repo.Spec.Content == "PackageRevision" {
```
yuwenma, line 156: Just curious in case I miss something: I guess we haven't launched "porch" api resources yet. why do we need this backward compatibility handler?

martinmaly: Because folks experimenting with the system may have stale Repository resources in their cluster and they'll be surprised when they reload Porch and their packages disappear.

There is a issue open to remove this code

justinsb: Makes sense - thanks @martinmaly .  We could also create a dedicated migration function and run it on startup; I'm sure this won't be the last time.

justinsb (approved):

porch/apiserver/pkg/registry/porch/background.go
yuwenma: I see. Thanks for the explanation. Yeah, I like the idea of migration function since it separate this logic from source code.

martinmaly: Good idea! I'll file it away and on next need we'll go with the migration function. I can already see how it evolves into a migration "init pod" :) Fun!

PR #2993, opened 2022-04-12
fix:  `kpt fn eval --save` shouldn't add `name` field for the function

yuwenma:
the `name` is for merging key, not function identifier.

Output after the fix:
```bash
kpt fn eval yu -s -i set-namespace:v0.3.2 -- namespace=yu
[RUNNING] "gcr.io/kpt-fn/set-namespace:v0.3.2"
[PASS] "gcr.io/kpt-fn/set-namespace:v0.3.2" in 700ms
  Results:
    [info]: namespace updated 
adding function to Kptfile
Kptfile updated
```

```yaml
# Kptfile
apiVersion: kpt.dev/v1
kind: Kptfile
metadata:
  name: yu
  annotations:
    config.kubernetes.io/local-config: "true"
info:
  description: sample description
pipeline:
  mutators:
    - image: gcr.io/kpt-fn/set-namespace:v0.3.2
      configMap:
        namespace: yu
```

### DIFF
--- a/e2e/testdata/fn-eval/save-fn/exec/.expected/diff.patch
+++ b/e2e/testdata/fn-eval/save-fn/exec/.expected/diff.patch
@@ -1,8 +1,8 @@
 diff --git a/Kptfile b/Kptfile
-index d9e2f05..3d2827b 100644
+index d9e2f05..daff2a6 100644
 --- a/Kptfile
 +++ b/Kptfile
-@@ -2,3 +2,8 @@ apiVersion: kpt.dev/v1
+@@ -2,3 +2,7 @@ apiVersion: kpt.dev/v1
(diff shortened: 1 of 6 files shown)


martinmaly (approved):
Should we have a test for the --save functionality? (presumably change like this would cause test failure and necessitate test update).

yuwenma:
> Should we have a test for the --save functionality? (presumably change like this would cause test failure and necessitate test update).

Wow, you are so fast. I was updating the e2e tests while you've already finished the review. 

Yes, we have pretty good test coverage for the --save, but the local result won't give the same git ref for `e2e/XXX/.expected/diff.patch` as the e2e tests, so I normally let the e2e test run first and use the output to update mine.

PR #2997, opened 2022-04-13
support reusing pods in pod evaluator

mengqiy:
Added a goroutine to manage the cache and handle the coordination.
Use to labels to retrieve old pods.
Set last use time in annotation.
Add GC to delete old pods.

I have manually tested it. Created the PR to get some early feedback.
I'm working on finding a way to test it without making the test setup too complex.

### DIFF
--- a/porch/func/internal/podevaluator.go
+++ b/porch/func/internal/podevaluator.go
@@ -17,9 +17,11 @@ package internal
 import (
 	"context"
 	"fmt"
+	"net"
 	"path"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/GoogleContainerTools/kpt/porch/func/evaluator"
@@ -30,6 +32,7 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -37,25 +40,26 @@ import (
 )
 
 const (
-	defaultWrapperServerPort = 9446
+	defaultWrapperServerPort = "9446"
 	volumeName               = "wrapper-server-tools"
 	volumeMountPath          = "/wrapper-server-tools"
 	wrapperServerBin         = "wrapper-server"
 	gRPCProbeBin             = "grpc-health-probe"
+	krmFunctionLabel         = "fn.kpt.dev/image"
+	lastUseTimeAnnotation    = "fn.kpt.dev/last-use"
+
+	channelBufferSize = 128
 )
 
 type podEvaluator struct {
-	// kubeClient is the kubernetes client
-	kubeClient client.Client
-	// namespace holds the namespace where the executors run
-	namespace string
-	// wrapperServerImage is the image name of the wrapper server
-	wrapperServerImage string
+	requestCh chan *clientConnRequest
+
+	podCacheManager *podCacheManager
 }
 
 var _ Evaluator = &podEvaluator{}
 
-func NewPodEvaluator(namespace, wrapperServerImage string) (Evaluator, error) {
+func NewPodEvaluator(namespace, wrapperServerImage string, interval, ttl time.Duration) (Evaluator, error) {
 	restCfg, err := config.GetConfig()
 	if err != nil {
 		return nil, fmt.Errorf("failed to get rest config: %w", err)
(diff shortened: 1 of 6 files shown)


porch/func/internal/podevaluator.go
```diff
@@ -65,30 +74,109 @@ func NewPodEvaluator(namespace, wrapperServerImage string) (Evaluator, error) {
...
+
+	pe := &podEvaluator{
 		kubeClient:         cl,
 		namespace:          namespace,
 		wrapperServerImage: wrapperServerImage,
-	}, nil
+		gcScanInternal:     time.Duration(interval) * time.Second,
+		podTTL:             time.Duration(ttl) * time.Second,
+		imageEntrypoint:    map[string][]string{},
+		grpcConnectionPool: map[string]*grpc.ClientConn{},
```
martinmaly, line 85: is the key for both maps the same?
If so, we could combine them into a single map, keyed by image and it contains the cached entrypoint info and the gRPC client to then call the function.

porch/func/internal/podevaluator.go
```diff
@@ -65,30 +74,109 @@ func NewPodEvaluator(namespace, wrapperServerImage string) (Evaluator, error) {
...
+	}
+	go pe.garbageCollector()
```
martinmaly, line 87: OK to add TODO for now and implement in a follow-up.
we should have a graceful shutdown for this goroutine.

porch/func/internal/podevaluator.go
```diff
@@ -65,30 +74,109 @@ func NewPodEvaluator(namespace, wrapperServerImage string) (Evaluator, error) {
...
+	return pe, nil
 }
 
-func (pe *podEvaluator) EvaluateFunction(ctx context.Context, req *pb.EvaluateFunctionRequest) (*pb.EvaluateFunctionResponse, error) {
-	ref, err := name.ParseReference(req.Image)
-	if err != nil {
-		return nil, err
+// TODO: We can use Watch + periodically reconciliation to manage the pods,
```
martinmaly, line 91: indeed!

porch/func/internal/podevaluator.go
```diff
@@ -65,30 +74,109 @@ func NewPodEvaluator(namespace, wrapperServerImage string) (Evaluator, error) {
...
 	}
-	img, err := remote.Image(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain))
-	if err != nil {
-		return nil, err
+}
+
+func (pe *podEvaluator) retrieveOrCreatePod(ctx context.Context, image string) (client.ObjectKey, error) {
+	// Try to retrieve the pod first.
+	// Lookup the pod by label to see if there is a pod that can be reused.
+	// Looking it up locally may not work if there are more than one instance of the function runner,
```
martinmaly, line 144: although we should probably check local cache first, and if not found, check for live pods (created by another runner), and if not found, create a new pod?

porch/func/internal/podevaluator.go
```diff
@@ -65,30 +74,109 @@ func NewPodEvaluator(namespace, wrapperServerImage string) (Evaluator, error) {
...
-	var fnCmd []string
-	if len(cfg.Entrypoint) != 0 {
-		fnCmd = cfg.Entrypoint
-	} else {
-		fnCmd = cfg.Cmd
+
+	// Create pod otherwise.
+
+	fnCmd, found := pe.imageEntrypoint[image]
+	if !found {
```
martinmaly, line 162: should we do this check before listing the pods above?

porch/func/internal/podevaluator.go
```diff
@@ -65,30 +74,109 @@ func NewPodEvaluator(namespace, wrapperServerImage string) (Evaluator, error) {
...
+	// Looking it up locally may not work if there are more than one instance of the function runner,
+	// since the pod may be created by one the other instance and the current instance is not aware of it.
+	// TODO: It's possible to set up a Watch in the fn runner namespace, and always try to maintain a up-to-date local cache.
+	podList := &corev1.PodList{}
+	err := pe.kubeClient.List(ctx, podList, client.InNamespace(pe.namespace), client.MatchingLabels(map[string]string{krmFunctionLabel: transformImageName(image)}))
+	if err == nil && len(podList.Items) > 0 {
+		pod := podList.Items[0]
+		klog.Infof("reusing KRM function evaluator pod %q for %q", pod.Name, image)
+		// Update the last-use annotation in the pod
+		if err = patchPodWithUnixTimeAnnotation(context.Background(), pe.kubeClient, &pod); err != nil {
```
martinmaly, line 153: all of these operations will add latency to the function evaluation. We'll want a fast path to fetch gRPC client from a cache and make a call through it, and then secondary slow paths that handle cases when the client or pod don't exist.

As for last use update, we could also do it after the call (outside of the critical request/response path).

porch/func/internal/podevaluator.go
```diff
@@ -65,30 +74,109 @@ func NewPodEvaluator(namespace, wrapperServerImage string) (Evaluator, error) {
...
+					}
+					continue
+				}
+				if time.Unix(lu, 0).Add(pe.podTTL).Before(time.Now()) {
+					// deleting the grpc connection from the pool
+					image := pod.Spec.Containers[0].Image
+					cc, found := pe.grpcConnectionPool[image]
+					if found {
+						klog.Infof("deleting grpc connection %v for %v", cc.Target(), image)
+						delete(pe.grpcConnectionPool, image)
```
martinmaly, line 126: we'll need synchronization between the garbageCollector and the request handler goroutines

this will need to handle cases not only of concurrent access and updates, but also concurrent cache misses with subsequent pod spin-up.

Ideally we need logic such as:

```
gRPCClient, found := findCachedClientForFunction(image)
if !found {
  gRPCClient = coordinatePodSpinUpAndWakeMeWhenClientAvailable(image)
}
```

the coordinated spin-up should block all but one request as they wait for the pod to come up, while the running goroutine does the spin-up.

One way to do it is to have a goroutine dedicated to managing the image cache and goroutines handling the requests get the gRPC client(s) via channels. That way the synchronization may be easiest.

porch/func/internal/podevaluator.go
```diff
@@ -160,59 +255,70 @@ func (pe *podEvaluator) EvaluateFunction(ctx context.Context, req *pb.EvaluateFu
...
-	podIP := pod.Status.PodIP
+func (pe *podEvaluator) EvaluateFunction(ctx context.Context, req *pb.EvaluateFunctionRequest) (*pb.EvaluateFunctionResponse, error) {
+	podKey, err := pe.retrieveOrCreatePod(ctx, req.Image)
+	if err != nil {
+		return nil, err
+	}
+	podIP, err := podIpIfRunningAndReady(ctx, pe.kubeClient, podKey)
+	if err != nil {
+		return nil, err
+	}
```
martinmaly, line 298: I'd recommend making this the 'slow path' of the request, should we get a cache miss on "which gRPC client is handling this function?" If we don't find the client we can then spin up the pod and get its client.

There is also potential for 'thundering herd' problem - if multiple clients call the same (uncached) function we'll start spinning up multiple pods in parallel.

martinmaly: We can do much of this as follow-ups

martinmaly (approved):
Some comments, but many of them can be addressed in follow-ups.

porch/func/internal/podevaluator.go
mengqiy: > we'll need synchronization between the garbageCollector and the request handler goroutines

Yes, you are right!

Re. coordinated spin-up, there are 2 levels of coordination we may need to spin up pods:
- Coordination between goroutines In one function runner
- Coordination between function runners

To achieve the later coordination, we may want to use k8s's coordination API: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#lease-v1-coordination-k8s-io

porch/func/internal/podevaluator.go
```diff
@@ -52,11 +55,17 @@ type podEvaluator struct {
...
+	podTTL         time.Duration
+
+	imageEntrypoint    map[string][]string
+	grpcConnectionPool map[string]*grpc.ClientConn
 }
 
 var _ Evaluator = &podEvaluator{}
 
-func NewPodEvaluator(namespace, wrapperServerImage string) (Evaluator, error) {
+func NewPodEvaluator(namespace, wrapperServerImage string, interval, ttl int) (Evaluator, error) {
```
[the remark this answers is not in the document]
mengqiy: Done

porch/func/server/server.go
```diff
@@ -37,6 +37,8 @@ var (
 	functions          = flag.String("functions", "./functions", "Path to cached functions.")
 	config             = flag.String("config", "./config.yaml", "Path to the config file.")
 	podNamespace       = flag.String("pod-namespace", "porch-fn-system", "Namespace to run KRM functions pods.")
+	podTTL             = flag.Int("pod-ttl-seconds", 600, "TTL for pods before GC.")
```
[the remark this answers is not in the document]
mengqiy: Done

porch/func/internal/podevaluator.go
mengqiy: Made patching asynchronous

porch/func/internal/podevaluator.go
mengqiy: It should have be resolved now. PTAL

porch/func/internal/podevaluator.go
```diff
@@ -64,40 +68,312 @@ func NewPodEvaluator(namespace, wrapperServerImage string) (Evaluator, error) {
...
 	} else {
-		fnCmd = cfg.Cmd
+		entrypoint = cfg.Cmd
+	}
+	pm.entrypointCache.Store(image, entrypoint)
+	return entrypoint, nil
+}
+
+func (pm *podManager) retrieveOrCreatePod(ctx context.Context, image string) (client.ObjectKey, error) {
+	// Try to retrieve the pod first.
```
martinmaly, line 345: the language referring to ordering (here in few other places) seems out-of-date since (I assume) refactoring created independent functions.

Maybe `// Try to retrieve the pod.`

porch/func/internal/podevaluator.go
```diff
@@ -163,59 +446,48 @@ func (pe *podEvaluator) EvaluateFunction(ctx context.Context, req *evaluator.Eva
...
+	return client.ObjectKeyFromObject(pod), nil
+}
 
+// podIpIfRunningAndReady waits for the pod to be running and ready and returns the pod IP and a potential error.
+func (pm *podManager) podIpIfRunningAndReady(ctx context.Context, podKey client.ObjectKey) (ip string, e error) {
+	var pod corev1.Pod
 	// Wait until the pod is Running
-	if err = wait.PollImmediate(100*time.Millisecond, 60*time.Second, func() (done bool, err error) {
-		err = pe.kubeClient.Get(ctx, client.ObjectKeyFromObject(pod), pod)
+	if e := wait.PollImmediate(100*time.Millisecond, 60*time.Second, func() (done bool, err error) {
```
martinmaly, line 461: oh, I need to start using wait.PollImmediate. Nice!

martinmaly (approved):
Thank you! This is super cool

Feel free to handle the feedback in a follow-up PR(s); I'll leave it up to you which comments you address before merging and which after in follow-ups.

mengqiy:
I will address the comments in a follow-up PR.

PR #2999, opened 2022-04-14
Store tasks in git

justinsb:
We put messages into commits, similar to what we do with OCI.

We could also move this to notes in future.

### DIFF
--- a/pkg/debug/formatter.go
+++ b/pkg/debug/formatter.go
@@ -1,0 +1,40 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package debug
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// deferredJSON is a helper that delays JSON formatting until/unless it is needed.
+type deferredJSON struct {
+	o interface{}
+}
+
+// String is the method that is called to format the object.
+func (d deferredJSON) String() string {
+	b, err := json.Marshal(d.o)
+	if err != nil {
+		return fmt.Sprintf("<error: %v>", err)
+	}
+	return string(b)
+}
+
+// JSON is a helper that prints the object in JSON format.
+// We use lazy-evaluation to avoid calling json.Marshal unless it is actually needed.
+func JSON(o interface{}) deferredJSON {
+	return deferredJSON{o}
+}
--- a/porch/engine/pkg/engine/engine.go
+++ b/porch/engine/pkg/engine/engine.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	"github.com/GoogleContainerTools/kpt/internal/fnruntime"
+	"github.com/GoogleContainerTools/kpt/pkg/debug"
 	"github.com/GoogleContainerTools/kpt/pkg/fn"
 	api "github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
 	configapi "github.com/GoogleContainerTools/kpt/porch/api/porchconfig/v1alpha1"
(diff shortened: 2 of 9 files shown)


porch/engine/pkg/engine/render.go
```diff
@@ -65,7 +65,7 @@ func (m *renderPackageMutation) Apply(ctx context.Context, resources repository.
 		Type: "eval",
 		Eval: &api.FunctionEvalTaskSpec{
 			Image:     "render",
-			ConfigMap: map[string]string{},
+			ConfigMap: nil,
```
justinsb, line 68: This is subtle ... Turns out `map[string]string{}` != `nil`, at least as far as reflect.DeepEqual is concerned.

porch/repository/pkg/git/draft.go
```diff
@@ -51,7 +53,21 @@ func (d *gitPackageDraft) UpdateResources(ctx context.Context, new *v1alpha1.Pac
 		ch.storeFile(path.Join(d.path, k), v)
 	}
 
-	message := fmt.Sprintf("Intermittent commit: %s", change.Type)
+	var message string
+	if change != nil {
+		message = fmt.Sprintf("Intermittent commit: %s", change.Type)
```
justinsb, line 58: Self-reviewing, and wondering if this should be "Intermediate commit..."

porch/engine/pkg/engine/engine.go
```diff
@@ -245,7 +246,7 @@ func (cad *cadEngine) UpdatePackageRevision(ctx context.Context, repositoryObj *
 			mutations = append(mutations, mutation)
 
 		default:
-			return nil, fmt.Errorf("updating task of type %q not supported", newTask.Type)
+			return nil, fmt.Errorf("updating task of type %q not supported (oldTask=%s, newTask=%s)", newTask.Type, debug.JSON(oldTask), debug.JSON(newTask))
```
justinsb, line 249: I added this logic mostly because I couldn't understand what was going on... turned out to be the ConfigMap nil vs empty problem (immediately below)

porch/repository/pkg/git/draft.go
martinmaly: Yes! sorry ...

porch/engine/pkg/engine/render.go
martinmaly: Not surprising, but yes, subtle.

porch/repository/pkg/git/git.go
```diff
@@ -754,3 +779,50 @@ func (r *gitRepository) pushAndCleanup(ctx context.Context, ph *pushRefSpecBuild
 	}
 	return nil
 }
+
+func (r *gitRepository) loadTasks(ctx context.Context, startCommit *object.Commit) ([]v1alpha1.Task, error) {
```
martinmaly, line 783: Is the idea that even published packages have tasks? Currently when publishing package, history of the intermediate commits is thrown away. We only maintain it for draft authoring; on approval it is collapsed away.

go-git has some commit-tree walking iterators too that can be helpful, I _think_ even including iterators that avoid specified ancestor paths...

porch/repository/pkg/git/git.go
```diff
@@ -754,3 +779,50 @@ func (r *gitRepository) pushAndCleanup(ctx context.Context, ph *pushRefSpecBuild
...
+		for _, line := range strings.Split(current.Message, "\n") {
+			line = strings.TrimSpace(line)
+			if strings.HasPrefix(line, "kpt:") {
+				task := v1alpha1.Task{}
+				b := []byte(strings.TrimPrefix(line, "kpt:"))
+				if err := json.Unmarshal(b, &task); err != nil {
+					klog.Warningf("failed to unmarshal task command %q: %w", line, err)
+					continue
+				}
+				tasks = append(tasks, task)
```
martinmaly, line 813: if we walk all commits, we will need to filter out commits that belong to different package, to associate tasks with the correct package

... unless we redesign the repository shape and each package gets its own branch. I was hoping to not go there because such repository structure is not what a person using Git to manage packages would create... but maybe they would and we should switch.

pkg/debug/formatter.go
```diff
@@ -0,0 +1,26 @@
+package debug
```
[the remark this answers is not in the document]
justinsb: Added!

porch/repository/pkg/git/git.go
justinsb: I think it's a good question re published packages.  If we think the information is valuable during authoring and approval (which I do), I think we should keep the information.  e.g. Imagine this scenario: package update goes through, deployed instances are updated, one of them breaks, we want to understand why: how was the bad yaml introduced, was it adequately presented in review, how can we prevent this happening again?  We could archive the information if we wanted to squash, but I'm not sure what it buys us really!  Also I think when a deployed instance gets a request to update, it will want to know what the upstream changes are.  My view is that the tasks are an important way to make sense of what has changed, adding useful insight over the raw YAML changes.

I propose adding a TODO re published packages, we can see how it feels in actual usage.

I will take a look at the iterators - it's a great point.

martinmaly: TODO for published packages is a good idea. I mostly want to make sure the code is self-consistent (i.e. we don't have some parts of the code assume that the history for published packages exists while it in actuality doesn't)

martinmaly (approved):

PR #3000, opened 2022-04-14
Special-case render so it doesn't appear as a task

justinsb:
This avoids problems when we are approving packages with tasks.

### DIFF
--- a/porch/repository/pkg/git/draft.go
+++ b/porch/repository/pkg/git/draft.go
@@ -41,7 +41,7 @@ type gitPackageDraft struct {
 
 var _ repository.PackageDraft = &gitPackageDraft{}
 
-func (d *gitPackageDraft) UpdateResources(ctx context.Context, new *v1alpha1.PackageRevisionResources, change *v1alpha1.Task) error {
+func (d *gitPackageDraft) UpdateResources(ctx context.Context, new *v1alpha1.PackageRevisionResources, task *v1alpha1.Task) error {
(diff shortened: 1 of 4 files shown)


porch/repository/pkg/git/draft.go
```diff
@@ -51,7 +51,13 @@ func (d *gitPackageDraft) UpdateResources(ctx context.Context, new *v1alpha1.Pac
...
 	}
 
-	message := fmt.Sprintf("Intermittent commit: %s", change.Type)
+	var message string
+	if task != nil {
+		message = fmt.Sprintf("Intermittent commit: %s", task.Type)
+	} else {
+		// TODO: Safe to assume it's always a render?
+		message = "Internal commit (render)"
+	}
```
martinmaly, line 60: why not instead return more general type from the Apply function; it can either carry information to add a task, or internal message for non-task things.

btw, the non-task changes seem to break the model - what should constitute a task and what shouldn't?

justinsb: Good idea - I'll give that a go.

If we're always going to run render, I don't know that we should show it.  (What happens if a user deletes it - do we just add it back?)  I think in general we should show everything, but specifically here for a task we automatically add it makes the synchronization between client state and server state more complicated - the value I get on the server after an apply has an extra task.  There's plenty of precedent for that in kubernetes (e.g. sidecar injection), but it does make the objects harder to work with (e.g. I think it breaks `kubectl diff`)

PR #3001, opened 2022-04-14
Create patch tasks when we edit the resources

justinsb:
Define real spec for patch tasks, generate patch tasks when making resource changes


porch/engine/pkg/engine/patchgen.go
```diff
@@ -0,0 +1,135 @@
...
+		switch patchSpec.PatchType {
+		case api.PatchTypeCreateFile:
+			if _, found := result.Contents[patchSpec.File]; found {
+				// TODO: We should be able to tolerate this.  Either do a merge or create as a different filename "-2"
+				return result, nil, fmt.Errorf("patch wants to create file %q but already exists", patchSpec.File)
+			}
+			result.Contents[patchSpec.File] = patchSpec.Contents
+		case api.PatchTypeDeleteFile:
+			if _, found := result.Contents[patchSpec.File]; !found {
+				// TODO: I don't think this should be an error, but maybe file manipulation should be replaced with objects
```
martinmaly, line 72: I think it would make sense to do the patch at the level of resources rather than files

martinmaly (approved):

porch/engine/pkg/engine/patchgen.go
justinsb: I agree ... file patches are more universal (they support comment changes etc), but if we can do an object patch that's much better.  I wanted to start simple though.  I'll add a TODO!

PR #3002, opened 2022-04-14
fix: Fix status reporting for ConfigConnectorContext resources

mortent:
Update the status rules for Config Connector resources to handle ConfigConnectorContext correctly. Unlike the other ConfigConnector resources, it does not have the `status.observedGeneration` field or conditions.


martinmaly (approved):
LGTM. my only surprise is that the status.healthy (boolean, or rather tri-state if we count absence as a valid state) is the main indicator, but I suppose it is sufficient.

mortent:
@martinmaly For most other ConfigConnector resources we do rely on conditions to determine reconcile status, but this one is an exception.

PR #3003, opened 2022-04-14
feat: Clean up package resources in kpt alpha sync del command

mortent:
This updates the `kpt alpha sync del` command to remove the resources belonging to a package when the RootSync resource is removed from the cluster. It does this by first updating the repo in the RootSync to an empty repo which will cause Config Sync to prune all resources. Once this is completed (we know by setting up a watch and waiting for the RootSync resource to reconcile), we delete the RootSync resource and optionally the secret.

### DIFF
--- a/internal/cmdsync/delete/command.go
+++ b/internal/cmdsync/delete/command.go
@@ -17,6 +17,7 @@ package delete
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/GoogleContainerTools/kpt/internal/errors"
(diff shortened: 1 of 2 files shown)


martinmaly (approved):

sdowell (approved):

mortent:
@martinmaly @sdowell Updated the code here to first wait for the RootSync reconciliation status (meaning Config Sync has made the call to delete all the resources) and then look at the ResourceGroup resource to make sure that the resources are actually removed from the cluster and not blocked by finalizers.

PR #3004, opened 2022-04-14
Use Create in repo reg command

justinsb:
We were using apply previously, which meant we weren't preventing
accidentally overwriting an existing repo.


justinsb:
I came _this close_ to parsing the protobuf format in our mock :-)

martinmaly (approved):

PR #3006, opened 2022-04-14
Make tests name-independent

martinmaly:
Tests no longer rely on specific package names generated by the server to
allow the server to change the name creation algorithm.


porch/apiserver/pkg/registry/porch/packagerevision.go
```diff
@@ -98,23 +98,17 @@ func (r *packageRevisions) Create(ctx context.Context, runtimeObject runtime.Obj
...
 	}
 
-	if name != expectedName {
-		return nil, apierrors.NewBadRequest(fmt.Sprintf("name should be %q", expectedName))
-	}
-
-	nameTokens, err := ParseName(name)
-	if err != nil {
-		return nil, apierrors.NewBadRequest(fmt.Sprintf("invalid name %q", name))
+	repositoryName := obj.Spec.RepositoryName
```
justinsb, line 105: Should we check that spec.repositoryName matches the value encoded name?  Or is the plan to just move away that restriction entirely?

porch/repository/pkg/git/git_test.go
```diff
@@ -521,29 +526,33 @@ func TestListPackagesDrafts(t *testing.T) {
...
 	}
 
-	got := map[string]v1alpha1.PackageRevisionLifecycle{}
+	got := map[repository.PackageRevisionName]v1alpha1.PackageRevisionLifecycle{}
 	for _, r := range revisions {
 		rev, err := r.GetPackageRevision()
 		if err != nil {
 			t.Errorf("GetPackageRevision failed for %q: %v", r.Name(), err)
 		}
-		got[r.Name()] = rev.Spec.Lifecycle
```
justinsb, line 546: If we do this a lot, we could define a method `Key()` which returns 

```
repository.PackageRevisionName{
			Repository: rev.Spec.RepositoryName,
			Package:    rev.Spec.PackageName,
			Revision:   rev.Spec.Revision,
		}
```

porch/repository/pkg/git/git_test.go
```diff
@@ -647,7 +660,11 @@ func TestDeletePackages(t *testing.T) {
 	for len(all) > 0 {
 		// Delete one of the packages
 		deleting := all[0]
-		name := deleting.Name()
+		pr, err := deleting.GetPackageRevision()
+		if err != nil {
+			t.Fatalf("GetPackageRevision of %q faile: %v", deleting.Name(), err)
```
justinsb, line 665: Supernit: failed

porch/repository/pkg/git/testing_repo.go
```diff
@@ -183,37 +183,45 @@ func findFile(t *testing.T, tree *object.Tree, path string) *object.File {
...
 
-func findPackage(t *testing.T, revisions []repository.PackageRevision, name string) repository.PackageRevision {
+func findPackage(t *testing.T, revisions []repository.PackageRevision, name repo.PackageRevisionName) repository.PackageRevision {
 	for _, r := range revisions {
-		if r.Name() == name {
+		rev, err := r.GetPackageRevision()
+		if err != nil {
+			t.Fatalf("GetPackageRevision of %q failed: %v", r.Name(), err)
+		}
+		if rev.Spec.RepositoryName == name.Repository &&
```
justinsb, line 192: An advantage of `Key()` could be that I think we could then just do a `==` check against the two structs.

porch/apiserver/pkg/registry/porch/packagerevision.go
justinsb: For Create, I think we expect it normally to be called with a Name, and I don't think we should default it if not set - I think that would be surprising and would likely break validation.  The functionality is available in GenerateName if we do want it, though it isn't used very often AFAIK

porch/apiserver/pkg/e2e/suite.go
```diff
@@ -637,3 +638,16 @@ func normalizeYamlOrdering(t *testing.T, contents string) string {
 	}
 	return stable.String()
 }
+
+func FindPackageRevision(t *testing.T, packages *porchapi.PackageRevisionList, name repository.PackageRevisionName) *porchapi.PackageRevision {
```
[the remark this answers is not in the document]
martinmaly: done

porch/repository/pkg/git/git_test.go
martinmaly: thanks!

justinsb (approved):

porch/repository/pkg/git/testing_repo.go
martinmaly: done

porch/apiserver/pkg/registry/porch/packagerevision.go
martinmaly: will add TODO with the 2nd PR

PR #3007, opened 2022-04-15
Use hash for package names

martinmaly:
Kubernetes resources have constrained naming convention and we cannot
therefore reflect package path in the name correctly. Therefore we compute
a hash for package identity and convey the package path, repository, and revision
in the `PackageRevision.Spec`


justinsb (approved):
We should have a TODO to figure out how we're going to store the name going forwards, but LGTM

PR #3008, opened 2022-04-15
Stricter error handling in git package discovery

justinsb:
We find nested packages, we also return errors more strictly.


martinmaly (approved):
Thank you!

PR #3010, opened 2022-04-15
Add ability to read file to commit helper

justinsb:
We can use this e.g. for manipulating the Kptfile or other metadata files.


martinmaly (approved):
Nice!

PR #3012, opened 2022-04-15
git CommitHelper: a few small fixes

justinsb:
Rationalizing some of the codepaths and adding some comments per my
understanding.

### DIFF
--- a/porch/repository/pkg/git/commit.go
+++ b/porch/repository/pkg/git/commit.go
@@ -29,7 +29,6 @@ import (
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/filemode"
 	"github.com/go-git/go-git/v5/plumbing/object"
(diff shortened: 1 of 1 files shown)


porch/repository/pkg/git/commit.go
```diff
@@ -274,14 +287,19 @@ func storeBlob(store storer.EncodedObjectStorer, value string) (plumbing.Hash, e
 		return plumbing.Hash{}, err
 	}
 
-	_, err = w.Write(data)
-	w.Close()
-	if err != nil {
+	if _, err := w.Write(data); err != nil {
 		return plumbing.Hash{}, err
```
martinmaly, line 291: I _think_ this return will skip close on line 294

martinmaly (approved):
one issue worth addressing, otherwise LGTM
Thanks for doing this!

porch/repository/pkg/git/commit.go
justinsb: Good call - thank you!

PR #3015, opened 2022-04-18
Detect Latest Package Revision

martinmaly:
The latest package revision (in semver order) is marked by a k8s resource tag
to make identifying it easier for clients.


porch/repository/pkg/cache/repository.go
```diff
@@ -213,3 +246,60 @@ func (r *cachedRepository) pollOnce(ctx context.Context) {
 		klog.Warningf("error polling repo functions %s: %v", r.id, err)
 	}
 }
+
+func ToCachedPackageRevisionSlice(revisions []repository.PackageRevision) []*cachedPackageRevision {
```
justinsb, line 250: Nit: Do we need to export this?

porch/repository/pkg/cache/repository.go
```diff
@@ -213,3 +246,60 @@ func (r *cachedRepository) pollOnce(ctx context.Context) {
...
+			PackageRevision:  revisions[i],
+			isLatestRevision: false,
+		}
+		result[i] = current
+	}
+	IdentifyLatestRevisions(result)
+	return result
+}
+
+func IdentifyLatestRevisions(result []*cachedPackageRevision) {
```
justinsb, line 263: Nit: Do we need to export this one also?

porch/repository/pkg/cache/repository.go
```diff
@@ -213,3 +246,60 @@ func (r *cachedRepository) pollOnce(ctx context.Context) {
...
+	// Compute the latest among the different revisions of the same package.
+	// The map is keyed by the package name; Values are the latest revision found so far.
+	latest := map[string]*cachedPackageRevision{}
+	for _, current := range result {
+		current.isLatestRevision = false // Clear all values
+
+		// Check if the current package revision is more recent than the one sen so far.
```
justinsb, line 270: Typo: s/sen/seen

porch/repository/pkg/cache/repository.go
```diff
@@ -213,3 +246,60 @@ func (r *cachedRepository) pollOnce(ctx context.Context) {
...
+		if current.Lifecycle() != v1alpha1.PackageRevisionLifecyclePublished {
+			continue
+		}
+
+		currentKey := current.Key()
+		if previous, ok := latest[currentKey.Package]; ok {
+			previousKey := previous.Key()
+			switch cmp := semver.Compare(currentKey.Revision, previousKey.Revision); {
+			case cmp == 0:
+				// Same revision.
```
justinsb, line 281: Perhaps warn?  Do we expect this?

justinsb (approved):
A few nits, but lgtm

porch/repository/pkg/cache/repository.go
```diff
@@ -138,7 +165,9 @@ func (r *cachedRepository) CreatePackageRevision(ctx context.Context, obj *v1alp
 }
 
 func (r *cachedRepository) UpdatePackage(ctx context.Context, old repository.PackageRevision) (repository.PackageDraft, error) {
-	created, err := r.repo.UpdatePackage(ctx, old)
+	// Unwrap
+	unwrapped := old.(*cachedPackageRevision).PackageRevision
```
mortent, line 169: Why do we need to unwrap here? It seems like the `UpdatePackage` accepts a parameter of type `PackageRevision` which the `*cachedPackageRevision` implements.

mortent (approved):
Looks good to me. I just had one question.

porch/repository/pkg/cache/repository.go
martinmaly: no; I'll hide it. good catch!

porch/repository/pkg/cache/repository.go
martinmaly: ditto. will hide

porch/repository/pkg/cache/repository.go
martinmaly: thanks!

porch/repository/pkg/cache/repository.go
martinmaly: Warning seems like a good idea.

porch/repository/pkg/cache/repository.go
martinmaly: the Git repository will cast the interface to *gitPackageRevision.

We could expand the interface and avoid the cast but Git is getting at some fields not exposed via the interface.

porch/repository/pkg/cache/repository.go
```diff
@@ -86,14 +113,14 @@ func (r *cachedRepository) getPackages(ctx context.Context, forceRefresh bool) (
...
+		packages = ToCachedPackageRevisionSlice(p)
 
 		r.mutex.Lock()
-		r.cachedPackages = p
+		r.cachedPackages = packages
 		r.mutex.Unlock()
 	}
 
-	return packages, nil
+	return ToPackageRevisionSlice(packages), nil
```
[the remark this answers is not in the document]
martinmaly: Maybe the names are not ideal. The cache now stores `cachedPackages []*cachedPackageRevision` so on the 'way out' it needs to be converted to `[]PackageRevision`. So this call really just "changes the type of the slice".
When we store the packages, we wrap then and compute the latest one; then we return them, still wrapped, but re-typed to []PackageRevision.

We _could_ store the cache as []PackageRevision and perform casts; I thought this was more explicit given all the other casting that is happening.

PR #3017, opened 2022-04-18
Respect Repository Registration Directory

martinmaly:
If a sub-directory is specified at repository registration, the repository will
only include package revisions in the specified directory. Likewise, it will
only create packages in the specified subdirectory.

In order to maintain stable resource names within the repo (regardless of the
registration parameters), the package names are unaffected by the directory
projection.

### DIFF
--- a/porch/repository/pkg/git/dir.go
+++ b/porch/repository/pkg/git/dir.go
@@ -1,0 +1,33 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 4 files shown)


porch/repository/pkg/git/dir.go
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package git
+
+import "strings"
+
+// Determines whether a package specified by a path is in a directory given.
+func packageInDirectory(pkg, dir string) bool {
```
mortent, line 20: If the inputs here are always file paths, could we use the `path/filepath` package in go instead of relying on string functions? It seems like maybe the [Rel](https://pkg.go.dev/path/filepath#Rel) function could be used.

martinmaly: Good question. I'd say that these paths are closer to what the `path` package deals with than `filepath`. I am not 100% sure whether Git on Windows switches polarity of slashes (for refs etc) but the internals do seem to use `/` (example: https://github.com/go-git/go-git/blob/master/plumbing/object/tree.go#L133). So the `filepath` package is n/a.
That's why I opted for implementing our own (`path` package, sadly, doesn't have anything reusable either).

mortent: Agree, the `filepath` package is not appropriate here.

mortent (approved):

PR #3020, opened 2022-04-18
Update Repository Resource Status

martinmaly:
When repository interactions encounter errors, we update status
of the repository with a failure condition. Likewise, on successful
resync of the repository we set a success condition.

### DIFF
--- a/porch/api/porchconfig/v1alpha1/types.go
+++ b/porch/api/porchconfig/v1alpha1/types.go
@@ -21,7 +21,7 @@ import (
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:resource:path=repositories,singular=repository
-//+kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=='Available')].status`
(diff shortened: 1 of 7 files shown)


porch/api/porchconfig/v1alpha1/types.go
```diff
@@ -142,6 +144,16 @@ type FunctionRef struct {
 	Name string `json:"name"`
 }
 
+const (
+	// Type of the Repository condition.
+	RepositoryAvailable = "Available"
```
mortent, line 149: Does this condition reflect the reconciliation status of the library (or somewhat equivalent to that?)

If so, I think we should try to adopt one of the two patterns supported by the [kstatus library](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus#conditions). This means that our other tools will automatically be able to understand the conditions.

martinmaly: Oh, nice. I was looking for some prior art and the k8s meta only mentioned "Available" as standard.

I would recommend we go with the `Ready` clause: https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus#the-ready-condition

which would be achieved simply by renaming using `Ready` instead of `Available`.

mortent (approved):

Issue #3014, opened 2022-04-17
User friendly `--type` flag behavior


yuwenma:
### Current behavior 
`--type` flag is defined in `kpt fn eval`, default to "mutator". It has autocompletion to suggest "mutator" and "validator". It has two functionalities
1. Used with `--save` flag to store the function to the corresponding Kptfile `pipeline.mutators` or `pipeline.validators` list
2. Used with `--keywords` or `--image` autocompletion to filter suggested functions by type.
  

### Problems
- Problem 1. 
`kpt fn eval` already has too many flags that sort of correlate to each other. `--keywords`, `--type`, `--exec`, `--image`. The usage of `--type` is determined and determines other behavior. I don't want to give users too much burden to understand all of these.  it may be even not a good idea to use --type flag.

- Problem 2. 
`--type` only really matters when used with "--save". At which point "mutator" is a valid default. But when used as a filter to functions, I'm not quite confident this default value is the right filter. 

### Proposal
-  `--type` is default to "" (rather than "mutator") with autocompletion enabled to suggest "validator" and "mutator".
- Matching functions by type accepts the "--type" to be empty string "", in such case suggest all functions. 
- Only when "--save" is true, will kpt validate "--type". It accepts value `["", "mutator", "validator"]`, and reassign `""` to `"mutator"`.

Pros:
The advantage is that users are guided to only care the `--type` when they want to write function to Kptfile `pipeline.validators`. All other cases the `--type` is a nice to have thing.  
Cons:
these assumptions might confuse users.

yuwenma:
This is from the previous PR discussion https://github.com/GoogleContainerTools/kpt/pull/2956#discussion_r851551772
CC @mengqiy @droot

mengqiy:
>--type is default to "" (rather than "mutator") with autocompletion enabled to suggest "validator" and "mutator".
Matching functions by type accepts the "--type" to be empty string "", in such case suggest all functions.

These 2 make sense.

>Only when "--save" is true, will kpt validate "--type". It accepts value ["", "mutator", "validator"], and reassign "" to "mutator".

IMO when `--save` is specified, we should require users to set `--type` explicitly. Defaulting it to `mutator` when it's empty might be a problematic defaulting.

yuwenma:
> > --type is default to "" (rather than "mutator") with autocompletion enabled to suggest "validator" and "mutator".
> > Matching functions by type accepts the "--type" to be empty string "", in such case suggest all functions.
> 
[...]

PR #3021, opened 2022-04-19
user friendly --type behavior

yuwenma:
Related to #3014

- --type is default to "" (rather than "mutator") with autocompletion enabled to suggest "validator" and "mutator".
- Matching functions by type accepts the "--type" to be empty string "", in such case suggest all functions.
- Only when "--save" is true, will require and validate "--type". 

Output 1: type required if save
```bash
kpt fn eval -s -i set-namespace:v0.3.3 
error: --type is required if saving functions to Kptfile (--save=true)
```
Outout 2: invalid type
```bash
kpt fn eval -s -t bad -i set-namespace:v0.3.3
error: --type can only be `mutator` or `validator`
```
Output 3: Autocomplete all
```bash
kpt fn eval -k
annotation   gcp          helm         label        name-suffix  replacement  terraform    
config-sync  generate     image        name         namespace    setters      viewer       
gatekeeper   generator    kubeval      name-prefix  project-id   starlark 
```
Output 4: Automplete filter by validator
```bash
kpt fn eval -t validator -k
gatekeeper  kubeval     starlark
```

### DIFF
--- a/thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
+++ b/thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
@@ -50,11 +50,11 @@ func GetEvalFnRunner(ctx context.Context, parent string) *EvalFnRunner {
 		return cmdutil.SuggestFunctions(cmd), cobra.ShellCompDirectiveDefault
 	})
 	r.Command.Flags().StringArrayVarP(
-		&r.Keywords, "keywords", "k", []string{}, "filter functions that match one or more keywords")
+		&r.Keywords, "keywords", "k", nil, "filter functions that match one or more keywords")
 	_ = r.Command.RegisterFlagCompletionFunc("keywords", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return cmdutil.SuggestKeywords(cmd), cobra.ShellCompDirectiveDefault
 	})
-	r.Command.Flags().StringVarP(&r.FnType, "type", "t", "mutator",
+	r.Command.Flags().StringVarP(&r.FnType, "type", "t", "",
 		"`mutator` (default) or `validator`. tell the function type for autocompletion and `--save` flag")
 	_ = r.Command.RegisterFlagCompletionFunc("type", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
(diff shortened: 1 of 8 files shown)


thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -361,19 +361,46 @@ func checkFnConfigPathExistence(path string) error {
...
+func (r *EvalFnRunner) validateOptionalFlags() error {
 	// Let users know that --include-meta-resources is no longer necessary
 	// since meta resources are included by default.
 	if r.IncludeMetaResources {
 		return fmt.Errorf("--include-meta-resources is no longer necessary because meta resources are now included by default")
 	}
+	// SaveFn stores function to Kptfile. If not enabled, only make in-place changes.
+	if r.SaveFn {
+		if r.FnType == "" {
+			return fmt.Errorf("--type is required if saving functions to Kptfile (--save=true)")
```
droot, line 373: nit: try to be consistent with the eror style guide https://github.com/GoogleContainerTools/kpt/blob/main/docs/style-guides/errors.md

An alternate language:

function type (validator or mutator) must be specified using `--type`

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -361,19 +361,46 @@ func checkFnConfigPathExistence(path string) error {
...
+		}
+		if r.FnType != "mutator" && r.FnType != "validator" {
+			return fmt.Errorf("--type can only be `mutator` or `validator`")
```
droot, line 376: Suggestion for language:

`--type` must be either `mutator` or `validator`.

droot (changes requested):
minor language suggestions.

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
yuwenma: Thanks!

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
yuwenma: Wow, didn't know we even have an error guide. Updated

droot (approved):

PR #3024, opened 2022-04-19
Require Registered Branch To Exist

martinmaly:
When working with empty repository, Porch may push the first branch to
the remote. GitHub makes the first branch the default which cannot be
easily deleted.

For more predictable user experience, require the branch the client
provided at registration (defaulting to `main`) to exist.

Improve the test repository archive command to set consistent file
timestamps and delete more unnecessary content. Add documentation for all test
repositories.


justinsb (approved):
lgtm

PR #3026, opened 2022-04-19
update kustomize dependencies

natasha41575:
Pick up the performance boost from https://github.com/kubernetes-sigs/kustomize/pull/4568


natasha41575:
I found https://github.com/kubernetes/client-go/issues/1075, which describes the issue that we are getting here.

The problem is that [github.com/googleapis/gnostic](http://github.com/googleapis/gnostic) was renamed to [github.com/google/gnostic](http://github.com/google/gnostic)
And different dependencies are pulling in different names, so we get a bunch of errors because Go thinks that types from [github.com/google/gnostic](http://github.com/google/gnostic) aren't the same as [github.com/googleapis/gnostic](http://github.com/googleapis/gnostic)

We had to make this name change in kustomize due to integration in kubectl https://github.com/kubernetes-sigs/kustomize/pull/4538/

Ideally we would be able to upgrade all the dependencies to make this work, but it appears that we still have some (like `k8s.io/cli-runtime`) that still depend on the old name.

I'm not sure if there is any solution for this besides a replace statement.

mengqiy:
We need https://github.com/kubernetes-sigs/cli-utils/pull/577

PR #3029, opened 2022-04-20
Implement fieldSelectors on PackageRevision

justinsb:
This lets us query objects by their "key fields", such as package /
revision / repository.

This is more important now that the name is no longer restricted.

### DIFF
--- a/porch/apiserver/pkg/registry/porch/fieldselector.go
+++ b/porch/apiserver/pkg/registry/porch/fieldselector.go
@@ -1,0 +1,103 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package porch
+
+import (
+	"fmt"
+
+	"github.com/GoogleContainerTools/kpt/porch/repository/pkg/repository"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/selection"
+)
+
+// convertPackageRevisionFieldSelector is the schema conversion function for normalizing the the FieldSelector for PackageRevision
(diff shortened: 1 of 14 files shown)


porch/apiserver/pkg/registry/porch/fieldselector.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+
+	return filter, nil
+}
+
+// parsePackageRevisionResourcesFieldSelector parses client-provided fields.Selector into a packageFilter
+func parsePackageRevisionResourcesFieldSelector(fieldSelector fields.Selector) (packageFilter, error) {
+	// TOOD: This is a little weird, because we don't have the same fields on PackageRevisionResources.
+	// But we probably should have the key fields
```
martinmaly, line 101: I was just noticing that; we may want to duplicate at least repo, name, revision

porch/repository/pkg/cache/repository.go
```diff
@@ -312,10 +313,12 @@ func identifyLatestRevisions(result []*cachedPackageRevision) {
...
-func toPackageRevisionSlice(cached []*cachedPackageRevision) []repository.PackageRevision {
-	result := make([]repository.PackageRevision, len(cached))
-	for i := range cached {
-		result[i] = cached[i]
+func toPackageRevisionSlice(cached []*cachedPackageRevision, filter repository.ListPackageRevisionFilter) []repository.PackageRevision {
+	result := make([]repository.PackageRevision, 0, len(cached))
+	for _, p := range cached {
+		if filter.Matches(p) {
+			result = append(result, p)
+		}
```
martinmaly, line 321: This worked out nicely!

porch/repository/pkg/cache/repository.go
```diff
@@ -119,8 +120,8 @@ func (r *cachedRepository) getPackages(ctx context.Context, forceRefresh bool) (
 
 	if packages == nil {
 		// TODO: Avoid simultaneous fetches?
-		var p []repository.PackageRevision
-		p, err = r.repo.ListPackageRevisions(ctx)
+		// TODO: Push-down partial refresh?
```
martinmaly, line 123: what is meant by "push-down"?

martinmaly (approved):

porch/apiserver/pkg/registry/porch/fieldselector.go
justinsb: Yes, I concur :-)  I can tackle that next!

porch/repository/pkg/cache/repository.go
justinsb: Ah, by push-down I mean pushing the predicates down into the lower levels, so that they can also optimize their fetching to just return the filtered results.  It's not obviously a win here, because if we push-down the filter, then we can't cache the results (well, not without _much_ more complex caching!)

PR #3031, opened 2022-04-20
Rationalize git package construction

justinsb:
We now have a single codepath for discovering git packages in the
tree; this allows us to include tasks in future.

### DIFF
--- a/porch/repository/pkg/git/package_tree.go
+++ b/porch/repository/pkg/git/package_tree.go
@@ -1,0 +1,170 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


porch/repository/pkg/git/package_tree.go
```diff
@@ -0,0 +1,162 @@
...
+	packages map[string]*packageListEntry
+}
+
+// packageListEntry is a single package found in a git repository
+type packageListEntry struct {
+	// parent is the packageList of which we are part
+	parent *packageList
+
+	// Path is the relative path to the root of the package (directory containing the Kptfile)
+	Path string
```
martinmaly, line 45: Is `Path` intentionally exported?

martinmaly (approved):

porch/repository/pkg/git/package_tree.go
justinsb: Ah ... it was used at one stage, but you're right, it's no longer used.  I will unexport till we need it.

PR #3032, opened 2022-04-20
Add lots more tracing

justinsb:
Also better wire up opentelemetry


martinmaly (approved):
once tests pass

I can't wait to play with otel and see the traces!

PR #3033, opened 2022-04-20
Remove Unused Error Results

martinmaly:
GetPackageRevision had an error return value but it was never used.
Since it performs a straightforward operation, removing the error return
value to simplify the interface.


justinsb (approved):
Nice!  lgtm

PR #3034, opened 2022-04-20
Cleanup Function Errors

martinmaly:
Return apiserver errors from Function registry to return clearer errors to the
client, rather than Internal errors.

### DIFF
--- a/porch/apiserver/pkg/registry/porch/functions.go
+++ b/porch/apiserver/pkg/registry/porch/functions.go
@@ -16,13 +16,13 @@ package porch
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
 	configapi "github.com/GoogleContainerTools/kpt/porch/api/porchconfig/v1alpha1"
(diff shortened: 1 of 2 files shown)


porch/apiserver/pkg/registry/porch/functions.go
```diff
@@ -16,13 +16,13 @@ package porch
...
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
 	configapi "github.com/GoogleContainerTools/kpt/porch/api/porchconfig/v1alpha1"
 	"github.com/GoogleContainerTools/kpt/porch/engine/pkg/engine"
+	"k8s.io/apimachinery/pkg/api/errors"
```
justinsb, line 25: Nit: in OSS we tend to import this as `apierrors`, just to avoid the conflict with errors in stdlib

justinsb (approved):
lgtm!

porch/apiserver/pkg/registry/porch/functions.go
martinmaly: renamed. thank you!

PR #3035, opened 2022-04-21
Bump dependencies in kpt and porch

mengqiy:
The 1st commit vendor some packages from cli-utils to the `thirdparty` directory.
The 2nd commit update the dependencies for kpt and porch.


porch/go.mod
```diff
@@ -32,22 +32,22 @@ require (
...
-	k8s.io/api v0.23.5
-	k8s.io/apimachinery v0.23.5
-	k8s.io/apiserver v0.23.5
-	k8s.io/client-go v0.23.5
-	k8s.io/component-base v0.23.5
+	k8s.io/api v0.24.0-rc.0
+	k8s.io/apimachinery v0.24.0-rc.0
+	k8s.io/apiserver v0.24.0-rc.0
+	k8s.io/client-go v0.24.0-rc.0
+	k8s.io/component-base v0.24.0-rc.0
```
martinmaly, line 45: what drives the change to take dependency on RC build?

yuwenma: Not an expert on k8s lib release versions. But the RC release contains the right version which will resolve the dependency conflicts that blocks [bumping builtin functions in Porch](https://github.com/GoogleContainerTools/kpt/pull/3028), and blocks [bumping kyaml in kpt](https://github.com/GoogleContainerTools/kpt/pull/3026/files).  

The former is needed for the next fishfooding and the RC seems the most stable version we can use. @mengqiy correct me if I'm wrong. 

More context is here: https://github.com/kubernetes/client-go/issues/1075

mengqiy: We want to include a kyaml change (https://github.com/kubernetes-sigs/kustomize/pull/4568) which contains a performance improvement for both kpt and porch. This change requires https://github.com/kubernetes/kube-openapi/pull/283 which is merged after breaking change https://github.com/kubernetes/kube-openapi/pull/285.
The kube-openapi breaking change was imported to k8s after v0.24.0 (https://github.com/kubernetes/kubernetes/pull/108644). It means that NO 0.23.X k8s packages (including apimachinery and client-go) work with our desired kyaml change.

According to k8s [1.24 release schedule](https://github.com/kubernetes/sig-release/tree/master/releases/release-1.24), 0.24.0-rc.1 is expected on 04/26 and 0.24.0 is expected on 05/03.

pkg/live/inventoryrg.go
```diff
@@ -376,9 +376,7 @@ func (rgi *ResourceGroupInstaller) InstallRG(ctx context.Context) error {
 				return e.Error
 			case pollevent.ResourceUpdateEvent:
 				if e.Resource.Status == kstatus.CurrentStatus {
-					if !meta.MaybeResetRESTMapper(mapper) {
-						return fmt.Errorf("failed to reset RESTMapper: %T", mapper)
-					}
+					meta.MaybeResetRESTMapper(mapper)
```
droot, line 379: so this doesn't throw any errors anymore ?

mengqiy: It doesn't throw error any more.

droot (approved):

yuwenma:
SWEEEET!!!

PR #3038, opened 2022-04-21
wrapper-server: Log time taken to run functions

justinsb:


mengqiy:
I added something similar in function-runner: https://github.com/GoogleContainerTools/kpt/blob/c1e49b975c7de78413121eac6b567fe879098fe4/porch/func/internal/podevaluator.go#L98-L102

porch/func/wrapper-server/main.go
```diff
@@ -95,14 +96,16 @@ func (e *singleFunctionEvaluator) EvaluateFunction(ctx context.Context, req *pb.
...
+
 	err := cmd.Run()
 	var exitErr *exec.ExitError
 	if err != nil && !errors.As(err, &exitErr) {
 		return nil, status.Errorf(codes.Internal, "Failed to execute function %q: %s (%s)", req.Image, err, stderr.String())
 	}
 
 	outbytes := stdout.Bytes()
-	klog.Infof("Evaluated %q: stdout length: %d\nstderr:\n%v", req.Image, len(outbytes), stderr.String())
+	klog.Infof("Evaluated %q: duration=%v\nstdout length: %d\nstderr:\n%v", req.Image, time.Since(start).Round(time.Millisecond), len(outbytes), stderr.String())
```
mengqiy, line 108: This latency is not the user-facing latency. When we calculate the time here, pod is already running.
If the KRM function is not in the cache, we will need to create a pod.

mengqiy:
If your goal is to measure the latency of the wrapper-server, I'm OK merging this PR.

mortent:
Maybe overkill in this PR, but should we expose this through a proper metrics endpoint instead of relying on logs? We are already using OpenTelemetry for tracing.

liamfallon:
This PR is on Porch. Porch has been moved to the Nephio project.

In the Nephio project, the approach to running kpt pipelines is being discussed.

PR #3039, opened 2022-04-21
Custom table rendering of PackageRevision resources

martinmaly:


porch/apiserver/pkg/registry/porch/table.go
```diff
@@ -0,0 +1,150 @@
...
+			{Name: "Latest", Type: "boolean"},
+			{Name: "Lifecycle", Type: "string"},
+			{Name: "Repository", Type: "string"},
+		},
+	}
+)
+
+func isLatest(pr *api.PackageRevision) bool {
+	labels := pr.Labels
+	if labels == nil {
```
justinsb, line 117: Nit: You don't need to check for nil when _reading_ a slice or map; the nil value is treated as the empty slice/map.

(This is also true when appending to an nil slice.  It is sadly not true when writing to an nil map)

porch/apiserver/pkg/registry/porch/table.go
```diff
@@ -0,0 +1,150 @@
...
+		return false
+	}
+	return labels[api.LatestPackageRevisionKey] == api.LatestPackageRevisionValue
+}
+
+func notAcceptableError(ctx context.Context, resource schema.GroupResource) error {
```
justinsb, line 123: Nit (for future PR): move these into their own file (errors.go)?

justinsb (approved):

porch/apiserver/pkg/registry/porch/table.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	return &table, nil
+}
+
+var (
+	packageRevisionTableConvertor = tableConvertor{
+		resource: porch.Resource("packagerevisions"),
+		cells: func(obj runtime.Object) []interface{} {
+			pr, ok := obj.(*api.PackageRevision)
```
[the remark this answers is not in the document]
martinmaly: we don't ... the caller will interpret this as an error and return an error. I wanted the per-resource logic minimal so I hoisted the error handling out.

porch/apiserver/pkg/registry/porch/table.go
martinmaly: You are right! indeed, the asymmetry of map got me there.

porch/apiserver/pkg/registry/porch/table.go
martinmaly: did it now. thanks!

PR #3037, opened 2022-04-21
Introduce generator as option for base

justinsb:

PR #3041, opened 2022-04-21
WIP: Registry for function discovery

justinsb:
- Introduce generator as option for base
- Use a CRD registry to query function images
- Start example configuration for CRD -> function discovery

Builds on #3037


liamfallon:
The Function support was removed the Function functionality from Porch.

PR #3046, opened 2022-04-21
Fix set-namespace function version in tests

martinmaly:


yuwenma:
Thanks, the fix is in already.

PR #3051, opened 2022-04-22
kpt rpkg get uses Table list output if appropriate

martinmaly:
This enables better user experience with kpt CLI

### DIFF
--- a/e2e/porch_test.go
+++ b/e2e/porch_test.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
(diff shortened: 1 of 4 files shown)


e2e/porch_test.go
```diff
@@ -70,6 +71,7 @@ func runTestCase(t *testing.T, git string, tc porch.TestCaseConfig) {
 	}
 
 	for i := range tc.Commands {
+		time.Sleep(5 * time.Second)
```
martinmaly, line 74: k8s apiserver seems to be throttling and rejecting requests, hence this timeout.
It is a temporary solution but it reduces flakiness.

droot (approved):

PR #3052, opened 2022-04-22
Fix duplicate drafts in package revision list

martinmaly:
The code should append individual draft when it matches the filter, not
all of them.


justinsb (approved):
lgtm

PR #3054, opened 2022-04-22
Improve kpt CLI `alpha get` command output

martinmaly:
- Show more information about repository
- Use Table output for `kpt alpha repo get`
- Update kpt CLI tests

### DIFF
--- a/internal/cmdrepoget/command.go
+++ b/internal/cmdrepoget/command.go
@@ -16,16 +16,16 @@ package cmdrepoget
 
 import (
 	"context"
+	"strings"
 
 	"github.com/GoogleContainerTools/kpt/internal/errors"
 	"github.com/GoogleContainerTools/kpt/internal/util/porch"
-	configapi "github.com/GoogleContainerTools/kpt/porch/api/porchconfig/v1alpha1"
 	"github.com/spf13/cobra"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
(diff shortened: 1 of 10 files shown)


internal/cmdrepoget/command.go
```diff
@@ -67,62 +67,79 @@ func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner
...
 func (r *runner) runE(cmd *cobra.Command, args []string) error {
 	const op errors.Op = command + ".runE"
 
-	var objs []runtime.Object
+	// For some reason our use of k8s libraries result in error when decoding
+	// RepositoryList when we use strongly typed data. Therefore for now we
+	// use unstructured communication.
+	// The error is: `no kind "RepositoryList" is registered for the internal
+	// version of group "config.porch.kpt.dev" in scheme`. Of course there _is_
+	// no such kind since CRDs seem to have only versioned resources.
```
droot, line 96: Trying to remember from my `kubebuilder` days, we had to register list-kind also to the scheme. Here is an example from a kubebuilder same project: https://github.com/jetstack/kubebuilder-sample-controller/blob/master/api/v1beta1/mykind_types.go#L77

droot (approved):

internal/cmdrepoget/command.go
martinmaly: Thank you! the List types are registered in the scheme. Problem is that the k8s implementation is looking for _unversioned_ type but those don't exist for CRDs (as far as I can tell).

PR #3060, opened 2022-04-25
Support editing of packages

mortent:


porch/api/porch/v1alpha1/types_packagerevisions.go
```diff
@@ -102,6 +103,7 @@ type Task struct {
 	Init  *PackageInitTaskSpec  `json:"init,omitempty"`
 	Clone *PackageCloneTaskSpec `json:"clone,omitempty"`
 	Patch *PackagePatchTaskSpec `json:"patch,omitempty"`
+	Edit  *PackageEditTaskSpec  `json:"edit,omitempty"`
```
martinmaly, line 106: The test failure is likely because corresponding API change needs to be made to the unversioned API in the parent directory.

mortent: Yeah, that fixed the issue.

mortent:
I'll add test in a follow-up. I think there is some interesting questions around how we can best test this without having to stub out several interfaces.

PR #3061, opened 2022-04-26
Include Repo Address in Repository Table

martinmaly:


martinmaly:
This is a bit of an experiment; JsonPath makes it a bit hard to express this so the JsonPath expression seems suboptimal. Not attached to thisone, just an idea how to make the kpt repo get output better.

porch/api/porchconfig/v1alpha1/config.porch.kpt.dev_repositories.yaml
```diff
@@ -42,6 +42,9 @@ spec:
     - jsonPath: .status.conditions[?(@.type=='Ready')].status
       name: Ready
       type: string
+    - jsonPath: .spec['git','oci']['repo','registry']
```
martinmaly, line 45: We _could_ change the API and instead of `repo` and `registry` for Git and OCI respectively, use Address.

It would create inconsistency with Kptfile design though (https://github.com/GoogleContainerTools/kpt/blob/main/pkg/api/kptfile/v1/types.go#L132) so I would not recommend going that route.

PR #3063, opened 2022-04-27
Test for edit package

mortent:
This adds a unit test for the `editPackageMutation`. I'm looking to add some additional tests for error conditions, but want to verify the approach first.

This required adding stub implementations of several interfaces. I'm open to ideas about how we can make this easier.

### DIFF
--- a/porch/engine/pkg/engine/edit_test.go
+++ b/porch/engine/pkg/engine/edit_test.go
@@ -1,0 +1,133 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


martinmaly (approved):
This seems like an approach worth trying.
So far we have been testing mostly with real repositories (see git_test or e2e_test) but I think we need to try some lower-level unit tests and this seems like a good way to go about it.

I imagine that the fake repository can be extracted into a reusable test building block. Then we can see whether using (and maintaining it) is more work than using Git repository directly. We have pretty comprehensive utilities built around the git repositories which make it very easy to spin up, seed the repositories with initial state, and then perform tests.

porch/engine/pkg/engine/edit_test.go
```diff
@@ -0,0 +1,131 @@
...
+apiVersion: kpt.dev/v1
+kind: Kptfile
+metadata:
+  name: test-package
+  annotations:
+    config.kubernetes.io/local-config: "true"
+info:
+  description: sample description
+	`)
+	if want, found := expected, strings.TrimSpace(res.Contents[kptfile.KptFileName]); want != found {
```
martinmaly, line 92: for multi-line diffs I like to use https://github.com/google/go-cmp

martinmaly (approved):

PR #3064, opened 2022-04-27
Add kpt command for package edit

mortent:
Add a kpt cli command to edit packages. Ref https://github.com/GoogleContainerTools/kpt/pull/3060

### DIFF
--- a/internal/cmdrpkgedit/command.go
+++ b/internal/cmdrpkgedit/command.go
@@ -1,0 +1,152 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmdrpkgedit
+
+import (
+	"context"
+	"fmt"
+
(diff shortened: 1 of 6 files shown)


internal/cmdrpkgedit/command.go
```diff
@@ -0,0 +1,152 @@
...
+	return newRunner(ctx, rcg).Command
+}
+
+func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner {
+	r := &runner{
+		ctx: ctx,
+		cfg: rcg,
+	}
+	c := &cobra.Command{
+		Use:     "edit SOURCE_PACKAGE NAME",
```
martinmaly, line 66: consider using the same identifiers here and line 33

internal/cmdrpkgedit/command.go
```diff
@@ -0,0 +1,152 @@
...
+func (r *runner) preRunE(cmd *cobra.Command, args []string) error {
+	const op errors.Op = command + ".preRunE"
+	client, err := porch.CreateClient(r.cfg)
+	if err != nil {
+		return errors.E(op, err)
+	}
+	r.client = client
+
+	if len(args) < 2 {
+		return errors.E(op, fmt.Errorf("SOURCE_PACKAGE and NAME are required positional arguments; %d provided", len(args)))
```
martinmaly, line 104: depending on which symbolic name is used for the 2nd positional parameter, this may need to be updated too

martinmaly (approved):
LGTM with couple of nits. let's add a test (https://github.com/GoogleContainerTools/kpt/tree/main/e2e/testdata/porch)

UPDATE_GOLDEN_FILES=1 will be helpful (that way one needs to only write the commands and the stdin/stdout are populated, modulo manual confirmation of correctness)

internal/cmdrpkgedit/command.go
mortent: Fixed.

PR #3065, opened 2022-04-27
support preloading cache in pod evaluator

mengqiy:
The function runner spins up the functions that are passed in through configmap.
The function runner blocks until all the required functions are running and ready.

### DIFF
--- a/porch/config/deploy/2-function-runner.yaml
+++ b/porch/config/deploy/2-function-runner.yaml
@@ -45,10 +45,24 @@ spec:
             - /server --config=/config.yaml --functions=/functions --pod-namespace=porch-fn-system --wrapper-server-image=gcr.io/example-google-project-id/porch-wrapper-server:latest
           ports:
             - containerPort: 9445
+          # Add grpc readiness probe to ensure the cache is ready
+          readinessProbe:
+            exec:
+              command:
+                - /grpc-health-probe
+                - -addr
+                - localhost:9445
           resources:
             requests:
               memory: "64Mi"
               cpu: "125m"
+          volumeMounts:
+            - mountPath: /pod-cache-config
+              name: pod-cache-config-volume
+      volumes:
+        - name: pod-cache-config-volume
+          configMap:
+            name: pod-cache-config
 
 ---
 apiVersion: v1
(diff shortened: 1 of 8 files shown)


porch/config/deploy/2-function-runner.yaml
```diff
@@ -63,3 +77,31 @@ spec:
...
       targetPort: 9445
+
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: pod-cache-config
+  namespace: porch-system
+data:
+  pod-cache-config.yaml: |
```
mengqiy, line 88: This configmap will be generated by a KRM function.

porch/go.mod
```diff
@@ -38,6 +38,7 @@ require (
 	google.golang.org/grpc v1.44.0
 	google.golang.org/protobuf v1.28.0
 	gopkg.in/yaml.v2 v2.4.0
+	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
```
martinmaly, line 41: interesting that we use both versions. Possible to consolidate?

martinmaly (approved):
Thank you! Just one nit

porch/func/internal/podevaluator.go
```diff
@@ -163,6 +172,44 @@ type imagePodAndGRPCClient struct {
...
+		return err
+	}
+
+	var wg sync.WaitGroup
+	for fnImage, ttlStr := range podsTTL {
+		wg.Add(1)
+		go func(img, ttlSt string) {
+			klog.Infof("preloading pod cache for function %v with TTL %v", img, ttlSt)
+			defer wg.Done()
+			ctx, _ := context.WithTimeout(context.Background(), time.Minute)
```
mortent, line 196: I think we should do:
```
ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
defer cancel()
```

porch/go.mod
mengqiy: Done. Using `gopkg.in/yaml.v2` now.

porch/func/internal/podevaluator.go
mengqiy: Fixed!
Thanks for catching this.

PR #3067, opened 2022-04-28
higher rate limit to prevent unnecessary client-side throttling

mengqiy:
Fixes b/230656405


porch/func/internal/podevaluator.go
```diff
@@ -68,6 +68,12 @@ func NewPodEvaluator(namespace, wrapperServerImage string, interval, ttl time.Du
 	if err != nil {
 		return nil, fmt.Errorf("failed to get rest config: %w", err)
 	}
+	// Give it a slightly higher QPS to prevent unnecessary client-side throttling.
+	if restCfg.QPS < 30 {
+		restCfg.QPS = 30.0
+		restCfg.Burst = 45
+	}
```
droot, line 75: oh.. so it was client-side throttling... interesting :)

droot (approved):

PR #3072, opened 2022-04-30
Clean up ResourceGroup with sync del

mortent:
Config Sync orphans the ResourceGroup object when the RootSync is deleted. This adds a temporary fix to make sure it is deleted.

### DIFF
--- a/internal/cmdsync/delete/command.go
+++ b/internal/cmdsync/delete/command.go
@@ -132,12 +132,8 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
 		Namespace: namespace,
 		Name:      name,
 	}
-	rs := unstructured.Unstructured{
-		Object: map[string]interface{}{
-			"apiVersion": rootSyncGVK.GroupVersion().String(),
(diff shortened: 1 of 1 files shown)


internal/cmdsync/delete/command.go
```diff
@@ -182,7 +178,15 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
 	}
 
 	if err := r.client.Delete(r.ctx, &rs); err != nil {
-		return errors.E(op, err)
+		return errors.E(op, fmt.Errorf("failed to clean up RootSync: %w", err))
```
martinmaly, line 181: I wonder if we should continue deleting when we encounter (some) errors and report aggregate errors at the end.

martinmaly (approved):

internal/cmdsync/delete/command.go
mortent: In this case there isn't much more we can do if we can't delete this one. CS will recreate the RG if we delete it.

PR #3074, opened 2022-05-01
docs: Update documentation for alpha commands

mortent:
Updates the reference docs on the kpt.dev site with the `kpt alpha repo` commands. This also uses the mdtogo tags to generate the cobra output for those commands from the site.

Wanted to get feedback on whether this is the right approach before doing the `rpkg` and `sync` commands.

We also need to figure out a good way to handle the flags about logging, talking to the apiserver, and printing. Explaining them all in every command seems unnecessary.


scripts/generate_site_sidebar/sidebar_template.md.tmpl
```diff
@@ -25,6 +25,11 @@
       - [install-resource-group](reference/cli/live/install-resource-group/)
       - [migrate](reference/cli/live/migrate/)
       - [status](reference/cli/live/status/)
+    - [alpha](reference/cli/alpha/)
+      - [repo](reference/cli/alpha/repo/)
+        - [get](reference/cli/alpha/repo/get/)
+        - [reg](reference/cli/alpha/repo/reg/)
+        - [unreg](reference/cli/alpha/repo/unreg/)
```
droot, line 32: I find `reg/unreg` unintuitive. should we use `repo add|remove` instead ?

site/reference/cli/alpha/repo/get/README.md
```diff
@@ -0,0 +1,48 @@
...
+type: docs
+description: >
+  List registered repositories.
+---
+
+<!--mdtogo:Short
+    List registered repositories.
+-->
+
+`get` lists registered repositories.
```
droot, line 13: I am thinking if we should rename this to `list` :) 

`kpt alpha repo list` ?

droot:
> We also need to figure out a good way to handle the flags about logging, talking to the apiserver, and printing. Explaining them all in every command seems unnecessary.

One way would adding a reference for `global flags` and linking to it from the commands that use global flags. And for command help, we might have to concatenate global flags help with the command help. WDYT ?

site/reference/cli/alpha/repo/get/README.md
martinmaly: `kubectl` uses `get` though :(
also, `list` is an alias in place for `kpt alpha repo get` (https://github.com/GoogleContainerTools/kpt/blob/main/internal/cmdrepoget/command.go#L52)

droot:
@mortent overall the approach looks good to me.

site/reference/cli/alpha/repo/get/README.md
droot: Good to know that we have an alias :)

For some reason, my brain associates `get` with fetching single entity, but it could be just me. list alias solves my problem then so will stop bikeshedding :).

scripts/generate_site_sidebar/sidebar_template.md.tmpl
martinmaly: I'd recommend collecting all CLI surface questions and addressing them together. Renaming the reg/unreg or add/remove command is a good suggestion but since renames such as this are disruptive (we need to update docs, design doc, tests, ...) I'd recommend doing them at once.

martinmaly (approved):
lgtm though before merging we should figure out process to finalize the CLI surface.

mortent:
@martinmaly Yeah, I plan to add docs for all the commands in this PR. Then we can figure out which changes to the API surface we want to make and then block this PR until we have addressed them.

martinmaly:
Thank you, @mortent!

To your questions:
* until we introduce the Package resource in the future, I do believe that using language of `package revision` consistently is a good idea.
* +1 on the default mode for the `edit/copy` command which will create new revision in the same repository
* finalizing and publishing are the same. We used 'finalize' term first but switched to publish as a more commonly used term for that activity. So we should remove use of 'finalize' in favor of 'publish'.

mortent:
@martinmaly Updated the PR to use package revision most places (still need some work here to clarify the terminology in some cases, but I'll do that in a follow-up), updated the docs for the copy command, and removed references to 'finalize' and instead use 'publish'.

commands/repocmd.go
```diff
@@ -32,8 +33,8 @@ func NewRepoCommand(ctx context.Context, version string) *cobra.Command {
 	repo := &cobra.Command{
 		Use:     "repo",
 		Aliases: []string{"repository"},
-		Short:   "[Alpha] Manage package repositories.",
-		Long:    "[Alpha] The `repo` command group contains subcommands for managing package repositories.",
+		Short:   "[Alpha]" + repodocs.RepoShort,
+		Long:    "[Alpha]" + repodocs.RepoLong,
```
martinmaly, line 37: this should probably be `"[Alpha] " + repodocs.RepoXYZ` so we get a space between [Alpha] and the doc comment

site/reference/cli/alpha/repo/unreg/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+REPOSITORY_NAME:
+  The name of a repository.
+```
+
+#### Flags
+
+```
+--keep-auth-secret:
+  Keep the Secret object with auth information referenced by the repository. By
```
martinmaly, line 34: super nit: I'd break line before `By`

site/reference/cli/alpha/rpkg/copy/README.md
```diff
@@ -0,0 +1,57 @@
...
+  Create a new package revision from an existing one.
+---
+
+<!--mdtogo:Short
+    Create a new package revision from an existing one.
+-->
+
+`copy` creates a new package revision from an existing one. The new
+revision will be identical to the existing one but with a different
+name.
```
martinmaly, line 15: `name or revision` ?

(since we are considering the default behavior to keep name the same and only rev the revision)

martinmaly (approved):
Only a couple of nits, they can be handled in follow-up.

scripts/generate_site_sidebar/sidebar_template.md.tmpl
droot: Ack.

droot (approved):

commands/repocmd.go
mortent: Good catch. Fixed.

site/reference/cli/alpha/rpkg/copy/README.md
mortent: Agree revision is a better word here. Updated.

PR #3075, opened 2022-05-02
Support server-side-apply creation of new objects

justinsb:
This requires effectively a new code path in the update logic.

### DIFF
--- a/porch/Makefile
+++ b/porch/Makefile
@@ -172,6 +172,11 @@ build-images:
 	IMAGE_NAME="$(PORCH_FUNCTION_RUNNER_IMAGE)" WRAPPER_SERVER_IMAGE_NAME="$(PORCH_WRAPPER_SERVER_IMAGE)" make -C func/ build-image
 	IMAGE_NAME="$(TEST_GIT_SERVER_IMAGE)" make -C test/ build-image
 
+.PHONY: dev-server
+dev-server:
+	docker buildx build --push --tag $(IMAGE_REPO)/$(PORCH_SERVER_IMAGE):$(IMAGE_TAG) -f ./build/Dockerfile.porch "$(KPTDIR)"
+	kubectl set image -n porch-system deployment/porch-server porch-server=$(IMAGE_REPO)/$(PORCH_SERVER_IMAGE):${IMAGE_TAG}
+
 .PHONY: apply-dev-config
 apply-dev-config:
 	# TODO: Replace with KCC (or self-host a registry?)
--- a/porch/pkg/registry/porch/packagecommon.go
+++ b/porch/pkg/registry/porch/packagecommon.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 
+	unversionedapi "github.com/GoogleContainerTools/kpt/porch/api/porch"
 	api "github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
 	configapi "github.com/GoogleContainerTools/kpt/porch/api/porchconfig/v1alpha1"
 	"github.com/GoogleContainerTools/kpt/porch/pkg/engine"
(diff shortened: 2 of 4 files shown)


porch/Makefile
```diff
@@ -163,6 +163,11 @@ build-images:
 	IMAGE_NAME="$(PORCH_FUNCTION_RUNNER_IMAGE)" WRAPPER_SERVER_IMAGE_NAME="$(PORCH_WRAPPER_SERVER_IMAGE)" make -C func/ build-image
 	IMAGE_NAME="$(TEST_GIT_SERVER_IMAGE)" make -C test/ build-image
 
+.PHONY: dev-server
```
martinmaly, line 166: nice!

porch/pkg/registry/porch/packagecommon.go
```diff
@@ -141,30 +145,71 @@ func (r *packageCommon) updatePackageRevision(ctx context.Context, name string,
 
 	oldPackage, err := r.getPackage(ctx, name)
 	if err != nil {
-		return nil, false, err
+		if apierrors.IsNotFound(err) {
+			// For server-side apply, we create new objects where they don't already exist
```
martinmaly, line 149: should we be checking `forceAllowCreate` here to see if creation on update is disabled?

(I honestly don't know whether the `false` value means "do not create on update" or  whether it means "do not force" which seems to read very permissive "just do whatever :)")

porch/pkg/registry/porch/packagecommon.go
```diff
@@ -141,30 +145,71 @@ func (r *packageCommon) updatePackageRevision(ctx context.Context, name string,
...
+		} else {
+			return nil, false, err
+		}
+	}
+
+	isCreate := false
```
martinmaly, line 155: [nit] I found this line surprising given 149 determined that we may proceed to creation.

I'd probably go for:

var isCreate bool
if ... {
  isCreate = true
} else {
}

or add a comment, or move the definition above line 146

porch/pkg/registry/porch/packagecommon.go
```diff
@@ -141,30 +145,71 @@ func (r *packageCommon) updatePackageRevision(ctx context.Context, name string,
...
+				klog.Infof("update failed create validation: %v", err)
+				return nil, false, err
+			}
+		}
+
+		// TODO: ValidateCreate function ?
+		// fieldErrors := r.updateStrategy.ValidateCreate(ctx, newRuntimeObj, oldRuntimeObj)
+		// if len(fieldErrors) > 0 {
+		// 	return nil, false, apierrors.NewInvalid(api.SchemeGroupVersion.WithKind("PackageRevision").GroupKind(), name, fieldErrors)
+		// }
```
martinmaly, line 210: or, like core k8s we should have CreateStrategy interface
[context](https://github.com/kubernetes/apiserver/blob/master/pkg/registry/rest/create.go#L42)
[k8s code](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go#L564-L575)

martinmaly (approved):
lgtm with some nits. IMO cleanest approach would be to add
`SimpleRESTCreateStrategy` similar to https://github.com/GoogleContainerTools/kpt/blob/main/porch/pkg/registry/porch/rest.go#L26

porch/pkg/registry/porch/packagecommon.go
justinsb: Good idea - I added a check for it, and it seems to work as expected.

I agree it's not very clear what the intent of the flag is...

porch/pkg/registry/porch/packagecommon.go
justinsb: Good point; reorganized the code to clean this up

porch/pkg/registry/porch/packagecommon.go
justinsb: That's #3090 so I'm going to merge this one!

PR #3077, opened 2022-05-02
applyset: introduce abstraction over objects

justinsb:
This lets us use typed and untyped objects.

### DIFF
--- a/porch/controllers/remoterootsync/pkg/applyset/interfaces.go
+++ b/porch/controllers/remoterootsync/pkg/applyset/interfaces.go
@@ -1,0 +1,29 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


justinsb:
cc @apelisse

porch/controllers/remoterootsync/pkg/applyset/interfaces.go
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+package applyset
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type ApplyableObject interface {
```
martinmaly, line 22: This seems reasonable. I do wonder about some of the prior art interfaces that are similar but not quite the same. for example

[runtime.Object](https://github.com/kubernetes/apimachinery/blob/v0.23.6/pkg/runtime/interfaces.go#L302) has slightly different structure.

I prefer the straightforward simplicity of the ApplyableObject but it is worth checking whether there is a reason to go with some prior art already in place.

martinmaly (approved):

porch/controllers/remoterootsync/pkg/applyset/interfaces.go
apelisse: Yeah, we typically get this information with the `meta.Accessor` from the `runtime.Object`

justinsb: Indeed, but the intent is that this isn't necessarily a runtime.Object.  It is supposed to be easy to take an unstructured object or a typed object and use it though!

PR #3079, opened 2022-05-02
Recognize latest package revision on publishing

martinmaly:
When package is published, make sure to recognize its latest status


justinsb (approved):
lgtm

PR #3080, opened 2022-05-02
use kpt git parsing in `rpkg clone`

natasha41575:
For `kpt alpha rpkg clone`, make using the `--ref` and `--directory` flags optional when using git URLs, instead allow something like 

```
kpt alpha rpkg clone https://github.com/GoogleCloudPlatform/blueprints.git/catalog/bucket@main cloned-bucket \
  --repository=deployments \
  --namespace=default
```

Throw errors if the ref/directory are both present in the URL and set via flags, and they don't match.

### DIFF
--- a/e2e/testdata/porch/rpkg-clone/config.yaml
+++ b/e2e/testdata/porch/rpkg-clone/config.yaml
@@ -11,9 +11,7 @@ commands:
       - rpkg
       - clone
       - --namespace=rpkg-clone
-      - https://github.com/platkrm/test-blueprints.git
-      - --directory=basens
-      - --ref=basens/v1
+      - https://github.com/platkrm/test-blueprints.git/basens@basens/v1
       - --repository=git
       - --revision=v0
       - basens-clone
--- a/internal/cmdrpkgclone/command.go
+++ b/internal/cmdrpkgclone/command.go
@@ -43,9 +43,11 @@ SOURCE_PACKAGE:
     * oci://oci-repository/package-name
     * http://git-repository.git/package-name
     * package-revision-name
-  With git URLs containing the ".git" suffix, you can optionally include the package directory 
-  and ref in the source package:
-    * http://git-repository.git/package-name/directory@main
+
+  With git URLs, you can either include the package directory and ref in SOURCE_PACKAGE, or
+  you can individually specify the repo, ref, and directory using flags:
+    * kpt alpha rpkg clone http://git-repository.git/package-name@main target --repository=repo
+    * kpt alpha rpkg clone http://git-repository.git target --directory=package-name --ref=main --repository=repo
 
 NAME:
   Target package revision name (downstream package)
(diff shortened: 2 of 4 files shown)


internal/cmdrpkgclone/command.go
```diff
@@ -154,11 +145,15 @@ func (r *runner) preRunE(cmd *cobra.Command, args []string) error {
 		}
 
 	case strings.Contains(source, "/"):
+		gitArgs, err := parse.GitParseArgs(context.Background(), args)
```
martinmaly, line 148: will this try to clone the repository locally (or try to contact it in any way), or is this purely syntax-level parsing?

natasha41575: From what I can tell, `GitParseArgs` will try to contact the repository only if there is no ref/version provided in the URL. It tries to make contact in this case to determine what the default branch is. 

What is the correct behavior for `kpt alpha rpkg clone` if no version is provided? I can extract the syntax-level parsing out of `GitParseArgs` and use that here instead if we want to do something different.

martinmaly: empty ref would be ok, or we can default to `main`. Porch defaults to `main` in other places where ref is not specified.

Error may be OK too.

I actually thought (though may be wrong) that another case where we contact the server is if the .git/ part is missing and the parser can't tell where the repository path ends and directory path begins:

domain.com/some/path/here

is the `some/path` repo? or `some` ?

regardless, the rpkg commands should not interact with the repository and only operate at the level of syntax of the URL provided, even if there are cases we can't handle and that will result in errors (IMO)

natasha41575: > empty ref would be ok, or we can default to main. Porch defaults to main in other places where ref is not specified.

I'll default to `main` to be consistent with other places. 

> I actually thought (though may be wrong) that another case where we contact the server is if the .git/ part is missing and the parser can't tell where the repository path ends and directory path begins

Ah you are correct, I missed that it contacts the repo while attempting to convert the url to a URL containing `.git`.

Is it a requirement for `kpt alpha rpkg clone` that the URL contains `.git`? I'm thinking about moving the logic in `GitParseArgs` that handles URLs with a .git suffix out to its own helper function so that we can use it here without worrying about communication with the repo.

martinmaly: Presence of the `.git` part of the URL is not requirement. I'd say the requirement is that the parser doesn't _modify_ the URL.

So if customer gave us https://github.com/org/repo.git/path/to/package
we should include the `.git`

if they provided https://github.com/org/repo/path/to/package
we should not artificially introduce `.git` part into the URL.

internal/cmdrpkgclone/command.go
```diff
@@ -96,8 +100,8 @@ func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner
 
 	c.Flags().StringVar(&r.strategy, "strategy", string(porchapi.ResourceMerge),
 		"update strategy that should be used when updating this package; one of: "+strings.Join(strategies, ","))
-	c.Flags().StringVar(&r.directory, "directory", "/", "Directory within the repository where the upstream package is located.")
-	c.Flags().StringVar(&r.ref, "ref", "main", "Branch in the repository where the upstream package is located.")
+	c.Flags().StringVar(&r.directory, "directory", "", "Directory within the repository where the upstream package is located.")
+	c.Flags().StringVar(&r.ref, "ref", "", "Branch in the repository where the upstream package is located.")
```
natasha41575, line 104: I removed these defaults because otherwise it is impossible to tell whether the user set the flag explicitly or not. Since we are trying to infer the ref from the URL, we need to be able to throw an error if the user sets the flag and also includes the ref in the URL and they contradict each other. I set the default manually below in L182-L187 after this error check occurs.

e2e/testdata/porch/rpkg-clone/config.yaml
```diff
@@ -11,9 +11,7 @@ commands:
       - rpkg
       - clone
       - --namespace=rpkg-clone
-      - https://github.com/platkrm/test-blueprints.git
-      - --directory=basens
-      - --ref=basens/v1
+      - https://github.com/platkrm/test-blueprints.git/basens@basens/v1
```
natasha41575, line 14: Updated this test to demonstrate the new inferring behavior. There is another similar command in `rpkg-edit/config.yaml` that uses the `--directory` and `--ref` flags which I haven't modified,  demonstrating that the user can still use the flags when needed.

martinmaly (approved):

natasha41575:
Thanks for the reviews and answering all my questions!

PR #3084, opened 2022-05-04
Heal Comments on Package Resources Update

martinmaly:
- Heal
- Fix golden files & flags

### DIFF
--- a/porch/pkg/engine/engine.go
+++ b/porch/pkg/engine/engine.go
@@ -17,6 +17,7 @@ package engine
 import (
 	"context"
 	"fmt"
+	"io"
 	"io/fs"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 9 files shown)


porch/pkg/engine/engine.go
```diff
@@ -546,3 +554,76 @@ func (m *mutationReplaceResources) Apply(ctx context.Context, resources reposito
...
+			}
+
+			var copyID kio.FilterFunc = func(r []*yaml.RNode) ([]*yaml.RNode, error) {
+				for _, n := range r {
+					for _, original := range items {
+						if n.GetNamespace() == original.GetNamespace() &&
+							n.GetName() == original.GetName() &&
+							n.GetApiVersion() == original.GetApiVersion() &&
+							n.GetKind() == original.GetKind() {
+							id, err := original.Pipe(yaml.GetAnnotation(kioutil.IdAnnotation))
```
martinmaly, line 579: @justinsb what algorithm did you use to match resources in the manual implementation of comment healing? I tried to reuse kpt machinery here and this seemed ok-ish to start with.

alternative would be to assign IDs when reading package resources though kpt libraries may not play nicely with it.

mortent:
The solution looks fine and seems to be reusing the functionality in kyaml already used for functions. I see that this can be unexpected for users since we can't distinguish between comments that have been deliberately deleted from what has been unintentionally lost as part of the process. But we can probably look at feedback here and see what we should do.

mortent (approved):

martinmaly:
I agree we'll have to figure out some way to heal comments when the client editor (i.e. UI) drops them but allow for their deletion. Maybe we'll need a flag. I'll merge this and we'll evolve from here.

PR #3093, opened 2022-05-05
Clarify Pluggable Deployment Mechanism

martinmaly:
In addition to Config Sync (which is the default deployment
mechanism used with Porch), clarifying that other GitOps-based
components can be used as well.


mortent (approved):

bgrant0607:
Thanks. Looks pretty good.

FWIW, "i.e." is "that is" and "e.g." is "for example", so ideally I'd change the "i.e."s in the images to "e.g."s, but this is much clearer already if you want to merge this first.

bgrant0607 (approved):

martinmaly:
Thank you, @bgrant0607! Fixed the images.

PR #3094, opened 2022-05-05
Rename edit command to copy

mortent:
This renames the `kpt alpha rpkg edit` command to `kpt alpha rpkg copy`. Since each package revision can be edited using the `kpt alpha rpkg push` command, it is a bit confusing that we have an edit command that creates a new package revision (and can even create a completely new package).

### DIFF
--- a/internal/cmdrpkgcopy/command.go
+++ b/internal/cmdrpkgcopy/command.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package cmdrpkgedit
+package cmdrpkgcopy
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


internal/cmdrpkgcopy/command.go
```diff
@@ -63,10 +63,10 @@ func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner
 		cfg: rcg,
 	}
 	c := &cobra.Command{
-		Use:     "edit SOURCE_PACKAGE NAME",
+		Use:     "copy SOURCE_PACKAGE NAME",
```
martinmaly, line 66: we can consider keeping edit as an alias (though we can start with copy and seeing where it gets us)

martinmaly (approved):

internal/cmdrpkgcopy/command.go
mortent: Yeah, adding edit as an alias is a good idea.

PR #3095, opened 2022-05-06
Update kpt definition on landing page

bgrant0607:
More about the approach and the value of the suite and less about the mechanics of the CLI.
cc @justinsb


bgrant0607:
cc @selfmanagingresource

justinsb (approved):
Thanks - lgtm!

yuwenma (approved):

PR #3097, opened 2022-05-06
Support creation of package directly from resources

justinsb:
A slightly awkward implementation, but this allows for direction
creation of packages without a base.


porch/pkg/registry/porch/packagerevisionresources.go
```diff
@@ -170,3 +172,89 @@ func (r *packageRevisionResources) Update(ctx context.Context, name string, objI
...
+		return nil, apierrors.NewInternalError(fmt.Errorf("error getting repository %v: %w", repositoryID, err))
+	}
+
+	fieldErrors := r.createStrategy.Validate(ctx, runtimeObject)
+	if len(fieldErrors) > 0 {
+		return nil, apierrors.NewInvalid(api.SchemeGroupVersion.WithKind("PackageRevision").GroupKind(), obj.Name, fieldErrors)
+	}
+
+	// We currently create the PackageRevision, and then attach the resources
+	// TODO: Can we streamline this - create the package with its resources, or maybe return the draft?
```
mortent, line 218: Could we address this by creating a new task type that could add the resources to the package as apart of creating the initial revision?

justinsb: We could, but I worry that it would make the model more complex, in that now some resources are directly created in PackageRevision.  That said, we know that patches share some elements of this problem in the reclone-and-replay approach.

PR #3099, opened 2022-05-06
Update CODEOWNERS with the broader team

justinsb:


droot (approved):

bgrant0607:
It seems like the directories covered in the file are pretty arbitrary. Do we need them all? Are there some that are missing?

Also, I think all listed owners are spammed for reviews on PRs that touch those directories. I'd be happy not to be spammed on anything other than the site. :-)

justinsb:
So the original intent of the directories was that...

* There would be one approver / one voice for all docs.  A few packages referenced from the docs would also have tighter control.
* Schema changes would get specially checked
* Tests that enforce the function contract would get specially checked.

I'm proposing that we adopt more of a broader web of trust here, so all directories have the same permissions; given that I think it's less important about the directory coverage, but I do need to check about what happens to non-covered directories.

However, I ack that it is annoying if you get spammed.  i.e. we're not _just_ giving people permissions, we're also destroying their email.  I'll set you specifically as only on the docs site if you like @bgrant0607 , and if anyone else wants to opt out, please LMK.

PR #3101, opened 2022-05-06
Update the auth flags for registering repositories

mortent:
We will eventually add support for other auth types than just basic auth. Therefore we need a convention for how we name the flags used for authentication against package repositories. This updates the current flags to follow the following convention:
* Flags used the format: `repo-<auth type>-<input>`
  * Every flag use the `repo-` prefix to make it clear it pertains to repository auth (rather than cluster) and to avoid conflicts with existing flags we inherit from kubectl (examples are username and password).
  * The auth type section tells which type of authentication this flag is used for. Currently we only support basic auth.
  * The last part contains the name of data needed for the specified auth type. For example, for basic auth, we need username and password.

Following this convention, the flags for specifying username and password for basic auth results in the flags `repo-basic-username` and `repo-basic-password`.

### DIFF
--- a/docs/design-docs/07-package-orchestration.md
+++ b/docs/design-docs/07-package-orchestration.md
@@ -509,8 +509,8 @@ GITHUB_TOKEN=<GitHub Personal Access Token>
 
 $ kpt alpha repo register \
   --namespace default \
-  --repo-username=${GITHUB_USERNAME} \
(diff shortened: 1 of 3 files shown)


martinmaly (approved):
the name is a bit long but better start with accurate (albeit long) name than not. We can evolve as needed.

PR #3102, opened 2022-05-06
FAQ overhaul

bgrant0607:
Take into account the larger scope of kpt and eliminate some of the smaller technical details.

cc @justinsb @selfmanagingresource

### DIFF
--- a/site/faq/README.md
+++ b/site/faq/README.md
@@ -29,7 +29,7 @@ _kpt_
 - Allows edits to the configuration in-place without creating complex patches.
 - Supports rebase with resource merge strategy allowing for edited config to
   be updated.
-- Enables workflows that combine programmatic changes (functions) with manual
+- Enables workflows that combine programmatic changes ([functions]) with manual
   edits.
 - Aims to support mutating and validating admission control on derived packages.
 - Also supports packages, [package orchestration], resource actuation, and GitOps.
--- a/site/faq/README.md
+++ b/site/faq/README.md
@@ -4,59 +4,45 @@
 
 Please visit the [roadmap document] and the [kpt milestones].
 
-### What does kpt provide that git clone doesn't?
-
-kpt enables out-of-the-box workflows that `git clone` does not. Such as: cloning
-and versioning git subdirectories, updating from upstream by performing
-structured merges on resources, programmatically editing configuration (rather
-than with an editor), etc.
-
 ### How is kpt different from other solutions?
 
-Rather than expressing configuration as code, kpt represents configuration
-packages as data, in particular as YAML or JSON objects adhering to [The
-Kubernetes Resource Model].
+Rather than expressing configuration as code that generates configuration,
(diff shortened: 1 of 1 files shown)


site/faq/README.md
```diff
@@ -4,59 +4,45 @@
 
 Please visit the [roadmap document] and the [kpt milestones].
 
-### What does kpt provide that git clone doesn't?
-
-kpt enables out-of-the-box workflows that `git clone` does not. Such as: cloning
-and versioning git subdirectories, updating from upstream by performing
-structured merges on resources, programmatically editing configuration (rather
-than with an editor), etc.
```
yuwenma, line 12: Is this removal intended?
This section refers to the `kpt pkg` which is not described anywhere below. And git is so popular so it seems useful to tell the differences (even if it's verbose). Or do we want to de-emphasize the package management feature from kpt?

bgrant0607: Yes, this removal was intended. This was the main emphasis when we first created the kpt packaging mechanism, but isn't now. For instance, we plan to support container images as storage.

I'm trying to up-level the FAQ to the whole toolchain and the wysiwyg experience.

site/faq/README.md
```diff
@@ -4,59 +4,45 @@
...
-- Allows you to edit the configuration in-place without creating complex patches.
-- Focuses on rebase with resource merge strategy allowing for edited config to
+- Optimizes for WYSIWYG configuration and in-place customization.
+- Allows edits to the configuration in-place without creating complex patches.
+- Supports rebase with resource merge strategy allowing for edited config to
   be updated.
-- Has a continuous learning curve as you usually start small with modifying
-  several YAML files using an editor and then want to scale with complexity of
-  the application.
 - Enables workflows that combine programmatic changes (functions) with manual
```
yuwenma, line 32: nit: square brackets around `functions`

bgrant0607: Thanks. Looks like that was a preexisting bug. Should be fixed now.

site/faq/README.md
yuwenma: ah, that makes sense.

site/faq/README.md
```diff
@@ -4,59 +4,45 @@
...
-(pkg, fn, live) independently if needed. For example:
+transformation, and actuation functionality. At the same time, kpt follows a modular
+design principle to make it possible to use each of its functionality
+independently if needed. For example:
 
 - You can use packaging without declaring functions
-- You can use imperative functions to operate on vanilla directories of k8s
+- You can use imperative functions to operate on vanilla directories of Kubernetes
   resources
 - You can use apply logic without buying into full the packaging story (still
```
yuwenma, line 47: oops, "the full packaging story"

yuwenma (approved):

site/faq/README.md
bgrant0607: Thanks. Fixed

PR #3106, opened 2022-05-06
redirecting the homepage to https://kpt.dev

droot:
This PR updates the `gh-pages` that powers old website at https://googlecontainertools.github.io/kpt/ to redirect to https://kpt.dev.

Note the non-root pages from old website will continue to serve and they have a header pointing to `kpt.dev`. So this isn't disruptive.


bgrant0607:
Thanks!

How do you test these kinds of changes?

mortent (approved):

droot:
> How do you test these kinds of changes?

Actually in this case, it was tricky because this is website for the alpha version and couldn't find relevant docs and our old bots to update the site are also not in operation. So I made this change manually and ran `python3 -m http.server` outside the kpt directory to simulate <github.com>/kpt. 

Since this is old website, I was okay to do this in hackish way. I found a doc that describes current website structure, I will create a public copy of that and  publish under docs in current repo so that we have some knowledge about current website.

PR #3109, opened 2022-05-07
added kpt pkg the hard way guide

droot:
This PR adds `kpt pkg the hard way` guide.

I am flexible in changing the title of the guide. Suggestions welcome.


site/guides/kpt-pkg-the-hard-way.md
```diff
@@ -0,0 +1,383 @@
...
+  name: app-admin
+  namespace: example
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: app-admin
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: Group
+```
```
bgrant0607, line 203: This appears to be missing a line

bgrant0607:
It looks like one of the yaml files was missing a line.

Title: I'd make it more informative and different from the other guide(s) we will likely have for the same scenario. "Namespace provisioning example via kpt CLI"?

droot:
> It looks like one of the yaml files was missing a line.
> 

That's a good catch.

> Title: I'd make it more informative and different from the other guide(s) we will likely have for the same scenario. "Namespace provisioning example via kpt CLI"?

I like it. Went with your suggestion. Updated it.

bgrant0607 (approved):
LGTM

site/guides/namespace-provisioning-example.md
```diff
@@ -0,0 +1,441 @@
+# Namespace provisioning example via kpt CLI
```
yuwenma, line 1: Suggest adding the link to the UI version guide, or simply call this "Namespace Provisioning Example"

yuwenma (approved):

site/guides/namespace-provisioning-example.md
droot: I am working with Chris to get UI guides published. I will take a second pass on reorganizing these once all guides are in.

Thanks for the suggestion.

PR #3114, opened 2022-05-10
Add Porch Book Chapter

martinmaly:
The chapter has a basic introduction to Package Orchestration
concepts and references the user guides that already exist,
rather than duplicating the information in the book.


bgrant0607 (approved):
Thanks for creating this section.

I suggest a merge-and-iterate approach.

This introduces the concepts of package repositories and package lifecycle stages, but it would help to mention the package authoring operations in order to understand how packages can be edited. 

We could also link to the design doc.

droot (approved):

PR #3116, opened 2022-05-10
Fix reference to value for name

aablsk:
Description: Example yaml shows `- name: wordpress-mysql` while text refers to "name `mysql`" which I believe is incorrect in this context.
Motivation: I believe documentation should be as concise as possible.


droot (approved):
Thank you @aablsk

PR #3123, opened 2022-05-10
Add a `kpt alpha live plan` command

mortent:
Early version of a `kpt live plan` command, currently positioned under the alpha command group.

The KRM output format allows users to write kpt functions that can validate the plan, which makes it easily extensible. I have a simple example of this [here](https://github.com/mortent/kpt-functions/blob/master/block-mutation/main.go#L67), which simply looks at whether the diff between current live state and the new state includes changes disallowed fields.

### DIFF
--- a/internal/cmdplan/command.go
+++ b/internal/cmdplan/command.go
@@ -1,0 +1,288 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmdplan
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/GoogleContainerTools/kpt/internal/util/argutil"
+	"github.com/GoogleContainerTools/kpt/pkg/api/plan/v1alpha1"
+	"github.com/GoogleContainerTools/kpt/pkg/live"
+	kptplanner "github.com/GoogleContainerTools/kpt/pkg/live/planner"
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/cli-utils/cmd/flagutils"
+	"sigs.k8s.io/cli-utils/pkg/common"
+	print "sigs.k8s.io/cli-utils/pkg/print/common"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+const (
+	TextOutput = "text"
+	KRMOutput  = "krm"
+
+	EntryPrefix   = "\t"
+	ContentPrefix = "\t\t"
+)
+
+func NewRunner(ctx context.Context, factory util.Factory, ioStreams genericclioptions.IOStreams) *Runner {
+	r := &Runner{
+		ctx:       ctx,
+		factory:   factory,
+		ioStreams: ioStreams,
+		serverSideOptions: common.ServerSideOptions{
+			ServerSideApply: true,
+		},
+	}
+	c := &cobra.Command{
+		Use:     "plan [PKG_PATH | -]",
+		PreRunE: r.PreRunE,
+		RunE:    r.RunE,
+	}
+	c.Flags().StringVar(&r.inventoryPolicyString, flagutils.InventoryPolicyFlag, flagutils.InventoryPolicyStrict,
+		"It determines the behavior when the resources don't belong to current inventory. Available options "+
+			fmt.Sprintf("%q and %q.", flagutils.InventoryPolicyStrict, flagutils.InventoryPolicyAdopt))
+	c.Flags().BoolVar(&r.serverSideOptions.ForceConflicts, "force-conflicts", false,
+		"If true, overwrite applied fields on server if field manager conflict.")
+	c.Flags().StringVar(&r.serverSideOptions.FieldManager, "field-manager", common.DefaultFieldManager,
+		"The client owner of the fields being applied on the server-side.")
+	c.Flags().StringVar(&r.output, "output", "text",
+		"The output format for the plan. Must be either 'text' or 'krm'. Default is 'text'")
+	r.Command = c
+
+	return r
+}
+
+func NewCommand(ctx context.Context, factory util.Factory, ioStreams genericclioptions.IOStreams) *cobra.Command {
+	return NewRunner(ctx, factory, ioStreams).Command
+}
+
+type Runner struct {
+	ctx       context.Context
+	Command   *cobra.Command
+	factory   util.Factory
+	ioStreams genericclioptions.IOStreams
+
+	inventoryPolicyString string
+	serverSideOptions     common.ServerSideOptions
+	output                string
+}
+
+func (r *Runner) PreRunE(c *cobra.Command, args []string) error {
+	return r.validateOutputFormat()
+}
+
+func (r *Runner) validateOutputFormat() error {
+	if !(r.output == "text" || r.output == "krm") {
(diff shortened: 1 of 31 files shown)


internal/cmdplan/command.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	c := &cobra.Command{
+		Use:     "plan [PKG_PATH | -]",
+		PreRunE: r.PreRunE,
+		RunE:    r.RunE,
+	}
+	c.Flags().StringVar(&r.inventoryPolicyString, flagutils.InventoryPolicyFlag, flagutils.InventoryPolicyStrict,
+		"It determines the behavior when the resources don't belong to current inventory. Available options "+
+			fmt.Sprintf("%q and %q.", flagutils.InventoryPolicyStrict, flagutils.InventoryPolicyAdopt))
+	c.Flags().BoolVar(&r.serverSideOptions.ServerSideApply, "server-side", true,
```
justinsb, line 61: I'd be OK with requiring SSA for plan.  We can always expose the option later as well...

pkg/api/plan/v1alpha1/types.go
```diff
@@ -0,0 +1,67 @@
...
+
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+var ResourceMeta = yaml.ResourceMeta{
+	TypeMeta: yaml.TypeMeta{
```
justinsb, line 23: Q: Why yaml.TypeMeta / yaml.ObjectMeta vs the apimachinery types?

internal/cmdplan/command.go
```diff
@@ -0,0 +1,285 @@
...
+		panic(fmt.Errorf("error reading text: %v", err))
+	}
+}
+
+func printKRM(plan *v1alpha1.Plan, ioStreams genericclioptions.IOStreams) error {
+	b, err := yaml.Marshal(plan)
+	if err != nil {
+		return fmt.Errorf("failed to marshall plan into yaml: %w", err)
+	}
+	rn, err := yaml.Parse(string(b))
```
justinsb, line 270: What's the difference vs just printing `b`?  (Can we add a comment to explain what we're doing here...)

pkg/api/plan/v1alpha1/types.go
```diff
@@ -0,0 +1,67 @@
...
+	Error     ActionType = "Error"
+)
+
+type Action struct {
+	Type      ActionType                 `json:"action,omitempty" yaml:"action,omitempty"`
+	Group     string                     `json:"apiVersion,omitempty" yaml:"apiVersion,omitempty"`
+	Kind      string                     `json:"kind,omitempty" yaml:"kind,omitempty"`
+	Name      string                     `json:"name,omitempty" yaml:"name,omitempty"`
+	Namespace string                     `json:"namespace,omitempty" yaml:"namespace,omitempty"`
+	Before    *unstructured.Unstructured `json:"before,omitempty" yaml:"before,omitempty"`
```
justinsb, line 64: Does the extra `object` here mean that we're not using `sigs.k8s.io/yaml`... should we be?

pkg/live/planner/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+
+	var actions []planv1alpha1.Action
+	var err error
+	for e := range eventCh {
+		if e.Type == event.InitType {
+			// This event includes all resources that will be applied, pruned or deleted, so
+			// we make sure we fetch all the resources from the cluster.
+			// TODO: See if we can update the actuation library to provide the pre-actuation
+			// versions of the resources as part of the regular run. This solution is not great
+			// as fetching all resources will take time.
```
justinsb, line 115: Agreed - it also means that we risk not getting a consistent view if there are concurrent changes.

pkg/live/planner/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+
+func (r *ClusterPlanner) dryRunForPlan(ctx context.Context, inv inventory.Info,
+	objects []*unstructured.Unstructured, o Options) ([]planv1alpha1.Action, error) {
+
+	eventCh := r.applier.Run(ctx, inv, objects, apply.ApplierOptions{
+		DryRunStrategy:    common.DryRunServer,
+		ServerSideOptions: o.ServerSideOptions,
+	})
+
+	var actions []planv1alpha1.Action
```
justinsb, line 107: Nit: might be easier to just use a map.

pkg/live/planner/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+func handlePruneEvent(e event.Event, a planv1alpha1.Action) planv1alpha1.Action {
+	if e.PruneEvent.Error != nil {
+		a.Type = planv1alpha1.Error
+		a.Error = e.PruneEvent.Error.Error()
+	} else {
+		switch e.PruneEvent.Operation {
+		case event.Pruned:
+			a.Type = planv1alpha1.Delete
+		case event.PruneSkipped:
+			a.Type = planv1alpha1.Skip
```
justinsb, line 189: Q: Why might prune get skipped?  A comment here or TODO to figure out why things are skipped would be great

pkg/live/planner/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+			index := indexForIdentifier(id, actions)
+			a := actions[index]
+			actions[index] = handleApplyEvent(e, a)
+		case event.PruneType:
+			id := e.PruneEvent.Identifier
+			index := indexForIdentifier(id, actions)
+			a := actions[index]
+			actions[index] = handlePruneEvent(e, a)
+		case event.DeleteType:
+			id := e.DeleteEvent.Identifier
```
justinsb, line 143: Q: What's the difference between a prune and a delete?

pkg/live/planner/cluster.go
```diff
@@ -0,0 +1,289 @@
...
+				Namespace: id.Namespace,
+				Before:    u,
+			})
+		}
+	}
+	return actions, nil
+}
+
+type resourceFetcher struct {
+	dc     dynamic.Interface
```
justinsb, line 236: Nit: I normally just use `client` or `dynamicClient` (or just embed the interface if it's intrinsic to the type, but that normally ends poorly!).  It just makes it a little easier to read when you come back to it after 6 months vs `dc`

pkg/live/planner/cluster_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+			if diff := cmp.Diff(tc.expectedPlan, plan); diff != "" {
+				t.Errorf("plan mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+type FakeApplier struct {
+	events []event.Event
```
justinsb, line 127: Observation: From a testing perspective, it would be great if we didn't have to construct the events, but could instead provide the pre-existing resources.  Not for this PR though!

justinsb (approved):
LGTM - a few comments / questions, but this is a great start and it's an alpha command so we should merge and iterate!

fmichaelobrien:
Thanks @justinsb - very nice - reviewing

pkg/api/plan/v1alpha1/types.go
mortent: I think either would work here. This just keeps it aligned with how the other types in the project is defined.

internal/cmdplan/command.go
mortent: This outputs the resources in the KRM function `ResourceList` resource. I'm not sure if this is the right way to do it, so I'll look into it. We do want to be able to output both the plan and the input resources, but that is something I am looking to add later.

internal/cmdplan/command.go
mortent: Fixed.

pkg/api/plan/v1alpha1/types.go
mortent: We are not using `sigs.k8s.io/yaml` for marshaling yaml. It is just uses serializes the `Plan` struct. We could build the plan yaml with the kyaml library directly instead.
If we don't provide a go type for the plan, clients will not have a type they can easily unmarshal the plan into. But I'm not sure if we can embed other `before` and `after` resources within another go type in a sane way anyway (i.e. avoid the unecessary `object` property) and the most common use for the plan is in functions, which rely on kyaml anyway.

pkg/live/planner/cluster.go
mortent: Added a comment here. Will add more information about why resources might be skipped in a follow-up PR.

pkg/live/planner/cluster.go
mortent: Legacy behavior from the actuation library.

pkg/live/planner/cluster.go
mortent: Fixed.

pkg/live/planner/cluster_test.go
mortent: Agree.

PR #3127, opened 2022-05-11
[WIP] kpt fn command to embed fn-config in a function image

mortent:
This is a POC for a `kpt fn` command that allows users to creates a new docker image that embeds the function config within the image. This allows for creating specific images from our more generic images available in the function catalog and avoid having to distribute the function config separately. It does this simply by adding a new layer to the function image that includes the function config in a well-known location (currently `/localconfig/fn-config.yaml`).

A function can then chose to use the local config if a different one is not provided in the input (or we could say that the local always win regardless). An example of this in [here](https://github.com/mortent/kpt-functions/blob/master/block-mutation/main.go#L67).


liamfallon:
@liamfallon to pull out into an issue

PR #3130, opened 2022-05-11
simplify rpkg copy: disallow editing outside repository and package name changes

natasha41575:
Simplify `kpt alpha rpkg copy`
- allow copying only within repository
- disallow package name change (so `copy` only creates new revision of the same package)
- added a TODO for defaulting the revision number to latest+1 which I will tackle in a followup PR. I had some questions about this task, which I left as a comment in the issue filed. 

Before:
`kpt alpha rpkg copy deployments-11ca1db650fa4bfa33deeb7f488fbdc50cdb3b82 foo --namespace default --repository deployments --revision v1`

After: 
`kpt alpha rpkg copy deployments-11ca1db650fa4bfa33deeb7f488fbdc50cdb3b82 --namespace default --revision v1`

PR to update docs: https://github.com/GoogleContainerTools/kpt/pull/3135


site/reference/cli/alpha/rpkg/copy/README.md
```diff
@@ -19,7 +19,7 @@ revision.
 <!--mdtogo:Long-->
 
 ```
-kpt alpha rpkg copy SOURCE_PACKAGE_REV_NAME TARGET_PACKAGE_NAME [flags]
+kpt alpha rpkg copy SOURCE_PACKAGE_REV_NAME [flags]
```
natasha41575, line 22: I probably shouldn't update the docs until release. So I'll split this out into another PR.

internal/cmdrpkgcopy/command.go
```diff
@@ -111,22 +117,59 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
...
 	if err := r.client.Create(r.ctx, pr); err != nil {
 		return errors.E(op, err)
 	}
-
 	fmt.Fprintf(cmd.OutOrStdout(), "%s created", pr.Name)
 	return nil
 }
+
+func (r *runner) getPackageRevisionSpec() (*porchapi.PackageRevisionSpec, error) {
+	res := resource.NewBuilder(r.cfg).
```
mortent, line 131: Do we need to fetch the data here in table format? It seems like just using client-go directly would be easier.

internal/util/rpkgutil/rpkgutil.go
```diff
@@ -0,0 +1,20 @@
+package rpkgutil
```
mortent, line 1: This file is missing the license header.

internal/cmdrpkgcopy/command.go
natasha41575: changed to client-go! Would appreciate if someone could take a look at it carefully, as I'm not familiar with client-go.

internal/cmdrpkgcopy/command.go
```diff
@@ -111,22 +112,43 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
...
+		return nil, err
+	}
+
+	result := porchapi.PackageRevision{}
+	err = restClient.
+		Get().
+		Namespace(*r.cfg.Namespace).
+		Resource("packagerevisions").
+		Name(r.copy.Source.Name).
+		Do(context.Background()).
```
yuwenma, line 140: we want to use the same context here, suggest changing to
`r.ctx`

mortent:
Other than the small issues mentioned by @yuwenma, this looks good to me.

internal/cmdrpkgcopy/command.go
natasha41575: done! Thanks

yuwenma (approved):
looking forward to the revision PR

PR #3133, opened 2022-05-11
support podman as a runtime for kpt CLI

mengqiy:
Introduced a new flag `--runtime` in `kpt fn render|eval`. It currently support 2 values: docker and podman.
There is a plan to add a built-in container runtime in kpt CLI in the near future. We can add new value under this flag when we want to do it.


Makefile
```diff
@@ -74,15 +74,24 @@ test:
 
 # This target is used to run Go tests that require docker runtime.
 # Some tests, like pipeline tests, need to have docker available to run.
+# It by default tests all the available runtimes. If you only want to test
+# against one specific runtime, you can set environment variable KPT_E2E_FN_RUNTIME.
+# e.g. KPT_E2E_FN_RUNTIME=docker
```
droot, line 79: will the tests complain if a runtime is not available on the machine ? (like on my mac, I don't have podman).

Thinking from a kpt developers perspective:
- in my dev setup, I would like to run all tests with one fn-runtime (sensible defaults for all platforms, but should be override with a personal setting in Makefile that can be preserved without checking in)
- in CI, tests should run for all the supported platforms
- Bonus would be: since these tests take the longest time, should run only if the change affects runtime

e2e/fn_test.go
```diff
@@ -18,12 +18,21 @@
...
+	"github.com/GoogleContainerTools/kpt/internal/fnruntime"
+	"github.com/GoogleContainerTools/kpt/internal/util/cmdutil"
 	"github.com/GoogleContainerTools/kpt/pkg/test/runner"
 )
 
+// This environment is for testing only
+const fnRuntimeEnv = "KPT_E2E_FN_RUNTIME"
+
+var availableFnRuntimes = []fnruntime.ContainerRuntime{fnruntime.DockerRuntime, fnruntime.PodmanRuntime}
+
```
droot, line 35: bikeshedding :)

fnruntime.Docker, fnruntime.Podman reads better :)

e2e/fn_test.go
```diff
@@ -51,26 +60,34 @@ func runAllTests(t *testing.T, path string) {
...
-			if err != nil {
-				t.Fatalf("failed to create test runner: %s", err)
-			}
-			if r.Skip() {
-				t.Skip()
-			}
-			err = r.Run()
-			if err != nil {
-				t.Fatalf("failed when running test: %s", err)
+	runtime := strings.ToLower(os.Getenv(fnRuntimeEnv))
```
droot, line 63: may be the environment variable should take list of runtimes (comma separated) instead of just one.

I can override the list if I want to as developer (and test runner can construct one by checking what all runtimes are available by default).

e2e/fn_test.go
```diff
@@ -51,26 +60,34 @@ func runAllTests(t *testing.T, path string) {
...
+	runtimesToTest := availableFnRuntimes
+	if runtime != "" {
+		runtimesToTest = []fnruntime.ContainerRuntime{cmdutil.StringToContainerRuntime(runtime)}
+	}
+	t.Logf("runtimes to test: %v", runtimesToTest)
+	for i := range runtimesToTest {
+		for _, c := range *cases {
+			c := c // capture range variable
+			if c.Config.Sequential != sequential {
+				continue
```
droot, line 73: Couldn't understand why is it continuing ?

e2e/testdata/fn-eval/image-pull-policy-never/.expected/config.yaml
```diff
@@ -16,4 +16,6 @@ testType: eval
 image: gcr.io/kpt-fn/not-exist:v0.1
 imagePullPolicy: Never
 exitCode: 1
-stdErr: "No such image"
+# The output in stderr is different depending on the runtime.
+# "No such image" is from docker and "image not known" is from podman.
+stdErrRegEx: No such image|image not known
```
droot, line 21: hmm... thinking if fnrunner should encapsulate the common error type like this.

e2e/testdata/fn-eval/save-fn/custom-pkg-path/.expected/diff.patch
```diff
@@ -12,23 +12,16 @@ index d9e2f05..93c547b 100644
...
+index 384a7ad..670db5a 100644
 --- a/sub-pkg/fn-config.yaml
 +++ b/sub-pkg/fn-config.yaml
-@@ -11,10 +11,10 @@
- # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- # See the License for the specific language governing permissions and
- # limitations under the License.
--
- apiVersion: v1
+@@ -15,5 +15,6 @@ apiVersion: v1
```
droot, line 18: Thanks for cleaning this up.

internal/cmdrender/cmdrender.go
```diff
@@ -50,6 +50,8 @@ func NewRunner(ctx context.Context, parent string) *Runner {
 		fmt.Sprintf("output resources are written to provided location. Allowed values: %s|%s|<OUT_DIR_PATH>", cmdutil.Stdout, cmdutil.Unwrap))
 	c.Flags().StringVar(&r.imagePullPolicy, "image-pull-policy", string(fnruntime.IfNotPresentPull),
 		fmt.Sprintf("pull image before running the container. It must be one of %s, %s and %s.", fnruntime.AlwaysPull, fnruntime.IfNotPresentPull, fnruntime.NeverPull))
+	c.Flags().StringVar(&r.runtime, "runtime", string(fnruntime.DockerRuntime),
```
droot, line 53: I am still thinking if we should use environment variable to control this. One advantage I see is: package users will always use `kpt fn render` without worrying about the runtimes and platform team can control the function runner by changing the environment variable without changing the `kpt fn` commands. It sort of decouples the team managing the runtime environments.

internal/fnruntime/container.go
```diff
@@ -112,7 +132,36 @@ func (f *ContainerFn) Run(reader io.Reader, writer io.Writer) error {
...
+			}
+		}
+		return fmt.Errorf("unexpected function error: %w", err)
+	}
+
+	if errSink.Len() > 0 {
+		f.FnResult.Stderr = filterPodmanCLIOutput(&errSink)
+	}
+	return nil
+}
```
droot, line 160: We are repeating this code for both. Wondering if there is a better way to structure this code.

internal/fnruntime/container.go
```diff
@@ -254,3 +303,44 @@ func isdockerCLIoutput(s string) bool {
...
+		strings.Contains(s, "Getting image source signatures") ||
+		strings.Contains(s, "Copying blob sha256:") ||
+		strings.Contains(s, "Copying config sha256:") ||
+		strings.Contains(s, "Writing manifest to image destination") ||
+		strings.Contains(s, "Storing signatures") ||
+		sha256Matcher.MatchString(s) {
+		return true
+	}
+	return false
+}
```
droot, line 346: Haven't thought a lot, but there is scope for reducing the duplication here.

droot (commented):
Awesome! I have some comments.

internal/cmdrender/cmdrender.go
mengqiy: I agree with your reasoning about env var is a better solution here.

I just realized that completely relying env var makes parallel testing impossible in golang. See: https://pkg.go.dev/testing#T.Setenv and https://stackoverflow.com/questions/58356031/how-to-run-test-in-parallel-requiring-different-values-for-env-variables

I see 2 options:
- Make all e2e test sequential. E2e tests will take longer to run.
- Support both env var and flag. And we use flag in the e2e tests.

@droot WDYT?

droot: That's a good point.

I think supporting both env var/flag could get confusing for the users, so going with an explicit command-line flag is probably ok. So, let's keep it as is then.

e2e/fn_test.go
mengqiy: Changed it to using env var

internal/fnruntime/container.go
mengqiy: Updated

mengqiy:
I updated it to use env var to configure runtime and tweaked he e2e tests to make it work and still run in parallel.

mengqiy:
@droot PTAL

Makefile
mengqiy: Created https://github.com/GoogleContainerTools/kpt/pull/3194

mengqiy:
Rebased to resolve merge conflict

.github/workflows/go.yml
```diff
@@ -19,8 +19,11 @@ on:
 
 jobs:
   build:
-    name: Build
+    name: Build-test-kpt-CLI
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        runtime: [docker, podman]
```
droot, line 26: sweet!

e2e/testdata/fn-eval/image-pull-policy-always/.expected/setup.sh
```diff
@@ -19,5 +19,5 @@ set -eo pipefail
 # function gcr.io/kpt-fn-demo/bar:v0.1 prints "bar" to stderr.
 # We intentionally tag a wrong image as gcr.io/kpt-fn-demo/foo:v0.1, since we
 # expect the correct image to be pulled and override the wrong image.
-docker pull gcr.io/kpt-fn-demo/bar:v0.1
-docker tag gcr.io/kpt-fn-demo/bar:v0.1 gcr.io/kpt-fn-demo/foo:v0.1
+${KPT_FN_RUNTIME} pull gcr.io/kpt-fn-demo/bar:v0.1
```
droot, line 22: is this always present ? or it needs to be defaulted.

e2e/testdata/fn-eval/image-pull-policy-if-not-present/.expected/setup.sh
```diff
@@ -18,5 +18,5 @@ set -eo pipefail
 # Function gcr.io/kpt-fn-demo/foo:v0.1 prints "foo" to stderr and
 # function gcr.io/kpt-fn-demo/bar:v0.1 prints "bar" to stderr.
 # We intentionally tag a wrong image as pull gcr.io/kpt-fn-demo/bar:v0.1
-docker pull gcr.io/kpt-fn-demo/foo:v0.1
-docker tag gcr.io/kpt-fn-demo/foo:v0.1 gcr.io/kpt-fn-demo/bar:v0.1
+${KPT_FN_RUNTIME} pull gcr.io/kpt-fn-demo/foo:v0.1
```
droot, line 21: ideally we should have a test where this is not specified and default runtime is used.

droot (changes requested):
Looks solid except a minor comment about testing the default runtime without specifying the environment variable.

mengqiy:
Updated. @droot PTAL

droot (approved):

droot:
@mengqiy Looking good. Can you pl. ensure that we are covering a test-case where `kpt fn render` uses docker if environment variable `KPT_FN_RUNTIME` isn't specified.

mengqiy:
Added one for render and one for eval

droot (approved):
Excellent!

Also, let's add documentation about supported function-runtime somewhere on `kpt.dev`.

PR #3154, opened 2022-05-13
feat: cache upstream during update

bharathkkb:
fixes https://github.com/GoogleContainerTools/kpt/issues/2988

This introduces caching for upstream repos so we can skip some git exec commands like `ls-remote` and `fetch` for repos that are already upto date. Looking for some early feedback on the approach and happy to add additional tests/benchmarks. Currently packages are not updated in parallel so I haven't investigated a thread safe approach.

### DIFF
--- a/internal/gitutil/gitutil.go
+++ b/internal/gitutil/gitutil.go
@@ -396,23 +396,25 @@ loop:
 		_, resolved := gur.ResolveRef(s)
 		// fetchedRefWithURI is the cache key created from repo URI SHA and ref
 		fetchedRefWithURI := fmt.Sprintf("%s-%s", uriSha, s)
+		// check if ref was previously fetched
 		_, fetched := gur.fetchedRefs[fetchedRefWithURI]
 		switch {
+		case fetched:
+			// skip refetching if previously fetched
+			break
 		case resolved || validFullSha:
 			// If the ref references a branch or a tag, or is a valid commit
 			// sha and has not already been fetched, we can fetch just a single commit.
-			if !fetched {
-				if _, err := gitRunner.RunVerbose(ctx, "fetch", "origin", "--depth=1", s); err != nil {
-					AmendGitExecError(err, func(e *GitExecError) {
-						e.Repo = uri
-						e.Command = "fetch"
-						e.Ref = s
-					})
-					return "", errors.E(op, errors.Git, fmt.Errorf(
-						"error running `git fetch` for ref %q: %w", s, err))
-				}
-				gur.fetchedRefs[fetchedRefWithURI] = true
+			if _, err := gitRunner.RunVerbose(ctx, "fetch", "origin", "--depth=1", s); err != nil {
+				AmendGitExecError(err, func(e *GitExecError) {
+					e.Repo = uri
+					e.Command = "fetch"
+					e.Ref = s
+				})
+				return "", errors.E(op, errors.Git, fmt.Errorf(
+					"error running `git fetch` for ref %q: %w", s, err))
 			}
+			gur.fetchedRefs[fetchedRefWithURI] = true
 		default:
 			// In other situations (like a short commit sha), we have to do
 			// a full fetch from the remote.
@@ -432,6 +434,7 @@ loop:
 				return "", errors.E(op, errors.Git, fmt.Errorf(
 					"error verifying results from fetch: %w", err))
 			}
+			gur.fetchedRefs[fetchedRefWithURI] = true
 			// If we did a full fetch, we already have all refs, so we can just
 			// exit the loop.
 			break loop
--- a/internal/gitutil/gitutil.go
+++ b/internal/gitutil/gitutil.go
@@ -137,12 +137,25 @@ func (g *GitLocalRunner) run(ctx context.Context, verbose bool, command string, 
 	}, nil
 }
 
-// NewGitUpstreamRepo returns a new GitUpstreamRepo for an upstream package.
-func NewGitUpstreamRepo(ctx context.Context, uri string) (*GitUpstreamRepo, error) {
-	const op errors.Op = "gitutil.NewGitUpstreamRepo"
+type NewGitUpstreamRepoOption func(*GitUpstreamRepo)
 
+func WithFetchedRefs(a map[string]bool) NewGitUpstreamRepoOption {
+	return func(g *GitUpstreamRepo) {
+		g.fetchedRefs = a
+	}
+}
+
+// NewGitUpstreamRepo returns a new GitUpstreamRepo for an upstream package.
+func NewGitUpstreamRepo(ctx context.Context, uri string, opts ...NewGitUpstreamRepoOption) (*GitUpstreamRepo, error) {
+	const op errors.Op = "gitutil.NewGitUpstreamRepo"
 	g := &GitUpstreamRepo{
 		URI: uri,
+	}
+	for _, opt := range opts {
+		opt(g)
+	}
+	if g.fetchedRefs == nil {
+		g.fetchedRefs = map[string]bool{}
 	}
 	if err := g.updateRefs(ctx); err != nil {
 		return nil, errors.E(op, errors.Repo(uri), err)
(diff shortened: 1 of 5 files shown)


bharathkkb:
@droot can you PTAL?

droot:
sorry @bharathkkb for the delay. Will take a look at it today.

internal/gitutil/gitutil.go
```diff
@@ -378,19 +394,24 @@ loop:
...
+				if _, err := gitRunner.RunVerbose(ctx, "fetch", "origin", "--depth=1", s); err != nil {
+					AmendGitExecError(err, func(e *GitExecError) {
+						e.Repo = uri
+						e.Command = "fetch"
+						e.Ref = s
+					})
+					return "", errors.E(op, errors.Git, fmt.Errorf(
+						"error running `git fetch` for ref %q: %w", s, err))
+				}
+				gur.fetchedRefs[fetchedRefWithURI] = true
```
droot, line 414: Do we need to update it in case of `default` case of this switch statement (below) ?

internal/util/fetch/fetch.go
```diff
@@ -94,21 +94,62 @@ func (c Command) validate(kf *kptfilev1.KptFile) error {
 	return nil
 }
 
+type Cloner struct {
```
droot, line 97: Would be good to add some context or background thinking how this new struct will be used and especially any deviation from current workflow.

internal/util/fetch/fetch.go
```diff
@@ -94,21 +94,62 @@ func (c Command) validate(kf *kptfilev1.KptFile) error {
...
+	// repoSpec spec to clone
+	repoSpec *git.RepoSpec
+
+	// cachedRepos
+	cachedRepo map[git.RepoSpec]*gitutil.GitUpstreamRepo
+
+	// fetchedRefs keeps track of refs already fetched from remote
+	fetchedRefs map[string]bool
```
droot, line 105: cachedRepo above also has `fetchedRefs` stored ...right ?

internal/util/fetch/fetch.go
```diff
@@ -94,21 +94,62 @@ func (c Command) validate(kf *kptfilev1.KptFile) error {
...
 	const op errors.Op = "fetch.cloneAndCopy"
 	pr := printer.FromContextOrDie(ctx)
 
-	err := ClonerUsingGitExec(ctx, r)
+	err := c.ClonerUsingGitExec(ctx)
 	if err != nil {
 		return errors.E(op, errors.Git, types.UniquePath(dest), err)
 	}
-	defer os.RemoveAll(r.Dir)
+	defer os.RemoveAll(c.repoSpec.Dir)
```
droot, line 149: what does this means wrt caching (do the fetchedRefs need to be cleaned up ?)

internal/util/fetch/fetch.go
```diff
@@ -129,46 +170,51 @@ func cloneAndCopy(ctx context.Context, r *git.RepoSpec, dest string) error {
...
 
 	// Create a local representation of the upstream repo. This will initialize
 	// the cache for the specified repo uri if it isn't already there. It also
 	// fetches and caches all tag and branch refs from the upstream repo.
-	upstreamRepo, err := gitutil.NewGitUpstreamRepo(ctx, repoSpec.CloneSpec())
-	if err != nil {
-		return errors.E(op, errors.Git, errors.Repo(repoSpec.CloneSpec()), err)
+	upstreamRepo, exists := c.cachedRepo[*c.repoSpec]
+	if !exists {
+		newUpstreamRemp, err := gitutil.NewGitUpstreamRepo(ctx, c.repoSpec.CloneSpec(), gitutil.WithFetchedRefs(c.fetchedRefs))
```
droot, line 181: still trying to parse this. So we are creating a new repo cache because the repo doesn't exists but we are initializing with fetchedRefs (which meant we fetched the repo before ?) ?

internal/util/fetch/fetch.go
```diff
@@ -129,46 +170,51 @@ func cloneAndCopy(ctx context.Context, r *git.RepoSpec, dest string) error {
...
 	gitRunner, err := gitutil.NewLocalGitRunner(dir)
 	if err != nil {
-		return errors.E(op, errors.Git, errors.Repo(repoSpec.CloneSpec()), err)
+		return errors.E(op, errors.Git, errors.Repo(c.repoSpec.CloneSpec()), err)
 	}
 
 	// Find the commit SHA for the ref that was just fetched. We need the SHA
 	// rather than the ref to be able to do a hard reset of the cache repo.
-	commit, found := upstreamRepo.ResolveRef(repoSpec.Ref)
+	commit, found := upstreamRepo.ResolveRef(c.repoSpec.Ref)
```
droot, line 215: so is caching being realized at this step ?

droot (commented):
Hey, I will need help to understand it fully. A quick walkthrough is needed :), lets block 30 mins to go over this.

Thank you for working on this.

internal/gitutil/gitutil.go
bharathkkb: Yeah we can expand caching to default case since we don't need to rerun after initial fetch.

internal/util/fetch/fetch.go
bharathkkb: Added some context

internal/util/fetch/fetch.go
bharathkkb: Right, I have removed this.

internal/util/fetch/fetch.go
bharathkkb: Missed this, we should clean up

internal/util/fetch/fetch.go
bharathkkb: Yeah we don't need fetch refs here. It was left over from my debugging, updated now.

internal/gitutil/gitutil.go
```diff
@@ -378,20 +394,27 @@ loop:
 		// commit sha.
 		validFullSha := s == strings.TrimSpace(rr.Stdout)
 		_, resolved := gur.ResolveRef(s)
-
+		// fetchedRefWithURI is the cache key created from repo URI SHA and ref
+		fetchedRefWithURI := fmt.Sprintf("%s-%s", uriSha, s)
```
droot, line 398: we should probably reverse these `ref-SHA` (`main-sldksls`, `main-hkdldl`)

droot: URI SHA is confusing with the `git SHA`. May be let's use `repoCacheKey` term for `URI SHA` and fetchedRefWithURI can be renamed to `refCacheKey`.

droot (commented):
Will take a detailed look at it.

internal/gitutil/gitutil.go
bharathkkb: @droot based on our conversation i did some digging. https://github.com/GoogleContainerTools/kpt/blob/72ef5b29cb372ed1d28e1380782651f366e2d14f/internal/gitutil/gitutil.go#L301-L307
The current `uriSHA` was an md5 of GitUpstreamRepo.URI so it's the same for each ref. I have refactored to just use the ref as the cache key instead.

bharathkkb:
@droot added some tests like we discussed.
I did some rough benchmarking updating https://github.com/bcongdon/many-subpackages-blueprint-example
- without fix 0m52.085s is spent updating 30 pkgs from same upstream before hanging
- with fix 0m4.533s is spent updating all 102 pkgs from same upstream

So roughly ~ 95% decrease in update time for a pkg if cache is hit.

droot (approved):

droot:
> @droot added some tests like we discussed. I did some rough benchmarking updating https://github.com/bcongdon/many-subpackages-blueprint-example
> 
> * without fix 0m52.085s is spent updating 30 pkgs from same upstream before hanging
> * with fix 0m4.533s is spent updating all 102 pkgs from same upstream
> 
> So roughly ~ 95% decrease in update time for a pkg if cache is hit.

Thank you @bharathkkb . This is great.

PR #3165, opened 2022-05-13
align video in kpt.dev home page

yuwenma:
Super nit, the current video alignment squeezes the "Get Started" and "Why kpt" and makes it feel a little bit crowded in the centre.

Before
Screen Shot 2022-05-13 at 12 49 25
After
Screen Shot 2022-05-13 at 12 49 10

### DIFF
--- a/site/coverpage.md
+++ b/site/coverpage.md
@@ -6,8 +6,8 @@
 > by manipulating declarative Configuration as Data, 
 > separated from the code that transforms it.
 
-<iframe id="ytplayer" type="text/html" width="550" height="360"
-    src="https://www.youtube.com/embed/L_x7z4CXHDw" allowFullScreen="allowFullScreen"> </iframe>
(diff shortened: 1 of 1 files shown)


bgrant0607:
I agree it looks crowded. I haven't looked at it on mobile at all yet, but I agree that on a laptop/desktop some things could go side by side. I thought about putting the logo next to the text instead of above it.

bgrant0607 (approved):
This is an improvement, so approving, and rerunning failed build.

droot:
> This is an improvement, so approving, and rerunning failed build.

checksite github action seems to be causing more trouble than helping. @mengqiy is looking into it. @mengqiy let's disable this till we find a fix.

Re: alignment
I was trying to play around but docsify framework doesn't provide much options (I think we will have to replace `coverpage.md` with index.html page to get more flexibility on layout). About this change, I agree this is an improvement, so let's get this in.

droot (approved):

PR #3166, opened 2022-05-13
chore: update kind to v0.13.0

karlkfi:
- Update e2e tests to support k8s v1.21 - v1.24
- Update default k8s test version to v1.21 (latest GKE stable)
- Update test workflows to all consistently use kind v0.13.0
- Update e2e tests to the latest k8s minor releases for each major release
- Disable e2e fast-fail to determine when errors are version specific
- Fix e2e test failure specific to K8s 1.24:
  ServiceAccounts no longer immediately generate an auth token secret.
- Use tee to print and capture output in e2e tests, to aid debugging.


karlkfi:
Figured out why this is erroring in K8s 1.24.

From https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md:
> The LegacyServiceAccountTokenNoAutoGeneration feature gate is beta, and enabled by default. When enabled, Secret API objects containing service account tokens are no longer auto-generated for every ServiceAccount. Use the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API to acquire service account tokens, or if a non-expiring token is required, create a Secret API object for the token controller to populate with a service account token by following this [guide](https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets). (https://github.com/kubernetes/kubernetes/pull/108309, [@zshihang](https://github.com/zshihang))

So the tests that switch to use a ServiceAccount with reduced permissions isn't working, because the SA doesn't have an auto-generated secret.

karlkfi:
Fixed by manually creating a service account secret:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-service-account-api-token

If this were production, an ephemeral token might be better, but for testing on a disposable kind cluster... it's not a big deal. And it's less change than using the TokenRequest API.

.github/workflows/e2eEnvironment.yml
```diff
@@ -27,8 +27,9 @@ jobs:
   kind:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        version: [1.19, 1.20]
+        version: ["1.19", "1.20", "1.21", "1.22", "1.23", "1.24"]
```
mortent, line 32: I think we can remove some of the older version here. 1.21 is the oldest version available for new GKE clusters, so I think we can remove 1.19 and 1.20.

mortent (approved):

.github/workflows/e2eEnvironment.yml
karlkfi: It’s not a speed issue since they run in parallel. So the only reason to limit it is cost, but I’m not sure if we’re paying GitHub for it or getting free actions for being open source.

droot:
> Figured out why this is erroring in K8s 1.24.
> 
> From https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md:
> 
> > The LegacyServiceAccountTokenNoAutoGeneration feature gate is beta, and enabled by default. When enabled, Secret API objects containing service account tokens are no longer auto-generated for every ServiceAccount. Use the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API to acquire service account tokens, or if a non-expiring token is required, create a Secret API object for the token controller to populate with a service account token by following this [guide](https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets). ([kubernetes/kubernetes#108309](https://github.com/kubernetes/kubernetes/pull/108309), [@zshihang](https://github.com/zshihang))
> 
> So the tests that switch to use a ServiceAccount with reduced permissions isn't working, because the SA doesn't have an auto-generated secret.

Sweet! Thanks for sharing these details. Resolves the mystery :)

.github/workflows/e2eEnvironment.yml
droot: Yeah, we don't have any policy defined around the support k8s versions. We can keep them for now and adjust it in a separate PR by looking at k8s OSS and GKE version policy.

droot (approved):

liggitt:
FWIW, the secrets field in the service account object is *only* for listing secrets to mount into pods. There was never a guarantee the first item in the list would be a service account token secret. Opened https://github.com/GoogleContainerTools/kpt/pull/3199 to unconditionally use the mechanism of declaring a secret for the purposes of obtaining a token for use outside a pod.

PR #3172, opened 2022-05-14
Use external resourcegroup object file to store inventory

mortent:
Replacement for https://github.com/GoogleContainerTools/kpt/pull/2802. This has the following changes:
* Updates the `kpt live init` command to generate RG resource with inventory information instead of putting it in the Kptfile.
* The `apply`, `destroy` commands look for inventory information in both the Kptfile and any RG resources (root package and in a stream).
* Validation that the inventory resource (Kptfile or RG resource) has an non-empty inventoryID.
* The `kpt live init` commands accepts a `--rg-filename` flag that allows users to customize the name of the file with the ResourceGroup resource.
* `kpt live init` and `kpt live migrate` commands' Run logic updated to remove feature gating
* Updated integration tests to pass
* Updated e2e tests to pass
* Added new e2e test cases to test the new feature (init, apply, destroy and migrate)


internal/cmdliveinit/cmdliveinit.go
```diff
@@ -80,9 +81,10 @@ func NewRunner(ctx context.Context, factory k8scmdutil.Factory,
 	r.Command = cmd
 
 	cmd.Flags().StringVar(&r.Name, "name", "", "Inventory object name")
-	cmd.Flags().BoolVar(&r.Force, "force", false, "Set inventory values even if already set in Kptfile")
+	cmd.Flags().BoolVar(&r.Force, "force", false, "Set inventory values even if already set in Kptfile or ResourceGroup file")
 	cmd.Flags().BoolVar(&r.Quiet, "quiet", false, "If true, do not print output message for initialization")
 	cmd.Flags().StringVar(&r.InventoryID, "inventory-id", "", "Inventory id for the package")
+	cmd.Flags().StringVar(&r.RGFileName, "rg-filename", rgfilev1alpha1.RGFileName, "Name of the file holding the ResourceGroup resource.")
```
yuwenma, line 87: nit: Shall we name this to `rg-file` to align with other flag names like `log-file`?

internal/cmdliveinit/cmdliveinit.go
```diff
@@ -137,7 +146,7 @@ func (r *Runner) runE(_ *cobra.Command, args []string) error {
 		Quiet:       r.Quiet,
 		Name:        r.Name,
 		InventoryID: r.InventoryID,
-		RGFileName:  r.RGFile,
+		RGFileName:  r.RGFileName,
```
yuwenma, line 149: 👍 good catch

pkg/live/load.go
```diff
@@ -154,16 +113,25 @@ func readInvInfoFromStream(in io.Reader) (kptfilev1.Inventory, error) {
...
+		return kptfilev1.Inventory{}, &pkg.MultipleKfInv{}
+	}
+	// Multiple ResourceGroup inventories found.
+	if len(rgFilter.Inventories) > 1 {
+		return kptfilev1.Inventory{}, &pkg.MultipleResourceGroupsError{}
+	}
+	// Multiple inventories found in Kptfile and ResourceGroup objects.
+	if len(invFilter.Inventories) > 0 && len(rgFilter.Inventories) > 0 {
+		return kptfilev1.Inventory{}, &pkg.MultipleInventoryInfoError{}
 	}
```
yuwenma, line 128: I'm thinking that after separating inventory from kptfile, the kptfile is no longer required by `kpt live apply` so the content of Kptfile is no longer needed to be verified. 

Also, verifying the Kptfile and RG is complicated: e.g. an inventory error requires migration plans, if more than one inventory or/and RG are detected, it's hard for both `kpt live migrate` and users to make good decisions to recover. So I think an easier approach would be: 
- `kpt live apply` ignores what Kptfile says (this also resolves the problems that nested Kptfile files may have different inventory)
- resourceGroup is a meta resource, but it is passed into `kpt live` as a normal KRM resource.  One and only one RG is required in `kpt live apply`

mortent: So even with this change we want to let users continue to use inventory from the Kptfile if they want to. We don't want to make this a change that requires users to do the migration in order to do a successful apply. For this particular check, I think we should fail if we find inventory information in both the Kptfile and the resourcegroup file. There isn't any good way to know which one if the correct one, so we need to user to clean it up.

internal/cmdliveinit/cmdliveinit.go
mortent: Fixed the name. I'm not sure if we need to point out the format of the file since the file is being generated by kpt. The user doesn't have to create the file themselves.

pkg/live/load.go
yuwenma: The feedback I got from skaffold (when introducing the inventory and have inventory-id conflict) is that cleaning up the duplicate inventory is challenging for users because 1. kpt inits the inventory by default, so users may lack knowledge about how RG is created and what it is used for. 2. users do not know what's the current in-use RG is and it is hard to find it out from the cluster. 3. If user force a new RG, they do not know how to generate a new one (is it random string or has to much some specific format). 
I'd propose make users life easier by giving some direct suggestions like "delete the inventory in Kptfile" or "use --force to override" etc.

yuwenma (approved):
One comment about real user feedback. Not blockers.

pkg/live/load.go
mortent: Yeah, agree the error messages could be better. We can provide custom error handlers to provide more information. I'll handle that in a follow-up since this is already a pretty massive PR.

Issue #3178, opened 2022-05-14
Fix site landing page for mobile


bgrant0607:
#3165 put the "Why kpt" and "Get started" buttons next to the embedded video. It looks better on my laptop.

However, on my Android phone, the buttons are squeezed off the page, with just "Why" and "Get" appearing at the edge.

We probably need to play with the layout more. For example:

K          text
buttons video

or

K  buttons
text
video

cc @yuwenma

yuwenma:
Approved on https://responsivedesignchecker.com/checker.php?url=https%3A%2F%2Fkpt.dev&width=375&height=667

The "Why kpt" and "Get Started" should fit into all listed computer, tablet and phones. 
This is the Android pixel look

 
<img width="261" alt="Screen Shot 2022-05-14 at 17 18 21" src="https://user-images.githubusercontent.com/5606098/168452268-82864c00-653d-4ef0-9b55-e3545d02ef7d.png">

PR #3179, opened 2022-05-14
fix: Adjust kpt.dev homepage video size to fit mobile display

yuwenma:
This shall improve the Mobile display problem as described in  #3178 

To better improve the responsive design of kpt.dev, we can also consider using [iframe-resize](https://github.com/davidjbradshaw/iframe-resizer), which may gives a better display, but requires much more work and may need some Frontend expertise to help.

Before
Screen Shot 2022-05-14 at 16 48 55
 
After
Screen Shot 2022-05-14 at 17 00 00


mortent (approved):

yuwenma:
thank you @mortent !

PR #3182, opened 2022-05-16
Simplify Plan resource

mortent:
Follow up on some of the comments from https://github.com/GoogleContainerTools/kpt/pull/3123:

* The property containing the resource from before the apply operation has been renamed from `before` to `original`, and the one containing the resource after the apply (i.e. the resource returned by the APIServer on a dry-run) is renamed from `after` to `updated`.
* The `original` and `updated` properties no longer contains a nested `object` property (as a result of marshaling an Unstructured object), but contains the resources directly underneath.
* The KRM output is generated directly with kyaml and we no longer provide an exported `Plan` type.


internal/cmdplan/command.go
```diff
@@ -265,24 +264,82 @@ func printWithPrefix(text, prefix string, ioStreams genericclioptions.IOStreams)
...
 // printKRM outputs the plan inside a ResourceList so the output format
 // follows the KRM function wire format.
-func printKRM(plan *v1alpha1.Plan, ioStreams genericclioptions.IOStreams) error {
-	b, err := yaml.Marshal(plan)
+func printKRM(
+	plan *kptplanner.Plan,
+	ioStreams genericclioptions.IOStreams,
+) error {
+	planResource, err := yaml.Parse(strings.TrimSpace(`
+apiVersion: config.google.com/v1alpha1
```
droot, line 272: Should we use `kpt.dev` :)

mortent: Agree. Updated to use `kpt.dev/v1alpha1`.

internal/cmdplan/command.go
```diff
@@ -265,24 +264,82 @@ func printWithPrefix(text, prefix string, ioStreams genericclioptions.IOStreams)
...
+  annotations:
+    config.kubernetes.io/local-config: true	
+`))
 	if err != nil {
-		return fmt.Errorf("failed to marshall plan into yaml: %w", err)
+		return fmt.Errorf("unable to create yaml document: %w", err)
 	}
-	rn, err := yaml.Parse(string(b))
+
+	sNode, err := planResource.Pipe(yaml.LookupCreate(yaml.SequenceNode, "spec", "actions"))
```
justinsb, line 283: Q: Why not use "normal" marshalling?  Is it because we want to preserve comments in the yaml?  Can we add a (code) comment?

justinsb (approved):
One question about why we're not doing normal marshalling, but LGTM

PR #3187, opened 2022-05-17
docs: updated meta resources related behavior in kpt fn docs

droot:
`meta` resources are no longer treated in any special way across `kpt` workflow.  And `meta resources` are now included by default in `kpt fn` commands.

Updated the docs to reflect this new behavior.

### DIFF
--- a/site/reference/cli/fn/eval/README.md
+++ b/site/reference/cli/fn/eval/README.md
@@ -104,7 +104,7 @@ fn-args:
 
 --include-meta-resources, m:
   (DEPRECATED) include-meta-resources is no longer necessary because meta
-  resources are now included by default.
+  resources are included by default with kpt version v1.0.0-beta.15+.
 
 --mount:
   List of storage options to enable reading from the local filesytem. By default,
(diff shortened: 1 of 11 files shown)


droot:
@natasha41575 Can you pl. take a look at this ?

site/reference/cli/fn/eval/README.md
```diff
@@ -103,8 +103,8 @@ fn-args:
   If using never, kpt will only use images from the local cache.
 
 --include-meta-resources, m:
-  If enabled, meta resources (i.e. `Kptfile` and `functionConfig`) are included
-  in the input to the function. By default it is disabled.
+  (DEPRECATED) include-meta-resources is no longer necessary because meta
+  resources are now included by default.
```
natasha41575, line 107: Should we list what version they started being included by default? Here and below

natasha41575 (approved):
minor question. Other than that, LGTM

site/reference/cli/fn/eval/README.md
droot: Good suggestion. included the version now. And also improved the `kpt pkg get` reference to include `--for-deployment`.

Thanks for the review.

PR #3194, opened 2022-05-18
Don't trigger kpt CLI e2e test when it's doc only changes

mengqiy:
Make kpt CLI e2e test CI ignores some doc related directories.

Related discussion: https://github.com/GoogleContainerTools/kpt/pull/3133#discussion_r871673713

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -15,6 +15,16 @@
 name: Go
 on:
   pull_request:
+    paths-ignore:
+      - "Formula/**"
+      - "demos/**"
+      - "docs/**"
+      - "firebase/**"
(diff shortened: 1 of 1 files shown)


mengqiy:
@droot  It seems one of the admins need to mark `Build` as optional. Otherwise, it is skipped by GHA but still blocks merging.

mengqiy:
@droot PTAL

droot:
> @droot It seems one of the admins need to mark `Build` as optional. Otherwise, it is skipped by GHA but still blocks merging.

I took a quick look and couldn't find a way to do it. Do you happen to know -- where do I see that option ?

droot (approved):

PR #3195, opened 2022-05-19
preserve comments in `kpt fn eval --save`

natasha41575:
There is some existing code in kyaml to preserve comments and field order. This PR utilizes it for preserving those in `kpt fn eval --save`.

Fixes https://github.com/GoogleContainerTools/kpt/issues/2952

### DIFF
--- a/thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
+++ b/thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
@@ -25,9 +25,11 @@ import (
 	"github.com/GoogleContainerTools/kpt/thirdparty/kyaml/runfn"
 	"github.com/google/shlex"
 	"github.com/spf13/cobra"
+	"sigs.k8s.io/kustomize/kyaml/comments"
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 	"sigs.k8s.io/kustomize/kyaml/fn/runtime/runtimeutil"
+	"sigs.k8s.io/kustomize/kyaml/order"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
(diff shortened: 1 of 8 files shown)


natasha41575:
cc @yuwenma @droot

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -255,16 +258,47 @@ func (r *EvalFnRunner) SaveFnToKptfile() {
...
+	if err != nil {
+		return nil, fmt.Errorf("could not parse Kptfile: %v", err)
+	}
+	// preserve comments and sync field order
+	if err := comments.CopyComments(kfAsRNode, mutatedKfAsRNode); err != nil {
+		return nil, fmt.Errorf("could not preserve Kptfile comments: %v", err)
+	}
+	if err := order.SyncOrder(kfAsRNode, mutatedKfAsRNode); err != nil {
+		return nil, fmt.Errorf("could not preserve Kptfile field order %v", err)
+	}
```
yuwenma, line 298: Cool for SyncOrder.
In the cases that 1. we append a new function to mutatedKfAsRNode 2. replace a function with different args in  mutatedKfAsRNode, what order change shall we expect?

yuwenma (approved):
Sweet! LGTM with a minor question

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
natasha41575: In general what I've seen in kustomize is that the order of map nodes sometimes changes after going through encoding/decoding. I'm not sure if there will actually be any order changes in either case that you described, but I saw this SyncOrder in kyaml is used when RNodes are mutated, so thought it was best to include it here too.

PR #3199, opened 2022-05-20
Fix method of obtaining service account token

liggitt:
The .secrets field within service accounts is only for listing secrets
to be mounted into pods, not for extracting secrets for other uses.
There is no guarantee the first item in the list will be a token secret.

In 1.24+, this field is not populated by default.

This unconditionally makes the test script create a dedicated secret for obtaining a token

### DIFF
--- a/e2e/live/end-to-end-test.sh
+++ b/e2e/live/end-to-end-test.sh
@@ -1187,16 +1187,11 @@ wait 2
 # Setup: use the service account just created. It does not have permissions
 # on the default namespace, so it will give a permissions error on apply
 # for anything attempted to apply to the default namespace.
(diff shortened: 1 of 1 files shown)


liggitt:
cc @karlkfi

karlkfi:
Thanks. How far back does support for that Secret annotation go?

liggitt:
to 1.0. That's always been the way to obtain a token manually

liggitt:
For reference, see 1.1-era docs at https://github.com/kubernetes/kubernetes/blob/release-1.1/docs/user-guide/service-accounts.md#manually-create-a-service-account-api-token

droot:
Thanks for the PR @liggitt. This totally makes sense.

droot (approved):

PR #3203, opened 2022-05-20
Add OCI Lock Support

martinmaly:
Include OCI upstream lock in Kptfile schema.
Implement OCI lock support in OCI repository adapter.


martinmaly:
This is something I had in my repo from a while back, feel free to incorporate if it is useful, or discard if not. Just wanted to share it in case it is helpful.

PR #3207, opened 2022-05-21
Docs: Fix code sample in `Developing in Go`

dnguyenzd:
Fix code sample in [Developing in Go](https://kpt.dev/book/05-developing-functions/02-developing-in-Go?id=write-the-krm-function-logic)

api version for deployment is `apps/v1` instead of `v1`. Previous code doesn't match and doesn't add the annotation


dnguyenzd:
close for now, will need to confirm CLA from my company first

yuwenma:
Thank you for catching this @dnguyenzd. I'll update the example on behalf of you.

PR #3229, opened 2022-05-23
tag configMap data nodes as 'string'

natasha41575:
Fixes https://github.com/GoogleContainerTools/kpt/issues/3193

First commit moves some code around and adds a (broken) test demonstrating current behavior. Second commit provides the fix for https://github.com/GoogleContainerTools/kpt/issues/3193.

When underlying go-yaml sees non-string values (e.g. integers or booleans), they must be explicitly tagged as strings in order to be quoted and processed correctly by other yaml-processing code.

With the following input:

```
data := map[string]string{
    "normal string": "abc",
    "integer":       "8081",
    "float":         "1.23",
    "bool":          "true",
}
m, err := NewConfigMap(data)
fmt.Println(m.MustString())
```

Before this fix:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: function-input
data: {normal string: abc, integer: 8081, float: 1.23, bool: true} # this is invalid
```

After this fix:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: function-input
data: {normal string: abc, integer: "8081", float: "1.23", bool: "true"}
```

### DIFF
--- a/internal/fnruntime/runner.go
+++ b/internal/fnruntime/runner.go
@@ -508,19 +508,7 @@ func newFnConfig(fsys filesys.FileSystem, f *kptfilev1.Function, pkgPath types.U
 		// directly use the config from file
 		return node, nil
 	case len(f.ConfigMap) != 0:
-		node = yaml.NewMapRNode(&f.ConfigMap)
-		if node == nil {
-			return nil, nil
-		}
-		// create a ConfigMap only for configMap config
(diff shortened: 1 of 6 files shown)


natasha41575:
@droot PTAL

natasha41575:
@droot PTAL

droot (approved):

PR #3287, opened 2022-06-03
Update packages in porch

mortent:
This updates porch and the kpt cli for updates.

In order to update a package, the client needs to update the upstream reference in the clone task and update the package revision resource in porch. 
In this change, updating a package will lead to an additional clone task (followed by a render task) in the task history for the package. Thus, the task list for a package revision can contain multiple clone tasks. When updating a package, it is required that the last clone task in the package is updated, otherwise it is an error. I'm not sure if this is a good way to do this, but we probably do need some kind of task to associate with the commit corresponding to the update.

### DIFF
--- a/internal/cmdrpkgupdate/command.go
+++ b/internal/cmdrpkgupdate/command.go
@@ -1,0 +1,175 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmdrpkgupdate
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/GoogleContainerTools/kpt/internal/errors"
+	"github.com/GoogleContainerTools/kpt/internal/util/porch"
+	porchapi "github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	command = "cmdrpkgupdate"
+)
+
+func NewCommand(ctx context.Context, rcg *genericclioptions.ConfigFlags) *cobra.Command {
+	return newRunner(ctx, rcg).Command
+}
+
+func newRunner(ctx context.Context, rcg *genericclioptions.ConfigFlags) *runner {
+	r := &runner{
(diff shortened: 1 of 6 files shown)


internal/cmdrpkgupdate/command.go
```diff
@@ -0,0 +1,174 @@
...
+
+	if len(args) < 1 {
+		return errors.E(op, fmt.Errorf("SOURCE_PACKAGE is a required positional argument"))
+	}
+	if len(args) > 1 {
+		return errors.E(op, fmt.Errorf("too many arguments; SOURCE_PACKAGE is the only accepted positional arguments"))
+	}
+
+	if r.revision == "" {
+		return errors.E(op, fmt.Errorf("revision is required"))
```
natasha41575, line 78: Not needed for MVP, but could we consider defaulting to latest in the future?

internal/cmdrpkgupdate/command.go
```diff
@@ -0,0 +1,174 @@
...
+		Namespace: *r.cfg.Namespace,
+		Name:      args[0],
+	}, &pr)
+	if err != nil {
+		return errors.E(op, err)
+	}
+
+	cloneTask, found := r.findCloneTask(&pr)
+	if !found {
+		err := fmt.Errorf("no clonetask found in package revision")
```
natasha41575, line 98: I'm wondering how this will sound to users... are the tasks part of the api that users are aware of?

porch/pkg/engine/engine.go
```diff
@@ -455,13 +469,30 @@ func (m *updatePackageMutation) Apply(ctx context.Context, resources repository.
...
-	// TODO: This is a hack
-	packageName := filepath.Base(m.task.Clone.Upstream.Git.Directory)
-	packageName = strings.TrimPrefix(packageName, ".git")
+	var ref string
+	var packageDir string
+	upstream := m.task.Clone.Upstream
+	switch upstream.Type {
+	case api.RepositoryTypeGit:
+		ref = upstream.Git.Ref
+		packageName := filepath.Base(m.task.Clone.Upstream.Git.Directory)
```
natasha41575, line 478: Naive question: is it always true that the packageName is the same as the directory? Is that still true with subpackages?

internal/cmdrpkgupdate/command.go
mortent: Yeah, we should. I'll add a TODO on this.

internal/cmdrpkgupdate/command.go
droot: Yeah, this is more of an internal diagnostic error.

internal/cmdrpkgupdate/command.go
droot: Actually, now that I have invoked this command so many times :) I would love following behavior where when we invoke `kpt alpha rpkg update <pkg>` without specifying a version, it should: list the available versions for the upstream package with last published as selected by default and if I just press enter and it runs the update with the default selected.

ofc, not scoped to this PR.

porch/pkg/engine/engine.go
droot: Just a note that this code goes away with my PR. My PR ensures that update is supported only for porch-registered packages.

droot (approved):

natasha41575 (approved):

PR #3289, opened 2022-06-06
porch: improve package update machinery

droot:
This PR does the following:
 - Ensures `#kpt-merge` comments are added to package resources on `cloning` . This is required for `pkg update` to establish identity of resources across downstream/upstream packages.
 - improves the package update logic ( the `updatePackageMutation` code) in following ways:
   - support for upgrading to porch registered upstream packages with better error checking etc.
   - no longer depends on `git` command line to fetch the upstream package resources.
   - Ensures `merge-key` is added upon 'update` for new resources
   - Introduces `PkgUpdater` interface (and default implementation) that makes `updater` to be pluggable and testable. Added unit tests for it. This should help us evolve update functionality without impacting rest of the porch machinery. 
 - Misc renaming/refactoring around self-lock and upstream-lock information

### DIFF
--- a/porch/pkg/engine/engine.go
+++ b/porch/pkg/engine/engine.go
@@ -22,7 +22,6 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
-	"strings"
 
 	"github.com/GoogleContainerTools/kpt/internal/fnruntime"
 	"github.com/GoogleContainerTools/kpt/pkg/debug"
@@ -34,6 +33,7 @@ import (
 	"github.com/GoogleContainerTools/kpt/porch/pkg/repository"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/trace"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/kustomize/kyaml/comments"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
@@ -279,9 +279,11 @@ func (cad *cadEngine) UpdatePackageRevision(ctx context.Context, repositoryObj *
 			}
 			mutation := &updatePackageMutation{
 				task:              newTask,
+				oldTask:           oldTask,
 				cad:               cad,
 				referenceResolver: cad.referenceResolver,
 				namespace:         repositoryObj.Namespace,
+				pkgName:           oldObj.GetName(),
 			}
 			mutations = append(mutations, mutation)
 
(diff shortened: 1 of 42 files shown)


droot:
@mortent @natasha41575 I have cleaned up the code, so the core update logic is fairly abstracted now and can be evolved independently.

porch/pkg/engine/engine.go
```diff
@@ -265,9 +265,11 @@ func (cad *cadEngine) UpdatePackageRevision(ctx context.Context, repositoryObj *
 			}
 			mutation := &updatePackageMutation{
 				task:              newTask,
+				oldTask:           oldTask,
```
natasha41575, line 268: still new to this code, do you mind explaining what is `oldTask` and `newTask`?

droot: Good question.

So the way a downstream package is created today, is by doing a `clone` operation on an upstream package. `Clone` info is stored in a Task struct that is part of the `Spec` of the PackageRevision API. Now, for updating a downstream package, the plan is to update the Upstream Ref in the existing clone task of the API.

The code that you are looking here gets invoked when PackageRevision is updated by updating the upstream bits in the clone task of the PackageRevision API. New bits are part of the `newTask`. When we are performing an package upgrade (i.e. `updatePackageMutation`), we need the current upstream ref and the target upstream ref. current upstream ref is stored in the current clone task that I am storing as `oldTask`. Now that I think of it, I don't need the whole task but just the upstream info (and that is also just the PackageRevisionRef), I could store just that and that may also be more clear from code/maintenance perspective as well.

porch/pkg/engine/engine.go
```diff
@@ -417,41 +422,91 @@ func (cad *cadEngine) ListFunctions(ctx context.Context, repositoryObj *configap
...
 	}
 
-	return loaded, m.task, nil
+	return updatedResources, m.task, nil
+}
+
+// Currently assumption is that downstream packages will be forked from a porch package.
+// As per current implementation, upstream package ref is stored in an old clone task, but this may
+// change so the logic of figuring out current upstream will live in this function.
+func (m *updatePackageMutation) currUpstream() (*api.PackageRevisionRef, error) {
```
mortent, line 501: So this information would also be available in the `UpstreamLock` of the Kptfile in the package right? I'm thinking we might want to consider that the SoT, but we need to figure out how we want to handle the tasks.

porch/pkg/engine/engine.go
```diff
@@ -417,41 +422,91 @@ func (cad *cadEngine) ListFunctions(ctx context.Context, repositoryObj *configap
...
+	}
 
-	packageDir := filepath.Join(dir, packageName)
-	if err := kpt.PkgUpdate(ctx, ref, packageDir, kpt.PkgUpdateOpts{}); err != nil {
-		return repository.PackageResources{}, nil, err
+	klog.Infof("performing pkg upgrade operation for pkg %s resource counts local[%d] original[%d] upstream[%d]",
+		m.pkgName, len(resources.Contents), len(originalResources.Spec.Resources), len(upstreamResources.Spec.Resources))
+
+	// May be have packageUpdater part of engine to make it easy for testing ?
+	updatedResources, err := (&defaultPackageUpdater{}).Update(ctx,
```
mortent, line 475: We need to decide where we draw the line between what should be in the kpt library and what should be done in porch. Given the discussion earlier this week, my thinking is that we should see if we can make the porch layer as thin as possible. I think that means that the update logic will happen based on git or oci, and unaware of the concept of porch packages (which will use either git or oci underneath).

porch/pkg/engine/engine.go
droot: hmm..currently no. As per current implementation that @natasha41575 has in progress, the upstream info will be contain upstream Git(repo/dir/commit) or OCI(...) info etc, but it doesn't contain mapping to the porch packagerevision unless we store that as annotation in the `Kptfile` of the package. 

Or we do lookup of the packagerevision on the basis of upstream info in Kptfile.

porch/pkg/engine/engine.go
droot: I think the challenging part has been mostly around the components that interact with storage (git/oci and credentials) and local filesystem. and we can slowly get there.

porch/pkg/engine/engine.go
natasha41575: We could still look at the Kptfile though, right? We could look at the Kptfile in the `resources repository.PackageResources` that are  available in the `Apply` method above to analyze the Kptfile and extract the upstream information. (I'm assuming that those are the package resources.) We don't necessarily need to know the packagerevision to look at the Kptfile.

droot: > Although, I realize that the UpstreamRef that we're looking for might not be the same thing that is stored in the Kptfile's UpstreamLock.Ref.

That's is precisely the reason I am avoiding using the upstream info from the `Kptfile`. And as per our discussion today, I think it is reasonable to assume that every downstream package will have `clone` Task as the first task. So retrieving upstream info from `clone` task (like it is being done currently) seems to be a reasonable approach.

droot:
@mortent @natasha41575 I managed to add support for `merge-key` directive while creating a downstream package and also rebased the PR. I also did some end to end testing and also the original e2e update test is also passing now. I think it's very close to getting merged now, so take a look at it.

porch/pkg/engine/engine.go
```diff
@@ -420,61 +423,91 @@ func (cad *cadEngine) ListFunctions(ctx context.Context, repositoryObj *configap
...
+		cad:               m.cad,
+		referenceResolver: m.referenceResolver,
+	}).FetchRevision(ctx, targetUpstream.UpstreamRef, m.namespace)
+	if err != nil {
+		return repository.PackageResources{}, nil, fmt.Errorf("error fetching revision for target upstream %s", targetUpstream.UpstreamRef.Name)
+	}
+	upstreamResources, err := (&PackageFetcher{
+		cad:               m.cad,
+		referenceResolver: m.referenceResolver,
+	}).FetchResources(ctx, targetUpstream.UpstreamRef, m.namespace)
```
mortent, line 467: Might be easier to just fetch the resources from the upstreamRevision here with `upstreamRevision.GetResources(ctx)`.

porch/pkg/engine/update.go
```diff
@@ -0,0 +1,109 @@
...
+		return repository.PackageResources{}, err
+	}
+	return loaded, nil
+}
+
+// PkgUpdate is a wrapper around `kpt pkg update`, running it against the package in packageDir
+func (m *defaultPackageUpdater) do(ctx context.Context, localPkgDir, originalPkgDir, upstreamPkgDir string) error {
+	// TODO: Printer should be a logr
+	pr := printer.New(os.Stdout, os.Stderr)
+	ctx = printer.WithContext(ctx, pr)
```
mortent, line 87: We should probably revisit use of the printer in kpt here. When we want to use the functionality as a library it doesn't really make sense having to provide a printer. But this should be ok for now.

porch/pkg/engine/engine.go
```diff
@@ -138,6 +138,7 @@ func (cad *cadEngine) CreatePackageRevision(ctx context.Context, repositoryObj *
 		if err != nil {
 			return nil, err
 		}
+		mutations = append(mutations, &addMergeKeyMutation{})
```
mortent, line 141: Why do this in a separate mutation rather than as part of the update? Since this is essentially an implementation detail of the merge logic, I don't think we need to make this a separate mutation (which leads to tasks and separate commits).

porch/pkg/engine/engine.go
droot: Good pt. Done.

porch/pkg/engine/engine.go
```diff
@@ -440,61 +442,93 @@ func (cad *cadEngine) ListFunctions(ctx context.Context, repositoryObj *configap
...
+		},
+		repository.PackageResources{
+			Contents: upstreamResources.Spec.Resources,
+		})
 	if err != nil {
-		return repository.PackageResources{}, nil, err
+		return repository.PackageResources{}, nil, fmt.Errorf("error updating the package to revision %s", targetUpstream.UpstreamRef.Name)
+	}
+
+	newUpstream, newUpstreamLock, err := upstreamRevision.GetUpstreamLock()
```
natasha41575, line 504: Does this still work with the changes I made to `GetUpstreamLock` in https://github.com/GoogleContainerTools/kpt/pull/3301?

droot: Good point. This introduces a bug :). This won't work. I need the upstreamRevision's (revision and revisionLock info) instead of upstreamRevision's upstream here.  It exists in the code, need a way to plumb it through.

droot: Addressed this in the latest commit: https://github.com/GoogleContainerTools/kpt/pull/3289/commits/2f6b7bd384c78b1a0069c29c18d12bb6501af669 PTAL

porch/pkg/git/package.go
```diff
@@ -203,6 +203,10 @@ func (p *gitPackageRevision) GetUpstreamLock() (kptfile.Upstream, kptfile.Upstre
 	return *kf.Upstream, *kf.UpstreamLock, nil
 }
 
+func (p *gitPackageRevision) GetLock() (kptfile.Upstream, kptfile.UpstreamLock, error) {
```
natasha41575, line 206: I suggest using this new `GetLock` to replace the lines here:

https://github.com/GoogleContainerTools/kpt/blob/919b09de3b8c3a6663557a997a604f2379ae9232/porch/pkg/engine/clone.go#L92-L100

Currently, it is clearing the Kptfile and then making a call to `GetUpstreamLock`, which reduces down to just `p.findParent()`.

If we change the clone task to use this instead of `GetUpstreamLock`, we can also change line 201 in this file to just return empty structs. That will simplify the code a bit - `GetUpstreamLock` will only look at the Kptfile and `GetLock` will only look at the parent.

porch/pkg/git/package.go
```diff
@@ -203,6 +203,10 @@ func (p *gitPackageRevision) GetUpstreamLock() (kptfile.Upstream, kptfile.Upstre
...
+	return p.findParent()
```
natasha41575, line 207: do we still need to keep `findParent`? Seems like an extra layer for no particular reason, suggest just folding the `findParent` code into this method.

porch/pkg/repository/repository.go
```diff
@@ -59,6 +59,10 @@ type PackageRevision interface {
 
 	// GetUpstreamLock returns the kpt lock information.
 	GetUpstreamLock() (kptfile.Upstream, kptfile.UpstreamLock, error)
+
+	// GetLock returns the current revision's lock information.
+	// This will be the upstream info for downstream revisions.
```
natasha41575, line 64: I think this is a little bit misleading. As we discussed in our previous offline chat, `parent` only points to the upstream info if the current revision is a fresh clone. If the current revision was produced by editing/copying, then it points to the revision it was copied from (which can be different from upstream). 

Could we capture that distinction, either in this comment or in a comment above the git package implementation of this method?

porch/pkg/git/package.go
natasha41575: Following up, @ChristopherFry reported that he's finding that packages created from scratch are getting an upstreamLock pointing to itself. I think this is because the `GetUpstreamLock` method currently points to `parent` as a backup (which is wrong), and I think my suggestion here will fix the bug.

droot: Ack.

porch/pkg/repository/repository.go
natasha41575: Discussed offline, renaming `parent` sounds like a great idea :)

porch/pkg/git/package.go
droot: Addressed this in the most recent commit.

porch/pkg/git/package.go
droot: Removed `findParent`.

porch/pkg/repository/repository.go
droot: Cleaned it up and with some renaming the new code needs very less explanation. PTAL.

droot:
@ChristopherFry if you want to test the `pkg upgrade`, then this PR is up-to-date. Feel free to give it a try if you want :)

droot:
@mortent @natasha41575 This is all ready now for final look.

droot:
> @droot Can you update the description of the PR and give a summary of the changes? It makes it easier to review the code.

Thanks for the reminder. Started with `POC` and forgot to update the description while refining it. Just updated it.

mortent (approved):

porch/pkg/git/package.go
```diff
@@ -196,19 +190,17 @@ func (p *gitPackageRevision) GetUpstreamLock() (kptfile.Upstream, kptfile.Upstre
...
-		return p.findParent()
+		// the package doesn't have any upstream package.
+		return kptfile.Upstream{}, kptfile.UpstreamLock{}, nil
 	}
 	return *kf.Upstream, *kf.UpstreamLock, nil
 }
 
+// GetLock returns the self version of the package. Think of it as the Git commit information
+// that represent the package revision of this package. Please note that it uses Upstream types
+// to represent this information but it has no connection with the associated upstream package (if any).
```
natasha41575, line 201: very helpful comment 😊

natasha41575 (approved):

PR #3291, opened 2022-06-06
Add task annotation to commits for published packages

mortent:
This updates Porch to store the tasks for Package Revisions in commits for Published packages. Up until now, the tasks were not included in the commit generated as part of publishing a package.

This achieves this by adding the list of tasks to the commit annotation, so while it keeps the tasks, the changes caused by each task is no longer separated into separate commits. It also adds the revision of the package revision to the annotation, which allows us to only find the tasks that belong to a specific package revision. If we want to look at the tasks further back, those will be available by following the reference to the upstream package in the clone task.

This doesn't preserve the individual commits after approval, but I think we can address that separately.

### DIFF
--- a/internal/cmdrpkgcopy/command.go
+++ b/internal/cmdrpkgcopy/command.go
@@ -26,9 +26,8 @@ import (
 	porchapi "github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
-	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -98,6 +97,7 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return errors.E(op, err)
 	}
+
 	pr := &porchapi.PackageRevision{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "PackageRevision",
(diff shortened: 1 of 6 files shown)


internal/cmdrpkgcopy/command.go
```diff
@@ -154,7 +155,7 @@ func (r *runner) getPackageRevisionSpec() (*porchapi.PackageRevisionSpec, error)
 		Revision:       r.revision,
 		RepositoryName: packageRevision.Spec.RepositoryName,
 	}
-	spec.Tasks = []porchapi.Task{{Type: porchapi.TaskTypeEdit, Edit: &r.copy}}
+	spec.Tasks = packageRevision.Spec.Tasks
```
droot, line 158: Do you happen to know why we end up creating `restClient` in this method instead of using the `client` in the runner ?

droot (commented):
Just drive by comments.

internal/cmdrpkgcopy/command.go
mortent: I think I suggested using client-go instead of the kubectl builder in https://github.com/GoogleContainerTools/kpt/pull/3130#discussion_r871725895. Agree that using the controller-runtime client would be better. We can clean that up at some point.

droot: Yes, can be cleaned up later.

droot (approved):
Thanks @mortent . I know we are a bit uncertain about the tasks, but I think it's reasonable to  merge this PR so that we can have update workflow in user's hand and learn more about the model.

My other PR is also in good shape and very close to getting merged. I think we are very close.

internal/cmdrpkgcopy/command.go
mortent: Replaced the `restClient` with the controller-runtime client.

droot: Sweet.

droot (approved):

droot:
🚢

PR #3293, opened 2022-06-07
feat: disable client-side throttling

karlkfi:
- Disable client-side throttling when server-side throttling is enabled.
  Otherwise default to 30 QPS (60 burst QPS).
  This is an increase from the rest.Config default of 5 QPS.

Based on my local performance testing, it looks like kpt v1.0.0-beta.16 has a performance regression that defaults the applier to 5 QPS. I haven't quite figured out why, but earlier versions (like v1.0.0-beta.15) are much faster.

This change both fixes the regression and removes the QPS limit on servers where server-side throttling is enabled. The beta feature flag is enabled by default on K8s v1.20+.

### DIFF
--- a/commands/factory.go
+++ b/commands/factory.go
@@ -15,18 +15,25 @@
 package commands
 
 import (
+	"context"
 	"flag"
 	"fmt"
+	"time"
 
 	"github.com/GoogleContainerTools/kpt/internal/util/cfgflags"
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


karlkfi:
My performance testing results:

| kpt version | kpt live apply | kpt live destroy |
| ---- | ---- | ---- |
| v1.0.0-beta.15 | 4m38.855s | 6m39.967s |
| v1.0.0-beta.16 | 36m16.383s | 7m42.252s |
| pull/3293 | 4m27.818s | 0m55.147s |

These times are for applying 1,000 Nginx Deployments with 1 replica each into an existing Namespace on a Kind v0.12.0 cluster running K8s v1.23.4.

droot (approved):

PR #3295, opened 2022-06-08
function test harness option to skip tests locally

natasha41575:
Add option to function test config: `skipIfLocal`. If this is set, then the test runner will check for an env variable `IN_CI`. If `IN_CI` is not set to true, then we assume that we are running the test locally and skip the test.

This also sets `IN_CI=true` to the CI job.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -46,7 +46,7 @@ jobs:
         run: |
           git config --global user.email you@example.com
           git config --global user.name Your Name
-          make all
-          make test-docker
+          IN_CI=true make all
+          IN_CI=true make test-docker
         env:
           KPT_FN_RUNTIME: ${{ matrix.runtime }}
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -46,7 +46,7 @@ jobs:
         run: |
           git config --global user.email you@example.com
           git config --global user.name Your Name
-          IN_CI=true make all
-          IN_CI=true make test-docker
+          make all
+          make test-docker
         env:
           KPT_FN_RUNTIME: ${{ matrix.runtime }}
--- a/pkg/test/runner/README.md
+++ b/pkg/test/runner/README.md
@@ -22,6 +22,7 @@ in `.expected`:
     used to run the test. Default: `render`.
   - `exitCode`: The expected exit code for the command. Default: 0.
   - `skip`: Runner will skip the test if `skip` is set to true. Default: false.
+  - `skipIfLocal`: Runner will skip the test if the env var IN_CI is "false" or not set. Default: false.
   - `sequential`: This test case should be run sequentially. Default: false.
   - `runtimes`: If the current runtime doesn't match any of the desired runtimes
     here, the test case will be skipped. Valid values are `docker` and `podman`.
--- a/pkg/test/runner/README.md
+++ b/pkg/test/runner/README.md
@@ -22,7 +22,7 @@ in `.expected`:
     used to run the test. Default: `render`.
   - `exitCode`: The expected exit code for the command. Default: 0.
   - `skip`: Runner will skip the test if `skip` is set to true. Default: false.
-  - `skipIfLocal`: Runner will skip the test if the env var IN_CI is "false" or not set. Default: false.
+  - `skipLocal`: Runner will skip the test if the env var CI is "false" or not set. Default: false.
   - `sequential`: This test case should be run sequentially. Default: false.
   - `runtimes`: If the current runtime doesn't match any of the desired runtimes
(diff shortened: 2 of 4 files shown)


natasha41575:
cc @mengqiy

pkg/test/runner/README.md
```diff
@@ -22,6 +22,7 @@ in `.expected`:
     used to run the test. Default: `render`.
   - `exitCode`: The expected exit code for the command. Default: 0.
   - `skip`: Runner will skip the test if `skip` is set to true. Default: false.
+  - `skipIfLocal`: Runner will skip the test if the env var IN_CI is "false" or not set. Default: false.
```
mengqiy, line 25: I'd suggest to make the 2 names more consistent in terms of naming. For example, `skipIfNotInCI` with `KPT_E2E_IN_CI`.

.github/workflows/go.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         run: |
           git config --global user.email you@example.com
           git config --global user.name Your Name
-          make all
-          make test-docker
+          IN_CI=true make all
```
mengqiy, line 49: Does `make all` need it?

.github/workflows/go.yml
```diff
@@ -46,7 +46,7 @@ jobs:
...
+          IN_CI=true make test-docker
```
mengqiy, line 50: Do we have any tests in kpt repo relying on this?

.github/workflows/go.yml
natasha41575: I think you're right that `make all` doesn't need it, will remove

.github/workflows/go.yml
natasha41575: There are no tests in the kpt repo that use this; I did some manual testing. For further testing maybe either you could pull this branch to do your own manual testing or I could commit a dummy test just to exercise the new option?

pkg/test/runner/README.md
natasha41575: done!

natasha41575:
@mengqiy PR is updated, PTAL

natasha41575:
cc @yuwenma

pkg/test/runner/config.go
```diff
@@ -85,6 +85,9 @@ type TestCaseConfig struct {
 	// Skip means should this test case be skipped. Default: false
 	Skip bool `json:"skip,omitempty" yaml:"skip,omitempty"`
 
+	// SkipIfNotInCI means should this test case be skipped if the env var KPT_E2E_IN_CI is "false" or not set. Default: false
```
yuwenma, line 88: "means this test case should be..."

yuwenma:
What is the scenarios that requires this condition? The solution seems to be a temporary workaround.

natasha41575:
> What is the scenarios that requires this condition? The solution seems to be a temporary workaround.

This is needed for the function catalog render-helm-chart private repo tests. The credentials are populated by Github secrets stored in CI, so it won't work when running locally.

pkg/test/runner/config.go
natasha41575: Thanks for the catch! Copy paste mistake. Fixed.

yuwenma:
> > What is the scenarios that requires this condition? The solution seems to be a temporary workaround.
> 
> This is needed for the function catalog render-helm-chart private repo tests. The credentials are populated by Github secrets stored in CI, so it won't work when running locally.

My main concern is that the SKIP and SKIPIfNotInCI config have overlapping use cases so it makes the code hard to maintain and extend. 
 
From the render-helm-charts test, do you think we can be simplified it as:
 ```go
func TestYourRenderHelmCharts():
      if testutil.IsDevLocal():
            pass

func IsDevLocal():
      _, githubCI := os.LookupEnv("CI")
      return !githubCI
```
[CI](https://docs.github.com/en/actions/learn-github-actions/environment-variables) is the GitHub default env var and is always set to true. So no manual change is needed and we do not complicate the runner skip config. WDYT?

natasha41575:
> CI is the GitHub default env var and is always set to true. So no manual change is needed and we do not complicate the runner skip config. WDYT?

I didn't know GH had a default CI var! That's very useful. That sounds good to me, I will update it.

yuwenma:
I see. Then that's fair to me. Can we change `skipIfNotInCI` to something like `SkipLocal`?  "skip" is like a NOT condition, and we can always improve code readability by avoiding double negation. :)

natasha41575:
> I see. Then that's fair to me. Can we change skipIfNotInCI to something like SkipLocal? "skip" is like a NOT condition, and we can always improve code readability by avoiding double negation. :)

Done!

natasha41575:
@yuwenma based on our chat offline today, I don't think we need this anymore. Closing.

PR #3301, opened 2022-06-09
rewrite and expose the git implementation of finding the upstreamLock

natasha41575:
The git implementation of `GetUpstreamLock()` seems to assume that the upstream repo is stored in the `parent` field. But with my testing, this isn't always the case.

For example, say we have an upstream `blueprint` and from that we create downstream package `deployment` with revision `v0`. Then we do `rpkg edit` on the downstream package, creating a new downstream package `deployment` with a revision `v1`. The upstream of the new package should still be `blueprint`, but the `parent` instead points at the `deployment v0` package it was copied from. 

I think in most cases we have to look at the Kptfile to determine the upstreamLock, so this code modifies `GetUpstreamLock` to do that. It's only when we clone a package for the first time that we will need to look at `parent` so that we can populate the Kptfile with the right upstream information.

### DIFF
--- a/porch/pkg/engine/clone.go
+++ b/porch/pkg/engine/clone.go
@@ -89,6 +89,11 @@ func (m *clonePackageMutation) cloneFromRegisteredRepository(ctx context.Context
 		return repository.PackageResources{}, fmt.Errorf("cannot read contents of package %q: %w", ref.Name, err)
 	}
 
+	// If the upstream we cloned from has its own upstream information, we need to clear and replace it
+	if err := kpt.UpdateKptfileUpstream(m.name, resources.Spec.Resources, v1.Upstream{}, v1.UpstreamLock{}); err != nil {
+		return repository.PackageResources{}, fmt.Errorf("failed to clear upstream lock to package %q: %w", ref.Name, err)
+	}
+
 	upstream, lock, err := revision.GetUpstreamLock()
 	if err != nil {
 		return repository.PackageResources{}, fmt.Errorf("cannot determine upstream lock for package %q: %w", ref.Name, err)
@@ -96,7 +101,7 @@ func (m *clonePackageMutation) cloneFromRegisteredRepository(ctx context.Context
 
 	// Update Kptfile
 	if err := kpt.UpdateKptfileUpstream(m.name, resources.Spec.Resources, upstream, lock); err != nil {
-		return repository.PackageResources{}, fmt.Errorf("failed to apply upstream lock to pakcage %q: %w", ref.Name, err)
+		return repository.PackageResources{}, fmt.Errorf("failed to apply upstream lock to package %q: %w", ref.Name, err)
 	}
 
 	return repository.PackageResources{
--- a/porch/pkg/git/package.go
+++ b/porch/pkg/git/package.go
@@ -170,6 +170,16 @@ func (p *gitPackageRevision) GetResources(ctx context.Context) (*v1alpha1.Packag
 	}, nil
 }
 
+// GetUpstreamLock returns the upstream for a package revision. The logic here is a little convoluted because
+// the upstream can come from different places depending on the task.
+// When cloning, the upstream is not available yet in the Kptfile, and is instead stored in the `parent` pointer.
+// When editing, the upstream is available in the Kptfile, and the `parent` pointer doesn't point
+// to the upstream (it points to the previous package revision that we have edited/copied from).
+// What we do here for now is to treat the Kptfile as the source of truth, and only look at `parent` if we
+// cannot get the upstream information from the Kptfile. When we do a clone task, we clear the Kptfile's upstream
+// field so that when we call GetUpstreamLock we end up looking at `parent`.
+// We may want to revisit this implementation in the future to make it easier to understand what is happening.
+// See discussion here https://github.com/GoogleContainerTools/kpt/pull/3301/files#r896190293.
 func (p *gitPackageRevision) GetUpstreamLock() (kptfile.Upstream, kptfile.UpstreamLock, error) {
 	resources, err := p.GetResources(context.Background())
 	if err != nil {
--- a/porch/pkg/git/package.go
+++ b/porch/pkg/git/package.go
@@ -20,8 +20,10 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
+	"github.com/GoogleContainerTools/kpt/internal/pkg"
 	kptfile "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
 	"github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
 	"github.com/GoogleContainerTools/kpt/porch/pkg/repository"
@@ -71,6 +73,21 @@ func (p *gitPackageRevision) uid() types.UID {
 func (p *gitPackageRevision) GetPackageRevision() *v1alpha1.PackageRevision {
 	key := p.Key()
 
+	_, lock, _ := p.GetUpstreamLock()
+
+	// TODO: Use kpt definition of UpstreamLock in the package revision status
+	// when https://github.com/GoogleContainerTools/kpt/issues/3297 is complete.
+	// Until then, we have to translate from one type to another.
+	lockCopy := &v1alpha1.UpstreamLock{
+		Type: v1alpha1.OriginType(lock.Type),
+		Git: &v1alpha1.GitLock{
+			Repo:      lock.Git.Repo,
+			Directory: lock.Git.Directory,
+			Commit:    lock.Git.Commit,
+			Ref:       lock.Git.Ref,
+		},
+	}
+
 	return &v1alpha1.PackageRevision{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "PackageRevision",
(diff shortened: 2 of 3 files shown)


porch/pkg/git/package_test.go
```diff
@@ -1,126 +0,0 @@
...
-
-	v1 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
-	"github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
-	configapi "github.com/GoogleContainerTools/kpt/porch/api/porchconfig/v1alpha1"
-	"github.com/GoogleContainerTools/kpt/porch/pkg/repository"
-	"github.com/go-git/go-git/v5/plumbing"
-	"github.com/google/go-cmp/cmp"
-)
-
-func (g GitSuite) TestUpstreamLock(t *testing.T) {
```
natasha41575, line 30: I'm not 100% sure how this works, but this test seems to be using this file here https://github.com/GoogleContainerTools/kpt/blob/main/porch/pkg/git/testdata/drafts-repository.tar which is a zipped up .git file. `GetResources()` doesn't appear to work on the mock repo created from it, so we can't read the Kptfiles. 

I'm not sure the best way to create a mock repo from which we can read resources, and given that I don't see any tests for `GetResources`, it seems like there isn't a precedent for this.

porch/api/go.mod
```diff
@@ -3,10 +3,11 @@ module github.com/GoogleContainerTools/kpt/porch/api
 go 1.17
 
 require (
-	k8s.io/apimachinery v0.23.5
-	k8s.io/client-go v0.23.5
-	k8s.io/code-generator v0.23.5
-	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65
+	github.com/GoogleContainerTools/kpt v1.0.0-beta.16
```
natasha41575, line 6: Is this dependency okay?

natasha41575:
Ah, I see that there is already an e2e test that depends on the current implementation. I'll explore some options.

porch/api/go.mod
mortent: I agree this might cause issues. I don't necessarily think it is a problem that the porch API has a dependency on kpt since porch is built upon the kpt concepts, and the `UpstreamLock` is a pretty important part of the kpt api.

I think splitting the kpt library from the kpt cli as I suggest in https://github.com/GoogleContainerTools/kpt/issues/3297#issuecomment-1151462512 would address the problem with any circular dependencies. The kpt cli can have dependencies on the porch api, but the kpt library should not.

porch/pkg/git/package.go
```diff
@@ -152,36 +157,27 @@ func (p *gitPackageRevision) GetResources(ctx context.Context) (*v1alpha1.Packag
...
-	if p.ref == nil {
-		return kptfile.Upstream{}, kptfile.UpstreamLock{}, fmt.Errorf("cannot determine package lock; package has no ref")
+	// get the upstream package URL from the Kptfile
+	var kf *kptfile.KptFile
+	for k, v := range resources.Spec.Resources {
+		if k == kptfile.KptFileName {
+			var err error
+			kf, err = pkg.DecodeKptfile(strings.NewReader(v))
+			if err != nil {
+				return kptfile.Upstream{}, kptfile.UpstreamLock{}, fmt.Errorf("cannot read package lock from Kptfile: %w", err)
```
mortent, line 172: This error message maybe seems a bit misleading, but the nested error might explain to the user that the format of the Kptfile is somehow invalid so it can't be decoded.

porch/pkg/git/package.go
```diff
@@ -152,36 +157,27 @@ func (p *gitPackageRevision) GetResources(ctx context.Context) (*v1alpha1.Packag
...
+			}
+		}
 	}
 
-	ref, err := refInRemoteFromRefInLocal(p.ref.Name())
-	if err != nil {
-		return kptfile.Upstream{}, kptfile.UpstreamLock{}, fmt.Errorf("cannot determine package lock for %q: %v", p.ref, err)
+	if kf.Upstream == nil || kf.UpstreamLock == nil {
```
mortent, line 177: Does this need to be an error? I don't think packages must have an upstream (some packages need to be on top of the hierarchy) and it seems reasonable that a package might not have the UpstreamLock either. If we think that the process of updating a package might be async, it would be possible to have a package with an upstream but no UpstreamLock.

porch/api/go.mod
droot: hmm...if this change will be blocked on the pkg restructuring, then we will have to go back to the plan of doing this step in the CLI, right ?

natasha41575: An intermediate solution would be to define a duplicate `UpstreamLock` type in the porch API. In the client side I can do a trivial translation, and leave a TODO in the code that the duplicate type should go away with https://github.com/GoogleContainerTools/kpt/issues/3297#issuecomment-1151462512.

porch/pkg/git/package.go
natasha41575: Makes sense to me! I changed the message to `cannot decode Kptfile`

porch/pkg/git/package.go
```diff
@@ -152,36 +157,27 @@ func (p *gitPackageRevision) GetResources(ctx context.Context) (*v1alpha1.Packag
...
 	if err != nil {
-		return kptfile.Upstream{}, kptfile.UpstreamLock{}, fmt.Errorf("cannot determine package lock: %w", err)
+		return kptfile.Upstream{}, kptfile.UpstreamLock{}, fmt.Errorf("cannot determine package lock; cannot retrieve resources: %w", err)
 	}
 
-	if p.ref == nil {
-		return kptfile.Upstream{}, kptfile.UpstreamLock{}, fmt.Errorf("cannot determine package lock; package has no ref")
+	// get the upstream package URL from the Kptfile
+	var kf *kptfile.KptFile
+	for k, v := range resources.Spec.Resources {
```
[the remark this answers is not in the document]
natasha41575: 🤦‍♀️ I'm so used to parsing ResourceLists, I forgot that there are easier ways to find resources in porch. Fixed, thanks!

porch/api/go.mod
droot: I like option 1 of duplicating `UpstreamLock` because it will unblock the `pkg update`. It helps us unblock update overall.

We might hit some unknowns in pkg restructuring or it might take a bit longer risking `pkg update` timelines.

natasha41575: Cool, I updated the PR to create duplicate `UpstreamLock` type.

porch/pkg/engine/clone.go
```diff
@@ -89,14 +89,19 @@ func (m *clonePackageMutation) cloneFromRegisteredRepository(ctx context.Context
...
+
 	upstream, lock, err := revision.GetUpstreamLock()
 	if err != nil {
 		return repository.PackageResources{}, fmt.Errorf("cannot determine upstream lock for package %q: %w", ref.Name, err)
 	}
 
 	// Update Kptfile
 	if err := kpt.UpdateKptfileUpstream(m.name, resources.Spec.Resources, upstream, lock); err != nil {
-		return repository.PackageResources{}, fmt.Errorf("failed to apply upstream lock to pakcage %q: %w", ref.Name, err)
+		return repository.PackageResources{}, fmt.Errorf("failed to apply upstream lock to package %q: %w", ref.Name, err)
```
droot, line 104: Was trying to trace this logic for three level deep inheritance and wasn't sure about this logic. Can you pl. run through the following scenario. Happy to discuss offline if needed.

```
P1 (originated in porch, so no upstream)

P1' (cloned from P1) ==> upstream(P1') should point to the git commit corresponding to the publishing of P1. 

P1'' (cloned from P1') ==> upstream(P1'') should point to the git commit corresponding to the publishing of P1'.
```

droot: > Lines 92-95 of this file should clear them. Then, porch makes a call to GetUpstreamLock() which should fail to find the information in the Kptfile because we cleared it, and should look at the parent pointer (pointing back to P1', where we cloned from).

Aah, I see it now. I was confused about why we are resetting the upstream info to empty (line 93) if we are going to update the upstream anyways at line 103 anyways. So, we are resetting the upstream here so that the GetUpstreamLock() can return the commit of the parent. Reading your comment again, I realize that we always want the parent commit, then why not just get the parent info and use that in the Kptfile ?

natasha41575: Parent does not always point to the upstream (I give an example in my PR description above with the edit/copy task)

Since the correct upstream information is _always_ available in the Kptfile with the only exception being a fresh clone, I am treating the Kptfile as the source of truth with the only exception being the clone task, which will get the upstream information from the parent. 

If we only look at the parent in GetUpstreamLock (which was the previous implementation), we’ll need a different method to look at the Kptfile when we do an update task because parent is not reliable. I thought about doing that, but it seemed weird to me to have a method called GetUpstreamLock that wasn’t actually getting the upstream in all cases, which is why I changed it. 

I’m happy to restructure the code somehow but looking at just the parent doesn’t seem to be the right thing to do.

droot: As discussed offline, let's capture the difference between `editing a package` and `cloning package` as a comment and move on till we have a better idea to resolve it.

natasha41575: Thanks! I added the comment. Will merge once the tests pass.

PR #3315, opened 2022-06-19
Remove the edit task and rely on replay instead

mortent:
Removes the edit task from porch. For creating new revisions, we instead rely on creating new revisions using the tasks from the source revision: https://github.com/GoogleContainerTools/kpt/blob/main/internal/cmdrpkgcopy/command.go#L144. This means we are essentially just replaying the tasks to create the new revision.


droot (approved):
Does it impact UI in anyway ?

mortent:
@droot Yes, so I'll hold off merging this PR until we know the UI is ready for it.

PR #3321, opened 2022-06-23
Preserve commits on package revision approval

mortent:
This updates the logic for approving packages to also capture the full history of the package, rather than squash the commits from the draft into a single commit on the main branch/tag. 
This also removes the logic for including multiple tasks into a single commit, as that is no longer needed after this change.

One known shortcoming of this implementation is that it doesn't capture the original author of each commit from the draft in the main branch. I plan on addressing that in a follow-up PR.


mortent:
@justinsb This is ready for review (finally)

porch/pkg/git/draft.go
```diff
@@ -245,36 +246,110 @@ func (r *gitRepository) commitPackageToMain(ctx context.Context, d *gitPackageDr
...
-	var tasks []*v1alpha1.Task
-	for i := range d.tasks {
-		t := d.tasks[i]
-		tasks = append(tasks, &t)
+	return commitHash, newPackageTreeHash, localTarget, nil
+}
+
+// TODO: This is an almost direct copy of
+// https://github.com/GoogleContainerTools/kpt/blob/3c3288af0c4c4a7e07ffeb6fe473d32afd81137b/porch/pkg/git/git.go#L860.
+// Fix it when we can use generics.
```
justinsb, line 311: We could just start using a comment like `// TODO:generics` or something like that

justinsb (approved):

PR #3326, opened 2022-06-24
Remove suffix from filenames in generated patches

mortent:
When we generate patches for updated resources, we add `@old` and `@new` suffixes to the filename in the patches. This prevents us from successfully applying the patches when we replay the task list. This removes those suffixes from the generated patches.

We should also add tests for this, but I can do that in a follow-up PR if merging this helps unblock other work.


justinsb:
If this works, that's great!  I think there are _more_ problems, but you're right - this definitely won't help.

/approve
/lgtm

mortent:
@justinsb Yeah, this doesn't fix the underlying issue with doing text-based patches.

This change just makes it possible to replay a package-revision using the same upstream, i.e. the patches will be applied to the same state as it was created from.

droot (approved):

PR #3335, opened 2022-06-30
Initial support for OCI package lifecycle

justinsb:
We store the lifecycle in the kpt.dev/lifecycle annotation.

### DIFF
--- a/porch/pkg/oci/mutate.go
+++ b/porch/pkg/oci/mutate.go
@@ -20,13 +20,13 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"path"
 	"strconv"
 	"time"
 
+	"github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
 	api "github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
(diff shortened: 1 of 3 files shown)


porch/pkg/oci/mutate.go
```diff
@@ -201,7 +201,31 @@ func (p *ociPackageDraft) Close(ctx context.Context) (repository.PackageRevision
 
 	klog.Infof("pushing %s", ref)
 
-	img, err := mutate.Append(p.base, p.addendums...)
+	addendums := append([]mutate.Addendum{}, p.addendums...)
+	if p.lifecycle != "" {
+		if len(addendums) == 0 {
```
mortent, line 206: When do we encounter this situation? Would this be when we approve an oci package, thus we want to update the lifecycle but there isn't any other changes to the content that requires a new layer?

mortent (approved):
Just a small question for clarification.

Also, should we add some tests for this?

porch/pkg/oci/mutate.go
justinsb: That's exactly right.  I think it will likely happen whenever we start approving packages.  I figure it's an important edge case that we're going to have to handle, but right now as I don't think we have test coverage I just surface an error.  I'd like to improve the test coverage here e.g. running all tests against both git & oci Repositories, as the API behaviour should be equivalent.

PR #3336, opened 2022-06-30
Support reclone and replay for updating clone tasks

justinsb:
When we see an update that changes the first Clone task, we treat that
as a signal that the user wants to use the "reclone and replay"
strategy for updating the package.

See https://github.com/GoogleContainerTools/kpt/issues/3329 for
background.


porch/pkg/git/git.go
```diff
@@ -784,6 +784,8 @@ func (r *gitRepository) pushAndCleanup(ctx context.Context, ph *pushRefSpecBuild
 		RefSpecs:          specs,
 		Auth:              auth,
 		RequireRemoteRefs: require,
+		// TODO(justinsb): Need to ensure this is a compare-and-swap
+		Force: true,
```
natasha41575, line 788: What is this for? I don't see it used anywhere

natasha41575 (approved):
Logic seems straightforward and makes sense to me, but should probably also get approval from @mortent or @droot

porch/pkg/git/git.go
justinsb: Ah, it's because when we update a package now it isn't necessarily a fast-forward, because we're effectively doing a `git rebase`, which needs a force when we `git push`.  We can do better than `git push -f` because we want to detect concurrent activities, but we need to track and pass the current expected version on git.

yuwenma (approved):

PR #3338, opened 2022-06-30
Add new update task to support merge updates

natasha41575:
Half of https://github.com/GoogleContainerTools/kpt/issues/3329


porch/api/porch/types_packagerevisions.go
```diff
@@ -128,6 +131,23 @@ type PackageCloneTaskSpec struct {
...
+	Upstream UpstreamPackage `json:"upstreamRef,omitempty"`
+
+	// 	Defines which strategy should be used to update the package. It defaults to 'resource-merge'.
+	//  * resource-merge: Perform a structural comparison of the original /
+	//    updated resources, and merge the changes into the local package.
+	//  * fast-forward: Fail without updating if the local package was modified
+	//    since it was fetched.
+	//  * force-delete-replace: Wipe all the local changes to the package and replace
+	//    it with the remote version.
+	Strategy PackageMergeStrategy `json:"strategy,omitempty"`
```
natasha41575, line 148: I'm planning to change this struct to be empty and use the Upstream and Strategy from the package revision's clone task instead.

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/GoogleContainerTools/kpt
 
 go 1.17
 
+replace github.com/GoogleContainerTools/kpt/porch/api => ./porch/api
```
natasha41575, line 5: The current CLI seems to be incompatible with the changes I made on the backend. My temporary solution is to pin it to the local directory so that the CLI tests can pass, and then remove this replace statement immediately in a followup PR when this merges.

natasha41575:
@droot @mortent @justinsb this PR is cleaned up and ready for review. I just got confirmation from @ChristopherFry that it does what it needs to do for the UI.

go.mod
natasha41575, line 5: The current CLI is incompatible with the changes I made on the backend. My temporary solution is to make the needed changes on the CLI side and pin it to the local porch/api directory so that the CLI tests can pass, and then remove this replace statement immediately in a followup PR when this merges.

mortent, line 5: Should this be part of the PR? I think we prefer not having these checked in and instead rely on [golang workspaces](https://go.dev/doc/tutorial/workspaces) for managing the different modules.

internal/cmdrpkgupdate/command.go
```diff
@@ -134,8 +135,16 @@ func (r *runner) doUpdate(pr *porchapi.PackageRevision) error {
 		if newUpstreamPr == nil {
 			return fmt.Errorf("revision %s does not exist for package %s", r.revision, pr.Spec.PackageName)
 		}
-		cloneTask.Clone.Upstream.UpstreamRef.Name = newUpstreamPr.Name
+		newTask := porchapi.Task{
+			Type: porchapi.TaskTypeUpdate,
+			Update: &porchapi.PackageUpdateTaskSpec{
+				Upstream: cloneTask.Clone.Upstream,
```
mortent, line 141: Why reuse the previous one from the CloneTask instead of creating one from scratch? Doesn't really matter that much, but I was a bit confused at first.

porch/pkg/engine/engine.go
```diff
@@ -515,23 +521,33 @@ func (m *updatePackageMutation) Apply(ctx context.Context, resources repository.
...
 	}
-	upstream := m.oldTask.Clone.Upstream
+	upstream := m.cloneTask.Clone.Upstream
 	if upstream.Type == api.RepositoryTypeGit || upstream.Type == api.RepositoryTypeOCI {
 		return nil, fmt.Errorf("upstream package must be porch native package. Found it to be %s", upstream.Type)
 	}
 	return upstream.UpstreamRef, nil
 }
 
+func findCloneTask(pr *api.PackageRevision) *api.Task {
```
mortent, line 541: With this change, a clone task can only exist as the first task in the list. Every packagerevision must start with either an InitTask or a CloneTask

mortent (commented):
Mostly looks good to me. Just a few small comments.

I think we need at some point to discuss what exposing the `UpstreamLock` in the API means for supporting other package formats and whether relying in the CloneTask for doing updates is the right way to do it vs. finding the origin from the Kptfile. But lets merge and iterate.

internal/cmdrpkgupdate/command.go
natasha41575: `Upstream` is a fairly complex type with lots of nested fields. I'm aware that we're just looking at one field right now (`Upstream.UpstreamRef`), but for the future I wasn't sure if the update task would need to use other nested fields stored in the `Upstream` field; it felt safer to copy the whole thing from the clone task.

go.mod
natasha41575: I'll see if I can separate the CLI changes from the porch changes. I vaguely recall having an issue with doing that, but I need to refresh my memory

porch/pkg/engine/engine.go
natasha41575: Fixed

go.mod
natasha41575: Or were you suggesting we start using go workspaces now?

e2e/testdata/porch/rpkg-update/config.yaml
```diff
@@ -61,24 +61,3 @@ commands:
       git-f1fbc2b72ee70b2f34e64c2630031d0560e6cf48   basens-edit         v0         false    Draft       git
       git-894137a40c0792e88a9b844491a95a466b085ca1   basens-edit         v2         true     Published   git
       git-542e4d9fd8ca2ee94edf43aadb3ec31dc2fbd3a0   basens-edit-clone   v0         false    Draft       git
-  - args:
-      - alpha
-      - rpkg
-      - update
-      - --namespace=rpkg-update
-      - --discover
-      - git-542e4d9fd8ca2ee94edf43aadb3ec31dc2fbd3a0
```
natasha41575, line 70: @mortent The current CLI seems incompatible with the changes I made to the backend (see the test failures in https://github.com/natasha41575/kpt/runs/7224748986?check_suite_focus=true).

This removal is temporary, I add it back along with the CLI change in https://github.com/GoogleContainerTools/kpt/pull/3346, which updates the CLI to be compatible with the new backend.

go.mod
natasha41575: I split out the CLI changes into a separate PR https://github.com/GoogleContainerTools/kpt/pull/3346, but I had to temporarily remove a test to do so. 

See https://github.com/GoogleContainerTools/kpt/pull/3338#discussion_r915371488

mortent (approved):

PR #3117, opened 2022-05-10
Fix guide links in README

ChristopherFry:
This pull request corrects the guide links in the README. Currently the links refer to a `guides` subfolder although the guides exist within the same folder as the README. This pull request also removes a duplicate link from the README.

PR #3340, opened 2022-07-01
docs: fix broken links in porch guides

ChristopherFry:
This pull request fixes several broken links in the Porch guides.

### DIFF
--- a/site/guides/porch-installation.md
+++ b/site/guides/porch-installation.md
@@ -91,7 +91,7 @@ packagerevisionresources                       porch.kpt.dev/v1alpha1           
 packagerevisions                               porch.kpt.dev/v1alpha1                 true         PackageRevision
 ```
 
-You can start [using Porch](./porch-user-guide.md).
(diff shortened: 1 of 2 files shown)


ChristopherFry:
> Thanks for fixing - though do we know why relative links don't work on kpt.dev. That's both surprising, and I'd imagine that this update would break the github markdown browser?

Correct, updating these links will correct the links on kpt.dev, however it will break the links for anyone browsing the markdown files directly on GitHub, so this pull request prioritizes the links being correct on kpt.dev.

I looked into this quickly a few months ago (with #3117) and it looks like this is due to the library we are using to render kpt.dev. Let me see if there's an issue out already to fix this (agree links should work everywhere), and if not I can get an issue created. I'm willing to look into this overall issue more as time allows.

justinsb (approved):

justinsb:
It would be great if we could fix this, but I'm merging because this does fix kpt.dev, and that's (arguably) more important than github.

PR #3342, opened 2022-07-01
deepcopy-gen: hack to ensure directory name is set

justinsb:
When we use `--input-dirs ./...` deepcopy-gen / defaulter-gen cannot
determine the package name, it appears (unless GOPATH is set).

Instead use the fully-qualified path as we're using for other
generators.

`../<dirname>` also works but is not consistent with our other
generators.


mortent (approved):

justinsb:
Let's not merge yet, I'm trying to figure out all the intricacies here!

PR #3345, opened 2022-07-06
Add listing and filtering functionalities on 

chunglu-chou:
To have better support on monitoring status of packages, the following features are added:
1. Listing functionality that allows users to see status of currently applied packages by adding --list flag.
2. Filtering options on both inventory name and namespace, allowing users to check status of targeted resources more easily by using --inv-name and --namespace flags.
3. Filtering options on status, allowing users to figure out which packages are in certain status with --status command this only works for table format now but will support other output format in the future.


thirdparty/cli-utils/status/printers/table/table_printer.go
```diff
@@ -86,7 +123,9 @@ func (t *tablePrinter) runPrintLoop(coll *CollectorAdapter, stop <-chan struct{}
 		Columns:   columns,
 	}
 
-	linesPrinted := baseTablePrinter.PrintTable(coll.LatestStatus(), 0)
+	fmt.Printf("%c[H", common.ESC)
+	fmt.Printf("%c[J\r", common.ESC)
```
nan-yu, line 127: Be cautious about the flushing result.

karlkfi:
> @mortent @droot This PR is not a small changed to kpt's fork of cli-utils. What is our plan about this fork going forward?
> 
> IMO we should keep our fork as minimal as possible.

@chunglu-chou, This PR should be made against the https://github.com/kubernetes-sigs/cli-utils repo instead. 

It also needs tests, which there's scaffolding for in the examples repo of cli-utils.

Once we validate and merge it in cli-utils, we can cut a new release and make PR to bump it in kpt.

chunglu-chou:
@karlkfi Sure, I will discuss this with Nan and Mortent, since I wrap some existing functions in the cli-utils in this PR, so I would probably need to rewrite some of them and test it.

PR #3346, opened 2022-07-07
fix updates on CLI side

natasha41575:
This makes the CLI compatible with the changes in the backend from #3338. (First 3 commits here are the same as in #3338 but for some reason still show up after rebasing.)

Related: https://github.com/GoogleContainerTools/kpt/issues/3329

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,10 +2,8 @@ module github.com/GoogleContainerTools/kpt
 
 go 1.17
 
-replace github.com/GoogleContainerTools/kpt/porch/api => ./porch/api
-
 require (
-	github.com/GoogleContainerTools/kpt/porch/api v0.0.0-20220617221430-3c3288af0c4c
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/GoogleContainerTools/kpt
 
 go 1.17
 
+replace github.com/GoogleContainerTools/kpt/porch/api => ./porch/api
```
natasha41575, line 5: This will go away with https://github.com/GoogleContainerTools/kpt/pull/3338

justinsb:
LGTM; we'll likely want that update-clonetask-in-place logic when we merge #3336, but it looks like it's easy enough to add back so not an issue :-)

natasha41575:
> LGTM; we'll likely want that update-clonetask-in-place logic when we merge #3336, but it looks like it's easy enough to add back so not an issue :-)

Are we planning to only support reclone-and-replay in the CLI? Or will we have some way of supporting both that and merge updates?

PR #3360, opened 2022-07-12
add upstream identifier in pkg update, get and clone(porch)

yuwenma:
- This PR lets `kpt` to add `internal.kpt.dev/upstream-identifier` annotation together with `kpt-merge` comment. 
- This function is called in [kpt pkg get](https://github.com/GoogleContainerTools/kpt/blob/72ef5b29cb372ed1d28e1380782651f366e2d14f/internal/util/get/get.go#L117), [kpt pkg update](https://github.com/GoogleContainerTools/kpt/blob/6008784a2a6de63ecba903d16c1a317022287b6e/internal/util/update/update.go#L185) and porch [package clone](https://github.com/GoogleContainerTools/kpt/blob/604b6a483f434d599abb2616fd5b392ee768654b/porch/pkg/engine/clone.go#L95), porch [package update](https://github.com/GoogleContainerTools/kpt/blob/604b6a483f434d599abb2616fd5b392ee768654b/porch/pkg/engine/engine.go#L532)  
- As discussed offline, we currently continue to rely on `kpt-merge` comment to match upstream resource and establish three-way merge. 

- Fix an existing issue where `addmergecomment` detects the resource meta by kyaml `IsNilOrEmpty`. not the meta can never be NilorEmpty since helper annotations are always added.

### DIFF
--- a/internal/util/addmergecomment/addmergecomment.go
+++ b/internal/util/addmergecomment/addmergecomment.go
@@ -23,7 +23,20 @@ import (
 	"github.com/GoogleContainerTools/kpt/internal/util/merge"
 	"sigs.k8s.io/kustomize/kyaml/copyutil"
 	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/resid"
 	kyaml "sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// TODO(yuwenma): Those const vars are defined in kpt-functions-sdk/go/fn v0.0.0-20220706221933-7181f451a663+
+// we cannot import go/fn directly because the porch/set-namespace uses an older go/fn version. Bumping kpt module alone fails
+// kpt CI.
+// We should update porch/set-namespace once https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/885 is released.
+// and cleanup the const vars below
+const (
+	upstreamIdentifierFmt = "%s|%s|%s|%s"
+	upstreamIdentifier    = "internal.kpt.dev/upstream-identifier"
+	unknownNamespace      = "~C"
+	defaultNamespace      = "default"
 )
 
 // AddMergeComment adds merge comments with format "kpt-merge: namespace/name"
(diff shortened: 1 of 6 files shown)


yuwenma:
Can you review this change @justinsb @droot

internal/util/addmergecomment/addmergecomment.go
```diff
@@ -49,6 +62,35 @@ func Process(paths ...string) error {
 	return nil
 }
 
+// addUpstreamAnnotation adds internal.kpt.dev/upstream-identifier annotation to resource.
+func addUpstreamAnnotation(object *kyaml.RNode, mergeComment string) error {
+	annotations := object.GetAnnotations()
+	if _, ok := annotations[upstreamIdentifier]; ok {
+		// skip adding annotation if the upstream annotation is already present
```
droot, line 69: If I clone a package that already has these annotations (that is a case for any 3 level pkg chain), this will skip adding annotation ? with this, all descendant packages will have GVKNN of the resources in the root pkg.

internal/util/addmergecomment/addmergecomment.go
```diff
@@ -60,15 +102,33 @@ func (amc *AddMergeComment) Filter(object *kyaml.RNode) (*kyaml.RNode, error) {
...
+		//    internal.config.kubernetes.io/path: 'k8s-cli-982798852.yaml'
+		//    internal.config.kubernetes.io/seqindent: 'compact'
+		//    internal.config.kubernetes.io/annotations-migration-resource-id: '0'
 		return object, nil
 	}
-	if strings.Contains(mf.Key.YNode().LineComment, merge.MergeCommentPrefix) {
-		// skip adding merge comment if merge comment is already present
+
+	// Only add merge comment if merge comment does not present
+	if !strings.Contains(mf.Key.YNode().LineComment, merge.MergeCommentPrefix) {
```
droot, line 124: hmm... I thought the merge-key comment tracks the immediate upstream resource's identifier, but looking this it tracks the root resource's identifier.

Two options:
1) preserve existing behavior (merge-key tracks root resources'id)
2) or change it to use immediate upstream

Both of these will work for update, only difference is the number of descendant nodes affected when root change's name/namespace.

yuwenma: I think both options can give identifier functions reliable id and both are more reasonable than the other in certain scenarios. So it defers to future plans on which scenarios are more important.
I'll switch to immediate upstream and list the pros and cons on different options in the PR description.

internal/util/addmergecomment/addmergecomment.go
yuwenma: I changed the code to use immediate upstream meta as the origin. Also added a comment to mark the potential caveats in 3-way-merge

porch/pkg/engine/engine.go
```diff
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"reflect"
```
mengqiy, line 24: This change seems to be from another commit in a different PR. I suspect something messed up with rebase.

mengqiy (commented):
LGTM if others' comments are addressed.
I will deter to Sunil or Justin for the final approval

droot (approved):
LGTM.

Would be good to add any tests for a 3-level pkg chain if we can.

porch/pkg/engine/engine.go
yuwenma: Good catch! Resolved by another rebase.

PR #3367, opened 2022-07-13
Function SDK guide

yuwenma:
This guide focuses on explaining the KRM resource editing patterns.


site/guides/effective-go-krm-function.md
```diff
@@ -0,0 +1,212 @@
...
+# Set your KRM function name.
+export FUNCTION_NAME=<YOUR FUNCTION NAME>
+export GOPATH=$(go env GOPATH)
+export FUNCTION_PATH=github.com/<YOUR USERNAME>
+
+# Create and direct to your Go working directory
+mkdir -p $GOPATH/src/${FUNCTION_PATH} && cd $GOPATH/src/${FUNCTION_PATH}
+
+# Get the "get-started" package.
+kpt pkg get https://github.com/GoogleContainerTools/kpt-functions-sdk.git/go/get-started-runner@master ${FUNCTION_NAME}
```
justinsb, line 32: This line didn't work for me ... 

`Error: path "/go/get-started-runner" does not exist in repo "https://github.com/GoogleContainerTools/kpt-functions-sdk" `

Do I need to merge a PR in kpt-functions-sdk?

site/guides/effective-go-krm-function.md
```diff
@@ -0,0 +1,212 @@
+# Effective GO KRM functions
+
+This guide gives tips to effectively write a KRM function. 
+
+This guide is for advanced kpt function users who find the [catalog.kpt.dev] cannot fulfil their needs 
+and want to design their own KRM functions.
```
droot, line 6: I think `advanced kpt users` can confuse users and might discourage users to read further.

Maybe we can combine this line and previous line with something along the lines of ...

`
This guide describes best practices and patterns on how to structure your Go KRM functions.
`

site/guides/effective-go-krm-function.md
```diff
@@ -0,0 +1,212 @@
+# Effective GO KRM functions
+
+This guide gives tips to effectively write a KRM function. 
+
+This guide is for advanced kpt function users who find the [catalog.kpt.dev] cannot fulfil their needs 
+and want to design their own KRM functions. 
+
+Suggest reading [Developing in Go] first
```
droot, line 8: nit `.` missing at the end.

site/guides/effective-go-krm-function.md
```diff
@@ -0,0 +1,212 @@
...
+      // TODO: Add your KRM resource mutate or validate logic.
+}
+
+func main() {
+    if err := fn.AsMain(&FunctionX{}); err != nil {
+        os.Exit(1)
+    }
+}
+```
+`FunctionX` implements the [`Runner`] interface that can process the input KRM resources as [`fn.ResourceList`], it initializes `fn.KubeObject` to hold the KRM resources,
```
justinsb, line 63: Do we need to mention ResourceList?

site/guides/effective-go-krm-function.md
```diff
@@ -0,0 +1,212 @@
...
+so that you can use [`fn.KubeObject` and `fn.SubObject`] methods directly. After `Run`, it will convert the modified `fn.KubeObjects` to KRM resources.
```
justinsb, line 64: Instead of "it initializes `fn.KubeObject` to hold the KRM resources, so that you can use [`fn.KubeObject` and `fn.SubObject`] methods directly", I would just say that something like "the objects passed to the function are available in items.  These objects can be modified in place and are used as the output of the function".

site/guides/effective-go-krm-function.md
```diff
@@ -0,0 +1,212 @@
...
+
+This guide is for advanced kpt function users who find the [catalog.kpt.dev] cannot fulfil their needs 
+and want to design their own KRM functions. 
+
+Suggest reading [Developing in Go] first
+
+## Prerequisites
+
+- [Install kpt]
+- [Install Docker]
```
droot, line 13: It will be good to specify minimum version of Go. I think `Go 1.17+` is what we have been using for the toolset.

site/guides/effective-go-krm-function.md
```diff
@@ -0,0 +1,212 @@
...
+  FnConfigBool bool
+  FnConfigInt  int
+  FnConfigFoo string
+}
+```
+
+For example, define a `SetImage` and add two variables to compare-and-swap the image value will be like:
+```go
+type SetImage struct {
+  OldImage string // Existing image
```
justinsb, line 81: This is good - the example makes it much clearer.  We might consider just walking through SetImage, instead of FunctionX.  Start with no parameters, and then say "we want to change an image from an old image value to a new image value, so we need parameters so the user can specify the old image and the new image" or something like that

site/guides/effective-go-krm-function.md
droot, line 32: Assuming myself as a Go developer, I will find this step to be awkward :) and I will also be surprised to see files such as `Kptfile` etc. It might get confusing. Can we just use `git clone` or something to fetch the skeleton code ?

site/guides/effective-go-krm-function.md
```diff
@@ -0,0 +1,212 @@
...
+    if nodeSelector.GetString("disktype") != "ssd" {
+       nodeSelector.SetNestedStringOrDie("ssd", "disktype")
+    }
+}
+```
+
+### Copy `KubeObject` to a typed struct
+
+If you already have some struct to define a KRM resource (like `corev1.ConfigMap`), you can switch the `KubeObject` 
+to the other type via `As`
```
justinsb, line 182: I would add an explicit warning that this copies the object and thus you have to copy it back

site/guides/effective-go-krm-function.md
```diff
@@ -0,0 +1,212 @@
...
+    	obj.SetNestedStringOrDie("RollingUpdate", "spec", "strategy", "type")
+    }
+}
+```
+
+Besides the [unstructured.Unstructured] style, you can also run functions on each sub-field as a `fn.SubObject`  
+```go
+func (r *YourFunction) Run(context *fn.Context, functionConfig *fn.KubeObject, items fn.KubeObjects) {
+    // Get first deployment object.
+	deployment := items.Where(fn.IsGVK("apps", "v1", "Deployment")).Where(func(o *fn.KubeObject) bool{return o.GetName() == "nginx"})[0]
```
droot, line 165: I think it's okay to have this expression in multiple lines that helps with readability.

site/sidebar.md
```diff
@@ -107,5 +107,6 @@
   - [Namespace Provisioning UI](guides/namespace-provisioning-ui.md)
   - [Variant Constructor Pattern](guides/variant-constructor-pattern.md)
   - [Value Propagation Pattern](guides/value-propagation.md)
+  - [Effective Go KRM functions](guides/effective-go-krm-function.md)
```
droot, line 110: Would it makes sense to have it alongside `Getting Started` instead of `guides` ?

droot (commented):
Thanks for writing this. I have a few minor comments.

liamfallon:
@efiacor @CsatariGergely This PR has a draft of a functions SDK user guide.

PR #3371, opened 2022-07-14
Added a basic cert-manager kpt package in package-examples

droot:
This PR adds a basic cert-manager package under `package-examples`.

I have added a section in cert-manager-basic/README.md explaining how the package was created.

Earlier I was using helm chart to create the initial manifests, but then realized that the github release from cert-manager ships the manifests in the form I want, so using those now. I do have instructions for creating the package from helm chart, I will share those in a separate PR.

/cc @justinsb


bzub:
In case it's useful, here's some vague links to how I generate a package from github release assets:

https://github.com/bzub/kpt-pkgs/blob/7c2d86f4602e0bd9d5d8c8ae63c74413f491b543/docker-bake.hcl#L107
https://github.com/bzub/kpt-pkgs/blob/7c2d86f4602e0bd9d5d8c8ae63c74413f491b543/Dockerfile#L338

The implementation needs to be simplified but in the end you can do:
```sh
docker buildx bake --set='cert-manager.args.VERSION=v1.8.2' https://github.com/bzub/kpt-pkgs.git cert-manager
```

droot:
> In case it's useful, here's some vague links to how I generate a package from github release assets:
> 
> https://github.com/bzub/kpt-pkgs/blob/7c2d86f4602e0bd9d5d8c8ae63c74413f491b543/docker-bake.hcl#L107 https://github.com/bzub/kpt-pkgs/blob/7c2d86f4602e0bd9d5d8c8ae63c74413f491b543/Dockerfile#L338
> 
> The implementation needs to be simplified but in the end you can do:
> 
> ```shell
> docker buildx bake --set='cert-manager.args.VERSION=v1.8.2' https://github.com/bzub/kpt-pkgs.git cert-manager
> ```

Sweet @bzub . Will check it out and I may also have follow up questions as we are trying to create kpt packages for some popular cluster addons and application pakages to understand more. What's the best way to reach you ? are you active on kubernetes slack ?

package-examples/cert-manager-basic/README.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+### Apply the package
+
+```sh
+kpt live init cert-manager-basic
+kpt live apply cert-manager-basic --reconcile-timeout=2m --output=table
+```
+
+### Steps used to create the package
```
droot, line 28: FYI @justinsb @bgrant0607 @yuwenma This section lists the steps I followed to create the initial package. I am using `app` label to split out the manifests into respective directories.

bgrant0607 (approved):
Nice. This organization is definitely better. We can look at https://github.com/bzub/kpt-pkgs/tree/main/cert-manager also, but this seems like a fine starting point. It will also be useful to capture the deltas on top of this baseline, so I suggest we merge and iterate.

bzub:
> > In case it's useful, here's some vague links to how I generate a package from github release assets:
> > https://github.com/bzub/kpt-pkgs/blob/7c2d86f4602e0bd9d5d8c8ae63c74413f491b543/docker-bake.hcl#L107 https://github.com/bzub/kpt-pkgs/blob/7c2d86f4602e0bd9d5d8c8ae63c74413f491b543/Dockerfile#L338
> > The implementation needs to be simplified but in the end you can do:
> > ```shell
> > docker buildx bake --set='cert-manager.args.VERSION=v1.8.2' https://github.com/bzub/kpt-pkgs.git cert-manager
> > ```
> 
> Sweet @bzub . Will check it out and I may also have follow up questions as we are trying to create kpt packages for some popular cluster addons and application pakages to understand more. What's the best way to reach you ? are you active on kubernetes slack ?

Yup you can reach me with @bzub on slack.

droot:
> Nice. This organization is definitely better. We can look at https://github.com/bzub/kpt-pkgs/tree/main/cert-manager also, but this seems like a fine starting point. It will also be useful to capture the deltas on top of this baseline, so I suggest we merge and iterate.

Agreed.

I took a quick look at https://github.com/bzub/kpt-pkgs/tree/main/cert-manager and noted few points
 - It uses namespace based organization (`kube-system`, `cert-manager`, `cluster-scoped (at root)`)
 - resources within a namespaces are grouped together in a single file named with `kind` of the resources. So `deployment.yaml` has all the three deployment.
 
@bzub Would be great if you can share -- how you arrived at this layout ?

droot:
> Yup you can reach me with @bzub on slack.

Awesome. Will reach out. Thanks.

justinsb (approved):
lgtm!

bzub:
> > Nice. This organization is definitely better. We can look at https://github.com/bzub/kpt-pkgs/tree/main/cert-manager also, but this seems like a fine starting point. It will also be useful to capture the deltas on top of this baseline, so I suggest we merge and iterate.
> 
> Agreed.
> 
> I took a quick look at https://github.com/bzub/kpt-pkgs/tree/main/cert-manager and noted few points
> 
>     * It uses namespace based organization (`kube-system`, `cert-manager`, `cluster-scoped (at root)`)
> 
>     * resources within a namespaces are grouped together in a single file named with `kind` of the resources. So `deployment.yaml` has all the three deployment.
> 
> 
> @bzub Would be great if you can share -- how you arrived at this layout ?

I think this was a choice based on pains I experienced rapidly developing/changing the filenames within my packages. I'm not sure if an issue has been created yet, but I think when a filename changed in the upstream package it would require manual cleanup after `kpt pkg update` was run downstream at least. When the filenames/paths were stable I had fewer issues. I'll see if I can reproduce this though.

PR #3380, opened 2022-07-15
added ingress-nginx kpt package

droot:
This PR adds kpt package for `ingress-nginx` cluster addon.

README.md in the package documents the steps that we used to create the package.

Note: This time we are using `kpt fn` to transform the static manifests released by ingress-nginx project.

### DIFF
--- a/package-examples/ingress-nginx/README.md
+++ b/package-examples/ingress-nginx/README.md
@@ -1,0 +1,71 @@
+# ingress-nginx
+
+## Description
+
+Nginx Ingress Controller package
+
+## Usage
+
+### Fetch the package
+
+```sh
+kpt pkg get git@github.com:googlecontainertools/kpt.git/package-examples/ingress-nginx ingress-nginx
+```
+
+### View package content
+
+```sh
+
+kpt pkg tree ingress-nginx
+
+```
+
+### Apply the package
+
+```sh
+
+kpt live init ingress-nginx
+kpt live apply ingress-nginx --reconcile-timeout=2m --output=table
+
+```
+
+### How was this package created
+
+```sh
+
+# download the static manifests from the github releases
+wget -O ingress-nginx.yaml https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v1.3.0/deploy/static/provider/cloud/deploy.yaml
+
+```
+
+Starlark function to add `app.kubernetes.io/component` label if it doesn't exists.
+
+```yaml
+
+## default-label.yaml
+
+apiVersion: fn.kpt.dev/v1alpha1
+kind: StarlarkRun
+metadata:
+  name: set-cluster-label
+  annotations:
+source: |
+  # set the component label to cluster if not specified
+  def setlabel(resources):
+    for resource in resources:
+      curr_labels = resource.get("metadata").get("labels")
+      if "app.kubernetes.io/component" not in curr_labels:
+        resource["metadata"]["labels"]["app.kubernetes.io/component"] = ""
+  setlabel(ctx.resource_list["items"])
+```
+
+Create the package
+
+```sh
+
+mkdir ingress-nginx
+kpt pkg init ingress-nginx
+
+cat ingress-nginx.yaml |kpt fn eval - -o unwrap -i starlark:v0.4.3 --fn-config default-label.yaml| kubectl-slice  --template '{{ index "app.kubernetes.io/component" .metadata.labels }}/{{.kind | lower}}-{{.metadata.name|dottodash}}.yaml' -o ingress-nginx --dry-run
+
+```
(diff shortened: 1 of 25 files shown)


package-examples/ingress-nginx/README.md
```diff
@@ -0,0 +1,71 @@
...
+metadata:
+  name: set-cluster-label
+  annotations:
+source: |
+  # set the component label to cluster if not specified
+  def setlabel(resources):
+    for resource in resources:
+      curr_labels = resource.get("metadata").get("labels")
+      if "app.kubernetes.io/component" not in curr_labels:
+        resource["metadata"]["labels"]["app.kubernetes.io/component"] = ""
```
justinsb, line 58: Not entirely sure what this is doing ... from the comment, it looks like this should be "cluster"?

justinsb (approved):
One question, but lgtm

yuwenma (approved):
Looks good. I think we can also add the `set-labels` and `set-namespace:v0.3.4` in Kptfile.

package-examples/ingress-nginx/admission-webhook/job-ingress-nginx-admission-patch.yaml
```diff
@@ -0,0 +1,49 @@
...
+        - --namespace=$(POD_NAMESPACE)
+        - --patch-mutating=false
+        - --secret-name=ingress-nginx-admission
+        - --patch-failure-policy=Fail
+        env:
+        - name: POD_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        image: registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.1.1@sha256:64d8c73dca984af206adf9d6d7e46aa550362b1d7a01f3a0a91b20cc67868660
```
bgrant0607, line 36: We'll want to be able to use set-image to update the tag and digest.

package-examples/ingress-nginx/admission-webhook/job-ingress-nginx-admission-patch.yaml
```diff
@@ -0,0 +1,49 @@
...
+        imagePullPolicy: IfNotPresent
+        name: patch
+        securityContext:
+          allowPrivilegeEscalation: false
+      nodeSelector:
+        kubernetes.io/os: linux
```
bgrant0607, line 42: We'll want something that can extract OS and arch info from a container image so that a function could then inject appropriate selectors.

package-examples/ingress-nginx/admission-webhook/job-ingress-nginx-admission-patch.yaml
```diff
@@ -0,0 +1,49 @@
...
+      restartPolicy: OnFailure
+      securityContext:
+        fsGroup: 2000
+        runAsNonRoot: true
+        runAsUser: 2000
+      serviceAccountName: ingress-nginx-admission
```
bgrant0607, line 48: We'll want to explore how to make this pattern easier. 3 resources are required: the ServiceAccount, the Role, and the RoleBinding. Of those, only the Role requires user input. The UI could guide the user through the creation process and provide autocompletion, but that wouldn't address the CLI experience.

bgrant0607 (commented):
Just some thoughts on future improvements

package-examples/ingress-nginx/admission-webhook/job-ingress-nginx-admission-patch.yaml
droot: Yes.

Actually, for the kubebuilder based controllers, kubebuilder will read the permissions for the Role from the comments (annotations) defined above the API types (or reconciler function).

droot (commented):
Thanks for the feedback everyone.

package-examples/ingress-nginx/README.md
droot: Good point. I changed it and mapped it to the controller. Though these resources are clusterscoped but belong to the controller.

droot:
> Looks good. I think we can also add the `set-labels` and `set-namespace:v0.3.4` in Kptfile.

We don't need to use `set-namespace` for this because there will be only one add-on per cluster. Configuring anything other than the `ingress-nginx` namespace is going to be exception rather than norm (Note all the docs etc use the default namespace so less likely someone will do it), so will add functions for the most common customizations.

Regarding `set-labels`, yes, this can be done. I will add it in the next iteration. I will have to use function target selector to selectively apply labels because multiple components are involved so want to play with it first.

package-examples/ingress-nginx/admission-webhook/job-ingress-nginx-admission-patch.yaml
droot: So, this will be an imperative operation on this package, right.

Issue #3125, opened 2022-05-11
Figure out how to handle secrets


bgrant0607:
Forking the discussion from https://github.com/GoogleContainerTools/kpt/issues/2433#issuecomment-931896466

For usage in applications, there's https://secrets-store-csi-driver.sigs.k8s.io/. 

For cases where we do need to copy into Kubernetes Secrets, I'd recommend using a similar approach as GitOps, which is copying from the storage system, the secret manager in this case rather than git. In the GitOps layer that could be a special kind of sync request. In the kpt CLI we'll need to figure out how we want to stage secrets before the apply, and ensure we don't leak them.

There's also the issue of generation, for cases where users do want to pre-generate secrets and push them along with their other configuration.

bgrant0607:
Yes, definitely as much as is practical we should recommend automatically rotated secrets: service accounts, workload identity, managed certs, and so on.

bgrant0607:
https://dnastacio.medium.com/why-you-should-avoid-sealed-secrets-in-your-gitops-deployment-e50131d360dd
mentions https://external-secrets.io/

yuwenma:
An example discovered in [Ghost Application package](https://github.com/GoogleContainerTools/kpt/pull/3381/commits/136f777959d49b2947f34232f2636e8127de75db):
 
Two secrets are used. One for ghost-app and the other is for MariaDB. 

The secretes are referred by ghost-app `Deployment` and mariadb `StatefulSet` on `spec.template.spec.containers[].env[].valueFrom.secretKeyRef` fields. 

The secretes are also injected in these two resources' `spec.template.metadata.annotations[].checksum/secrets`. checksum/secrets is a helm feature for [auto roll deployment](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments). This ghost app is originally converted from a helm charts `bitnami/ghost`

bgrant0607:
For completeness, kustomize supports secret generation plugins.
https://github.com/kubernetes-sigs/kustomize/blob/master/examples/secretGeneratorPlugin.md

Example plugin:
https://github.com/viaduct-ai/kustomize-sops

bgrant0607:
External secrets was inducted into the CNCF sandbox:
https://lists.cncf.io/g/cncf-toc/message/7273

bgrant0607:
We probably want a mechanism to help users not commit Secrets to git.

PR #3381, opened 2022-07-17
Add a basic ghost package

yuwenma:
This package is originated from helm chart bitnami/ghost, which uses MariaDB.
By default, the ghost helmchart does not have the ghost, nor does it enable ingress, volumePermissions, networkPolicy, prometheus.

When rendering the ghost helm chart, we include the ghost host (`Deployment` and `PersistentVolumeClaim`), as well as the other services (`NetworkPolicy` `Ingress` `PrometheusRule`). 

Follow up work is tracked in https://github.com/GoogleContainerTools/kpt/issues/3379


package-examples/ghost/Kptfile
```diff
@@ -0,0 +1,12 @@
...
+kind: Kptfile
+metadata:
+  name: ghost
+  annotations:
+    config.kubernetes.io/local-config: "true"
+info:
+  description: sample description
+pipeline:
+  mutators:
+    - image: gcr.io/kpt-fn/set-namespace:v0.3.4
```
droot, line 11: Was curious we aren't using `set-namespace:v0.4.1` ?

package-examples/ghost/ghost/Kptfile
```diff
@@ -0,0 +1,12 @@
+apiVersion: kpt.dev/v1
+kind: Kptfile
+metadata:
+  name: ghost
```
droot, line 4: ghost pkg inside ghost confused me a bit :) Will think of an alternative as I go further in the PR.

package-examples/ghost/ghost/deployment-example-ghost.yaml
```diff
@@ -0,0 +1,108 @@
...
+      annotations:
+        checksum/secrets: 711677f8a87ff3df1d09863858613b5dbecb3f6b8ea51a8c37b40762aff66a20
+    spec:
+      affinity:
+        podAffinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - podAffinityTerm:
+                namespaces:
+                  - "example"
```
droot, line 21: This is interesting for `set-namespace`'s function perspective.

package-examples/ghost/ghost/secret-example-ghost.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: example-ghost
+  namespace: example
+type: Opaque
+data:
+  ghost-password: amRYdHhYYVFueg==
```
droot, line 8: We should not have this secret in this package. 

The Good thing is when user will deploy this package, the deployment that needs this secret will be blocked until the secret is available.

So for now secret can be provisioned out-of-band manually. And in later part of our CUJ, we can add a secret manager to manage the secret. WDYT ?

droot (commented):
Some quick comments. Feel free to merge it and address it as a follow up PR.

package-examples/ghost/ghost/deployment-example-ghost.yaml
```diff
@@ -0,0 +1,108 @@
...
+                topologyKey: kubernetes.io/hostname
+              weight: 1
+        nodeAffinity:
+      securityContext:
+        fsGroup: 1001
+      containers:
+        - name: example-ghost
```
bgrant0607, line 28: I wouldn't prefix the container name. There's no real point to it.

package-examples/ghost/ghost/deployment-example-ghost.yaml
```diff
@@ -0,0 +1,108 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: example-ghost
```
bgrant0607, line 4: Variants are being uniquified by name prefix as well as by namespace name?

package-examples/ghost/ghost/secret-example-ghost.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
bgrant0607, line 1: This is ok for the initial version, but we don't want actual Secret resources in packages, especially not in git. We'll want a configuration that specifies how to generate or access the Secret instead. See #3125

package-examples/ghost/Kptfile
yuwenma: `v0.4.1` does not accept duplicate GVKNN which means using `v0.4.1` users need to specify the set-namespace in the children package rather than using it in the parent. 
FYI, I think that is an unnecessary requirement. So the latest set-namespace (unreleased) removes this requirement.

package-examples/ghost/ghost/deployment-example-ghost.yaml
yuwenma: Yes, so as the namespaceSelector for podAffinityTerm.

package-examples/ghost/mariadb/configmap-example-mariadb.yaml
```diff
@@ -0,0 +1,36 @@
+# Source: ghost/charts/mariadb/templates/primary/configmap.yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: example-mariadb
+  namespace: example
+data:
+  my.cnf: |-
```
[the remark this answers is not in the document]
yuwenma: Yes! This package gives the example for issues  
https://github.com/GoogleContainerTools/kpt/issues/3118
https://github.com/GoogleContainerTools/kpt/issues/3119
https://github.com/GoogleContainerTools/kpt/issues/3210 (Augeas part)

package-examples/ghost/ghost/deployment-example-ghost.yaml
```diff
@@ -0,0 +1,108 @@
...
+  namespace: example
+spec:
+  replicas: 1
+  strategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      annotations:
+        checksum/secrets: 711677f8a87ff3df1d09863858613b5dbecb3f6b8ea51a8c37b40762aff66a20
```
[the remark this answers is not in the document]
bgrant0607: https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

package-examples/ghost/package-context.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kptfile.kpt.dev
```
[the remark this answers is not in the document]
yuwenma: Both `kpt pkg init` and `kpt pkg get --for-deployment` generate the package-context.yaml. `kpt pkg get` won't.

This package-contex.yaml is generated by `kpt pkg init`, which is how we suggest users to create blueprint from scratch in kpt CLI https://kpt.dev/guides/namespace-provisioning-cli?id=initialize-a-package. @droot shall we change kpt on `kpt pkg init` to not creating `package-context.yaml`? But then it will break the local testing. Or we can add one more step to delete this files manually.

package-examples/ghost/mariadb/configmap-example-mariadb.yaml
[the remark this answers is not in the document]
bgrant0607: Yep.

If this is INI format, I'd change the file suffix to .ini and change the corresponding config mount in the StatefulSet.

package-examples/ghost/mariadb/package-context.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kptfile.kpt.dev
+  annotations:
+    config.kubernetes.io/local-config: "true"
+data:
+  name: example
```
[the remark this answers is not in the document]
bgrant0607: A user in slack has encountered similar issues using Config Connector subpackages (e.g., CloudSQL).

package-examples/ghost/ghost/setlabels.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: fn.kpt.dev/v1alpha1
+kind: SetLabels
+metadata:
+  name: ghost
+  annotations:
+    config.kubernetes.io/local-config: "true"
+labels:
```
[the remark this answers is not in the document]
yuwenma: 😞 no spec. I think all custom functionConfig type do not follow the `spec` `status` rule. The argument is that those resources won't be deployed to the cluster so it is not a strict spec/status. Skipping spec saves some manual editing. 

I'm more happy to add the `spec` to `functionConfig`, because it does help answering questions like "what fields & resources a user could be interested in modifying"  https://kubernetes.slack.com/archives/C0155NSPJSZ/p1657561360040189

package-examples/ghost/ghost/secret-example-ghost.yaml
droot: We discussed and decided to treat secret as an external resources (not managed by our stack for now). So secrets will not be part of the package.

package-examples/ghost/ghost/secret-example-ghost.yaml
droot: This is discussed offline...treating secret as an external resources.

package-examples/ghost/ghost/setlabels.yaml
```diff
@@ -0,0 +1,10 @@
...
+  app.kubernetes.io/name: ghost
+  app.kubernetes.io/instance: example
```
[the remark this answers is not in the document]
droot: Yeah, I was wondering if the current package wires using the ApplyReplacements or Starlark.

package-examples/ghost/package-context.yaml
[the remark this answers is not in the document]
droot: @bgrant0607 your understanding is correct and current behavior is inline with that.

@yuwenma . No, we don't have to change the `kpt pkg init` behavior. 

Listing the scenario that we need to confirm:

kpt pkg get <ghost-super-pkg> app-1 --for-deployment

1. the app-1/package-context.yaml will contain `data.name: app-1`
2. app-1/ghost/package-context.yaml  ??
3. app-2/mariadb/package-context.yaml ??

And when you run `kpt fn render app-1`, the set-namespace will use the package-context at the top-level, so everything should be fine, but we need to confirm this.

droot (commented):
We discussed this offline and I captured the notes here.

package-examples/ghost/ghost/Kptfile
yuwenma: chose ghost-app. Updated.

package-examples/ghost/package-context.yaml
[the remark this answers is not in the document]
yuwenma: I added the full steps in README.yaml and approved this works well:

`kpt pkg get --for-deployment` will create package-context.yaml under each sub directory as well as the parent directory, with the directory name as the `data.name`, which are `ghost-app`, `mariadb` `root-dir`. Running set-namespace on root updates all namespaces to the value in `root-dir` package context `data.name`. Then, `kpt live apply root-dir` deploys all resources to the cluster successfully

yuwenma (commented):
Package updated, please take another look @bgrant0607 @droot 
Some key changes:
- Removed secrets and the references (env in pod template, checksum/secret annotation, livenessProbe).
- Added CRD for PrometheusRule under `crds`, this should be the only CRD needs to install.
- Dropped name prefix `example-` and updated the file names.
- Dropped the `app.kubernetes.io/instance` label, I leave a question about the kpt equivalence.
- Updated README to guide users get, hydrate and deploy the package.

package-examples/ghost/ghost/setlabels.yaml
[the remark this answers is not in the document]
bgrant0607: On the kpt side, the deployment package name is similar to the release name.

package-examples/ghost/ghost/Kptfile
bgrant0607: I've been thinking we need to provide the full package path of any subpackage as a string, such as toppkg-middlepkg-leafpkg. Maybe that's what should populate the subpackage package-context. Or maybe we should add an additional field to it containing the full name.

For example, if the top-level package name were "ghost" and this subpackage's name were "app", then then full name would be "ghost-app".

package-examples/ghost/ghost/deployment-example-ghost.yaml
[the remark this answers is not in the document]
bgrant0607: I prefer the way kustomize does this, by generating a new ConfigMap or Secret with a hash appended to the name. If the ConfigMap or Secret is modified in place, then rollout and rollback can't be controlled predictably.

That said, this really should be done server-side at this point.

bgrant0607:
Thanks. LGTM if the application works.

package-examples/ghost/README.md
```diff
@@ -0,0 +1,32 @@
...
+"Ghost is a powerful app for new-media creators to publish, share, and grow a business around their content. It comes with modern tools to build a website, publish content, send newsletters & offer paid subscriptions to members."
+https://ghost.org/
+
+### Quick start
+
+#### Get KPT Pacakge
+```bash
+export NAMESPACE=<YOUR NAMESPACE>
+# make sure the namespace is correct and exists. Otherwise, create the namespace
+kubectl create namespace ${NAMESPACE}
```
droot, line 12: why can't we add namespace to the package itself ? 

If you used helm chart to get the manifests, you can use `--create-namespace` arg to add it.

That will make the workflow truly declarative.

droot (approved):
I have a minor comment, but it is good to go.

package-examples/ghost/README.md
bgrant0607: The ability to create a namespace would imply that the app team had cluster admin privilege. It also would bring in the need for the flexibility of the namespace provisioning scenario we've been working on: resource quota, limit range, network policy, etc.

bgrant0607: We should have an example namespace blueprint for this, which could be used instead of kubectl create namespace.

droot: > The ability to create a namespace would imply that the app team had cluster admin privilege. It also would bring in the need for the flexibility of the namespace provisioning scenario we've been working on: resource quota, limit range, network policy, etc.

Aah.. completely missed that. Good pt.

Issue #3365, opened 2022-07-13
All in-cluster components should be installed using kpt


bgrant0607:
Porch, Backstage, Config Sync.

Installation instructions don't use kpt.
https://kpt.dev/installation/

Porch downloads a tarball and does kubectl apply.
https://kpt.dev/guides/porch-installation?id=run-released-version-of-porch

Config Sync is similar:
https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/docs/installation.md#installing-a-released-version

The Backstage plugin doesn't have a comparable release yet.

PR #3388, opened 2022-07-20
Fix the WI docs to include annotating the KSA

johnbelamaric:
The step of annotating the KSA was missing from the documentation for setting up Porch with Workload Identity.

### DIFF
--- a/site/guides/porch-installation.md
+++ b/site/guides/porch-installation.md
@@ -176,7 +176,7 @@ To set it up, create the necessary service accounts and give it the required rol
 ```sh
 GCP_PROJECT_ID=$(gcloud config get-value project)
 
-# Create GCP service account for Porch server.
+# Create GCP service account (GSA) for Porch server.
 gcloud iam service-accounts create porch-server
 
 # We want to create and delete images. Assign IAM roles to allow repository
@@ -188,6 +188,11 @@ gcloud projects add-iam-policy-binding ${GCP_PROJECT_ID} \
 gcloud iam service-accounts add-iam-policy-binding porch-server@${GCP_PROJECT_ID}.iam.gserviceaccount.com \
   --role roles/iam.workloadIdentityUser \
   --member "serviceAccount:${GCP_PROJECT_ID}.svc.id.goog[porch-system/porch-server]"
+
+# We need to associate the Kubernetes Service Account (KSA)
+# with the GSA by annotating the KSA.
+kubectl annotate serviceaccount porch-server -n porch-system \
+  iam.gke.io/gcp-service-account=porch-server@${GCP_PROJECT_ID}.iam.gserviceaccount.com
 ```
 
 Build Porch, push images, and deploy porch server and controllers using the


site/guides/porch-installation.md
```diff
@@ -188,6 +188,11 @@ gcloud projects add-iam-policy-binding ${GCP_PROJECT_ID} \
 gcloud iam service-accounts add-iam-policy-binding porch-server@${GCP_PROJECT_ID}.iam.gserviceaccount.com \
   --role roles/iam.workloadIdentityUser \
   --member "serviceAccount:${GCP_PROJECT_ID}.svc.id.goog[porch-system/porch-server]"
+
+# We need to associate the Kubernetes Service Account (KSA)
+# with the GSA by annotating the KSA.
+kubectl annotate serviceaccount porch-server -n porch-system \
```
bgrant0607, line 194: Could we use kpt to add this prior to deployment?

I'd like to start using kpt to configure and deploy our components. #3365

We have a set-annotations function.
https://catalog.kpt.dev/set-annotations/v0.1/

It would also be nice to eventually have a dedicated function for this that reads from the gcloud config.
https://github.com/GoogleContainerTools/kpt/pull/2715/files

cc @yuwenma

johnbelamaric: The whole thing should move to a kpt package.

johnbelamaric: Yes agree a dedicated function makes sense, to get the GSA name and do the annotation, based on the gcloud config.

bgrant0607: Ok, fair. We can do this then.

johnbelamaric: This does illustrate another dependency between apps being deployed and infrastructure - this time the right cloud service accounts. This is an area of significant pain when deploying apps and so has a lot of opportunity for an improved UX.

johnbelamaric: Kind of thinking aloud, I could imagine two packages - the app package itself, and a separate package containing the cloud provider SAs and associated roles and role bindings. When deploying the app to a given cloud provider, you deploy the associated cloud provider app to your infra cluster (CC in the case of GCP). We would want to somehow capture the dependency, and run the appropriate version of the workload identity function (the annotation names and values are different for different cluster providers).

mortent: I completely agree we need to provide kpt packages for our components. We should also look at hosting and managing our kpt packages through a porch setup.

Allowing users to use the gcloud information from the environment seems like a good solution here, although we also need a solution for users where they might need to specify a different project id. Similarly, users might want to use a different SA than the one provided. If so, it might be that editing the configuration directly is the best way to do it.

For porch setting up workload identity is something that will have to be configured on setup, while usage of this happens when someone registers a repo that can use WI as the auth method. For other auth schemes the user will provide the auth information when registring the repo.

johnbelamaric:
Yes, I did test it and it worked. Alas 16 hours later #3390

mortent:
Looked a bit more into this. Which command are you running? The `porch/scripts/create-deployment-blueprint.sh` scripts [should add the annotation to the SA](https://github.com/GoogleContainerTools/kpt/blob/940e3bb35cfe91eddcb7bd871cc0e8269179c5d2/porch/scripts/create-deployment-blueprint.sh#L104), and it is used by the `push-and-deploy` make target.  Do you have a project configured in your gcloud (i.e. `gcloud config get-value project` returns something)?

johnbelamaric:
I just followed the documentation. I didn't use a script or a make target. 

https://kpt.dev/guides/porch-installation?id=workload-identity

That section is, seemingly erroneously, under "Running a Custom Build of Porch", which I am not doing. I probably should have moved it out of that section. It really belongs in the section on registering repos (so does the OCI part). But I didn't want to undertake reorganizing all the docs, I just wanted to push a quick fix out.

johnbelamaric:
I see that this section is perhaps intended to be in there with the make push-and-install - but I don't see using WI and building a custom porch as related. Maybe we do need to re-org the docs rather than just add this line.

site/guides/porch-installation.md
yuwenma: +1. I can see how much happy users will experience from getting rid of this painful KSA, GSA, WI config. Even more, if config connector is installed, the previous steps like GSA creation, IAM binding can be managed by kpt (in the form of CNRM resources). I think @justinsb has done some prior art about that.

mortent:
Yeah, we need to restructure the docs a bit. But lets merge this now.

PR #3394, opened 2022-07-20
Deployable ghost pacakge

yuwenma:
This PR contains several changes to the ghost package:

1.  It adds a manual step which updates the ghost host IP defined in `Deployment` container env var after the first `kpt live apply`. The new ghost host IP is generated by Ghost `Service` which is LoadBalancer type. Later on, we can change that to use ingress (external package).
2. It removes the PodAffinity from both `StatefulSet` and `Deployment`. This is mainly to get rid of the step to update the namespaces in `podAffinityTerm`     
3. Added back the set-label functions to add `name`, which is required for deployment and statefulset. 
4. Turn on the bitnami ghost debug mode. It's harder to figure out the error on pod level.


droot (approved):
Thanks for getting this working.

PR #3398, opened 2022-07-21
flatten ghost package with better envs handling

yuwenma:
Related to https://github.com/GoogleContainerTools/kpt/issues/3343. 

The nested structure for ghost package is unnecessary. 

Nested package actually makes user configuration harder: 
`Deployment` and `StatefulSet` `spec.template.spec.containers[].env` have shared env configurations like `BITNAMI_DEBUG` `ALLOW_EMPTY_PASSWORD`. Users today need to update the config twice. This progress can be much simplified (and better protected) if we can change the `containers[].env` to use `configMapKeyRef`. 

This PR also fixes the MariaDB config name `my.cnf`, This file name is a MariaDB convention [doc](https://mariadb.com/kb/en/configuring-mariadb-with-option-files/#default-option-file-locations-on-linux-unix-mac)


justinsb (approved):
I like this direction!

package-examples/ghost/configmap-mariadb-ini.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: mariadb
   namespace: example
 data:
-  my.ini: |-
+  my.cnf: |-
```
droot, line 8: I remember some discussion about keepting the extension to `ini`. Is it intended change ?

package-examples/ghost/statefulset-mariadb.yaml
```diff
@@ -21,15 +21,9 @@ spec:
...
-              value: "true"
-            - name: MARIADB_USER
-              value: "bn_ghost"
-            - name: MARIADB_DATABASE
-              value: "bitnami_ghost"
-            - name: ALLOW_EMPTY_PASSWORD
-              value: "true"
+          envFrom:
+            - configMapRef:
+                name: ghost-env
```
droot, line 26: One disadvantage of using the same configmap here is that any config change to that configmap affects restarting ghost-app and mysql. ghost-app config is superset of mysql. Like changing the ghost-app's host is going to restart the mysql for now reason.

So an optimal solution could be, define two config-maps. one for db config and another for pure ghost-app config.

ghost-app's config will be provided by both the config-map and mysql one will be powered by just the db configmap.

package-examples/ghost/configmap-mariadb-ini.yaml
yuwenma: yeah, I added the reason in the PR description. The `my.cnf` is [MariaDB default config file name](https://mariadb.com/kb/en/configuring-mariadb-with-option-files/#default-option-file-locations-on-linux-unix-mac)

package-examples/ghost/statefulset-mariadb.yaml
yuwenma: > One disadvantage of using the same configmap here is that any config change to that configmap affects restarting ghost-app and mysql. ghost-app config is superset of mysql. Like changing the ghost-app's host is going to restart the mysql for now reason.
> 

That's true. Restarting both to align the env change is expected behavior. One issue I had today is that I only updated the env config for MariaDB (so we saw the MariaDB pod works) but not Ghost host. 

> So an optimal solution could be, define two config-maps. one for db config and another for pure ghost-app config.
> 
> ghost-app's config will be provided by both the config-map and mysql one will be powered by just the db configmap.

Sure, I can use two ConfigMaps since the `envFrom` accept a list

droot: Actually thought some more and realized, this looks nice from config organization perspective but functionally is inferior to previous approach of `inline config` in the deployment itself. Because now I have to worry about updating the deployment or statefulset otherwise new change will not rollout and in the previous approach it will happen automatically :)

yuwenma: Another thing I discovered is that the `envFrom` [cannot be updated](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2355). This seems not to be the model we are looking for.

Issue #3356, opened 2022-07-11
Come up with a short, descriptive tag line for the site


bgrant0607:
The current description is considered [by some](https://www.youtube.com/watch?v=Wavigyq7NdQ) to be long and academic.

A shorter, pithier but somewhat descriptive tag line for the landing page and repo description would be useful. We can put the longer, more accurate description deeper in the docs.

Kustomize's tag line was "template-free customization".

What should kpt's be?
* Configuration as Data
* WYSIWYG config
* Automate your configuration
* API-driven configuration
* KRM-native configuration
* KRM Platform Toolkit
Other ideas???

cc @justinsb @selfmanagingresource @droot @mortent

selfmanagingresource:
Well, he's not wrong - it is a bit long! Another question is whether we should mentioned GitOps in the short-tag line or will it be too much? 

A few thoughts:
- Of the ones listed above I think something akin to "automate your configuration for scale" is closest. WYSIWYG describes more the result you get with the toolkit. We've leaned in a lot into WYSIWYG for discussing what we enable that is challenging today, but other projects don't really focus on that in their tag lines.
- KRM Platform Toolkit has the benefit of actually explaining what KPT stands for.
- We can go all in on GitOps. Kubernetes GitOps Automation Toolkit

justinsb:
How about: "Kubernetes-first packages".  To me, the core idea is that we're embracing kubernetes ideas and kubernetes objects and kubernetes schemas.  We're different from the tooling that tries to build an abstraction layer that hides kubernetes - because we've found that these abstraction layers can be counterproductive.

bgrant0607:
Along with the tag line, we should explain the value prop. We tried to explain what CaD is and what the motivation was, but that's not the same as the user value, by role: platform team, app ops, app dev.

bgrant0607:
More ideas:
* Full-lifecycle configuration automation
* Day 0 to N configuration automation
* Operate on your configuration
* Generate, transform, validate
* The API for configuration
* Control and data plane for configuration
* Accelerate configuration management

I don't think we want users to think about kpt as just another package tool.

bgrant0607:
Another:
* The Configuration Automation Platform

Thinking about the GitOps angle:
* The frontend for GitOps
* The GitOps Platform

bgrant0607:
More:
* Simplify, accelerate, and scale configuration management and governance
* Low-code configuration management
* Transform configuration management

[...]

PR #3408, opened 2022-07-22
Improve explanations of what kpt is.

bgrant0607:
As discussed in #3356.

### DIFF
--- a/site/faq/README.md
+++ b/site/faq/README.md
@@ -6,12 +6,30 @@ Please visit the [roadmap document] and the [kpt milestones].
 
 ### How is kpt different from other solutions?
 
-Rather than expressing configuration as code that generates configuration,
-kpt represents [Configuration as Data], in particular as YAML or JSON objects
-adhering to [The Kubernetes Resource Model], and uses a transformation-based
-approach to customization. This enables interoperability of a variety of generators,
+Think of configuration as an API or data in a database. kpt can operate on configuration
+in storage, git or OCI.
+
+Rather than expressing configuration AS code or templates that generate configuration,
+kpt represents [Configuration as Data]. In particular, it represents configuration
+as YAML or JSON objects adhering to [The Kubernetes Resource Model], the same as the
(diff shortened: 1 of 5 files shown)


johnbelamaric:
This is definitely much better. It still reads a bit academic in some places - there's a fair number of assumptions about background knowledge in statements like "This enables interoperability of a variety of generators, transformers, and validators. One doesn’t have to make all changes through a monolithic generator implementation."

Similarly, I think WYSIWYG is great for those of use that have been around a while, but I am not sure most people (especially worldwide) are all that familiar with the term.

None of this blocks the PR, which like I said is much better. But I think we can do a lot more to make this clearer.

bgrant0607:
Thanks @johnbelamaric. I agree with you. I was trying to just update the tagline, add the explanation from my recent twitter thread, and make a few other edits in the time I had. 

I'd be happy to replace WYSIWYG with a term more widely in the current decade, but haven't thought of a good replacement yet. Suggestions?

We also need a glossary.

site/faq/README.md
```diff
@@ -6,13 +6,31 @@ Please visit the [roadmap document] and the [kpt milestones].
...
+
+kpt uses an in-place transformation approach to customization: 
+read configuration in, modify it, and write it back.
+
+This enables interoperability of a variety of generators,
 transformers, and validators. One doesn’t have to make all changes through a monolithic
 generator implementation.
 
+By storing the result of config generators and transformers, automated mutations can be separated in time
+from use and other modifications. This enables generation via a UI, for example.
```
droot, line 25: `separated in time` is difficult to grok :) we might need an example to explain it.

droot (approved):
This is an area where we will keep on iterating. This is definitely an improvement over what we had, so I am happy merging this.

yuwenma (approved):

site/faq/README.md
bgrant0607: Ack

PR #3410, opened 2022-07-25
Fix refresh of auth tokens for WI

mortent:
This updates the logic for communicating with git so that it will check the expiration time for the auth token and request a new one if it expires in less than a minute.

Fixes: https://github.com/GoogleContainerTools/kpt/issues/3390

### DIFF
--- a/porch/pkg/git/git.go
+++ b/porch/pkg/git/git.go
@@ -624,7 +624,7 @@ func (r *gitRepository) getAuthMethod(ctx context.Context) (transport.AuthMethod
 	r.mutex.Lock()
 	defer r.mutex.Unlock()
 
-	// If not secret is provided, we try without any auth.
+	// If no secret is provided, we try without any auth.
 	if r.secret == "" {
 		return nil, nil
 	}
--- a/porch/pkg/git/git.go
+++ b/porch/pkg/git/git.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	kptfilev1 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
@@ -33,7 +34,6 @@ import (
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/object"
 	"github.com/go-git/go-git/v5/plumbing/transport"
-	"github.com/go-git/go-git/v5/plumbing/transport/http"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/trace"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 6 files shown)


porch/pkg/registry/porch/secret.go
```diff
@@ -209,3 +197,24 @@ func (g *GcloudWIResolver) lookupGSAEmail(ctx context.Context, name, namespace s
...
+
+type GcloudWICredential struct {
+	token *oauth2.Token
+}
+
+var _ repository.Credential = &GcloudWICredential{}
+
+func (b *GcloudWICredential) Valid() bool {
+	timeLeft := time.Until(b.token.Expiry)
+	if timeLeft > time.Minute {
```
johnbelamaric, line 209: How long to these tokens typically last? If they are long I don’t see why we shouldn’t give ourselves more like an hour here. One minute doesn’t give enough time for retries on some temporary failure.

mortent: The tokens seems to last for an hour from what I can see. 

I did originally use five minutes here, but I couldn't think of any reason why we would need such a long grace period. For retries we will do this check for every attempt, so it seems like it is mostly time diffs that could potentially cause a problem here.

I'm ok with increasing the value somewhat, I'm just not sure if it is needed.

johnbelamaric:
I am imagining a short outage - say 5-10 minutes - in the infra for renewing the token. If the tokens lasted 12 hours or something and we had an hour buffer then such an outage would never impact that user, since we would have ample opportunity to refresh the token in that hour.

Since the tokens are so short-lived already, obviously we can't do an hour. But I think 5 minutes still reduces the likelihood of an upstream outage impacting our user.

porch/pkg/git/git.go
```diff
@@ -614,33 +618,26 @@ func (r *gitRepository) dumpAllRefs() {
...
+func (r *gitRepository) getAuthMethod(ctx context.Context) (transport.AuthMethod, error) {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
 
-	return &http.BasicAuth{
-		Username: string(username),
-		Password: string(password),
-	}, nil
-}
+	// If not secret is provided, we try without any auth.
```
justinsb, line 627: Nit: If no secret

justinsb (approved):

justinsb:
This LGTM.  Another strategy (which we could still use) is to automatically retry/reload whenever we get an error.

johnbelamaric:
> This LGTM. Another strategy (which we could still use) is to automatically retry/reload whenever we get an error.

Yes, agreed that we should be doing that as well, probably with exponential backoff...

mortent:
I did look into refreshing the tokens on error, but it seems like CSR returns a 400 if the token has expired rather than 401. So it turned out to be difficult to distinguish expired token from other errors.

I did implement refresh on error in https://github.com/GoogleContainerTools/kpt/pull/3411 when we get a 401. This works for github repos. Didn't implement the exp backoff in the PR, but did leave a TODO that we probably should add it: https://github.com/GoogleContainerTools/kpt/pull/3411/files#diff-f54c4501c5bde751d8aeb60dd4d5e484600f62e58b1c4198e0493a45f3378ecdR205

johnbelamaric:
*If I understand the whole of the code here*, we would hit this error in response to a user action, and retry on failure? In that case backoff isn't so critical, it's not going to hammer the upstream with failed attempts.

mortent:
@johnbelamaric Porch does have a job that syncs against git, so if the token refresh doesn't work, we will be retrying outside of user actions. I will try to update this PR with logic to do backoff.

mortent:
@justinsb @johnbelamaric Updated this PR to do backoff when requesting new WI tokens.

PR #3411, opened 2022-07-25
Refresh creds on auth fail

mortent:
This causes porch to refresh the auth credentials from the k8s secret if it gets an auth error when talking to git.

This must be merged after https://github.com/GoogleContainerTools/kpt/pull/3410

Fixes: https://github.com/GoogleContainerTools/kpt/issues/3406

### DIFF
--- a/porch/pkg/git/draft.go
+++ b/porch/pkg/git/draft.go
@@ -16,6 +16,7 @@ package git
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


justinsb (approved):

justinsb:
Nice - you implemented the "retry on error" logic :-)

PR #3412, opened 2022-07-25
docs: added instructions for tagging the package before deploying 

### LINKED ISSUE #3378: namespace-provisioning-cli tutorial needs to tag the deployment package
https://kpt.dev/guides/namespace-provisioning-cli doesn't tag the deployment package, so the package won't be visible via porch and the UI. Both the blueprint and the deployment package should be tagged with porch-compatible revisions.

droot:
Added instructions to tag the package before deploying so that it is visible in the CaD UI.

fixes #3378


droot:
@yuwenma can you pl. take a quick look at this one.

yuwenma (approved):
looks great!

Issue #3405, opened 2022-07-22
kpt-set mark removed from line after update to 1.0.0-beta.18


RafalMaleska:
### Expected behavior

`kpt pkg update`
 
```
apiVersion: v1
kind: Namespace
metadata: # kpt-merge: /some-app
  name: some-app # kpt-set: ${some-app}
  labels:
  annotations:
    ncp/static_snat_ip: 122.122.122.122 # kpt-set: ${gateway-snat-ip}
```

### Actual behavior

`kpt pkg update`

```
apiVersion: v1
kind: Namespace
metadata: # kpt-merge: /some-app
  name: some-app # kpt-set: ${some-app}
  labels:
  annotations:
    internal.kpt.dev/upstream-identifier: '|Namespace|default|some-app'
    ncp/static_snat_ip: 122.122.122.122
```

### Information

kpt version: 1.0.0-beta.18

After updating kpt to the newest version and using `kpt pkg update` a new annotation is added (as stated in the release notes). For some reason if there is another annotation which needs to be modified by a kpt function (kpt-set). The function mark is removed and the pipeline is ignoring this resource. 

Briefly: 

- before update the last line looked after kpt pkg update like: 
```
    ncp/static_snat_ip: 122.122.122.122 # kpt-set: ${gateway-snat-ip}
```
- after update to the newest version the last line is missing the # kpt-set

```
    ncp/static_snat_ip: 122.122.122.122 
```

Is it possible to not use the new kpt annotation? 
Is it a bug of kpt which causes the removal of this kpt-set mark?

### Steps to reproduce the behavior

kpt pkg update on an upstream resource which has an kpt-set function on an annotation. Tested this only with the namespace resource.

mortent:
This is probably related to https://github.com/GoogleContainerTools/kpt/pull/3360.
@yuwenma

yuwenma:
Thanks for reporting this! I'll take a look today.

yuwenma:
@RafalMaleska The fix should go in soon. 
Just a suggestion about your `ncp/static_snat_ip` annotation, `set-labels` is a preferred function than setters. 
You can try something like
```yaml
# Kptfile
pipeline:
    mutators:
         - image: gcr.io/kpt-fn/set-labels:v0.1.5
            configMap:
                 ncp/static_snat_ip: "122.122.122.122"
```
Or 
```yaml
# Kptfile
pipeline:
    mutators:
[...]

PR #3413, opened 2022-07-25
fix: preserve annotation comments

yuwenma:
Fix https://github.com/GoogleContainerTools/kpt/issues/3405
Comments are dropped when calling `rnode.SetAnnotations`. This PR adds back the comments after adding the upstream-identifier, and also sort the order.

From the example in #3405, before adding the upstream annotation we have
``` yaml
kind: Namespace
metadata: # kpt-merge: /some-app
  name: some-app # kpt-set: ${some-app}
  labels:
  annotations:
    ncp/static_snat_ip: 122.122.122.122 # kpt-set: ${gateway-snat-ip}
    config.kubernetes.io/index: '0'
    config.kubernetes.io/path: 'namespace.yaml'
    internal.config.kubernetes.io/index: '0'
    internal.config.kubernetes.io/path: 'namespace.yaml'
    internal.config.kubernetes.io/seqindent: 'compact'
    internal.config.kubernetes.io/annotations-migration-resource-id: '0'
```
After adding upstream annotation
```yaml
apiVersion: v1
kind: Namespace
metadata: # kpt-merge: /some-app
  name: some-app # kpt-set: ${some-app}
  labels:
  annotations:
    ncp/static_snat_ip: 122.122.122.122 # kpt-set: ${gateway-snat-ip}
    config.kubernetes.io/index: "0"
    config.kubernetes.io/path: namespace.yaml
    internal.config.kubernetes.io/index: "0"
    internal.config.kubernetes.io/path: namespace.yaml
    internal.config.kubernetes.io/seqindent: compact
    internal.config.kubernetes.io/annotations-migration-resource-id: "0"
    internal.kpt.dev/upstream-identifier: '|Namespace|default|some-app'
``` 

if not sort annotations, the output can be annoying to users in git diff.  
```yaml
apiVersion: v1
kind: Namespace
metadata: # kpt-merge: /some-app
  name: some-app # kpt-set: ${some-app}
  labels:
  annotations:
    config.kubernetes.io/index: "0"
    config.kubernetes.io/path: namespace.yaml
    internal.config.kubernetes.io/annotations-migration-resource-id: "0"
    internal.config.kubernetes.io/index: "0"
    internal.config.kubernetes.io/path: namespace.yaml
    internal.config.kubernetes.io/seqindent: compact
    internal.kpt.dev/upstream-identifier: '|Namespace|default|some-app'
    ncp/static_snat_ip: 122.122.122.122 # kpt-set: ${gateway-snat-ip}
```

### DIFF
--- a/internal/util/addmergecomment/addmergecomment.go
+++ b/internal/util/addmergecomment/addmergecomment.go
@@ -71,7 +71,6 @@ func Process(paths ...string) error {
 // One known caveat is that upstream meta change can cause downstream origin mismatch. This potentially causes the 3-way merge
 // to fail in pkg update step.
 func addUpstreamAnnotation(object *kyaml.RNode, mergeComment string) error {
-	annotations := object.GetAnnotations()
 	group, _ := resid.ParseGroupVersion(object.GetApiVersion())
 	var name, namespace string
 	if strings.Contains(mergeComment, merge.MergeCommentPrefix) {
@@ -90,8 +89,8 @@ func addUpstreamAnnotation(object *kyaml.RNode, mergeComment string) error {
 	} else if object.GetNamespace() == resid.TotallyNotANamespace {
 		namespace = unknownNamespace
 	}
-	annotations[upstreamIdentifier] = fmt.Sprintf(upstreamIdentifierFmt, group, object.GetKind(), namespace, name)
-	return object.SetAnnotations(annotations)
+	upstreamIdentifierValue := fmt.Sprintf(upstreamIdentifierFmt, group, object.GetKind(), namespace, name)
+	return object.PipeE(kyaml.SetAnnotation(upstreamIdentifier, upstreamIdentifierValue))
 }
 
 // Filter implements kyaml.Filter
--- a/internal/util/addmergecomment/addmergecomment_test.go
+++ b/internal/util/addmergecomment/addmergecomment_test.go
@@ -47,7 +47,7 @@ metadata: # kpt-merge: my-space/nginx-deployment
   name: nginx-deployment
   namespace: my-space
   annotations:
-    internal.kpt.dev/upstream-identifier: apps|Deployment|my-space|nginx-deployment
+    internal.kpt.dev/upstream-identifier: 'apps|Deployment|my-space|nginx-deployment'
 spec:
   replicas: 3
  `,
@@ -70,7 +70,7 @@ metadata: # kpt-merge: default/nginx-deployment
   name: nginx-deployment
   namespace: default
   annotations:
-    internal.kpt.dev/upstream-identifier: apps|Deployment|default|nginx-deployment
+    internal.kpt.dev/upstream-identifier: 'apps|Deployment|default|nginx-deployment'
 spec:
   replicas: 3
  `,
@@ -91,7 +91,7 @@ kind: Deployment
 metadata: # kpt-merge: /nginx-deployment
   name: nginx-deployment
   annotations:
-    internal.kpt.dev/upstream-identifier: apps|Deployment|default|nginx-deployment
+    internal.kpt.dev/upstream-identifier: 'apps|Deployment|default|nginx-deployment'
 spec:
   replicas: 3
  `,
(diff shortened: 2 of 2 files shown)


mortent:
So the root cause of the issues both with ordering and the comments is that we read the annotations of the resource into a golang map, which drops comments and provide no guarantee about preserving the order. Wouldn't it be easier to just use the kyaml library to make the changes directly on the kyaml structure?

justinsb:
@mortent do you mean using something like KubeObject?  https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/33aee7fe8a2e1bdf29d9455b5f0c919018432620/go/fn/object.go#L701

yuwenma:
@mortent yes, the KubeObject [GoogleContainerTools/kpt-functions-sdk@33aee7f/go/fn/object.go#L701](https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/33aee7fe8a2e1bdf29d9455b5f0c919018432620/go/fn/object.go#L701) can add annotations easily and preserve comments. Since we decide to not have kpt-functions-sdk/go/sdk specifically support RNode, calling KubeObject methods would require serializing the RNode and then parse it to a KubeObject. This seems not make the code any simpler though. 
Or are you suggesting to use the [YNode operation](https://github.com/GoogleContainerTools/kpt/blob/94bc0853629e9360cc5f5c390085ed1819604d5e/internal/util/addmergecomment/addmergecomment.go#L107)? I'm not sure if the kyaml is intended to let users use YNode this way. Mapping between a map or struct object to a slice (where the odd index is for key and the even index is for value) to preserve the comments, is more like some internal methods to better serve RNode, I think it can cause some unaware risk conditions and can be error prone. Some evidences to prove my assumption:
1. In kustomize, YNode is only called inside kyaml library, and not referred in the kustomize/kustomize and kustomize/api libraries where the kyaml is mainly used for.
2. We found a mis-use of MapNode which is converted from RNode via YNode operations, due to the unawareness of the meta resource itself.   https://github.com/GoogleContainerTools/kpt/blob/94bc0853629e9360cc5f5c390085ed1819604d5e/internal/util/addmergecomment/addmergecomment.go#L109-L122   

Based on this, I'm more leaning towards calling the RNode functions directly here.  I know kyaml is heavily used in kpt, I'm happy to discuss further plans if we want to switch to kpt-functinos-sdk/go/fn

yuwenma:
> @yuwenma I think using the RNode [Pipe/PipeE](https://github.com/kubernetes-sigs/kustomize/blob/2f2ba40876b3b6c5b33281e8dee503010a1bc537/kyaml/yaml/rnode.go#L333) functions and the [filters](https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/yaml/fns.go?rgh-link-date=2022-07-26T20%3A52%3A16Z) provided with kyaml should allow us to do this. The [implementation of the `SetAnnotation`](https://github.com/kubernetes-sigs/kustomize/blob/2f2ba40876b3b6c5b33281e8dee503010a1bc537/kyaml/yaml/rnode.go#L543) function actually does this in the implementation, but we can do this without going via the go map

I'm not a big fan of the kyaml PipeE functions in general especially since we have a preferred approach: the KRM resource centric KubeObject SDK , but I don't have a strong opinion in this case either. Updated the function according to @mortent suggestions.

internal/util/addmergecomment/addmergecomment.go
```diff
@@ -90,8 +89,9 @@ func addUpstreamAnnotation(object *kyaml.RNode, mergeComment string) error {
 	} else if object.GetNamespace() == resid.TotallyNotANamespace {
 		namespace = unknownNamespace
 	}
-	annotations[upstreamIdentifier] = fmt.Sprintf(upstreamIdentifierFmt, group, object.GetKind(), namespace, name)
-	return object.SetAnnotations(annotations)
+	upstreamIdentifierValue := fmt.Sprintf(upstreamIdentifierFmt, group, object.GetKind(), namespace, name)
+	_, err := object.Pipe(kyaml.SetAnnotation(upstreamIdentifier, upstreamIdentifierValue))
```
mortent, line 93: You can just use the `PipeE` function if you just want the error return value.

mortent (approved):

internal/util/addmergecomment/addmergecomment.go
yuwenma: Thanks! Done

PR #3414, opened 2022-07-25
docs: fix kustomize link on kpt rationale guide

ChristopherFry:
This pull request fixes the broken Kustomize link on the kpt rationale guide.


bgrant0607 (approved):
Thanks for the update. Looks like consolidation of kustomize docs has made progress.

PR #3416, opened 2022-07-26
Introduce alpha apply command

justinsb:
First task is a more compact table view, so start by duplicating the table printer logic.


mortent (approved):

karlkfi:
few questions:
- how is this an improvement?
- why is this in an alpha command?
- why isn’t this in cli-utils with the rest of the apply output code?
- got example output?

PR #3425, opened 2022-08-01
Simplified printing for live apply

justinsb:
We aim to create one column that shows the current activity,
to give users a single column to watch.

### DIFF
--- a/internal/alpha/printers/table/printer.go
+++ b/internal/alpha/printers/table/printer.go
@@ -6,11 +6,14 @@ package table
 import (
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"sigs.k8s.io/cli-utils/pkg/apply/event"
 	"sigs.k8s.io/cli-utils/pkg/common"
+	pollingevent "sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
 	printcommon "sigs.k8s.io/cli-utils/pkg/print/common"
 	"sigs.k8s.io/cli-utils/pkg/print/table"
 )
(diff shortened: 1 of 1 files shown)


yuwenma (approved):
+1 to the coloring and formatting idea

internal/alpha/printers/table/printer.go
```diff
@@ -72,87 +75,23 @@ func (t *Printer) Print(ch <-chan event.Event, _ common.DryRunStrategy, _ bool)
...
 		table.MustColumn("resource"),
-		actionColumnDef,
-		table.MustColumn("status"),
-		reconciledColumnDef,
-		table.MustColumn("conditions"),
-		table.MustColumn("age"),
-		table.MustColumn("message"),
+
+		// We are trying out a "single column" model here
+		unifiedStatusColumnDef,
```
droot, line 94: Does this change applies only to `kpt alpha live apply ?` or `kpt live apply` as well.

If this applies to `kpt alpha` only, then we don't have to think much about any surprises this may cause to the end users. Thinking more, since `--output=table` is not default, so maybe it is ok.

droot (approved):
Would be good to get quick look from @mortent or @karlkfi

droot:
> I attached a screenshot so people can follow along:
> 
> image
> 
> A few things that I intend to tackle moving forwards:
> 
> * I don't know if it makes sense to combine name + kind, but keep namespace separate.  It looks a little weird on cluster-scoped objects.  But TBD!
> * We aren't capturing errors from apply
> * I'd like to come up with a better "all is well" state than Current.  AppliedAndHealthy or just Done, maybe.  I probably also wouldn't print the conditions in this case either.
> * Generally I want to de-emphasize healthy/done objects, and draw more attention to the failures (and "medium" attention to things that are in-progress).
> * When things fail, we have _more_ room now, but we probably still don't have enough room.  I think we may need to dedicate multiple rows to failed objects, and/or maybe on a Control-C we print the full errors for failed objects.

Thanks, the list looks promising. I am surprised about `we aren't capturing errors from Apply`, really ?

internal/alpha/printers/table/printer.go
justinsb: Only alpha live apply (the previous commit split them), and only --output=table

I wonder if we should also make table the default for apply (alpha initially) and maybe also server-side the default also.  We can also make the default switch based on whether the output is a TTY or not (I think).  But those changes would be in future PRs!

justinsb:
> I am surprised about we aren't capturing errors from Apply, really ?

We are capturing that there _was_ an error, but not what the error was.  So KCC immutable fields, for example, are hard to understand.  But a fix is coming next :-)

mortent:
It looks like this removes the `message` column. How do we provide more information to users about why a resource isn't successfully reconciling? While ideally we should have everything based off of conditions, that isn't actually the case.

It also seems like this removes the tree view which includes child resources. Don't we want to have that? It does have some overhead in terms of fetching the state of those resources, but it is pretty useful when things goes wrong.

It looks like the action column is also removed. I think that can be pretty useful for users to know whether resources are being applied or pruned.

internal/alpha/printers/table/printer.go
droot: experimenting with the UX in `alpha` makes sense. @mortent raises some very good point about why we have these additional fields.

FWIW, I found `--output=table` to be more useful and I end up specifying that always, so +1 to making it default.

droot:
> > I am surprised about we aren't capturing errors from Apply, really ?
> 
> We are capturing that there _was_ an error, but not what the error was. So KCC immutable fields, for example, are hard to understand. But a fix is coming next :-)

Got it. Thanks

justinsb:
@mortent We aren't going to remove any useful information.  Rather the intent is to combine all the "status" columns into one column, because what I've observed is that often only one column is showing important information at a time (or one is showing the status and the other is showing the details).  By doing so, we help the user (well, me) understand what's going on, and we get a little more space to play on the screen.

mortent:
The issue that the error is not being printed seems like something we should fix in the regular command as well as in the alpha command.

mortent (approved):

PR #3426, opened 2022-08-01
add set-namespace v0.3.4 to porch dockerfile

natasha41575:
cc @mengqiy 

I'm trying to make set-namespace v0.3.4 easily available to porch. Not sure if this is the right place, lmk if there's another file that needs to be updated as well. 

Last thing to do before closing https://github.com/GoogleContainerTools/kpt/issues/3419

Marking this as a draft until I try it out


mengqiy (approved):

natasha41575:
@mengqiy qq, when I run porch locally, I still get this error:

```
Error: Internal error occurred: fn.render: pkg /:
        pkg.render:
        pipeline.run: func eval "gcr.io/kpt-fn/set-namespace:v0.3.4" failed: rpc error: code = NotFound desc = function "gcr.io/kpt-fn/set-namespace:v0.3.4" not found 
```

Is that expected?

PR #3429, opened 2022-08-02
porch should ignore tags that are not associated with any package

natasha41575:
Proposed fix for https://github.com/GoogleContainerTools/kpt/issues/3407

Basically ignore tags that don't have a package name in front of the version, and ignore tags that throw errors. As we see in the kpt repo, a tag like `v0.33.0` is a release tag and not associated with any package. In the kustomize repo, we have release tags like `kustomize/v1.2.3` that look like package revision tags, but aren't. If users have repos with similar release tagging schemas, porch should not throw an error when it encounters those tags.

An alternative would be to change the porch tagging scheme, making it have a unique `porch` prefix or something.


natasha41575:
ping @mortent

mortent (approved):

PR #3430, opened 2022-08-02
[WIP] feat: Enable sharding of ResourceGroup objects

rquitales:
This commit enables sharding of ResourceGroup objects to scale the
number of GKNN resources tracked within an inventory. Prior to sharding,
the number of objects tracked is limited to the size of the
ResourceGroup object and is bottlenecked by etcd size limits. This
modification allows us to distribute GKNNs to multiple 'paginated'
ResourceGroups.


pkg/live/inventoryrg.go
```diff
@@ -249,78 +308,432 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+	})
 	if err != nil {
 		return err
 	}
+	resourcegroups := resourceList.Items
+
+	// Internal flag to indicate whether current ResourceGroup should be updated to include new labels.
+	var shouldUpdate bool
+
+	// Fallback to using name lookup for backwards compatibility.
```
karlkfi, line 339: Why do we need this fallback? Upgrades? Maybe expand the comment.

pkg/live/inventoryrg.go
```diff
@@ -249,78 +308,432 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+		// Reshuffle GKNNs to the start of ResourceGroup object series.
+		return icm.collapse(dc, mapper)
+	}
+
+	return nil
 }
 
-func (icm *InventoryResourceGroup) ApplyWithPrune(dc dynamic.Interface, mapper meta.RESTMapper, statusPolicy inventory.StatusPolicy, _ object.ObjMetadataSet) error {
-	invInfo, namespacedClient, err := icm.getNamespacedClient(dc, mapper)
+// collapse is a utility function to rebalance objects stored within sharded ResourceGroups.
```
karlkfi, line 547: collapse and rebalance mean different things to me. Perhaps this is more accurate: "collapse the set of ResourceGroups into as few ResourceGroups as possible, given the MaxResourceCount setting."

pkg/live/inventoryrg.go
```diff
@@ -46,8 +49,22 @@ import (
...
+	// shardedRGLabelFmt describes the format used for the label key used on sharded
+	// Resourcegroup objects. The label value should be formatted as `<resourcegroup_name>/id`.
+	shardedRGLabelFmt = "%s/id"
+	// DisableSharding indicates the value ResourceCount should be if sharding should not be done.
+	DisableSharding = -1
 )
 
+// ResourceCount describes the number of resources to be stored within a single ResourceGroup object, before tracked
+// resources needed to be sharded into another ResourceGroup.
+var ResourceCount = 1000
```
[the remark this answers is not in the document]
karlkfi: Maybe the field name should be `MaxResourceCount`

pkg/live/inventoryrg.go
```diff
@@ -249,78 +308,432 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+func createShardedName(name string, idx int) string {
+	return fmt.Sprintf(shardedRGNameFmt, name, len(name), idx)
+}
+
+// nextOf returns a new unstructured object with metadata that attributes it to the
+// next sharded ResourceGroup object.
+func (icm *InventoryResourceGroup) nextOf(curr *unstructured.Unstructured) *unstructured.Unstructured {
+	// Copy over metadata from the current ResourceGroup to use as the base for the new ResourceGroup.
+	newRG := curr.DeepCopy()
+	newRG.SetResourceVersion("")
```
karlkfi, line 455: Does setting it to empty string mean the same thing as removing it from the map?

Maybe not... I know it's different for server-side apply (Apply type Patch), but maybe not Create.

pkg/live/inventoryrg.go
```diff
@@ -249,78 +308,432 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+
+		var rg *unstructured.Unstructured
+		switch totalStored {
+		case 0: // Initial RG in series.
+			rg = icm.inv
+			rg.SetResourceVersion("")
+		default:
+			rg = icm.inv.DeepCopy()
+			rg.SetResourceVersion("")
+			rg.SetName(fmt.Sprintf(shardedRGNameFmt, rg.GetName(), len(rg.GetName()), totalStored/ResourceCount))
```
karlkfi, line 647: use `createShardedName`?

pkg/live/inventoryrg.go
```diff
@@ -249,78 +308,432 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
 	}
 	if invInfo == nil {
-		return nil, nil, fmt.Errorf("attempting to create a nil inventory object")
+		return nil, errors.New("attempting to interact with a nil inventory object")
 	}
 
 	mapping, err := mapper.RESTMapping(invInfo.GroupVersionKind().GroupKind(), invInfo.GroupVersionKind().Version)
 	if err != nil {
-		return nil, nil, err
+		return nil, fmt.Errorf("unable to create rest mapper for %s ResourceGroup: %w", invInfo.GroupVersionKind().Version, err)
```
karlkfi, line 680: Inaccurate error message. It couldn't create a REST "mapping" for a specific resource, not a "mapper". The mapper is the struct that resolves mappings.

Also, I suspect the RESTMapping includes the GVK in the error message, but you could check. You're only adding the version here, which seems weird.

pkg/live/inventoryrg.go
```diff
@@ -249,78 +308,432 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+// to an ObjectMetadata.
+func idFromFieldInterface(fields interface{}) (object.ObjMetadata, error) {
+	item, ok := fields.(map[string]interface{})
+	if !ok {
+		return object.ObjMetadata{}, fmt.Errorf("unable to cast field to the required map type for: %v", fields)
+	}
+
+	id := object.ObjMetadata{
+		GroupKind: schema.GroupKind{
+			Group: item["group"].(string),
```
karlkfi, line 699: These with panic if not strings.
I suspect you want to use NestedString to retrieve them and return any errors.
It shouldn't error if it came from the CRD with the right scheme, but you never know.

pkg/live/inventoryrg.go
```diff
@@ -249,78 +308,432 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+	namespacedClient, err := icm.getNamespacedClient(dc, mapper)
 	if err != nil {
 		return err
 	}
 
-	// Update the cluster inventory object.
-	appliedObj, err := namespacedClient.Update(context.TODO(), invInfo, metav1.UpdateOptions{})
+	// Get all cluster ResourceGroup objects using label selector.
+	resourceList, err := namespacedClient.List(context.TODO(), metav1.ListOptions{
+		LabelSelector: fmt.Sprintf(shardedRGLabelFmt, icm.Name()),
```
karlkfi, line 561: You're doing `fmt.Sprintf(shardedRGLabelFmt, icm.Name())` a few times. Maybe pull it out to function.

pkg/live/inventoryrg.go
```diff
@@ -249,78 +308,432 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+	// Re-balance GKNN objects in sharded ResourceGroups.
+	objIdx := 0
+	for objIdx < count {
+		var curr []interface{}
+		if objIdx+ResourceCount < len(items) {
+			curr = items[objIdx : objIdx+ResourceCount]
+		} else {
+			curr = items[objIdx:]
+		}
+		err := unstructured.SetNestedSlice(resourcegroups[objIdx/ResourceCount].UnstructuredContent(), curr, "spec", "resources")
```
karlkfi, line 592: This seems unsafe. Maybe pull out `resourcegroups[objIdx/ResourceCount]` and/or `objIdx/ResourceCount` to variables?

pkg/live/inventoryrg.go
```diff
@@ -249,78 +308,432 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+		if err != nil {
+			return err
+		}
+		_, err = namespacedClient.Update(context.TODO(), &resourcegroups[objIdx/ResourceCount], metav1.UpdateOptions{})
```
karlkfi, line 596: Update will update the object in place, in the resourcegroups. Is that what you want here?

pkg/live/inventoryrg.go
```diff
@@ -249,78 +308,432 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+	}
+
+	// totalStored tracks the total number of resources stored within all sharded ResourceGroups.
+	totalStored := 0
+
+	for totalStored < len(icm.objMetas) {
+		var items []interface{}
+
+		switch ResourceCount {
+		case DisableSharding:
```
karlkfi, line 628: Not really loving this disabled = -1 as an enum pattern. Maybe just use `ResourceCount > 0` for enabled and <= 0 for disabled?

pkg/live/inventoryrg.go
```diff
@@ -46,8 +49,22 @@ import (
 const (
 	applyRGTimeout      = 10 * time.Second
 	applyRGPollInterval = 2 * time.Second
+
```
karlkfi, line 52: Would like to see some unit tests for this file specifically.

internal/cmddestroy/cmddestroy.go
```diff
@@ -152,10 +153,10 @@ func (r *Runner) runE(c *cobra.Command, args []string) error {
 		}
 	}
 
-	return r.destroyRunner(r, invInfo, dryRunStrategy)
+	return r.destroyRunner(r, invInfo, inv, dryRunStrategy)
 }
 
-func runDestroy(r *Runner, inv inventory.Info, dryRunStrategy common.DryRunStrategy) error {
+func runDestroy(r *Runner, inv inventory.Info, kptInv kptfilev1.Inventory, dryRunStrategy common.DryRunStrategy) error {
```
karlkfi, line 159: Are there unit tests for this that can be extended for this to show it still works without regression?

thirdparty/cli-utils/status/cmdstatus.go
```diff
@@ -128,7 +128,7 @@ func (r *Runner) runE(c *cobra.Command, args []string) error {
 		return err
 	}
 
-	invInfo, err := live.ToInventoryInfo(inv)
+	invInfo, err := live.ToInventoryInfo(inv, inventory.NameStrategy)
```
karlkfi, line 131: I don't like that this has to be set as default in every command. Is there a way to set it as default in the struct so that it uses the NameStrategy if strategy is not set?

karlkfi (changes requested):

pkg/live/inventoryrg.go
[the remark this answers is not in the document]
rquitales: The global variable isn't really needed if we go with making ResourceCount a configurable field on the InventoryResourceGroup. The default would always be to not shard (ie, MaxResourceCount is unlimited).

Have removed the global variable to set resourceCount as an internal variable within InventoryResourceGroup.

pkg/live/inventoryrg.go
```diff
@@ -79,14 +97,14 @@ func WrapInventoryObj(obj *unstructured.Unstructured) inventory.Storage {
 	if obj != nil {
 		klog.V(4).Infof("wrapping Inventory obj: %s/%s\n", obj.GetNamespace(), obj.GetName())
 	}
-	return &InventoryResourceGroup{inv: obj}
+	return &InventoryResourceGroup{inv: obj, strategy: inventory.NameStrategy}
```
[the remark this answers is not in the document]
rquitales: kpt has only supported name strategy, and this PR doesn't change that behaviour, as we used to always hardcode strategy to Name. I'm taking advantage of the label strategy for destroy commands to allow deleting all sharded ResourceGroups when the command is run. Added a function comment to document this. We only need it as a function argument for WrapInventoryInfoObj as that is the one caller in the destroy command that needs it to be of label strategy.

e2e/testdata/live-apply/json-output/pre-apply/rg.yaml
```diff
@@ -17,6 +17,7 @@ kind: ResourceGroup
 metadata:
   labels:
     cli-utils.sigs.k8s.io/inventory-id: json-output
+    json-output/id: ""
```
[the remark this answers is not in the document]
rquitales: Erroneous added line from earlier testing.

pkg/live/inventoryrg.go
rquitales: Yes, we do want to update the resourcegroup in place as the update will only add more GKNNs from subsequent sharded resource groups. This means that we will have duplicated items in the inventories, until the resourcegroups with duplicated GKNNs are deleted in case we have an error collapsing the ResourceGroups.

pkg/live/inventoryrg.go
rquitales: Refactored to have resourceCount as an internal variable, with a new method `shardingDisabled()` to check `<=0` for disabled.

thirdparty/cli-utils/status/cmdstatus.go
rquitales: Refactored and no longer set in every command.

internal/cmddestroy/cmddestroy.go
rquitales: The existing unit tests should already be capturing this.

pkg/live/inventoryrg.go
```diff
@@ -249,78 +298,624 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+// ResourceGroup object name.
+func sortResourceGroups(resourcegroups []unstructured.Unstructured) {
+	sort.Slice(resourcegroups, func(i, j int) bool {
+		return resourcegroups[i].GetName() < resourcegroups[j].GetName()
+	})
+}
+
 // Apply is a Storage interface function implemented to apply the inventory
 // object.
 func (icm *InventoryResourceGroup) Apply(dc dynamic.Interface, mapper meta.RESTMapper, statusPolicy inventory.StatusPolicy) error {
```
karlkfi, line 311: This Apply function now has a LOT more code in it. Can you add some unit tests to make sure it does what you expect?

karlkfi: Same for ApplyWithPrune.

pkg/live/inventoryrg.go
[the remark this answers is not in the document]
karlkfi: The `WrapInventoryObj` func satisfies an interface required to pass it to `inventory.NewClient`. 

I thought `WrapInventoryInfoObj` also used to satisfy an interface for a different inventory function. Does it not?

If `WrapInventoryInfoObj` is only used in `ToInventoryInfo` could `wrapInventoryInfoObj` be a private function?

karlkfi (changes requested):
Looking MUCH cleaner and more readable now, thanks!

Just a few more suggestions and it needs some unit tests for all that sharding logic in Apply and ApplyWithPrune.

pkg/live/inventoryrg.go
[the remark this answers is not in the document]
rquitales: `WrapInventoryInfoObj` does not need to satisfy interface requirements unlike `WrapInventoryObj`. The function is still an exported public function since it is used in the live migrate command `internal/cmdmigrate.go`.

pkg/live/inventoryrg.go
rquitales: Added tests for both funcs.

pkg/live/inventoryrg_test.go
```diff
@@ -240,3 +280,729 @@ func TestIsResourceGroupInventory(t *testing.T) {
...
+
+			managedResources := make(map[object.ObjMetadata]struct{}, len(irg.objMetas))
+			for _, objMeta := range tc.objMetas {
+				managedResources[objMeta] = struct{}{}
+			}
+			allResources := len(managedResources)
+			var rgObjects int
+
+			for _, rg := range resourcegroups {
+				objs, err := irg.readResourceGroupObjects(&rg)
```
karlkfi, line 675: Since we need to use it to test with, should this be a public package function?
Won't we also need it for e2e tests?

pkg/live/inventoryrg_test.go
```diff
@@ -240,3 +280,729 @@ func TestIsResourceGroupInventory(t *testing.T) {
...
+		},
+	}
+
+	for tn, tc := range testCases {
+		t.Run(tn, func(T *testing.T) {
+			dc, mapper := fakeClient()
+			irg := InventoryResourceGroup{
+				inv:           inventoryObj,
+				objMetas:      tc.objMetas,
+				strategy:      inventory.NameStrategy,
```
karlkfi, line 938: Need LabelStrategy tests too, right?

karlkfi (changes requested):

pkg/live/inventoryrg_test.go
```diff
@@ -37,7 +76,8 @@ var inventoryObj = &unstructured.Unstructured{
 			"name":      inventoryObjName,
 			"namespace": testNamespace,
 			"labels": map[string]interface{}{
-				common.InventoryLabel: testInventoryLabel,
+				common.InventoryLabel:                 testInventoryLabel,
```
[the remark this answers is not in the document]
rquitales: Good catch. The inv object here isn't actually used by Apply and ApplyWithPrune, so it has no side effect. Removed the label anyways.

pkg/live/inventoryrg.go
```diff
@@ -60,33 +63,47 @@ var ResourceGroupGVK = schema.GroupVersionKind{
...
 // object metadata (inventory) to and from the wrapped ResourceGroup.
 type InventoryResourceGroup struct {
-	inv       *unstructured.Unstructured
-	objMetas  []object.ObjMetadata
-	objStatus []actuation.ObjectStatus
+	inv           *unstructured.Unstructured
+	objMetas      []object.ObjMetadata
+	objStatus     map[object.ObjMetadata]actuation.ObjectStatus
+	strategy      inventory.Strategy
+	resourceCount int
```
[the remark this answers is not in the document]
rquitales: Updated to `maxObjectsPerShard`

pkg/live/inventoryrg.go
```diff
@@ -249,78 +293,734 @@ func (icm *InventoryResourceGroup) GetObject() (*unstructured.Unstructured, erro
...
+// The name should be formatted as `<resourcegroup_name>-<name-length-hash>-<index>`.
+// An example sharded name is: "resourcegroup-13-1"
+func (irg *InventoryResourceGroup) shardedName(idx int) string {
+	return fmt.Sprintf("%s-%d-%d", irg.Name(), len(irg.Name()), idx)
+}
+
+// shardedLabel genereates the formatted label name required for sharded ResourceGroups.
+// The label value should be formatted as `<resourcegroup_name>/id`.
+func (irg *InventoryResourceGroup) shardedLabel() string {
+	return fmt.Sprintf("%s/id", irg.Name())
```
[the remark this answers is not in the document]
rquitales: The label format for sharded ResourceGroups is of the format `<resourcegroup_name>/id` with an empty string value.

pkg/live/inventoryrg_test.go
```diff
@@ -240,3 +280,729 @@ func TestIsResourceGroupInventory(t *testing.T) {
...
+			}
+
+			err = irg.Apply(dc, mapper, inventory.StatusPolicyNone)
+
+			if err != nil {
+				t.Error(err)
+			}
+
+			clusterRGList, err := namespacedClient.List(context.TODO(), v1.ListOptions{
+				LabelSelector: irg.shardedLabel(),
```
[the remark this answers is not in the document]
rquitales: kpt only supports NameStrategy, and has always had strategy hard-coded as NameStrategy. LabelStrategy is only used in the destroy command since we want the destroyer to delete all sharded ResourceGroups by its common label.

pkg/live/inventoryrg_test.go
rquitales: Good idea, exporting this helper function!

pkg/live/inventoryrg.go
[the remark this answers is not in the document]
karlkfi: Should this new label be namespaced?
kpt uses `config.kubernetes.io/`: https://kpt.dev/reference/annotations/
cli-utils uses `cli-utils.sigs.k8s.io/`
/cc @nan-yu @janetkuo

pkg/live/inventoryrg_test.go
rquitales: kpt does not use LabelStrategy.

pkg/live/inventoryrg_test.go
```diff
@@ -15,20 +15,59 @@
...
+			},
+		},
+	}
+
+	var allObjs []interface{}
+	for _, obj := range objs {
+		allObjs = append(allObjs, idToUnstructuredMap(obj))
+	}
+
+	err := unstructured.SetNestedSlice(u.Object, allObjs, "spec", "resources")
```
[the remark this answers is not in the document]
rquitales: Ah yes!

pkg/live/inventoryrg_test.go
```diff
@@ -240,3 +280,729 @@ func TestIsResourceGroupInventory(t *testing.T) {
...
+			}
+
+			clusterRGList, err := namespacedClient.List(context.TODO(), v1.ListOptions{
+				LabelSelector: irg.shardedLabel(),
+			})
+			if err != nil {
+				t.Error(err)
+			}
+
+			// Check we applied the ResourceGroups correctly.
```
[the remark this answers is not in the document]
rquitales: Updated

pkg/live/inventoryrg.go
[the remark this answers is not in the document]
janetkuo: Yes, unprefixed keys are reserved for end-users; all labels/annotations need to be prefixed: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions

 The proposed prefix is `resourcegroup_name`. If we choose a different prefix, the resource group could be the value of the label.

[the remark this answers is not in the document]
karlkfi: I suggest using a conventional pattern for truncation. If the name is longer than N characters, hash the full name with sha1, and add the hash as the suffix. So is `1234567890` is the name and the max is 9, then the name becomes `1-<hash>`, where the hash is the first 7 characters of the full hash (git --short default). This doesn't guarantee uniqueness. You still have to check for collisions with other names in the namespace, but it does make it much less likely there are collisions. and when there are, you log an error and set an error status condition that says there's a collision.

karlkfi:
/hold

Per offline discussion, the cli-utils ClusterClient needs refactoring to push more of the inventory lifecycle management into the provider/storage layer.

PR #3432, opened 2022-08-03
docs: update porch ui installation guide for gke

ChristopherFry:
This pull request updates the Porch UI Installation Guide for GKE with instructions on how to create Google OAuth Client.

### DIFF
--- a/site/guides/porch-ui-installation.md
+++ b/site/guides/porch-ui-installation.md
@@ -22,18 +22,50 @@ This setup assumes that you have a GKE cluster up and running with porch install
 your current kube context is set to that GKE cluster. We would welcome contributions or feedback
 from people that have set this up in other clouds outside of GKE.
 
-First, find a published image in the [kpt-dev/kpt-backstage-plugins container registry](https://console.cloud.google.com/gcr/images/kpt-dev/global/kpt-backstage-plugins/backstage-plugin-cad?project=kpt-dev).
-For this example, we will use `gcr.io/kpt-dev/kpt-backstage-plugins/backstage-plugin-cad:v0.1.0`.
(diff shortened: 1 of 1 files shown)


ChristopherFry:
cc @johnbelamaric

johnbelamaric:
LGTM

We should as we discussed move this to a kpt package. But this is start.

site/guides/porch-ui-installation.md
```diff
@@ -22,18 +22,50 @@ This setup assumes that you have a GKE cluster up and running with porch install
...
+5. Configure an OAuth consent screen, if required
+   - For scopes, select `openid`, `auth/userinfo.email`,
+     `auth/userinfo.profile`, and `auth/cloud-platform`.
+   - Add any users that will want access to the UI if using External user type
+6. Set **Application Type** to `Web Application` with these settings:
+   - `Name`: Config as Data UI (or any name you prefer)
+   - `Authorized JavaScript origins`: http://localhost:7007
+   - `Authorized redirect URIs`:
+     http://localhost:7007/api/auth/google/handler/frame
+7. Click Create
```
droot, line 51: Can a user do this step using `gcloud` CLI as well ? (not for this PR, but if it is possible, then we can share that command and save these steps).

droot (approved):
Nice. So now users can run UI for their own org.

site/guides/porch-ui-installation.md
johnbelamaric: No :( - I looked into it yesterday. Maybe, if they are using IAP. It's not clear to me how those things interact (IAP vs OAuth 2.0 API screens above).

PR #3437, opened 2022-08-03
update kyaml and kustomize/api dependencies

natasha41575:
Fixes https://github.com/GoogleContainerTools/kpt/issues/3419

This will pick up the change (https://github.com/kubernetes-sigs/kustomize/pull/4729)  that should fix the issue with porch complaining about colons in filenames in the cert-manager-basic package.

cc @ChristopherFry


natasha41575:
cc @droot

yuwenma (approved):

PR #3438, opened 2022-08-04
POC:`kpt pkg get` runs its bundled pipeline

yuwenma:


liamfallon:
We need to rebase or even reimplement this but the idea might be something interesting. We should pull this out into an issue as a possible improvement.

PR #3443, opened 2022-08-04
`kpt alpha rpkg update --discover`: option to find downstream packages that can be updated from an upstream package

natasha41575:
Related to https://github.com/GoogleContainerTools/kpt/issues/3202

Trying to tackle the second bullet point - finding out which downstream packages need an update based on the provided upstream package. 

This PR does the following:

What is currently supported by `kpt alpha rpkg update --discover` will change to `kpt alpha rpkg update --discover=upstream`:
```
$ kpt alpha rpkg update --discover=upstream
PACKAGE REVISION                                            UPSTREAM REPOSITORY   UPSTREAM UPDATES
kpt-samples-47e4436a7490df604f57e75a9d995c2ca7ba0a88                              No update available
kpt-samples-6b14817885cbbaa1144d5b19df3378144cc24551                              No update available
test-deployments-2f36052b298673c231e8debcdd1a4553200b45c7   kpt-samples           v1, v2
test-deployments-bfcd0186c8528d13068e6d71b73329bac504f30b   kpt-samples           v2
test-deployments-f21b3faa487feec5cd231c3aa8698bbd8c2dd88d   kpt-samples           v2
```

This PR adds another option `--discover=downstream` to tackle the reverse question. If a package revision has multiple downstream packages that need to be updated, those entries in the output table will be grouped together (like the middle two in the following output):
```
$ kpt alpha rpkg update --discover=downstream
PACKAGE REVISION                                       DOWNSTREAM PACKAGE                                          DOWNSTREAM UPDATE
kpt-samples-dc223acc4eeba1e3a41a90e9756b00872a51855d   test-deployments-2f36052b298673c231e8debcdd1a4553200b45c7   v0->v2
kpt-samples-b2d61a19750d212c61a1c64b0a1cf1ce0efddc28   test-deployments-bfcd0186c8528d13068e6d71b73329bac504f30b   v1->v2
kpt-samples-b2d61a19750d212c61a1c64b0a1cf1ce0efddc28   test-deployments-f21b3faa487feec5cd231c3aa8698bbd8c2dd88d   v1->v2
kpt-samples-ec01f2365a60ce92d9217225dc447da8e1e722c1   test-deployments-2f36052b298673c231e8debcdd1a4553200b45c7   v0->v1
```

You can optionally provide package revisions as arguments if you only want to look at specific ones:
```
$ kpt alpha rpkg update --discover=downstream kpt-samples-b2d61a19750d212c61a1c64b0a1cf1ce0efddc28
PACKAGE REVISION                                       DOWNSTREAM PACKAGE                                          DOWNSTREAM UPDATE
kpt-samples-b2d61a19750d212c61a1c64b0a1cf1ce0efddc28   test-deployments-bfcd0186c8528d13068e6d71b73329bac504f30b   v1->v2
kpt-samples-b2d61a19750d212c61a1c64b0a1cf1ce0efddc28   test-deployments-f21b3faa487feec5cd231c3aa8698bbd8c2dd88d   v1->v2
```

If there is nothing to output, we just print that all downstream packages are up to date:
```
$ kpt alpha rpkg update --discover=downstream kpt-samples-414c1509c368a46650e4a1098487096ed5c05d81
All downstream packages are up to date.
```

### DIFF
--- a/internal/cmdrpkgupdate/discover.go
+++ b/internal/cmdrpkgupdate/discover.go
@@ -181,7 +181,14 @@ func printdownstreamUpdates(downstreamUpdatesMap map[string][]porchapi.PackageRe
 		upstreamPkgRevName := split[0]
 		upstreamPkgRevNum := split[1]
 		for _, downstreamPkgRev := range downstreamPkgRevs {
-			downstreamUpdates = append(downstreamUpdates, []string{upstreamPkgRevName, upstreamPkgRevNum, downstreamPkgRev.Name, downstreamPkgRev.Spec.Revision})
+			// figure out which upstream revision the downstream revision is based on
+			lastIndex := strings.LastIndex(downstreamPkgRev.Status.UpstreamLock.Git.Ref, "v")
+			if lastIndex < 0 {
+				// this ref isn't formatted the way that porch expects
+				continue
+			}
+			downstreamRev := downstreamPkgRev.Status.UpstreamLock.Git.Ref[lastIndex:]
+			downstreamUpdates = append(downstreamUpdates, []string{upstreamPkgRevName, upstreamPkgRevNum, downstreamPkgRev.Name, downstreamRev})
 		}
 	}
 
@@ -206,7 +213,7 @@ func printdownstreamUpdates(downstreamUpdatesMap map[string][]porchapi.PackageRe
 			return err
 		}
 	} else {
-		if _, err := fmt.Fprintln(w, "UPSTREAM PACKAGE\tUPSTREAM REVISION\tdownstream PACKAGE\tdownstream REVISION"); err != nil {
+		if _, err := fmt.Fprintln(w, "UPSTREAM PACKAGE\tUPSTREAM REVISION\tDOWNSTREAM PACKAGE\tDOWNSTREAM REVISION"); err != nil {
 			return err
 		}
 		for _, pkgRev := range pkgRevsToPrint {
--- a/internal/cmdrpkgupdate/discover.go
+++ b/internal/cmdrpkgupdate/discover.go
@@ -29,7 +29,7 @@ import (
 func (r *runner) discoverUpdates(cmd *cobra.Command, args []string) error {
 	var prs []porchapi.PackageRevision
 	var errs []string
-	if len(args) == 0 {
+	if len(args) == 0 || r.discover == downstream {
 		prs = r.prs
 	} else {
 		for i := range args {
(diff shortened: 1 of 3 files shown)


internal/cmdrpkgupdate/discover.go
```diff
@@ -130,18 +147,81 @@ func (r *runner) getRepositories() (*configapi.RepositoryList, error) {
...
+	} else {
+		pkgRevsToPrint = downstreamUpdates
+	}
+
+	if len(pkgRevsToPrint) == 0 {
+		if _, err := fmt.Fprintln(w, "All downstream packages are up to date."); err != nil {
+			return err
+		}
+	} else {
+		if _, err := fmt.Fprintln(w, "UPSTREAM PACKAGE\tUPSTREAM REVISION\tDOWNSTREAM PACKAGE\tDOWNSTREAM REVISION"); err != nil {
```
natasha41575, line 216: The last column "DOWNSTREAM REVISION" isn't actually the package revision number of the downstream revision, it's the upstream revision number that the downstream package revision is based on - but I was struggling to think of a clear, concise way to specify that in the column name. It'll probably be good to document it somewhere what these column names actually mean.

internal/cmdrpkgupdate/discover.go
```diff
@@ -50,31 +50,48 @@ func (r *runner) discoverUpdates(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	var updates [][]string
+	var upstreamUpdates [][]string
+	downstreamUpdatesMap := make(map[string][]porchapi.PackageRevision) // map from the upstream package revision to a list of its downstream package revisions
+
```
mortent, line 55: Would it be easier to just create separate functions for the upstream and downstream cases here? The switch statement inside the loop will always have the same value for every iteration. I think it would also remove the need for two separate switch statements (finding updates and printing the results).

internal/cmdrpkgupdate/discover.go
mortent: Yeah, I agree the terminology is a bit difficult here. 

When the user specifies `--discovery=upstream`, we currently list packages (using just "PACKAGE" as the column heading). I think maybe we should just use the heading "PACKAGE" in the `--discovery=downstream` case as well, since having columns named both "UPSTREAM PACKAGE" and "DOWNSTREAM PACKAGE" becomes a bit confusing.

For the "DOWNSTREAM REVISION" column, could we maybe name it something like "FROM VERSION" or "INCLUDES VERSION" to make it clearer that this is the version of the upstream that is included in the downstream package.

natasha41575: > When the user specifies --discovery=upstream, we currently list packages (using just "PACKAGE" as the column heading). I think maybe we should just use the heading "PACKAGE" in the --discovery=downstream case as well, since having columns named both "UPSTREAM PACKAGE" and "DOWNSTREAM PACKAGE" becomes a bit confusing.

👍 I will make it consistent with `--upstream`

> For the "DOWNSTREAM REVISION" column, could we maybe name it something like "FROM VERSION" or "INCLUDES VERSION" to make it clearer that this is the version of the upstream that is included in the downstream package.

I like "FROM VERSION", but taking it a bit further, maybe we could combine the revision columns:

```
$ kpt alpha rpkg update --discover=downstream
PACKAGE REVISION                                    DOWNSTREAM PACKAGE                                          AVAILABLE UPDATE
kpt-samples-b2d61a19750d212c61a1c64b0a1cf1ce0efddc2 test-deployments-2f36052b298673c231e8debcdd1a4553200b45c7   v1->v2
```

WDYT?

internal/cmdrpkgupdate/discover.go
natasha41575: done! Thanks for the suggestion, the code is a lot cleaner this way.

internal/cmdrpkgupdate/discover.go
mortent: yeah, I like that idea.

mortent (approved):

PR #3450, opened 2022-08-09
add wasm runtimes (embedded and nodejs) in kpt

mengqiy:
This PR introduces 2 commands: `kpt alpha wasm push` and `kpt alpha wasm pull`:
- kpt alpha wasm push [path-to-local-wasm-file] [desired-OCI-image-name]
- kpt alpha wasm pull [OCI-image-name] [desired-local-path-to-store-the-wasm-file]

This PR added 2 wasm runtimes (one embedded wasmtime and node.js) in kpt:
- It is not a WASI runtime, since golang doesn't have a great support to compile go program to WASI and most of our KRM functions are written in go.
- It requires the wasm to register a js function called `processResourceList` in the globals.

To use it in Kptfile:
```
apiVersion: kpt.dev/v1
kind: Kptfile
metadata:
  name: app
pipeline:
  mutators:
    - wasm: gcr.io/kpt-fn-demo/set-namespace:v0.5.0 # Note: this image has a manifest with arch: wasm and os: js
      configMap:
        namespace: staging
    - wasm: gcr.io/kpt-fn-demo/set-labels:v0.2.0 # Note: this image has a manifest with arch: wasm and os: js
      configMap:
        foo: bar
```
Then run `kpt fn render --allow-alpha-wasm` with 2 functions. `KPT_FN_WASM_RUNTIME` can be used to select a runtime.

| Syntax      | embedded wasmtime | node.js | container|
| ----------- | ----------- | ----------- | ----------- |
| first run      | real    0m8.648s <br> user    0m44.559s <br> sys     0m1.047s       | TODO | TODO |
| subsequent run   | real    0m1.223s <br> user    0m1.125s <br> sys     0m0.498s | TODO | TODO |

This PR requires CGO, it seems cross compilation doesn't work nicely with CGO. We are using GitHub actions to ensure it can be compiled on other OS. https://github.com/GoogleContainerTools/kpt/issues/3463 blocks it from working on windows.

To produce wasm artifacts, please take a look at https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/898.

Depends on https://github.com/GoogleContainerTools/kpt/pull/3470


pkg/wasm/client.go
```diff
@@ -0,0 +1,260 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to parse tag %q: %v", imageName, err)
+	}
+
+	options := []remote.Option{
+		remote.WithAuthFromKeychain(gcrane.Keychain),
+		remote.WithContext(ctx),
+	}
+
+	rmt, err := remote.Get(tag, options...)
```
justinsb, line 63: I think you need to check the err != nil case as well, at least to check that the error is what you expect.

pkg/wasm/client.go
```diff
@@ -0,0 +1,260 @@
...
+	if err == nil {
+		if rmt.MediaType != types.OCIImageIndex && rmt.MediaType != types.DockerManifestList {
+			return fmt.Errorf("unexpected media type: %s, expect either %s or %s", rmt.MediaType, types.OCIImageIndex, types.DockerManifestList)
+		}
+	}
+
+	// Compress the wasm file.
+	tarReader, err := wasmFileToTar(wasmFile)
+	if err != nil {
+		return fmt.Errorf("unable to compress %v: %v", wasmFile, err)
```
justinsb, line 76: Super-nit: `...: %w`

pkg/wasm/client.go
```diff
@@ -0,0 +1,260 @@
...
+		Name: filepath.Base(filename),
+		Size: int64(blen),
+		Mode: 0644,
+	}); err != nil {
+		return nil, fmt.Errorf("failed to write tar header: %w", err)
+	}
+
+	if n, err := writer.Write(b); err != nil {
+		return nil, fmt.Errorf("failed to write tar contents: %w", err)
+	} else if n != blen {
```
justinsb, line 175: I don't think you need to check this, I think the io.Writer contract says that err != nil if it's a short-write (?)

porch/api/generated/clientset/versioned/scheme/register.go
```diff
@@ -35,14 +35,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
 // AddToScheme adds all types of this clientset into the given scheme. This allows composition
 // of clientsets, like in:
 //
-//   import (
-//     "k8s.io/client-go/kubernetes"
-//     clientsetscheme "k8s.io/client-go/kubernetes/scheme"
-//     aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
-//   )
+//	import (
```
justinsb, line 38: Is this because of go 1.19?

justinsb:
Some small nits, but this is pretty close I think.

I think sending the go 1.19 and ioutil -> os changes separately would help make this much more reviewable.

mengqiy:
> I think sending the go 1.19 and ioutil -> os changes separately would help make this much more reviewable.

These changes are really from https://github.com/GoogleContainerTools/kpt/pull/3447. I cherry picked the commits to make CI happy.

This PR requires CGO, but it seems we cross compilation doesn't work nicely with CGO. To workaround it, We may need to build it in container for darwin and windows. @justinsb Any suggestions or prior arts in other k8s projects?

go.mod
```diff
@@ -1,20 +1,26 @@
 module github.com/GoogleContainerTools/kpt
 
-go 1.17
+go 1.19
+
+replace github.com/GoogleContainerTools/kpt/porch v0.0.7 => ./porch
```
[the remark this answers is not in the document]
mengqiy: IIRC this may break `go get` and `go install` for kpt CLI. Thanks for catching it.
It seems we need to move something from porch to kpt to avoid this replace statement here.
It seems porch's go.mod already has `replace github.com/GoogleContainerTools/kpt => ../`.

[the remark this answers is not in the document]
mortent: What is the reason for doing this? If it is to be able to make changes across the different modules more easily, we should look at how we can leverage golang workspaces. 

Also, why would kpt needs to depend on porch? I would assume it should only need to depend on the API? If there are shared components, we should probably either move those to kpt or try to separate it out. But reluctant to add more modules in the repo until we have a better plan for the structure.

And yes, it breaks `go install`.

internal/fnruntime/exec.go
```diff
@@ -59,6 +61,10 @@ func (f *ExecFn) Run(r io.Reader, w io.Writer) error {
 	cmd.Stdout = w
 	cmd.Stderr = &errSink
 
+	for k, v := range f.Env {
+		cmd.Env = append(cmd.Env, fmt.Sprintf("%v=%v", k, v))
```
natasha41575, line 65: Are exec functions unable to access env vars today? Why is this change necessary for the wasm stuff?

internal/fnruntime/nodejs.go
```diff
@@ -0,0 +1,604 @@
...
+			if err = os.Mkdir(cacheDir, 0755); err != nil {
+				return
+			}
+		}
+		if err = os.WriteFile(jsPath, glueJSCode, 0666); err != nil {
+			return
+		}
+	})
+	if err != nil {
+		return nil, fmt.Errorf("unable to create the js glue file")
```
natasha41575, line 50: nit: We should surface `err` in the error message

internal/fnruntime/nodejs.go
```diff
@@ -0,0 +1,604 @@
...
+		},
+	}
+	return f, nil
+}
+
+func (f *WasmNodejsFn) Run(r io.Reader, w io.Writer) error {
+	return f.NodeJsRunner.Run(r, w)
+}
+
+var glueJSCode = []byte(`const fs = require('fs');
```
natasha41575, line 88: nit: maybe this var belongs in its own file since it's long and is actually its own code

internal/fnruntime/nodejs.go
```diff
@@ -0,0 +1,604 @@
...
+const {
+    performance
+} = require('perf_hooks');
+const crypto = require('crypto');
+// Code between line 9 to line 561 in the JS file are glue codes
```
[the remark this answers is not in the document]
natasha41575: > Can we link to where this came from?

+1

internal/fnruntime/wasm.go
```diff
@@ -0,0 +1,75 @@
...
+		nf, err := NewNodejsFn(wasm)
+		if err != nil {
+			return nil, err
+		}
+		return &WasmFn{
+			runtimeType: Nodejs,
+			nodejs:      nf,
+		}, nil
+	case string(Wasmtime):
+		fallthrough
```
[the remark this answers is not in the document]
natasha41575: I think it would be fine to default to "wasmtime" if the env var is not set at all. But if it is set to something besides "wasmtime" or "nodejs" (like "foo") we should probably throw an error.

internal/fnruntime/wasmtime.go
```diff
@@ -0,0 +1,222 @@
...
+	return err
+}
+
+// Write implements the wasmexec.fdWriter interface.
+func (f *WasmtimeFn) Write(fd int, b []byte) (n int, err error) {
+	switch fd {
+	case 1:
+		n, err = os.Stdout.Write(b)
+	case 2:
+		n, err = os.Stderr.Write(b)
```
natasha41575, line 211: I'm pretty sure you can just do

```
case 1, 2:
  n, err := os.Stderr.Write(b)
default: 
...
```

pkg/api/fnresult/v1/types.go
```diff
@@ -29,6 +29,8 @@ type Result struct {
 	// If user provides an executable file with commands, ExecPath should
 	// contain the entire input string.
 	ExecPath string `yaml:"exec,omitempty"`
+
+	Wasm string `yaml:"wasm,omitempty"`
```
natasha41575, line 33: Since this is going to be in alpha, we don't need to resolve this now. But I'm wondering if having a new field for each potential runtime might not be the best syntax. I wondering if it would be better to have a dedicated `runtime` field to better organize these details (I think this is similar to what is proposed in `Catalog`).

internal/cmdrender/cmdrender.go
```diff
@@ -52,6 +52,8 @@ func NewRunner(ctx context.Context, parent string) *Runner {
 		fmt.Sprintf("pull image before running the container. It must be one of %s, %s and %s.", fnruntime.AlwaysPull, fnruntime.IfNotPresentPull, fnruntime.NeverPull))
 	c.Flags().BoolVar(&r.allowExec, "allow-exec", false,
 		"allow binary executable to be run during pipeline execution.")
+	c.Flags().BoolVar(&r.allowWasm, "allow-wasm", false,
+		"allow wasm to be used during pipeline execution.")
```
natasha41575, line 56: The help message should probably specify that this feature is alpha

internal/util/render/executor.go
```diff
@@ -39,7 +39,10 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
-var errAllowedExecNotSpecified = fmt.Errorf("must run with `--allow-exec` option to allow running function binaries")
+var (
+	errAllowedExecNotSpecified = fmt.Errorf("must run with `--allow-exec` option to allow running function binaries")
+	errAllowedWasmNotSpecified = fmt.Errorf("must run with `--allow-wasm` option to allow running wasm functions")
```
natasha41575, line 44: Same with this error message, it should probably specify that this feature is alpha

internal/fnruntime/exec.go
mengqiy: The exec functions can access env vars today. But we want to provide additional env var when running with node.js runtime.
This change makes `ExecFn` not only to inherit env var from the host (current behavior) but also allow us to provide additional env var. It makes `ExecFn` more flexible to use as a library. This change will not affect end users directly.

commands/wasmcmd.go
```diff
@@ -0,0 +1,53 @@
...
+			if h {
+				return cmd.Help()
+			}
+			return cmd.Usage()
+		},
+	}
+
+	pf := repo.PersistentFlags()
+
+	pf.AddGoFlagSet(flag.CommandLine)
```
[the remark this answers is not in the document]
mengqiy: Removed

go.mod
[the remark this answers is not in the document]
mengqiy: I have moved the oci package a bit now. `replace` is no longer needed.

internal/cmdwasmpull/command.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/GoogleContainerTools/kpt/pkg/wasm"
+	"github.com/spf13/cobra"
+)
+
+const (
+	command = "cmdwasmpull"
+)
+
+func newRunner(ctx context.Context) *runner {
+	storeClient, err := wasm.NewClient(path.Join(os.TempDir(), "kpt"))
```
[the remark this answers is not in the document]
mengqiy: Updated

internal/cmdwasmpull/command.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
mengqiy: Updated

internal/fnruntime/nodejs.go
```diff
@@ -0,0 +1,604 @@
...
+		if err = os.WriteFile(jsPath, glueJSCode, 0666); err != nil {
+			return
+		}
+	})
+	if err != nil {
+		return nil, fmt.Errorf("unable to create the js glue file")
+	}
+
+	wasmFile := wasm
+	if !filepath.IsAbs(wasm) {
```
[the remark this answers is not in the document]
mengqiy: Updated. Use `image` and `exec` now.

internal/fnruntime/nodejs.go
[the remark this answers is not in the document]
mengqiy: Added it.

internal/fnruntime/wasm.go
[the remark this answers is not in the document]
mengqiy: Done

internal/fnruntime/wasmtime.go
```diff
@@ -0,0 +1,222 @@
...
+
+	spFn     *wasmtime.Func
+	resumeFn *wasmtime.Func
+}
+
+func NewWasmtimeFn(wasm string) (*WasmtimeFn, error) {
+	f := &WasmtimeFn{}
+	var data []byte
+	var err error
+	// If it looks like an absolute path, we read it from a file. Otherwise, we
```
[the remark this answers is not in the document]
mengqiy: Updated it by using an interface

internal/fnruntime/wasmtime.go
```diff
@@ -0,0 +1,222 @@
...
+	resourceList, err := io.ReadAll(r)
+	if err != nil {
+		return err
+	}
+	result, err := f.gomod.Call(jsEntrypointFunction, string(resourceList))
+	if err != nil {
+		return fmt.Errorf("unable to invoke %v: %v", jsEntrypointFunction, err)
+	}
+	resultStr := fmt.Sprintf("%s", result)
+	// TODO: figure out why it has `&{outputResourceList}`
```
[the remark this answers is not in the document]
mengqiy: Thanks for the tips.
I tried it and I added a comment to explain the type and reason.

internal/fnruntime/wasmtime.go
```diff
@@ -0,0 +1,222 @@
...
+	resultStr = resultStr[2 : len(resultStr)-1]
+	// Try to parse the output as yaml.
+	_, err = yaml.Parse(resultStr)
+	if err == nil {
```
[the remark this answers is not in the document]
mengqiy: Done

pkg/api/kptfile/v1/validation.go
```diff
@@ -67,10 +67,10 @@ func (p *Pipeline) validate(fsys filesys.FileSystem, pkgPath types.UniquePath) e
 }
 
 func (f *Function) validate(fsys filesys.FileSystem, fnType string, idx int, pkgPath types.UniquePath) error {
-	if f.Image == "" && f.Exec == "" {
+	if f.Image == "" && f.Exec == "" && f.Wasm == "" {
 		return &ValidateError{
 			Field:  fmt.Sprintf("pipeline.%s[%d]", fnType, idx),
-			Reason: "must specify a functon (`image` or `exec`) to execute",
+			Reason: "must specify a functon (`image`, `exec` or `wasm`) to execute",
```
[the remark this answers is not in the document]
mengqiy: Updated. Reusing image and exec now

pkg/wasm/client.go
mengqiy: Added it.

pkg/wasm/client.go
mengqiy: You are right! Removed.

internal/fnruntime/nodejs.go
mengqiy: Done

internal/fnruntime/wasmtime.go
mengqiy: Done

pkg/api/fnresult/v1/types.go
mengqiy: Removed it and reusing image and exec now.

internal/cmdrender/cmdrender.go
mengqiy: Updated it to `allow-alpha-wasm`

internal/util/render/executor.go
mengqiy: This is no longer needed. Removed now.

mengqiy:
I just realized that a few comments have not been addressed. I missed them until I resolved other comments. I will resolve them tomorrow.
Meanwhile the PR is mostly ready for another round of review.

I have move part of `porch/pkg/oci` package to `pkg/oci`. If you think I should split it out to its own PR. Please let me know. It should not be hard to do it.

commands/wasmcmd.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/spf13/cobra"
+)
+
+func NewWasmCommand(ctx context.Context, version string) *cobra.Command {
+	repo := &cobra.Command{
+		Use:   "wasm",
+		Short: "[Alpha] TODO", // + repodocs.RepoShort,
+		Long:  "[Alpha] TODO", // + repodocs.RepoLong,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			h, err := cmd.Flags().GetBool("help")
```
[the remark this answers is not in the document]
mengqiy: Removed it. It seems it is cobra handles it automatically.
But I found most commands in https://github.com/GoogleContainerTools/kpt/tree/main/commands have this. We should clean it up in a separate PR.

internal/fnruntime/nodejs.go
```diff
@@ -0,0 +1,604 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to load wasm content from image %v: %w", wasm, err)
+		}
+		defer wasmFileReadCloser.Close()
+		// Create a module out of the Wasm file.
+		data, err := io.ReadAll(wasmFileReadCloser)
+		if err != nil {
+			return nil, fmt.Errorf("unable to read wasm content from reader: %w", err)
+		}
+		wasmFile = filepath.Join(cacheDir, "fn.wasm")
```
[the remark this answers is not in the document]
mengqiy: Done

internal/fnruntime/nodejs.go
```diff
@@ -0,0 +1,604 @@
...
+type WasmNodejsFn struct {
+	NodeJsRunner *ExecFn
+}
+
+func NewNodejsFn(wasm string) (*WasmNodejsFn, error) {
+	var err error
+	cacheDir := filepath.Join(os.TempDir(), "kpt-wasm-fn")
+	jsPath := filepath.Join(cacheDir, "kpt-fn-wasm-glue-runner.js")
+	writeJsFile.Do(func() {
+		if _, err = os.Stat(cacheDir); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
mengqiy: Done

.github/workflows/go.yml
```diff
@@ -47,12 +51,50 @@ jobs:
           go-version: 1.18
         id: go
       - name: Check out code into the Go module directory
-        uses: actions/checkout@v1
+        uses: actions/checkout@v3
+        with:
```
justinsb, line 55: Why do we need this now, when we didn't need it before?  Is it because we've moved to checkout v3?

natasha41575 (approved):
lgtm, defer to @justinsb for final approval

internal/fnruntime/nodejs.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func NewNodejsFn(loader WasmLoader) (*WasmNodejsFn, error) {
+	cacheDir := filepath.Join(os.TempDir(), "kpt-wasm-fn")
+	tempDir, err := os.MkdirTemp(cacheDir, "nodejs-")
+	if err != nil {
+		return nil, fmt.Errorf("unable to create temp dir: %w", err)
+	}
+	jsPath := filepath.Join(tempDir, "kpt-fn-wasm-glue-runner.js")
+	if err = os.WriteFile(jsPath, []byte(golangWasmJSCode+glueCode), 0666); err != nil {
```
justinsb, line 42: Nit: I think 0644 or 600, because we don't want other people to write to the file.

justinsb (approved):

justinsb:
Two small things, but lgtm!

.github/workflows/go.yml
mengqiy: I was trying to follow how kops does it in https://github.com/kubernetes/kops/blob/master/.github/workflows/main.yml as you suggested. It uses a hash but it's checkout v3.
I tried it with checkout v1 and run into an issue. I switched to checkout v3, it works fine.
I didn't spend much time trying to figure out the difference v1 and v3. Do we prefer to use checkout v1?

PR #3451, opened 2022-08-09
surface function results on validation failure

natasha41575:
Fixes https://github.com/GoogleContainerTools/kpt/issues/3423

cc @ChristopherFry 

This PR surfaces the function `Results` from the function runner library when a function encounters a failure (rather than returning a generic error message). As a side effect, the format of the logs on the server side change slightly. 

### Before:
Output on CLI side:
```
$ kpt alpha rpkg push test-deployments-9ed61849dae3bbc9345afc7c61e8b5d10a81e176 ../foo -ndefault > /dev/null 
Error: Internal error occurred: fn.render: pkg /:
	pkg.render:
	pipeline.run: error: function failure
```

Output on server side:
```
I0809 20:41:49.839213 3642946 render.go:59] Package "/": 
I0809 20:41:49.839425 3642946 render.go:59] [RUNNING] "gcr.io/kpt-fn/starlark:v0.4.3"
I0809 20:41:49.839443 3642946 render.go:59] 
I0809 20:41:49.842271 3642946 render.go:59] [FAIL] "gcr.io/kpt-fn/starlark:v0.4.3" in 0s
I0809 20:41:49.842298 3642946 render.go:59]   Results:
    [error]: fail: validation always fails
```

### After:
Output on CLI side:
```
$ kpt alpha rpkg push test-deployments-9ed61849dae3bbc9345afc7c61e8b5d10a81e176 ../foo -ndefault
Error: Internal error occurred: fn.render: pkg /:
	pkg.render:
	pipeline.run: 
[RUNNING] "gcr.io/kpt-fn/starlark:v0.4.3"
[FAIL] "gcr.io/kpt-fn/starlark:v0.4.3" in 0s
  Results:
    [error]: fail: validation always fails
```

Output on server side:
```
I0809 22:15:45.327832 3668588 render.go:59] Package "/": 
I0809 22:15:45.331214 3668588 render.go:59] [RUNNING] "gcr.io/kpt-fn/starlark:v0.4.3"
[FAIL] "gcr.io/kpt-fn/starlark:v0.4.3" in 0s
  Results:
    [error]: fail: validation always fails
```


natasha41575:
cc @droot @mortent

natasha41575:
Will see if I can add a test for this

internal/fnruntime/runner.go
```diff
@@ -179,32 +179,35 @@ type FunctionRunner struct {
 
 func (fr *FunctionRunner) Filter(input []*yaml.RNode) (output []*yaml.RNode, err error) {
 	pr := printer.FromContextOrDie(fr.ctx)
+	var fnOutput string
+	defer func() {
+		pr.Printf(fnOutput)
```
mortent, line 184: Do we need to use a defer for this? I usually think of it more for error handling and cleanup.

internal/fnruntime/runner.go
```diff
@@ -179,32 +179,35 @@ type FunctionRunner struct {
 
 func (fr *FunctionRunner) Filter(input []*yaml.RNode) (output []*yaml.RNode, err error) {
 	pr := printer.FromContextOrDie(fr.ctx)
```
mortent, line 181: Maybe not for this PR, but we need to remove use of the printer from all of kpt that isn't the cli commands.

internal/fnruntime/runner.go
```diff
@@ -179,32 +179,35 @@ type FunctionRunner struct {
...
+	var fnOutput string
```
mortent, line 182: I suggest using the [`strings.Builder`](https://pkg.go.dev/strings#Builder) here instead of string concat.

e2e/testdata/fn-render/path-index-duplicate/.expected/config.yaml
```diff
@@ -15,4 +15,4 @@
 # One of the functions in the pipeline fails resulting in
 # non-zero exit code and no changes in the resources.
 exitCode: 1
-stdErr: 'resource at path "resources.yaml" and index "0" already exists'
+stdErr: 'resource at path \"resources.yaml\" and index \"0\" already exists'
```
natasha41575, line 18:
```
Because this change is to surface the function results whenever we encounter an error, some output of `kpt fn render` changes.

On main:
```
Package "path-index-duplicate": 
[RUNNING] "gcr.io/kpt-fn/starlark:v0.2.1"
[FAIL] "gcr.io/kpt-fn/starlark:v0.2.1" in 2.2s
Error: resource at path "resources.yaml" and index "0" already exists
```

With this PR:

```
$ kpt fn render e2e/testdata/fn-render/path-index-duplicate
Package "path-index-duplicate": 
[RUNNING] "gcr.io/kpt-fn/starlark:v0.2.1"
[FAIL] "gcr.io/kpt-fn/starlark:v0.2.1" in 1.7s
  Stderr:
    "resource at path \"resources.yaml\" and index \"0\" already exists"
```

If there had been any `Results`, this PR would have surfaced them while `main` would not have. 

I think this is a good change because we should be surfacing stderr in our structured results format wherever possible, as that's what caused the original issue in porch, but I did want to call out that this PR affects both kpt CLI and porch (which is not surprising considering that it is a shared library between them. 
```

internal/fnruntime/runner.go
natasha41575: Changed it to not use a defer

internal/fnruntime/runner.go
natasha41575: done! Thanks

natasha41575:
Ah I broke something in the most recent commit, working on it now

natasha41575:
Marking this as WIP because I want to run this through a few more scenarios

e2e/testdata/fn-eval/non-krm-resource/.expected/config.yaml
```diff
@@ -19,4 +19,5 @@ args:
 exitCode: 1
 stdErr: |
   [FAIL] "gcr.io/kpt-fn/set-namespace:v0.1.3" in 0s
-  error: input resource list must contain only KRM resources: nonkrm.yaml: resource must have `apiVersion`
+    Stderr:
+      "input resource list must contain only KRM resources: nonkrm.yaml: resource must have `apiVersion`"
```
natasha41575, line 23: The output on this changes as well with this PR.

Currently:
```
$ kpt fn eval e2e/testdata/fn-eval/non-krm-resource/ --image=gcr.io/kpt-fn/set-namespace:v0.1.3 -- namespace=staging 
[RUNNING] "gcr.io/kpt-fn/set-namespace:v0.1.3"
[FAIL] "gcr.io/kpt-fn/set-namespace:v0.1.3" in 0s
error: "input resource list must contain only KRM resources: nonkrm.yaml: resource must have `apiVersion`"
```

With this PR:
```
$ kpt fn eval e2e/testdata/fn-eval/non-krm-resource/ --image=gcr.io/kpt-fn/set-namespace:v0.1.3 -- namespace=staging 
error: [RUNNING] "gcr.io/kpt-fn/set-namespace:v0.1.3"
[FAIL] "gcr.io/kpt-fn/set-namespace:v0.1.3" in 0s
  Stderr:
    "input resource list must contain only KRM resources: nonkrm.yaml: resource must have `apiVersion`"
```

natasha41575:
@mortent In addition to changing the output for porch, this PR changes the output slightly for `kpt fn render` and `kpt fn eval`. More details in https://github.com/GoogleContainerTools/kpt/pull/3451#discussion_r942756230 and https://github.com/GoogleContainerTools/kpt/pull/3451#discussion_r942924045. I wanted to double check with you to see if you have any concerns about that.

internal/fnruntime/runner.go
```diff
@@ -179,32 +179,35 @@ type FunctionRunner struct {
...
-		printOpt := printer.NewOpt()
-		pr.OptPrintf(printOpt, "[FAIL] %q in %v\n", fr.name, time.Since(t0).Truncate(time.Millisecond*100))
-		printFnResult(fr.ctx, fr.fnResult, printOpt)
+		fnOutput.WriteString(fmt.Sprintf("[FAIL] %q in %v\n", fr.name, time.Since(t0).Truncate(time.Millisecond*100)))
+		fnOutput.WriteString(printFnResult(fr.fnResult))
 		var fnErr *ExecError
 		if goerrors.As(err, &fnErr) {
-			printFnExecErr(fr.ctx, fnErr)
+			fnOutput.WriteString(printFnExecErr(fnErr))
+			pr.Printf(fnOutput.String())
```
mortent, line 198: Why do we still need to print from here?

mortent:
@natasha41575 These changes seem fine to me, but I think we want @droot or @mengqiy to verify this.

natasha41575:
> @natasha41575 These changes seem fine to me, but I think we want @droot or @mengqiy to verify this.

👍 Will ask @droot to take a look when he comes back

internal/fnruntime/runner.go
natasha41575: This is a remnant from the previous implementation, which handled errors returned from the function differently from function execution errors. This PR removes the need to do a separate print, so I have updated the PR accordingly.

internal/fnruntime/runner.go
```diff
@@ -201,36 +201,38 @@ type FunctionRunner struct {
...
 		} else {
-			pr.Printf("[RUNNING] %q", fr.name)
+			fnOutput.WriteString(fmt.Sprintf("[RUNNING] %q", fr.name))
 		}
 		if fr.displayResourceCount {
-			pr.Printf(" on %d resource(s)", len(input))
+			fnOutput.WriteString(fmt.Sprintf(" on %d resource(s)", len(input)))
 		}
-		pr.Printf("\n")
+		fnOutput.WriteString("\n")
```
droot, line 214: If we don't write the `fnOutput` to the printer here, the end users in CLI case won't see the `[RUNNING]....` message before running the function. Instead, they will see it in the end. Seeing it before a function is actually invoked is important from the UX perspective especially in case of running a function for the first time (because it may take long time due to downloading the function image etc).

General rule: any diagnostic info about running something should be printed as real time as possible.

droot: Another thing I realized: This is important for a `kpt CLI` user because they see the feedback in realtime and may not be important for `porch` because a porch user sees the output at the end.

May be passing a simple printer (that uses buffered output) for porch invocation might work better leaving the current code untouched ?

internal/fnruntime/runner.go
```diff
@@ -201,36 +201,38 @@ type FunctionRunner struct {
...
 	}
 	t0 := time.Now()
 	output, err = fr.do(input)
 	if err != nil {
-		printOpt := printer.NewOpt()
-		pr.OptPrintf(printOpt, "[FAIL] %q in %v\n", fr.name, time.Since(t0).Truncate(time.Millisecond*100))
-		printFnResult(fr.ctx, fr.fnResult, printOpt)
+		fnOutput.WriteString(fmt.Sprintf("[FAIL] %q in %v\n", fr.name, time.Since(t0).Truncate(time.Millisecond*100)))
```
droot, line 219: Another thing I realized:
In case of `fn render` errors: `CaD UI` should use functionResult part of `fn render` response (resourcelist) to display the errors. It doesn't have to match the CLI output.

internal/fnruntime/runner.go
```diff
@@ -393,8 +395,7 @@ func populateResourceRef(item *yaml.RNode, resultItem *framework.Result) error {
 
 // printFnResult prints given function result in a user friendly
 // format on kpt CLI.
-func printFnResult(ctx context.Context, fnResult *fnresult.Result, opt *printer.Options) {
-	pr := printer.FromContextOrDie(ctx)
+func printFnResult(fnResult *fnresult.Result) string {
```
droot, line 398: nit: If we are taking the printer out, may be the name of the function should also be changed to `fnResultToStr` or something.

internal/fnruntime/runner.go
```diff
@@ -406,30 +407,23 @@ func printFnResult(ctx context.Context, fnResult *fnresult.Result, opt *printer.
...
-	pr := printer.FromContextOrDie(ctx)
-	printFnStderr(ctx, fnErr.Stderr)
-	pr.Printf("  Exit code: %d\n\n", fnErr.ExitCode)
+	return ""
 }
 
 // printFnStderr prints given stdErr in a user friendly format on kpt CLI.
-func printFnStderr(ctx context.Context, stdErr string) {
-	pr := printer.FromContextOrDie(ctx)
+func printFnStderr(stdErr string) string {
```
droot, line 416: nit: name should probably be changed to `fnStdErrToStr`

internal/util/render/executor.go
```diff
@@ -383,6 +383,11 @@ func (pn *pkgNode) runPipeline(ctx context.Context, hctx *hydrationContext, inpu
 	}
 
 	if err = pn.runValidators(ctx, hctx, mutatedResources); err != nil {
+		// if the error is a multiline string, prepend a newline for
+		// cleaner output
+		if len(strings.Split(err.Error(), "\n")) > 1 {
```
droot, line 388: `runMutators` above this will also need this.

droot (changes requested):
@natasha41575 I have some comments. I would like to explore a porch specific buffered printer to see if that can simplify the code. Happy to discuss offline.

internal/fnruntime/runner.go
```diff
@@ -54,7 +54,7 @@ func NewRunner(
 	pkgPath types.UniquePath,
 	fnResults *fnresult.ResultList,
 	imagePullPolicy ImagePullPolicy,
-	setPkgPathAnnotation, displayResourceCount, allowWasm bool,
+	setPkgPathAnnotation, displayResourceCount bool, allowWasm bool,
```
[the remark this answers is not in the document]
natasha41575: Ah this was from resolving a merge conflict, I'll fix that

internal/fnruntime/runner.go
natasha41575: > May be passing a simple printer (that uses buffered output) for porch invocation might work better leaving the current code untouched ?

I think this makes sense. I will look further into this.

natasha41575:
Chatted with @droot offline about this.

This PR tries to expose the fnResults as a string in the error message, but that doesn't make sense as a porch API. We think it makes more sense to expose the `fnResults` in the status field of the returned resources when render fails.  I plan to update this PR accordingly.

Issue #3396, opened 2022-07-20
Flesh out the input data model and patterns


bgrant0607:
Topic that needs more work.

We've figured out some aspects and requirements of package / function inputs:
* Packages are not encapsulated and don't have monolithic package-specific interfaces: https://kpt.dev/guides/rationale
* Setters are another form of manually specified parameters, and are not recommended: #3131. We also ran into problems with composition of multiple packages when using setters.
* Individual attributes should be able to just be edited in place, since we store the rendered output and update it in place, and functions should "patch" resources in general, rather than blow them away #2528.
* KRM functions can take their inputs from a specified "function config", which is either a ConfigMap or a client-side KRM type, such as ApplyReplacements, which we should be able to automatically map to functions #3339. 
* We've noticed with value transformers #3155, such as set-namespace and set-labels, that input values often need to be copied from their sources to the input structures expected by the functions.
* The [variant-constructor pattern](https://kpt.dev/guides/variant-constructor-pattern) can use the package name to provide distinct identities for variants.
* We've identified the need for a deployment target "context", such as cluster targets #3387, along the lines of kubeconfig, [gcloud config](https://github.com/GoogleContainerTools/kpt/pull/2715), terraform provider config, etc. 
* Standardization of input types, APIs, across packages increase the opportunity for plug-and-play-style automation
* There may be other properties associated with the variant "context" that we need to discover or users need to provide. If the latter, we may need to be able to identify those attributes automatically, so that we can prompt the user. Multiple sources of context are common, such as environment and application.
* We need to be able to find sets of input contexts in order to automatically generate corresponding sets of deployment packages. #3347 
* We need to be able to identify downstream inputs in order to implement a "replay" approach to package upgrades. #3329
* We want to support loosely coupled external dependencies, such as an application package like ghost requiring a namespace or a SQL database
* Fully dynamic data, like autoscaled replica counts, may not belong in config storage. Another common example is allocated IP addresses, for which service discovery systems and DNS are common. But there may be some that we want to "snapshot" and write to storage. GitOps [image updaters](https://argocd-image-updater.readthedocs.io/en/stable/basics/update-methods/#method-git) are an example.
* Some inputs may be reasonably self-contained, such as application config for ConfigMap generation #3119.

But we don't have a fully fleshed out model or recommended patterns yet.

kpt isn't the first config tool to encounter these issues. We should look at data-oriented, non-package-parameter-based models for inspiration.

Some examples:
[...]

PR #3456, opened 2022-08-11
WorkloadIdentityBinding operator

justinsb:
WIP, for discussion cc @mortent


johnbelamaric:
A few high-level comments, more-or-less documenting what we discussed in person:
* A README or some description of what this is for would be helpful in understanding the code. Also maybe some examples?
* This is a GCP only controller, I would expect that to be reflected in the naming or something. Or perhaps if we expect to have a number of these, we bundle them into a separate GCP-specific controller-manager.
* Why does this need to be a controller? Could it be a function instead?
* Does this only apply to WI for Porch itself? What about configuring WI for workloads running in other clusters for which Porch is just pushing packagerevisions to their deployment repos? This is more the scenario that would be useful for my use cases - a function I *think* could satisfy both of these (not sure though).

justinsb:
Thanks @johnbelamaric!

> A README or some description of what this is for would be helpful in understanding the code. Also maybe some examples?
Good idea - I'll add in a follow-on PR, as we don't want this to hold up the other PRs that build on it.

> This is a GCP only controller, I would expect that to be reflected in the naming or something. Or perhaps if we expect to have a number of these, we bundle them into a separate GCP-specific controller-manager.

Good plan also, although WorkloadIdentity is a actually a GCP name, other clouds use different terms.  It's alpha so I think we can iterate here.

> Why does this need to be a controller? Could it be a function instead?

In this case, we need the identity-namespace/project-id that we're running in.  That's not a piece of information that is really part of the package, it's more part of the deployment environment.  I'm experimenting with building these packages, and this one really helps us out.  Elsewhere I've been able to use a function.  If we find we need the identity namespace, then I think this could well be a function.  I actually suspect that this ends up in KCC (or a KCC-adjacent project), because really we're trying to make the KRM schema a lot more regular, vs encoding a lot of information in a "magic" pseudo-ServiceAccount name.

> Does this only apply to WI for Porch itself? What about configuring WI for workloads running in other clusters for which Porch is just pushing packagerevisions to their deployment repos? This is more the scenario that would be useful for my use cases - a function I think could satisfy both of these (not sure though).

It is generic in terms of binding Kubernetes ServiceAccounts to GCP ServiceAccounts.  It should run wherever the KSA is configured.  I suspect in most cases KCC is running in the same cluster, configuring the GCP ServiceAccount, but this isn't required.  (But, if something else is provisioning the GCP SA, probably that thing is a good candidate for configuring the KSA -> GCP SA IAM association).

mortent (approved):

johnbelamaric:
> That's not a piece of information that is really part of the package, it's more part of the deployment environment.

The deployment environment is known at the time of package configuration. For example we could annotate the deployment repo with this project ID. Or associate at “ClusterDenographics” or “ClusterGCPContext” resource with a repo. This is effectively what I did in the Nephio controller PoC with a CluaterScaleProfile.

rajatchowdhary:
Any ETA for the Porch release that will include this operator?

johnbelamaric:
> The deployment environment is known at the time of package configuration.

Follow up. This is the point of this bullet in #3396:

> We've identified the need for a deployment target "context", such as cluster targets https://github.com/GoogleContainerTools/kpt/issues/3387, along the lines of kubeconfig, https://github.com/GoogleContainerTools/kpt/pull/2715, terraform provider config, etc.

PR #3462, opened 2022-08-15
Add RootSyncSet controller

mortent:
This adds the RootSyncSet controller to porch. This also copies code for token exchange to support workload identity from `pkg/registry/porch/wi` to `pkg/tokenexchange`.

This builds on https://github.com/GoogleContainerTools/kpt/pull/3456, so it must be merged after.


porch/controllers/rootsyncset/api/v1alpha1/rootsyncset_types.go
```diff
@@ -0,0 +1,93 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// RootSyncSetSpec defines the desired state of RootSyncSet
+type RootSyncSetSpec struct {
+	ClusterRefs []*ClusterRef `json:"clusterRefs,omitempty"`
```
johnbelamaric, line 23: Embedding these directly doesn't work at scale. Are we sure we want to introduce the concept of "Cluster"?

porch/controllers/rootsyncset/api/v1alpha1/rootsyncset_types.go
```diff
@@ -0,0 +1,93 @@
...
+	Template    *RootSyncInfo `json:"template,omitempty"`
```
johnbelamaric, line 24: So we expect no variation between clusters?

johnbelamaric (changes requested):
I may be just out of the loop, but I am yet convinced on this concept at all. It's not clear to what problem it is solving. Can we get a use case description before we see code? I can't determine if this is the right approach before having a better understanding of the problem.

porch/controllers/rootsyncset/api/v1alpha1/rootsyncset_types.go
justinsb: I'd be fine with using "Target" instead (I think I'd personally prefer it), but I think we can iterate here.

porch/controllers/rootsyncset/api/v1alpha1/rootsyncset_types.go
justinsb: As I understand the plan, we're starting here and then going to figure out what variation we need.

porch/controllers/rootsyncset/api/v1alpha1/rootsyncset_types.go
johnbelamaric: I guess this controller uses direct API access so cluster makes sense

mortent: Yeah, it possible that this doesn't work at large scale. We can iterate on this when we get some experience with this.

porch/controllers/rootsyncset/api/v1alpha1/rootsyncset_types.go
mortent: Yeah, no variation is a starting point. We know that for some (maybe many or all) cases we do need more than this.

justinsb (approved):
Awesome - lgtm!

PR #3466, opened 2022-08-16
Expose a Compute function for Config Connector status

mortent:
This exposes a function from the configconnector package that allows for computing the reconciliation status for a passed in Config Connector resource. This just restructures the existing code so that it is possible to use the basic code without requiring the full functionality of the StatusReader implementations. 

The immediate use-case for this is to provide a way to handle status for Config Connector resources in the ResourceGroup controller: https://github.com/GoogleContainerTools/kpt-resource-group

### DIFF
--- a/pkg/status/configconnector.go
+++ b/pkg/status/configconnector.go
@@ -20,7 +20,7 @@ import (
 	"fmt"
 	"strings"
 
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


justinsb (approved):
lgtm

justinsb:
Looks good.  If we think there might be more of these, we could consider defining + implementing an interface (for use with the unstructured form).  But not a blocker.

mortent:
@justinsb Agree that there should be a specific interface for this. Just needed to unblock another effort so didn't want to start looking at that now.

PR #3470, opened 2022-08-17
move part of porch/pkg/oci to pkg/oci

mengqiy:
Split it from https://github.com/GoogleContainerTools/kpt/pull/3450 in case it's useful.


justinsb (approved):

justinsb:
LGTM

PR #3472, opened 2022-08-19
Provide the option of creating the package branch if it doesn't exist

mortent:
This adds a `CreateBranch` property on the `GitRepository` object of the `Repository` resource. It allows clients to specify whether porch should create the package branch if it doesn't already exist in the repository.

There will be a separate PR to add a flag for this to the `kpt alpha repo register` command.


porch/pkg/git/git.go
```diff
@@ -676,21 +686,84 @@ func (r *gitRepository) fetchRemoteRepository(ctx context.Context) error {
...
-		// Check existence of main branch. If it doesn't exist, fail.
-		// If the main branch doesn't exist, we could create draft or proposal branch
-		// which could become the default branch and those are hard to delete.
-		if _, err := r.repo.Reference(r.branch.RefInLocal(), false); err != nil {
+	if opts.MainBranchStrategy == SkipVerification {
+		return nil
+	}
+
+	if _, err := r.repo.Reference(r.branch.RefInLocal(), false); err != nil {
+		switch opts.MainBranchStrategy {
```
natasha41575, line 698: I see that this can be set through the code above in cache.go or the clone package mutation. Is it ever possible that this is not set? Should this switch statement have a default case?

porch/pkg/git/git_test.go
```diff
@@ -96,7 +96,7 @@ func (g GitSuite) TestOpenEmptyRepository(t *testing.T) {
 		t.Errorf("Unexpectedly succeeded opening empty repository with main branch validation enabled.")
 	}
 
-	if _, err := OpenRepository(ctx, name, namespace, repository, tempdir, GitRepositoryOptions{SkipMainBranchVerification: true}); err != nil {
+	if _, err := OpenRepository(ctx, name, namespace, repository, tempdir, GitRepositoryOptions{MainBranchStrategy: SkipVerification}); err != nil {
```
natasha41575, line 99: Can we also test the other `MainBranchStrategy` options here?

porch/pkg/git/git.go
```diff
@@ -121,10 +130,11 @@ func OpenRepository(ctx context.Context, name, namespace string, spec *configapi
 	}
 
 	if err := repository.fetchRemoteRepository(ctx); err != nil {
+		klog.Infof("Error fetching remote repository: %v", err)
```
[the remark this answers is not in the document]
mortent: Just a leftover log statement from debugging. I have removed it.

porch/pkg/git/git.go
mortent: It is always set, but might be the default value. The default value for the enum is `ErrorIfMissing`, just like it used to be.
All the values are covered in the code, but I've added an error message for the default case in the switch statement in case additional values are added at some point in the future.

porch/pkg/git/git_test.go
mortent: Added.

natasha41575 (approved):

PR #3477, opened 2022-08-20
Add status reader for the rollout object

chunglu-chou:
This change adds a customized status reader for the Rollout object.
The motivation is to integrate Config Sync with Argo Rollouts and track the status of the Rollout object with kpt, so the status reader is added in kpt instead of Config Sync.

### DIFF
--- a/pkg/status/rollout.go
+++ b/pkg/status/rollout.go
@@ -1,0 +1,216 @@
+package status
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strconv"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/engine"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
+	"sigs.k8s.io/cli-utils/pkg/object"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	ArgoGroup   = "argoproj.io"
+	Rollout     = "Rollout"
+	Degraded    = "Degraded"
+	Failed      = "Failed"
+	Healthy     = "Healthy"
+	Paused      = "Paused"
+	Progressing = "Progressing"
+)
+
+type RolloutStatusReader struct {
+	Mapper meta.RESTMapper
+}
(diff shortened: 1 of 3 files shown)


pkg/status/rollout.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	if !found {
+		e := fmt.Errorf("metadata.generation not found")
+		return &result, e
+	}
+
+	observedGeneration, found, err := unstructured.NestedInt64(u.Object, "status", "observedGeneration")
+	if err != nil {
+		var observedGenerationString string
+		observedGenerationString, found, err = unstructured.NestedString(u.Object, "status", "observedGeneration")
```
mortent, line 68: Add a comment as to why we are doing this. It is not obvious why we try to fetch the observed generation as a string.

pkg/status/rollout.go
```diff
@@ -0,0 +1,227 @@
...
+		if err != nil {
+			return &result, fmt.Errorf("looking up status.observedGeneration from resource: %w", err)
+		}
+		observedGeneration, err = strconv.ParseInt(observedGenerationString, 10, 64)
+		if err != nil {
+			return &result, fmt.Errorf("looking up status.observedGeneration from resource: %w", err)
+		}
+	}
+	if !found {
+		// We know that Config Connector resources uses the ObservedGeneration pattern, so consider it
```
mortent, line 78: This seems like a copy-paste error, although it seems like it would be true for Rollout as well.

pkg/status/rollout.go
chunglu-chou: Sure, that sounds reasonable. Also, could we read the observedGeneration field as a string without checking if it is an integer first? Since this is customized for the Rollout objects, it might be unnecessary to check if it is an integer, and we probably should return an error when it is an integer cause that might now be a valid Rollout object.

mortent: Yeah, that makes sense. This code should only run once we know we are dealing with a Rollout resource.

pkg/status/rollout.go
```diff
@@ -0,0 +1,228 @@
...
+
+	if generation != observedGeneration {
+		msg := fmt.Sprintf("%s generation is %d, but latest observed generation is %d", u.GetKind(), generation, observedGeneration)
+		result.Status = status.InProgressStatus
+		result.Message = msg
+		return &result, nil
+	}
+
+	var e error
+	e = nil
```
mortent, line 92: This isn't necessary. The default value of error is nil. Also, we should just use the name `err` for the variable as is the common use in go.

chunglu-chou: @mortent I wrote this because previously I thought the error should be returned after phase and conditions are parsed and loaded into the status result object, but looks like this logic is not necessary, so I think I can just return an empty status result with error right after an error is detected, and I can get rid of this variable.

mortent (approved):

PR #3481, opened 2022-08-22
Include timestamp and author for approved packagerevisions in the API

mortent:
This PR exposes the time and author for published packagerevisions. They are exposed as the new fields `publishedBy` and `publishedAs` on the status object of the packagerevision resource.

For published packages on a git tag, we fetch the author and timestamp from the latest commit on the tag. For packages on the main branch, we can't assume that the last commit is related to the package in question, so we scan through the git commit log to find the latest commit for the specific package, by looking for the gitAnnotations. If we can't find any annotations for the package, we just leave the fields empty.

This still needs a few more tests, but wanted feedback on the API changes.


justinsb (approved):

mortent:
I will add some more tests on this before merging.

PR #3483, opened 2022-08-23
Modify to use loader and inventory client interfaces in cli-utils

chunglu-chou:
Remove the forked status command from cli-utils by implementing the loader and inventory client interfaces and utilize the command Runner code in cli-utils


commands/livecmd.go
```diff
@@ -45,13 +46,15 @@ func GetLiveCommand(ctx context.Context, _, version string) *cobra.Command {
 	}
 
 	f := newFactory(liveCmd, version)
+	invFactory := live.NewClusterClientFactory()
```
mortent, line 49: I think we can just keep this inside the cmdstatus package?

pkg/live/fake-loader.go
```diff
@@ -0,0 +1,61 @@
+package live
```
mortent, line 1: I think we can keep this one too in the cmdstatus package. It might be that we want to reuse it in the other kpt live commands, but we can move it if that happens.

mortent (commented):
Mostly looks good. Great to get rid of the fork.

commands/livecmd.go
chunglu-chou: @Morten Do you mean the NewClusterClientFactory function and the struct related to it or you are just saying this line could be placed inside the cmdstatus?

chunglu-chou: I think the test utilizes the FakeClusterClient which is a different one, so the invFactory needs to be an input argument, otherwise the test will fail. But I think it's ok to move the implementation of the client factory into cmdstatus.

mortent (approved):

PR #3485, opened 2022-08-23
Fix controller configuration

mortent:
* Updated the `porch/scripts/create-deployment-blueprint.sh` script to include the CRDs for the WorkloadIdentityBindings and the RootSyncSet controllers.
* Changed the group for the WorkloadIdentityBindings CRD from `porch.kpt.dev` to `config.cloud.google.com` to avoid using the same group for the CRD and the types for the aggregated apiserver.

Longer term, we need to decide:
* Which group should we be using for kpt/porch types?
* We probably need to be able to enable/disable certain of the controllers.

### DIFF
--- a/porch/controllers/config/crd/bases/config.porch.kpt.dev_remoterootsyncsets.yaml
+++ b/porch/controllers/config/crd/bases/config.porch.kpt.dev_remoterootsyncsets.yaml
@@ -19,9 +19,9 @@ metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.8.0
   creationTimestamp: null
-  name: remoterootsyncsets.config.cloud.google.com
+  name: remoterootsyncsets.config.porch.kpt.dev
 spec:
-  group: config.cloud.google.com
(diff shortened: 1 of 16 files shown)


justinsb:
Thank you for fixing this!

LGTM other than the eternal issue of deciding the group ... I think we thought `config.porch.kpt.dev` to match the existing CRDs or `alpha.kpt.dev` to indicate that this is even more alpha!

mortent:
I think `config.porch.kpt.dev` seems ok, although I'm a little bit uncertain about including the name of the tool in the group. But I think using the same name as the CRDs seems like the right way to go. 

Including `alpha` in the group name seems tricky, since we would have to change the group when we (hopefully) can go to beta and GA. Also, the version should make it clear it is alpha, but not sure how many actually pay attention to that...

Updated the PR to use `config.porch.kpt.dev`.

justinsb (approved):

justinsb:
Thanks @mortent 

/approve
/lgtm

PR #3486, opened 2022-08-23
WorkloadIdentityBinding: better surface status.conditions

justinsb:
If objects fail to apply, surface that in conditions.


yuwenma (approved):
Great!

PR #3487, opened 2022-08-23
fix: rename example files to fix broken Go import

rquitales:
The presence of colons in the filenames prevented other Go projects from
importing kpt.


justinsb (approved):

justinsb:
LGTM but I think @droot should comment about how these were generated (i.e. do we want to update a makefile or similar)

rquitales:
Hunted down the PR that introduced these files #3371. Looks like an adhoc command was used to generate the yaml by grabbing the cert-manager uber yaml from the github release and splitting it using `kubectl-slice -f cert-manager.yaml --template '{{ index "app" .metadata.labels }}/{{.kind | lower}}-{{.metadata.name|dottodash}}.yaml' -o cert-manager-basic`. This is documented in the README file of the cert-manager package. `kubectl-slice` does not have helper template functions to also convert colons to dashes.

If it helps, here's the adhoc command I used to batch rename the yaml files:
```sh
for file in $(fdfind ":"); do mv $file $(echo $file | sed "s/:/-/"); done
```

droot (approved):

droot:
> The presence of colons in the filenames prevented other Go projects from
importing kpt

Interesting. 

Thanks for the fix and also digging into how these were produced.

PR #3497, opened 2022-08-25
WorkloadIdentityBinding: needs RBAC permissions to read namespaces

justinsb:
We need to get the project-id annotation from the namespace.

### DIFF
--- a/porch/controllers/config/rbac/role.yaml
+++ b/porch/controllers/config/rbac/role.yaml
@@ -27,6 +27,14 @@ rules:
   - create
   - patch
 - apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - config.porch.kpt.dev
   resources:
   - remoterootsyncsets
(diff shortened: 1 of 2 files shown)


justinsb:
cc @apelisse ... what are the permissions needed for SSA?  I couldn't find them documented anywhere.  

/assign @apelisse

porch/controllers/config/rbac/role.yaml
```diff
@@ -26,6 +26,14 @@ rules:
   verbs:
   - create
   - patch
+- apiGroups:
```
mortent, line 29: Currently the porch make targets doesn't use this file directly, but rather use the rbac from https://github.com/GoogleContainerTools/kpt/blob/main/porch/deployments/porch/9-controllers.yaml#L57. I'm meaning to fix this, but until that is fixed we need to update the rbac rules directly in that file.

porch/controllers/workloadidentitybinding/pkg/controllers/workloadidentitybinding/controller.go
```diff
@@ -44,6 +44,12 @@ type WorkloadIdentityBindingReconciler struct {
 //+kubebuilder:rbac:groups=config.porch.kpt.dev,resources=workloadidentitybindings/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=config.porch.kpt.dev,resources=workloadidentitybindings/finalizers,verbs=update
 
+// Needs to read namespace to get project-id annotation
+//+kubebuilder:rbac:groups="",resources=namespaces,verbs=get;list;watch
+
+// Creates iampolicymembers (using server-side-apply)
+//+kubebuilder:rbac:groups=iam.cnrm.cloud.google.com,resources=iampolicymembers,verbs=get;list;watch;create;update;patch;delete
```
yuwenma, line 51: oh, this looks really cool!

liamfallon:
This PR is on Porch. Porch has been moved to the Nephio project.

PR #3498, opened 2022-08-25
alpha apply table output: capture and render errors correctly

justinsb:
We were dropping errors by trying to copy them around; instead keep
the original results and format them directly.

### DIFF
--- a/internal/alpha/printers/table/printer.go
+++ b/internal/alpha/printers/table/printer.go
@@ -11,9 +11,11 @@ import (
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


internal/alpha/printers/table/printer.go
```diff
@@ -161,23 +162,30 @@ func getProgress(resInfo *resourceInfo) (string, string, error) {
...
-		case event.ApplyPending:
-			text = "PendingApply"
-		case event.ApplySuccessful:
-			text = "Applied"
-			printStatus = true
-		case event.ApplySkipped:
-			text = "SkippedApply"
-		case event.ApplyFailed:
-			text = "ApplyFailed"
+		{
```
mortent, line 165: Why the extra set of brackets around this switch statement and not the next one?

mortent (approved):

internal/alpha/printers/table/printer.go
justinsb: Good point - I love my unnecessary brackets!  Removed.

PR #3505, opened 2022-08-26
add resourcegroup vanity server

sdowell:


sdowell:
fyi @rquitales there is no prow here :)

mortent (approved):

PR #3510, opened 2022-08-28
Fix RootSyncSet permissions

mortent:
Updates the rbac permissions to the RootSyncSet controller to include the config connector and hub resources types needed. This also adds an example of how to use the RootSyncSet controller.

### DIFF
--- a/porch/controllers/rootsyncset/pkg/controllers/rootsyncset/controller.go
+++ b/porch/controllers/rootsyncset/pkg/controllers/rootsyncset/controller.go
@@ -43,7 +43,6 @@ import (
 var (
 	rootSyncNamespace  = "config-management-system"
 	rootSyncApiVersion = "configsync.gke.io/v1beta1"
-	rootSyncName       = "root-sync"
 	rootSyncKind       = "RootSync"
 )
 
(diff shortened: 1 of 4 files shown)


porch/controllers/rootsyncset/pkg/controllers/rootsyncset/controller.go
```diff
@@ -58,6 +57,11 @@ type RootSyncSetReconciler struct {
 //+kubebuilder:rbac:groups=config.porch.kpt.dev,resources=rootsyncsets,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=config.porch.kpt.dev,resources=rootsyncsets/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=config.porch.kpt.dev,resources=rootsyncsets/finalizers,verbs=update
+//+kubebuilder:rbac:groups=configcontroller.cnrm.cloud.google.com,resources=configcontrollerinstances,verbs=get;list
```
justinsb, line 60: I always add in watch, personally, even if we're not using it.  My 2c is that unless we're locking it down to one object by name, we should just request get,list,watch for read-only.

justinsb (approved):

justinsb:
One suggestion about adding watch, but not a blocker.

/approve
/lgtm

porch/controllers/rootsyncset/pkg/controllers/rootsyncset/controller.go
mortent: I agree. I have updated the PR.

PR #3519, opened 2022-08-30
function-runner should use WorkloadIdentity

justinsb:
If we're using private images, the function runner needs to pull
details of the image (the entrypoint in particular).

### DIFF
--- a/porch/func/internal/podevaluator.go
+++ b/porch/func/internal/podevaluator.go
@@ -26,7 +26,7 @@ import (
 	"time"
 
 	"github.com/GoogleContainerTools/kpt/porch/func/evaluator"
-	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/gcrane"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"google.golang.org/grpc"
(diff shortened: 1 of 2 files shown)


porch/func/internal/podevaluator.go
```diff
@@ -432,7 +432,7 @@ func (pm *podManager) imageDigestAndEntrypoint(image string) (*digestAndEntrypoi
 	if err != nil {
 		return nil, err
 	}
-	img, err := remote.Image(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain))
+	img, err := remote.Image(ref, remote.WithAuthFromKeychain(gcrane.Keychain))
```
yuwenma, line 435: Maybe add the `remote.WithContext(ctx)` as well? 

https://github.com/GoogleContainerTools/kpt/pull/3519/files#diff-0fd72eeec83b1c25062dfba7e2628742b684217250e4789bc48f4bd4f1ecacc3R215
https://github.com/google/go-containerregistry/blob/v0.11.0/pkg/gcrane/options.go#L89

yuwenma (approved):
A minor comment.

porch/func/internal/podevaluator.go
justinsb: Good call - plumbed through the context and added `remote.WithContext`

PR #3522, opened 2022-08-31
porch: Added renderstatus to packagerevisionresources update api

droot:
This PR exposes `kpt fn render` results in the porch API (`packagerevisionresources`).

Note: Since API no longer returns failure on render failures, the existing tests are failing, I am looking into it but the code is good for review. The change is split in two commits to make it easier to review.

/cc @mortent @justinsb @ChristopherFry 

Some follow ups:

- `rpkg push` should display renderstatus
- functionruntimes (builtin and podevaluator) should support structuredresults and fix exitCode.


porch/api/porch/v1alpha1/types_packagerevisions.go
```diff
@@ -127,6 +128,18 @@ type Task struct {
...
+type TaskResult struct {
+	Type         TaskType      `json:"type"`
+	RenderStatus *RenderStatus `json:"renderStatus,omitempty"`
+}
+
+// RenderStatus represents the result of performing render operation
+// on a package resources.
+type RenderStatus struct {
+	Result fnresult.ResultList `json:"result,omitempty"`
+	Err    string              `json:"err"`
```
justinsb, line 140: Nit: `json:"error"` I suggest

porch/pkg/engine/engine.go
```diff
@@ -846,86 +846,103 @@ func (cad *cadEngine) DeletePackage(ctx context.Context, repositoryObj *configap
...
 }
 
-func applyResourceMutations(ctx context.Context, draft repository.PackageDraft, baseResources repository.PackageResources, mutations []mutation) error {
+// applyResourceMutations mutates the resources and returns the most recent renderResult.
+func applyResourceMutations(ctx context.Context, draft repository.PackageDraft, baseResources repository.PackageResources, mutations []mutation) (applied repository.PackageResources, renderStatus *api.RenderStatus, err error) {
 	for _, m := range mutations {
-		applied, task, err := m.Apply(ctx, baseResources)
+		updatedResources, taskResult, task, err := m.Apply(ctx, baseResources)
+		if taskResult != nil && taskResult.Type == api.TaskTypeEval {
+			renderStatus = taskResult.RenderStatus
```
justinsb, line 928: Does this mean we only capture the last `eval` (i.e. the last function?)

porch/pkg/engine/init.go
```diff
@@ -57,13 +57,13 @@ func (m *initPackageMutation) Apply(ctx context.Context, resources repository.Pa
...
 	}
 
 	result, err := readResources(fs)
 	if err != nil {
-		return repository.PackageResources{}, nil, err
+		return repository.PackageResources{}, nil, nil, err
 	}
 
-	return result, m.task, nil
+	return result, nil, m.task, nil
```
justinsb, line 68: Nit: four retvals is likely where I start to think about a struct

justinsb (approved):
Looking good ... I vote to merge and iterate :-)

porch/pkg/engine/init.go
natasha41575: along these lines, perhaps there could be a struct that contains both `Task` and its `TaskResult`?

natasha41575 (approved):
lgtm

.gitignore
```diff
@@ -20,6 +20,7 @@ kpt/kpt
 *.exe
 *.gif
 bin/
+porch/cmd/porch/__debug_bin
```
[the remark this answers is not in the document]
droot: Found the wildcard `**` matches any subdirectories.

porch/api/porch/v1alpha1/types_packagerevisionresources.go
```diff
@@ -27,8 +27,8 @@ type PackageRevisionResources struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec PackageRevisionResourcesSpec `json:"spec,omitempty"`
-	// Status PackageRevisionResourcesStatuc `json:"status,omitempty"`
+	Spec   PackageRevisionResourcesSpec   `json:"spec,omitempty"`
+	Status PackageRevisionResourcesStatus `json:"status,omitempty"`
```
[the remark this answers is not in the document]
droot: Yes. PackageRevision is the ultimate home.

porch/api/porch/v1alpha1/types_packagerevisions.go
droot: Done.

porch/pkg/engine/engine.go
droot: Yes.

porch/pkg/engine/init.go
droot: Now returning `TaskResult` that contains `*Task`. Thanks.

porch/pkg/engine/engine.go
```diff
@@ -846,86 +845,107 @@ func (cad *cadEngine) DeletePackage(ctx context.Context, repositoryObj *configap
...
+			renderer: cad.renderer,
+			runtime:  cad.runtime,
+		}})
+	if err != nil {
+		// Render failure will not fail the overall API operation.
+		// The render error and result is captured as part of renderStatus above
+		// and is returned in packageresourceresources API's status field. We continue with
+		// saving the non-rendered resources to avoid losing user's changes.
+		// and supress this err.
+		err = nil
```
mortent, line 909: Do we need this block? If we don't use the `err` variable for anything, can't we just replace it with `_` in the call to `applyResourceMutations`?

porch/api/porch/v1alpha1/types_packagerevisionresources.go
```diff
@@ -54,3 +54,9 @@ type PackageRevisionResourcesSpec struct {
 	// Resources are the content of the package.
 	Resources map[string]string `json:"resources,omitempty"`
 }
+
+// PackageRevisionResourcesStatus represents state of the rendered package resources.
+type PackageRevisionResourcesStatus struct {
+	// RenderStatus contains the result of rendering the package resources.
+	RenderStatus RenderStatus `json:"renderStatus,omitempty"`
```
mortent, line 61: So this only contains the result after a push right? We don't actually store the information anywhere, so a subsequent `get` will not contain the error? I would expect that he results from the latest run would be returned on every `get`. Similar to the previous comment, I would expect this to be on the `PackageRevision` rather than the `PackageRevisionResources` type.

porch/pkg/engine/engine.go
mortent: But is the goal that it should contain the output/errors from every function in the pipeline? Or do we abort the run when a function return an error?

mortent (commented):
I agree with the merge and iterate approach, but we should be careful with the API. Probably easier to make changes in the types from the Aggregated APIserver than CRDs though.

Could we add a test of this to make sure we don't accidentally break it?

porch/pkg/engine/engine.go
droot: Good suggestion. Done. Moved the comment above it.

porch/api/porch/v1alpha1/types_packagerevisionresources.go
droot: subsequent `get` will not contain the error. I started with this API because this is the API that get's called when user edits a package and that's the place renderStatus is most useful. And I agree that `renderStatus` should be added to `packagerevision`'s status and should also be persisted so that we get it on every `get`. And that can be done as a follow up incrementally.

droot (commented):
Thanks for the feedback everyone. I have addressed all the feedback that I was planning to address in current cycle. There is definitely more follow ups but I think this is a reasonable checkpoint to get merged.

PR #3531, opened 2022-09-01
Add missing RBAC for WorkloadIdentityBinding controller

mortent:
Adds missing RBAC required for the WorkloadIdentityBinding controller to work. It also restructures the list of RBAC rules a bit to make it clearer which rule is needed by each controller.


yuwenma (approved):
Sweet.

PR #3537, opened 2022-09-06
rpkg get: clean up namespace handling

justinsb:
If we don't specify a namespace, we should only get the current namespace.

If we specify --all-namespaces or -A, we should print the namespace.

### DIFF
--- a/internal/cmdrpkgget/command.go
+++ b/internal/cmdrpkgget/command.go
@@ -22,15 +22,17 @@ import (
 	"github.com/GoogleContainerTools/kpt/internal/docs/generated/rpkgdocs"
 	"github.com/GoogleContainerTools/kpt/internal/errors"
 	"github.com/GoogleContainerTools/kpt/internal/util/porch"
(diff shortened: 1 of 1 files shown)


mortent (approved):

internal/cmdrpkgget/command.go
```diff
@@ -74,14 +80,29 @@ type runner struct {
...
+	packageName   string
+	revision      string
+	printFlags    *get.PrintFlags
+	AllNamespaces bool
 
 	requestTable bool
 }
 
 func (r *runner) preRunE(cmd *cobra.Command, args []string) error {
+	if *r.cfg.Namespace == "" {
```
mortent, line 92: Could we handle this in a single place for all the rpkg commands?

justinsb: Good idea - I'll look at refactoring it out!

justinsb: Done in #3541

PR #3540, opened 2022-09-07
Move cobra commands from internal package into hierarchical structure under commands

mortent:
This moves the kpt cobra commands from the flat structure under the `internal` package to a hierarchical structure in the `commands` package that reflects the way the commands are structured in the kpt cli. This makes it easier to navigate the commands and allows for simpler naming. 

I think of this as a first step towards better separation between the kpt cli and the kpt library as described in https://github.com/GoogleContainerTools/kpt/issues/3297


justinsb (approved):

justinsb:
I really like this - much more self-documenting and better architecturally.

I'm afraid I forced a conflict on you - sorry!  You probably want to get this in before #3541 (I can rebase that pretty easily!)

PR #3542, opened 2022-09-08
Declerative targeting of packages to clusters

mortent:
This adds the RootSyncRollout controller that provides declarative cluster targeting of packages built upon the RootSyncDeployment controller.


porch/controllers/rootsyncrollout/api/v1alpha1/rootsyncrollout_types.go
```diff
@@ -0,0 +1,77 @@
...
+
+// RootSyncRolloutSpec
+type RootSyncRolloutSpec struct {
+	Targets []ClusterTargetSelector `json:"targets,omitempty"`
+
+	Packages PackageSelector `json:"packages,omitempty"`
+}
+
+type ClusterTargetSelector struct {
+	ClusterType ClusterType `json:"clusterType,omitempty"`
```
justinsb, line 29: Maybe this should be a GVK or GK?  We can see how it goes though...

porch/controllers/rootsyncrollout/api/v1alpha1/rootsyncrollout_types.go
```diff
@@ -0,0 +1,77 @@
...
+
+type ClusterType string
+
+const (
+	ConfigConnectorCluster ClusterType = "configConnector"
+	LocalCluster           ClusterType = "local"
+)
+
+type PackageSelector struct {
+	Namespace string `json:"namespace,omitempty"`
```
justinsb, line 42: I guess we need to go cross-namespace on either packages or targets, and I agree that packages is less dangerous!

porch/controllers/rootsyncrollout/pkg/controllers/rootsyncrollout/controller.go
```diff
@@ -0,0 +1,317 @@
...
+				Name:      rootSyncSetName,
+				Namespace: rootSyncSetNamespace,
+			}, &rootsyncset)
+
+			if err != nil && !apierrors.IsNotFound(err) {
+				return ctrl.Result{}, err
+			}
+
+			if apierrors.IsNotFound(err) {
+				klog.Infof("Creating rootsyncset %s", rootSyncSetName)
```
justinsb, line 143: I wonder if we can express this as creating a set of "child objects", and then use applylib or similar.  Should be less code and take care of pruning.  Not a blocker though!

porch/controllers/rootsyncrollout/pkg/controllers/rootsyncrollout/controller.go
```diff
@@ -0,0 +1,317 @@
...
+	nextReconcile, found := r.NextScheduledReconcile[req.NamespacedName]
+	if !found {
+		nextReconcile = time.Time{}
+	}
+	if nextReconcile.After(time.Now().Add(2 * time.Second)) {
+		return ctrl.Result{}, nil
+	}
+
+	newReconcile := time.Now().Add(30 * time.Second)
+	r.NextScheduledReconcile[req.NamespacedName] = newReconcile
```
justinsb, line 179: Do we need to cache this information?  I've always seen it just returning RequeueAfter...

justinsb (approved):

porch/controllers/rootsyncrollout/api/v1alpha1/rootsyncrollout_types.go
mortent: Yeah, we could make this a GK/GVK. That will help with discovery for clusters that are represented as resources in the local cluster, but we still need logic to know how to work with those clusters, i.e. connecting to them. Also, there might be situations where we want to discover clusters through some other mean that listing resources, although we could argue that all APIs must reflect clusters as resources.

porch/controllers/rootsyncrollout/api/v1alpha1/rootsyncrollout_types.go
mortent: I think we need to allow for cross-namespace. I added this partly as a way to allow for restricting the controller to target a subset of packages, given that package revisions do not yet support labels.

porch/controllers/rootsyncrollout/pkg/controllers/rootsyncrollout/controller.go
mortent: Yeah, I did consider that approach, but in the end decided on the more standard controller approach.

porch/controllers/rootsyncrollout/pkg/controllers/rootsyncrollout/controller.go
mortent: Not sure I follow. Are you saying there is a way to query the workqueue to check when the next reconcile is scheduled for? This solution is just temporary until we have support for watches on packagerevisions and I've figured out how to set up watches correctly with labelselectors. That is the next steps here.

mortent:
I agree with the merge and iterate approach, but I think we should come up with an approach where all the "very alpha" controllers doesn't run by default for users trying out Porch. We had some users reporting issues due to our renaming of some of the CRDs in the latest release. We could disable them in the regular porch release and just ask users who want to try them out to build from source. Or even better, publish porch as a kpt package an allow users to opt-in to these controllers if they want to.

johnbelamaric:
> I agree with the merge and iterate approach, but I think we should come up with an approach where all the "very alpha" controllers doesn't run by default for users trying out Porch. We had some users reporting issues due to our renaming of some of the CRDs in the latest release. We could disable them in the regular porch release and just ask users who want to try them out to build from source. Or even better, publish porch as a kpt package an allow users to opt-in to these controllers if they want to.

I also think controllers with GCP-specific functionality need to be separated from the rest of the controllers and optionally deployed. Preferably in a single controller-manager. Alternatively they can be part of KCC, especially if they depend on KCC CRDs.

porch/controllers/rootsyncrollouts/pkg/controllers/rootsyncrollout/controller.go
```diff
@@ -0,0 +1,384 @@
...
+		Owns(&rsdapi.RootSyncDeployment{}).
+		Watches(
+			&source.Kind{Type: &pkgRev},
+			handler.EnqueueRequestsFromMapFunc(r.findRolloutsForPackageRevision),
+		).
+		Complete(r)
+}
+
+func (r *RootSyncRolloutReconciler) findRolloutsForPackageRevision(pkgRev client.Object) []reconcile.Request {
+	// There is not support for reverse lookup by label: https://github.com/kubernetes/kubernetes/issues/1348
```
justinsb, line 369: I think controller-runtime has some ability to build our own indexes, but I agree this is a pain.

justinsb (approved):

porch/controllers/rootsyncrollouts/pkg/controllers/rootsyncrollout/controller.go
mortent: Yeah, there is support for building indexes based on fields in controller-runtime, but I couldn't find support for indexes on labels.

PR #3545, opened 2022-09-09
Simple license scanner

justinsb:
Prototype of a simple license scanner based on the idea of building a
database of license info, rather than relying on detection.


tools/depscan/go.mod
```diff
@@ -0,0 +1,17 @@
+module anthoscli/vulnscan
```
droot, line 1: nit: copy paste

Go introduced db for vulnerabilities recently https://pkg.go.dev/golang.org/x/vulndb

So this looks very aligned with that in approach.

tools/depscan/main.go
```diff
@@ -0,0 +1,302 @@
...
+	buildInfo, err := buildinfo.ReadFile(opts.Binary)
+	if err != nil {
+		return fmt.Errorf("error reading binary info from %q: %w", opts.Binary, err)
+	}
+
+	var errors []error
+
+	var modules []*Module
+	for _, dep := range buildInfo.Deps {
+		for _, ignorePackage := range opts.IgnorePackage {
```
droot, line 103: curious about ignoring package. What led you to add support for this ?

tools/depscan/main.go
```diff
@@ -0,0 +1,302 @@
...
+type RunLicenseScanOptions struct {
+	Binary string
+
+	IgnorePackage []string
+
+	IncludeLicenses bool
+}
+
+func RunLicenseScan(ctx context.Context, opts RunLicenseScanOptions) error {
+	buildInfo, err := buildinfo.ReadFile(opts.Binary)
```
droot, line 94: buildInfo. This is great. TIL.

tools/depscan/main.go
```diff
@@ -0,0 +1,302 @@
...
+	mustShipCodeByLicense := map[string]bool{
+		"APACHE-2.0":       false,
+		"BSD-3-CLAUSE":     false,
+		"BSD-2-CLAUSE":     false,
+		"MIT":              false,
+		"UNICODE-DFS-2016": false,
+		"MPL-2.0":          true,
+		"PUBLICDOMAIN":     false,
+		"ISC":              false,
+	}
```
droot, line 170: you forgot GPL :)

droot (approved):
Thanks @justinsb.

This looks great (small minor nit about the module name).

Pl. merge this, then I can start the process internally for license review.

tools/depscan/main.go
justinsb: Added a note - it's to avoid internal libraries when they are detected as modules.

tools/depscan/main.go
justinsb: I added libraries that we're using in various binaries where I've tried this out ... no GPL yet!

tools/depscan/go.mod
justinsb: Fixed - and removed remnants vuln scanning because go vuln checker is much better there!

PR #3549, opened 2022-09-09
Fix null pointer deref in diff if no Upstream

Ralph7C2:
Running `kpt pkg diff` without an upstream set panics because we attempt
to dereference `kptFile.Upstream` without checking to see if it is nil. 

This commit adds a check and returns early if necessary.

### DIFF
--- a/internal/util/diff/diff.go
+++ b/internal/util/diff/diff.go
@@ -125,7 +125,7 @@ func (c *Command) Run(ctx context.Context) error {
 
 	// Return early if upstream is not set
 	if kptFile.Upstream == nil || kptFile.Upstream.Git == nil {
-		return errors.Errorf("no upstream set")
+		return errors.Errorf("package missing upstream in Kptfile at '%s'", c.Path)
 	}
 
 	// Create a staging directory to store all compared packages
--- a/internal/util/diff/diff.go
+++ b/internal/util/diff/diff.go
@@ -123,6 +123,11 @@ func (c *Command) Run(ctx context.Context) error {
 		return errors.Errorf("package missing Kptfile at '%s': %v", c.Path, err)
 	}
 
+	// Return early if upstream is not set
+	if kptFile.Upstream == nil || kptFile.Upstream.Git == nil {
+		return errors.Errorf("no upstream set")
+	}
+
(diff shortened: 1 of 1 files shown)


internal/util/diff/diff.go
```diff
@@ -123,6 +123,11 @@ func (c *Command) Run(ctx context.Context) error {
 		return errors.Errorf("package missing Kptfile at '%s': %v", c.Path, err)
 	}
 
+	// Return early if upstream is not set
+	if kptFile.Upstream == nil || kptFile.Upstream.Git == nil {
+		return errors.Errorf("no upstream set")
```
natasha41575, line 128: nit: I suggest including the package path in the error message. 

```suggestion
		return errors.Errorf("package missing upstream in Kptfile at '%s', c.Path")
```

natasha41575 (commented):
Thanks for the fix! I can reproduce the error on the main branch and I agree an error message is far better than panicking.

internal/util/diff/diff.go
```diff
@@ -123,6 +123,11 @@ func (c *Command) Run(ctx context.Context) error {
 		return errors.Errorf("package missing Kptfile at '%s': %v", c.Path, err)
 	}
 
+	// Return early if upstream is not set
+	if kptFile.Upstream == nil || kptFile.Upstream.Git == nil {
+		return errors.Errorf("package missing upstream in Kptfile at '%s', c.Path")
```
natasha41575, line 128: Agh sorry my bad

```suggestion
		return errors.Errorf("package missing upstream in Kptfile at '%s'", c.Path)
```

internal/util/diff/diff.go
Ralph7C2: Good call, thanks!

internal/util/diff/diff.go
Ralph7C2: :+1:

natasha41575 (approved):
LGTM (and built kpt from this PR to verify the fix). Will defer to @mortent or @droot (or another code owner) to give their final approval and merge.

droot:
Awesome! Merging this in.. Thanks.

PR #3551, opened 2022-09-11
Use merge commits on approval

mortent:
This changes the way we handle existing commits on approval of packagerevisions from recreating the commits to doing a merge commit. This simplifies the code and also makes sure that the original author and timestamps for each package commit is preserved.

This functionality is already covered by the existing test: https://github.com/GoogleContainerTools/kpt/blob/main/porch/pkg/git/git_test.go#L681

### DIFF
--- a/porch/pkg/git/commit.go
+++ b/porch/pkg/git/commit.go
@@ -267,7 +267,7 @@ func (h *commitHelper) readFile(path string) ([]byte, error) {
 }
 
 // commit stores all changes in git and creates a commit object.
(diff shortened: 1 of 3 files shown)


droot:
I am not super familiar with the git machinery in porch, so can't comment on the actual code, but quick look suggests the new code is indeed simpler. Also, I think this is inline with how an end-user also expects the commits to be handled on approval so overall I am very happy about this change.

droot (approved):

PR #3552, opened 2022-09-13
Annotate KSA with WorkloadIdentityBinding controller

mortent:
This updates the WorkloadIdentityBinding operator to not only create the IAMPolicyMember, but also add the `iam.gke.io/gcp-service-account` annotation to the referenced KSA.


porch/controllers/workloadidentitybinding/pkg/controllers/workloadidentitybinding/controller.go
```diff
@@ -45,46 +46,56 @@ type WorkloadIdentityBindingReconciler struct {
...
-	// 		// remove our finalizer from the list and update it.
-	// 		controllerutil.RemoveFinalizer(&subject, myFinalizerName)
-	// 		if err := r.Update(ctx, &subject); err != nil {
-	// 			return ctrl.Result{}, fmt.Errorf("failed to update %s after delete finalizer: %w", req.Name, err)
-	// 		}
-	// 	}
-	// 	// Stop reconciliation as the item is being deleted
-	// 	return ctrl.Result{}, nil
-	// }
+	myFinalizerName := "config.porch.kpt.dev/finalizer"
```
johnbelamaric, line 57: 1) should this be a package level const?
2) do we use `config.porch.kpt.dev` elsewhere? Why not just `porch.kpt.dev`?

porch/controllers/workloadidentitybinding/pkg/controllers/workloadidentitybinding/controller.go
```diff
@@ -45,46 +46,56 @@ type WorkloadIdentityBindingReconciler struct {
...
+			}
+		}
+	} else {
+		// The object is being deleted
+		if controllerutil.ContainsFinalizer(&subject, myFinalizerName) {
+			// // our finalizer is present, so lets handle any external dependency
+			if err := r.removeWiAnnotation(ctx, &subject); err != nil {
+				// if fail to delete the external dependency here, return with error
+				// so that it can be retried
+				return ctrl.Result{}, fmt.Errorf("have problem to delete external resource: %w", err)
```
johnbelamaric, line 75: Maybe "failed to delete external resource" would be a little better wording.

`removeWiAnnotation` may return IgnoreNotFound errors - which we then wrap in another error, so it won't be ignored. Should we just be returning the error directly, rather than wrapping it?

porch/controllers/workloadidentitybinding/pkg/controllers/workloadidentitybinding/controller.go
mortent: Yeah, moved this to a package level const.
We are using `config.porch.kpt.dev` group for CRDs since the `porch.kpt.dev` group is already in use by the aggregated apiserver.

porch/controllers/workloadidentitybinding/pkg/controllers/workloadidentitybinding/controller.go
```diff
@@ -45,46 +46,56 @@ type WorkloadIdentityBindingReconciler struct {
...
+			}
+			// remove our finalizer from the list and update it.
+			controllerutil.RemoveFinalizer(&subject, myFinalizerName)
+			if err := r.Update(ctx, &subject); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to update %s after delete finalizer: %w", req.Name, err)
```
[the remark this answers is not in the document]
mortent: Fixed.

droot (approved):
I think we can merge this so that nephio community can try it out.

porch/controllers/workloadidentitybinding/pkg/controllers/workloadidentitybinding/controller.go
```diff
@@ -29,6 +29,11 @@ import (
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	finalizerName = "config.porch.kpt.dev/finalizer"
```
johnbelamaric, line 36: Actually, we should change this to something like `config.porch.kpt.dev/workload-identity`. We don't need the word "finalizer" in there - we already know it's a finalizer because that's the list it is in. However, we may have multiple controllers that all use finalizers - the user should be able to tell which controller needs to clear the finalizer by looking at the name.

porch/controllers/workloadidentitybinding/pkg/controllers/workloadidentitybinding/controller.go
```diff
@@ -211,9 +230,55 @@ func (r *WorkloadIdentityBindingReconciler) BuildObjectsToApply(ctx context.Cont
...
 
+func (r *WorkloadIdentityBindingReconciler) addWiAnnotation(ctx context.Context, projectID string, subject *api.WorkloadIdentityBinding) error {
+	sa, err := r.getServiceAccount(ctx, subject)
+	if err != nil {
+		return client.IgnoreNotFound(err)
+	}
+
+	gsaEmail := fmt.Sprintf("%s@%s.iam.gserviceaccount.com", subject.Spec.ResourceRef.Name, projectID)
+
+	patch := client.MergeFrom(sa.DeepCopy())
```
justinsb, line 245: My understanding is that we should not do this now that we have server-side apply.

My understanding is that we should construct an object with only the fields we care about, either by using unstructured.Unstructured (my preference) or the new "for-apply" variants of the typed objects.

porch/controllers/workloadidentitybinding/pkg/controllers/workloadidentitybinding/controller.go
```diff
@@ -211,9 +230,55 @@ func (r *WorkloadIdentityBindingReconciler) BuildObjectsToApply(ctx context.Cont
...
 	return objects, nil
 }
 
+func (r *WorkloadIdentityBindingReconciler) addWiAnnotation(ctx context.Context, projectID string, subject *api.WorkloadIdentityBinding) error {
+	sa, err := r.getServiceAccount(ctx, subject)
+	if err != nil {
+		return client.IgnoreNotFound(err)
+	}
+
+	gsaEmail := fmt.Sprintf("%s@%s.iam.gserviceaccount.com", subject.Spec.ResourceRef.Name, projectID)
```
justinsb, line 243: I think we should to look up the referenced IAMServiceAccount and get the email from the status.  I don't think we can assume the project-id or the name.

porch/controllers/workloadidentitybinding/pkg/controllers/workloadidentitybinding/controller.go
```diff
@@ -45,46 +50,57 @@ type WorkloadIdentityBindingReconciler struct {
 //+kubebuilder:rbac:groups=config.porch.kpt.dev,resources=workloadidentitybindings/finalizers,verbs=update
 //+kubebuilder:rbac:groups=,resources=namespaces,verbs=get;list;watch
 //+kubebuilder:rbac:groups=iam.cnrm.cloud.google.com,resources=iampolicymembers,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=,resources=serviceaccounts,verbs=get;list;watch;create;update;patch;delete
```
justinsb, line 53: Unclear what permissions we need, but probably just get;list;watch and either update or patch.

porch/controllers/workloadidentitybinding/api/v1alpha1/workloadidentitybinding_types.go
```diff
@@ -31,7 +31,7 @@ type WorkloadIdentityBinding struct {
 	Status WorkloadIdentityBindingStatus `json:"status,omitempty"`
 }
 
-// WorkloadIdentityBindingSpec defines the desired state of RemoteRootSync
+// WorkloadIdentityBindingSpec defines the desired state of WorkloadIdentityBinding
 type WorkloadIdentityBindingSpec struct {
 	ServiceAccountRef ServiceAccountRef `json:"serviceAccountRef,omitempty"`
 	ResourceRef       ResourceRef       `json:"resourceRef,omitempty"`
```
justinsb, line 37: ResourceRef is an awkward name, if you have any better ideas :-)

justinsb:
LGTM conceptually, I think we need to be careful of some of the mechanics though.

porch/controllers/workloadidentitybinding/pkg/controllers/workloadidentitybinding/controller.go
mortent: Yeah, I changed this to `config.porch.kpt.dev/workloadidentitybindings`.

porch/controllers/workloadidentitybinding/pkg/controllers/workloadidentitybinding/controller.go
mortent: Updated this to use the SSA patch with unstructured.

porch/controllers/workloadidentitybinding/pkg/controllers/workloadidentitybinding/controller.go
mortent: Updated the PR to look up the IAMServiceAccount. As part of this we also change the field names in the CRD, so this is a very breaking change.

porch/controllers/workloadidentitybinding/pkg/controllers/workloadidentitybinding/controller.go
mortent: Fixed.

porch/controllers/workloadidentitybinding/api/v1alpha1/workloadidentitybinding_types.go
mortent: Updated the name.

porch/controllers/workloadidentitybindings/pkg/controllers/workloadidentitybinding/controller.go
```diff
@@ -213,9 +245,115 @@ func (r *WorkloadIdentityBindingReconciler) BuildObjectsToApply(ctx context.Cont
...
+	saRef := subject.Spec.KubernetesServiceAccountRef
+
+	saName := saRef.Name
+	saNamespace := saRef.Namespace
+	if saNamespace == "" {
+		saNamespace = subject.GetNamespace()
+	}
+
+	var sa unstructured.Unstructured
+	sa.SetGroupVersionKind(corev1.SchemeGroupVersion.WithKind("ServiceAccount"))
```
justinsb, line 311: Oooh this is a nice way of setting apiVersion/kind that I am going to steal shamelessly :-)

porch/controllers/workloadidentitybindings/pkg/controllers/workloadidentitybinding/controller.go
```diff
@@ -213,9 +245,115 @@ func (r *WorkloadIdentityBindingReconciler) BuildObjectsToApply(ctx context.Cont
...
+	return u, err
+}
+
+func (r *WorkloadIdentityBindingReconciler) updateServiceAccount(ctx context.Context, sa *unstructured.Unstructured, subject *api.WorkloadIdentityBinding) error {
+	data, err := json.Marshal(sa)
+	if err != nil {
+		return err
+	}
+
+	mapping, err := r.restMapper.RESTMapping(corev1.SchemeGroupVersion.WithKind("ServiceAccount").GroupKind())
```
justinsb, line 342: It's nice to have this code, but in this case we do know the GVR....

justinsb (approved):

justinsb:
LGTM!

PR #3557, opened 2022-09-18
Changes release URL from jib to kpt

openrory:
The current attestation file gives the following error: FAILED: SLSA verification failed: source used to generate the binary does not match provenance: expected source 'GoogleContainerTools/kpt', got 'GoogleContainerTools/jib' so I changed this file to point to the kpt release page.


mortent (approved):
Good catch. Thanks for fixing this.

PR #3558, opened 2022-09-19
Support allowlisting of reconcilers included in porch deployment and distribution

mortent:
This change allows us to specify at build time which reconcilers should be included with porch. The goal is to be able to develop new reconcilers and CRDs without them being available to regular users of Porch until they are sufficiently mature.

It does this by:
- The controller manager can be configured to enable a subset of the reconcilers either through a command line flag or through env variables.
- The rbac rules are now generated on a per-reconciler basis so we end up with one clusterrole and clusterrolebinding per reconciler. This allows us to include only the ones needed and it makes it clearer which reconciler requires each clusterrole.
- Reconcilers are specified by the plural name of the CRD, i.e. `rootsyncsets`, `workloadidentitybindings`, .. The folders under `porch/controllers` have been renamed to align with this, as it allows the release scripts to automatically create the correct paths and filenames for including CRDs and rbac rules.
- The `create-deployment-blueprint.sh` script takes a list of reconcilers as a parameter, which allows us to specify from the make target whether to enable all or a subset of the reconcilers.


porch/api/generated/clientset/versioned/scheme/register.go
```diff
@@ -35,14 +35,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
 // AddToScheme adds all types of this clientset into the given scheme. This allows composition
 // of clientsets, like in:
 //
-//	import (
-//	  "k8s.io/client-go/kubernetes"
-//	  clientsetscheme "k8s.io/client-go/kubernetes/scheme"
-//	  aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
-//	)
+//   import (
```
justinsb, line 38: It looks like this indentation might be flip-flopping between different authors.... do we have any idea why (not a blocker)?

justinsb:
LGTM!

PR #3561, opened 2022-09-19
WIP: Support labels & annotations

justinsb:
Initial experimental support for labels & annotations

### DIFF
--- a/porch/pkg/meta/annotation.go
+++ b/porch/pkg/meta/annotation.go
@@ -1,0 +1,58 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package meta
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
+)
+
+// Annotation is the structured data that we store with commits.
+// Currently this is stored as a json-encoded blob in the commit message,
+// in future we might use git notes or a similar mechanism.
+// TODO: Rationalize with OCI data structure?
+type Annotation struct {
+	// PackagePath is the path of the package we modified.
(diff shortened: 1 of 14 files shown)


porch/pkg/meta/annotation.go
```diff
@@ -0,0 +1,58 @@
...
+	// belongs to.
+	Revision string `json:"revision,omitempty"`
+
+	// Task holds the task we performed, if a task caused the commit.
+	Task *v1alpha1.Task `json:"task,omitempty"`
+
+	// Annotations holds the annotations.
+	// Note: pointer to a map so we can tell the difference between empty and nil
+	// TODO: Doesn't workk ... omitempty still includes the null
+	Annotations *map[string]string `json:"annotations,omitempty"`
```
johnbelamaric, line 44: maps are already pointers so this shouldn't be necessary. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required and  https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta

johnbelamaric, line 44: *map shouldn't be necessary since map is already a pointer.

porch/pkg/meta/annotation.go
```diff
@@ -0,0 +1,58 @@
...
+
+	// Labels holds the labels.
+	// Note: pointer to a map so we can tell the difference between empty and nil
+	// TODO: Doesn't workk ... omitempty still includes the null
+	Labels *map[string]string `json:"labels,omitempty"`
```
johnbelamaric, line 49: ditto

porch/pkg/meta/annotation.go
```diff
@@ -0,0 +1,58 @@
...
+	// This is useful for disambiguating which package we are modifying in a tree of packages,
+	// without having to check file paths.
+	PackagePath string `json:"package,omitempty"`
+
+	// Revision hold the revision of the package revision the commit
+	// belongs to.
+	Revision string `json:"revision,omitempty"`
+
+	// Task holds the task we performed, if a task caused the commit.
+	Task *v1alpha1.Task `json:"task,omitempty"`
```
johnbelamaric, line 39: Why not just include the PackageMeta struct?

porch/pkg/meta/meta.go
```diff
@@ -0,0 +1,43 @@
...
+// limitations under the License.
+
+package meta
+
+import "github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
+
+type PackageMeta struct {
+	Tasks       []v1alpha1.Task
+	Labels      map[string]string
+	Annotations map[string]string
```
johnbelamaric, line 22: We will also need ownerRefs and maybe finalizers.

johnbelamaric (commented):
Exciting!

PR #3567, opened 2022-09-21
Use controller-oriented RESTMapper in porch controllers

justinsb:
We want to try this out, should be much faster and lighter on the apiserver.

This is code that we're trying the "copy and paste" reuse approach until we find the right library, copied from https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/tree/master/pkg/restmapper .  That code is also Apache licensed and I am the original author.

### DIFF
--- a/porch/pkg/controllerrestmapper/caching.go
+++ b/porch/pkg/controllerrestmapper/caching.go
@@ -1,0 +1,129 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


porch/pkg/controllerrestmapper/caching.go
```diff
@@ -0,0 +1,129 @@
...
+		return nil, err
+	}
+
+	cached, found := kinds[kind]
+	if !found {
+		return nil, nil
+	}
+	return &meta.RESTMapping{
+		Resource:         c.gv.WithResource(cached.Resource),
+		GroupVersionKind: c.gv.WithKind(kind),
```
yuwenma, line 81: Ah, I was wondering the other day how to use the gvk functions with more flexibility.

yuwenma (approved):
LGTM. may defer to @mortent for final approval.

Issue #3448, opened 2022-08-08
Package dependencies: expressing ("my package requires") and fulfilling ("my package provides")


johnbelamaric:
### Describe your problem

As we have discussed a few times, packages often have dependencies. They may depend on particular resources existing in the cluster, or they may depend on availability of services running on an endpoint, or may have other types of dependencies. This issue is a spin off of #3351, to provide a single place to capture examples of these dependencies and discuss ways of managing them within `kpt`. It is not the same as `depends-on`, which is about `apply` ordering within a package's resources.

Some examples:
 * A package needs a namespace in which to be provisioned.
 * A package needs a particular CRD to be installed.
 * A package needs a Secret that must be provisioned out-of-band, OR satisfied via an ExternalSecret.
 * A package needs a value from the runtime instantiation of another package (e.g., a load balancer IP address).
 * A package needs a particular service running (e.g., a database), which may be provisioned in-cluster or out-of-cluster.
 * A package needs a particular shared resource to be available; for example, a shared Ingress Gateway.

Each of these may be satisfied in different ways, and some of them may need to be satisfied by a user with different privilege levels than the package deployer (e.g., deploying a namespace vs into a namespace, or deploying a CRD vs a CR).

We would like to be able to express the *requirement* a package has, without the package having an opinion about *how* the requirement is met. For example, if an application package needs a namespace, that could be provisioned manually, or via deployment of a namespace package, or could be added by the deployer to the application package itself. The application package needs to be able to simply say "I need a namespace", and then the Kpt suite of tools needs to identify whether or not that dependency has been satisfied, at the earliest possible time in the package lifecycle (see #3422).

To that end, packages also need a way to say "I satisfy some requirement". This allows the tooling to understand and match the "package requires" and "package provides". However, we need to figure out how that is expressed in each layer of the solution. For example, at runtime this model may be expressed using a "Claim" pattern; it's not clear that works at Config Time.

Another thought is how this relates to the idea of "mixins". One might consider something like Environment (#3280) as being expressed as a dependency - the package depends on some Environment package/mixin being deployed. The user would choose which (dev/stage/prod) one to use to resolve that dependency.

johnbelamaric:
@droot @yuwenma @bgrant0607 @justinsb

droot:
Thanks for starting this thread @johnbelamaric 

> A package needs a value from the runtime instantiation of another package (e.g., a load balancer IP address).

[...]

PR #3575, opened 2022-09-22
klippy: a controller that auto-instantiates packages

justinsb:
This controller will automatically create draft package instances where the bindings are satisfied.  If the bindings are sufficiently constraining, this can act as a guided workflow to help users do the next right thing.

### DIFF
--- a/porch/controllers/klippy/pkg/controllers/klippy/klippy_controller.go
+++ b/porch/controllers/klippy/pkg/controllers/klippy/klippy_controller.go
@@ -1,0 +1,443 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package klippy
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"path/filepath"
+	"strings"
+	"unicode"
+
+	api "github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
+	"github.com/GoogleContainerTools/kpt/porch/controllers/remoterootsyncsets/pkg/applyset"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/dynamic"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/yaml"
+)
+
+type Options struct {
+	// BindFunction is the image name for the "bind" function
+	BindFunction string
+}
+
+func (o *Options) InitDefaults() {
+	o.BindFunction = "bind"
+}
+
+func (o *Options) BindFlags(prefix string, flagset *flag.FlagSet) {
+	flagset.StringVar(&o.BindFunction, prefix+"bindFunction", o.BindFunction, "image name for the bind function")
+}
+
+// KlippyReconciler reconciles Klippy objects
+type KlippyReconciler struct {
+	Options
+
+	client client.Client
+
+	dynamicClient dynamic.Interface
+	restMapper    meta.RESTMapper
+}
+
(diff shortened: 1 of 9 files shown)


porch/controllers/klippy/pkg/controllers/klippy/klippy_controller.go
```diff
@@ -0,0 +1,436 @@
...
+	// TODO: We should be able to cache this
+	blueprints, err := r.parseBlueprints(ctx, &blueprintPackageRevisions)
+	if err != nil {
+		return err
+	}
+
+	parentObjects, err := parseObjects(ctx, parentResources)
+	if err != nil {
+		return err
+	}
```
droot, line 121: parseObjects and parseBlueprints seems like functions could be useful in other places (porchclient++).

porch/controllers/klippy/pkg/controllers/klippy/klippy_controller.go
```diff
@@ -0,0 +1,436 @@
...
+
+	var allProposals []*api.PackageRevision
+	for _, blueprint := range blueprints {
+		proposals, err := r.buildProposals(ctx, parent, parentObjects, blueprint)
```
droot, line 125: seems like this is the part we would like to externalize (delegate to proposal makers).

porch/controllers/klippy/pkg/controllers/klippy/klippy_controller.go
```diff
@@ -0,0 +1,436 @@
...
+				Eval: &api.FunctionEvalTaskSpec{
+					Image:  image,
+					Config: bindingConfig,
+				},
+			})
+
+		}
+
+		proposals = append(proposals, proposal)
+	}
```
droot, line 358: lots of binding stuff that I skipped over. I am curious, where is `bind` function used that was taken as input ?

droot (commented):
some drive by comments.

johnbelamaric:
This is quite similar to what I am looking for with #3448. The idea is that a local-config resource will describe the "my package requires", and another one in other packages will describe the "my package provides". I want a dependency resolution controller that sees these in Draft package revisions and resolves them. Dependencies themselves are also surfaced as conditions, so that we can identify when all dependencies have been satisfied.

porch/controllers/klippy/pkg/controllers/klippy/klippy_controller.go
justinsb: Yes - I feel like we should have them somewhere already.  I just don't know where...

porch/controllers/klippy/pkg/controllers/klippy/klippy_controller.go
justinsb: Ooops - I forgot to do that - fixed in the latest update!

justinsb:
> This is quite similar to what I am looking for with https://github.com/GoogleContainerTools/kpt/issues/3448. The idea is that a local-config resource will describe the "my package requires", and another one in other packages will describe the "my package provides". I want a dependency resolution controller that sees these in Draft package revisions and resolves them. Dependencies themselves are also surfaced as conditions, so that we can identify when all dependencies have been satisfied.

That's great news, thank @johnbelamaric .  I do expect we'll have multiple controllers making proposals.  I think it would be great if @yuwenma (or @natasha41575 ) tackled updating-the-container-image next.  I don't know whether your needs are close enough that we should extend this controller, or whether we should build a second controller - generally I think we'll evolve our understanding as we build out this functionality (better to have many specialized controllers, or fewer more generic ones?)

mortent (commented):
Looks good to me, but can we add a README file that describes what it does and includes a simple example for testing it? I've found that pretty useful in the other controllers when I need to make changes. So something like this: https://github.com/GoogleContainerTools/kpt/tree/main/porch/controllers/workloadidentitybindings/config/samples

porch/controllers/klippy/pkg/controllers/klippy/rbac.go
```diff
@@ -0,0 +1,21 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package klippy
+
+//+kubebuilder:rbac:groups=porch.kpt.dev,resources=packagerevisionresources,verbs=get;list;watch;create;update;patch;delete
```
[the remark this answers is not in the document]
justinsb: Was mostly just trying it out to see how I/we felt about it.   The RBAC generated is the same either way.  I'll move it into the main controller, to follow the existing patterns.

porch/controllers/klippy/pkg/controllers/klippy/klippy_controller.go
```diff
@@ -0,0 +1,436 @@
...
+func (r *KlippyReconciler) parseBlueprints(ctx context.Context, packageRevisions *api.PackageRevisionList) ([]*blueprint, error) {
+	log := log.FromContext(ctx)
+
+	var blueprints []*blueprint
+
+	for i := range packageRevisions.Items {
+		packageRevision := &packageRevisions.Items[i]
+
+		if !strings.HasPrefix(packageRevision.Spec.PackageName, "infra/") {
+			log.Info("HACK: ignoring non-infra package", "package", packageRevision.Spec.PackageName)
```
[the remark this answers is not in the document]
justinsb: Good call, I removed the hack.  Instead we only apply this where packages actually have bindings, which achieves much the same thing (scoping just to the packages that are using bindings).

justinsb:
Thanks @mortent - I added a README and was able to remove the hack.

mortent (approved):

PR #3578, opened 2022-09-23
Add status.deployment field to PackageRevision

justinsb:
Shows whether the package is in a deployment repo.


mortent:
I'm not necessarily against this change, but what is the use-case for this? Can we assume that clients either already know whether they are working against a deployment repo or they can check the field in the repo?

mortent (approved):

PR #3579, opened 2022-09-23
[WIP] Store metadata for package and packagerevisions in CRs

mortent:
This is still a work-in-progress.

So far:
* Adds a new CRD with group `internal.porch.kpt.dev` and kind `InternalPackageRevision`. Hope to find a better name eventually.
* Creates/Updates/Gets/Deletes the CRD to correspond with `PackageRevision` being accessed through the porch API.

Still remaining:
* Handling for `Package` and `PackageRevisionResources` types.
* Management of the internal CRD for packages that are added through scanning of a repository.
* Cleanup of CRs when Repos or Packages are deleted.


porch/pkg/meta/store.go
```diff
@@ -0,0 +1,174 @@
...
+	configapi "github.com/GoogleContainerTools/kpt/porch/api/porchconfig/v1alpha1"
+	internalapi "github.com/GoogleContainerTools/kpt/porch/internal/api/porchinternal/v1alpha1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type MetadataStore interface {
+	Get(ctx context.Context, name, namespace string) (*internalapi.InternalPackageRevision, error)
```
justinsb, line 29: Suggest taking `types.NamespacedName` or similar to avoid callers passing namespace / name params in the wrong order

porch/pkg/meta/store.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	internalPkgRev.Labels = labels
+	internalPkgRev.Annotations = annos
+	err = c.coreClient.Update(ctx, internalPkgRev)
+	return internalPkgRev, err
+}
+
+func (c *crdMetadataStore) Delete(ctx context.Context, name, namespace string) (*internalapi.InternalPackageRevision, error) {
+	var internalPkgRev internalapi.InternalPackageRevision
+	err := c.coreClient.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, &internalPkgRev)
```
justinsb, line 95: This somewhat breaks optimistic concurrency, but I think this naturally gets fixed if we make PackageRevision itself a CRD.

porch/pkg/meta/store.go
mortent: I've changed the interface for the MetadataStore.

porch/pkg/meta/store.go
mortent: Do you mean splitting the data for each PackageRevision across to types of storage breaks optimistic concurrency or the particular "get and then delete" here?

mortent:
@justinsb I've restructured this PR. It now relies on the `Engine` implementation as the place for combining information from a repository (git/oci) with data from the CRD. I think this seems like the most logical place to handle it. It did require adding a new set of types, which I wish wasn't necessary. But each of them does have a clear use-case:
* API types: Exposes the PackageRevision as a KRM resource.
* Engine types: Internal representation independent of the KRM API that includes both metadata and repo data.
* Repository types: Representation of PackageRevision data from a repository.

The background sync functionality is a bit awkward here, but it is in a weird spot as it works directly on the cache data structure.

porch/pkg/meta/store.go
```diff
@@ -0,0 +1,229 @@
...
+		labels = make(map[string]string)
+	}
+	labels[PkgRevisionRepoLabel] = repo.Name
+	internalPkgRev := internalapi.PkgRevMeta{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        pkgRevMeta.Name,
+			Namespace:   pkgRevMeta.Namespace,
+			Labels:      labels,
+			Annotations: pkgRevMeta.Annotations,
+			// We probably should make these owner refs point to the PackageRevision CRs instead.
```
natasha41575, line 131: Question, couldn't we have owner refs to both the PackageRevision and repo here?

porch/pkg/meta/store.go
```diff
@@ -0,0 +1,229 @@
...
+		Annotations: internalPkgRev.Annotations,
+	}, nil
+}
+
+func (c *crdMetadataStore) Delete(ctx context.Context, namespacedName types.NamespacedName) (PackageRevisionMeta, error) {
+	ctx, span := tracer.Start(ctx, "crdMetadataStore::Delete", trace.WithAttributes())
+	defer span.End()
+
+	var internalPkgRev internalapi.PkgRevMeta
+	err := c.coreClient.Get(ctx, namespacedName, &internalPkgRev)
```
natasha41575, line 207: why do we need to `Get` the object here (L207-L213) before deleting it?

porch/pkg/cache/repository.go
```diff
@@ -373,13 +398,48 @@ func (r *cachedRepository) refreshAllCachedPackages(ctx context.Context) (map[re
...
+			if _, err := r.metadataStore.Delete(ctx, namespacedName); err != nil {
+				return nil, nil, err
+			}
 			r.objectCache.notifyPackageRevisionChange(watch.Deleted, oldPackage)
 		}
 	}
 
+	// We make sure that any PackageRevisionMetas belonging to
+	// this repository does have a corresponding PackageRevision
+	// in git/oci. We
```
natasha41575, line 432: nit: unfinished sentence?

porch/pkg/engine/engine.go
```diff
@@ -42,18 +44,72 @@ type CaDEngine interface {
...
+func (p *PackageRevision) KubeObjectName() string {
+	return p.RepoPackageRevision.KubeObjectName()
+}
+
+func (p *PackageRevision) GetResources(ctx context.Context) (*api.PackageRevisionResources, error) {
+	return p.RepoPackageRevision.GetResources(ctx)
+}
+
+type Function struct {
+	RepoFunction repository.Function
```
natasha41575, line 104: Could we embed this `RepoFunction`, and likewise the `RepoPackage` in the above Package struct? It would reduce the amount of code needed in this file

porch/pkg/engine/engine.go
```diff
@@ -42,18 +44,72 @@ type CaDEngine interface {
...
+func (p *Package) GetPackage() *api.Package {
+	return p.RepoPackage.GetPackage()
+}
+
+func (p *Package) KubeObjectName() string {
+	return p.RepoPackage.KubeObjectName()
+}
+
+type PackageRevision struct {
+	RepoPackageRevision repository.PackageRevision
```
natasha41575, line 74: similar comment, I'm wondering if it would be beneficial to embed this `RepoPackageRevision` so that we don't need to separately define `KubeObjectName` and `GetResources`?

porch/Makefile
```diff
@@ -253,3 +254,17 @@ deploy-no-sa: deployment-config-no-sa
 
 .PHONY: push-and-deploy-no-sa
 push-and-deploy-no-sa: push-images deploy-no-sa
+
+.PHONY: run-in-kind
+run-in-kind:
```
droot, line 259: `kind` support is cool. Recently I ran into situation where I wanted to test using external function in function-runner in my dev setup and couldn't do so. This will help.

porch/internal/api/porchinternal/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,61 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package v1alpha1 contains API Schema definitions for the v1alpha1 API group
+// +kubebuilder:object:generate=true
+// +groupName=internal.porch.kpt.dev
```
droot, line 17: curious, did you use kb to generate these or copied (and edited) ?

porch/internal/api/porchinternal/v1alpha1/internal.porch.kpt.dev_pkgrevmetas.yaml
```diff
@@ -0,0 +1,65 @@
...
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.8.0
+  creationTimestamp: null
+  name: pkgrevmetas.internal.porch.kpt.dev
+spec:
+  group: internal.porch.kpt.dev
+  names:
+    kind: PkgRevMeta
```
droot, line 26: I think we can remove `Meta` from the name. `PkgRev` is shorter and can be used to store any info

porch/pkg/cache/repository.go
```diff
@@ -373,13 +398,48 @@ func (r *cachedRepository) refreshAllCachedPackages(ctx context.Context) (map[re
...
 		}
+		// If a new package is found that doesn't have a corresponding PackageRevisionMeta,
+		// we create one.
+		if _, found := existingPkgRevMetasMap[newPackage.KubeObjectName()]; !found {
+			pkgRevMeta := meta.PackageRevisionMeta{
+				Name:      newPackage.KubeObjectName(),
+				Namespace: newPackage.KubeObjectNamespace(),
+			}
+			if _, err := r.metadataStore.Create(ctx, pkgRevMeta, r.repoSpec); err != nil {
+				return nil, nil, err
```
droot, line 409: hmm...shouldn't we continue to refresh other packages ?

porch/pkg/cache/repository.go
```diff
@@ -373,13 +398,48 @@ func (r *cachedRepository) refreshAllCachedPackages(ctx context.Context) (map[re
...
 		if newPackageRevisionMap[k] == nil {
+			// If a package has been deleted from git/oci,
+			// we also make sure to clean up the corresponding
+			// PackageRevisionMeta.
+			namespacedName := types.NamespacedName{
+				Name:      oldPackage.KubeObjectName(),
+				Namespace: oldPackage.KubeObjectNamespace(),
+			}
+			if _, err := r.metadataStore.Delete(ctx, namespacedName); err != nil {
+				return nil, nil, err
```
droot, line 424: same comment as above.

droot (commented):
Awesome. The change is very well done.

lgtm (some minor questions), but will defer final approval to @justinsb who is more familiar with this machinery.

porch/pkg/engine/engine.go
mortent: I want the new types to be the interface exposed outside of the Engine, and it is a pretty limited number of functions needed. Embedding the types would mean that the full interface of those types would be available. I considered adding interfaces for the types, but it seemed unnecessary at this point.

porch/pkg/engine/engine.go
mortent: Same argument as above. I have however updated the PR so the members of the types are private rather than public.

porch/Makefile
mortent: Yeah, this is helpful for debugging the porch e2e tests locally.

porch/internal/api/porchinternal/v1alpha1/groupversion_info.go
mortent: This was copied from the other types, not generated by kb.

porch/internal/api/porchinternal/v1alpha1/internal.porch.kpt.dev_pkgrevmetas.yaml
mortent: Yeah, I've changed the name of the CRD to `PackageRev`. We want to avoid using the same name as the type for the aggregated apiserver. I've also moved it to the `config.porch.kpt.dev` group. The idea is that we over time want to use this CRD as a replacement for the `PackageRevision` type from the aggregated apiserver, and therefore avoid the internal group and the references to meta.

porch/pkg/cache/repository.go
mortent: I've updated the logic here to make it simpler and also handle automatic retries of creation of the `PackageRev` CRD on every sync.

porch/pkg/cache/repository.go
mortent: Same as above

porch/internal/api/porchinternal/v1alpha1/config.porch.kpt.dev_packagerevs.yaml
```diff
@@ -0,0 +1,65 @@
...
+# limitations under the License.
+
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.8.0
+  creationTimestamp: null
+  name: packagerevs.config.porch.kpt.dev
```
justinsb, line 22: Nit: It's only a temporary name, I guess, but we should generally avoid abbreviations IMO

porch/pkg/cache/repository.go
```diff
@@ -328,22 +338,38 @@ func (r *cachedRepository) refreshAllCachedPackages(ctx context.Context) (map[re
 	// TODO: Avoid simultaneous fetches?
 	// TODO: Push-down partial refresh?
 
+	// Look up all existing PackageRevCRs so we an compare those to the
```
justinsb, line 341: Nit: can

porch/pkg/meta/fake/memorystore.go
```diff
@@ -0,0 +1,101 @@
...
+	"context"
+
+	configapi "github.com/GoogleContainerTools/kpt/porch/api/porchconfig/v1alpha1"
+	"github.com/GoogleContainerTools/kpt/porch/pkg/meta"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// MemoryMetadataStore is an in-memory implementation of the MetadataStore interface. It
```
justinsb, line 27: Looks good.  If we need more complex functionality, we can also use mockkubeapiserver from kubebuilder-declarative-pattern

porch/pkg/meta/store.go
```diff
@@ -0,0 +1,229 @@
...
+					Kind:       configapi.RepositoryGVK.Kind,
+					Name:       repo.Name,
+					UID:        repo.UID,
+				},
+			},
+		},
+	}
+	if err := c.coreClient.Create(ctx, &internalPkgRev); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return c.Update(ctx, pkgRevMeta)
```
justinsb, line 147: Nit: ideally we would not do this, as it indicates that we probably aren't doing optimistic concurrency correctly.  But, fine to tackle later

justinsb (approved):

justinsb:
A few nits, but LGTM!  #ShipIt

porch/internal/api/porchinternal/v1alpha1/config.porch.kpt.dev_packagerevs.yaml
mortent: Yeah, I would rather have called it `packagerevisions`, but that leads to issues since it is already in use. I was thinking that if/when we can remove the `packagerevisions` resource from the aggregated apiserver, we could introduce an packagerevisions alias for this one.

porch/pkg/meta/store.go
mortent: Yeah, the optimistic concurrency is a bit tricky with the two separate sources.

PR #3586, opened 2022-09-28
Keep the repository an impelementation detail of the engine

mortent:
This removes the `OpenRepository` function from the `CaDEngine` interface. Instead of passing the repository to callers, we add the `ListPackageRevisions` and `ListPackages` functions to the interface and move callers to use those instead of the repository.


justinsb (approved):

justinsb:
Thanks - I like this PR and I like the direction.  lgtm

PR #3593, opened 2022-09-29
WIP: add description field to package revision and autoassign revision numbers

natasha41575:
Partially implements https://github.com/GoogleContainerTools/kpt/issues/3467#issuecomment-1236430901

- Adds a new `spec.description` field to the package revision API, which gets used as the branch name for proposed and draft package revisions, and is used to generate the metadata.name hash.
- Deprecates `spec.revision`, and adds `status.revision`. `status.revision` gets assigned to the package revision as the latest revision + 1 when the package revision is published.

TODO: Some additional testing


natasha41575:
cc @mortent @justinsb

porch/pkg/git/git.go
```diff
@@ -486,7 +509,12 @@ func (r *gitRepository) loadPackageRevision(ctx context.Context, version, path s
 
 	var ref *plumbing.Reference = nil // Cannot determine ref; this package will be considered final (immutable).
 
-	packageRevision, err := krmPackage.buildGitPackageRevision(ctx, version, ref)
+	description, err := getPkgDescription(commit, path)
```
mortent, line 512: I would expect that the description was a parameter to the `loadPackageRevision` function, as it is one of the properties that uniquely identifier a packagerevision (although in the draft state). Or how do we handle it since we effectively end up with somewhat different properties depending on whether a package is published or not.

porch/pkg/git/git.go
```diff
@@ -286,20 +286,42 @@ func (r *gitRepository) CreatePackageRevision(ctx context.Context, obj *v1alpha1
...
 
-	draft := createDraftName(obj.Spec.PackageName, obj.Spec.Revision)
+	// for backwards compatibility, if spec.revision is provided and spec.description is not, use spec.revision
+	// as spec.description
+	description := obj.Spec.Description
+	if description == "" {
+		if obj.Spec.Revision == "" {
+			return nil, fmt.Errorf("package revision with name %s in repo %s is missing spec.description", obj.Spec.PackageName, obj.Spec.RepositoryName)
+		}
+		description = obj.Spec.Revision
```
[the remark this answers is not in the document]
natasha41575: I wasn't sure the best way to handle the current CLI tests without this, as the CLI (e.g. kpt alpha rpkg copy) at the moment passes in a revision but not the description. That means that without this, the CLI tests will all fail as it won't be able to create any package revisions. I was thinking here to do this temporarily, patch up the CLI, then remove this. 

As an alternative, I could temporarily disable the broken CLI tests for this PR (I've already sort of done this for some of the update tests), and reenable those tests when I patch up the CLI. I plan to make the corresponding PR for the CLI in a close followup this week. Do you think this alternative would be better?

porch/pkg/git/git.go
```diff
@@ -927,7 +961,12 @@ func (r *gitRepository) loadTasks(ctx context.Context, startCommit *object.Commi
 		}
 
 		for _, gitAnnotation := range gitAnnotations {
-			if gitAnnotation.PackagePath == packagePath && gitAnnotation.Revision == revision {
+			packageMatches := gitAnnotation.PackagePath == packagePath
+			descriptionMatches := gitAnnotation.Description == description ||
+				// this is needed for porch package revisions created before the description field existed
+				(gitAnnotation.Revision == description && gitAnnotation.Description == "")
+
+			if packageMatches && descriptionMatches {
```
natasha41575, line 969: Here is the scenario in my mind:

I create a porch package draft (before the description field was added). I give it a revision of v1. I do some work on it, so that it has some commits with the annotation revision:v1.

This PR goes in. My porch package draft now has an empty revision but a description v1. I do some work so this package revision now has some commits with annotation description:v1. When I load up tasks, I need to look for:
- commits that have a revision:v1, but no description.
- commits with a description:v1. 

Then let's say I publish the package revision and it gets auto-assigned a revision of v3. When I load tasks, what I look for is unchanged:
- commits that have a revision:v1, but no description.
- commits with a description:v1. 

For porch packages revisions that were created after the description field was introduced, all git commit annotations should have a populated description field, so the check here is still valid (just match based on the description).

porch/pkg/git/draft.go
```diff
@@ -112,13 +114,22 @@ func (d *gitPackageDraft) Close(ctx context.Context) (repository.PackageRevision
 
 func (r *gitRepository) closeDraft(ctx context.Context, d *gitPackageDraft) (*gitPackageRevision, error) {
 	refSpecs := newPushRefSpecBuilder()
-	draftBranch := createDraftName(d.path, d.revision)
-	proposedBranch := createProposedName(d.path, d.revision)
+	draftBranch := createDraftName(d.path, d.description)
```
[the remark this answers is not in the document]
natasha41575: That's a good point, I hadn't implemented any restrictions here. What if we imposed restrictions similar to kubernetes object names?

- at most 63 characters long
- contain only lowercase alphanumeric characters or '-'
- start with an alphanumeric character
- end with an alphanumeric character

It's a somewhat arbitrary set of restrictions, but it should ensure that branch names are valid. I added this to the PR, but can change it if you have feedback.

porch/pkg/git/git.go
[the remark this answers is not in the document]
natasha41575: I took it out in the most recent commit (so that spec.revision is not being used), so we can see which CLI tests fail.

porch/pkg/git/git.go
```diff
@@ -927,7 +961,12 @@ func (r *gitRepository) loadTasks(ctx context.Context, startCommit *object.Commi
 		}
 
 		for _, gitAnnotation := range gitAnnotations {
-			if gitAnnotation.PackagePath == packagePath && gitAnnotation.Revision == revision {
+			packageMatches := gitAnnotation.PackagePath == packagePath
+			descriptionMatches := gitAnnotation.Description == description ||
+				// this is needed for porch package revisions created before the description field existed
+				(gitAnnotation.Revision == description && gitAnnotation.Description == "")
```
natasha41575, line 967: Here is the scenario:

I create a porch package draft (before the description field was added). I give it a revision of v1. I do some work on it, so that it has some commits with the annotation revision:v1.

This PR goes in. My porch package draft now has an empty revision but a description v1. I do some work so this package revision now has some commits with annotation description:v1. When I load up tasks, I need to look for:

commits that have a revision:v1, but no description.
commits with a description:v1.

Then let's say I publish the package revision and it gets auto-assigned a revision of v3. When I load tasks, what I look for is unchanged:

commits that have a revision:v1, but no description.
commits with a description:v1.

For porch packages revisions that were created after the description field was introduced, all git commit annotations should have a populated description field, so the check here is still valid (just match based on the description).

porch/pkg/git/git.go
natasha41575: Thanks for pointing this out, I took some more time to go through this code today.

It looks like the `version` that gets passed through here (through `GetPackageRevision` and then to `loadPackageRevision`) isn't actually necessarily a revision number, it can be any git ref (a branch name, tag, commit SHA, etc). It is `loadPackageRevision`'s job to deference this into a commit and package revision number. 

I think `loadPackageRevision` is incorrectly assuming that the version passed in is always just the revision number, which leads to incorrect behavior. For example, I invoked this code using something like `kpt alpha rpkg clone https://github.com/natasha41575/blueprints.git/foo -ndefault --repository=deployments test --ref=drafts/foo/v4 --revision=v14` to see what would happen when I passed in a ref with a draft prefix. Stepping through, I can see that the current `loadPackageRevision` actually returns a git package revision with a revision field of `drafts/foo/v4` in this case. Similarly, I ran `kpt alpha rpkg clone https://github.com/natasha41575/blueprints.git/foo -ndefault --repository=deployments test2 --ref=foo/v3 --revision=v15` (where `foo/v3` corresponds to a published tag). Stepping through `loadPackageRevision` again, and I can see that this results in a git package revision with a revision field of `foo/v3`. That doesn't seem correct to me.  

This code is only invoked when we clone directly from git and we are trying to fetch the package revision from upstream, so I don't think the incorrect revision/description makes a difference here. But I updated `loadPackageRevision` in my most recent commit so that we do actually parse the passed in ref to the right revision/description based on the prefix of the ref.

porch/pkg/cache/draft.go
```diff
@@ -31,6 +31,11 @@ func (cd *cachedDraft) Close(ctx context.Context) (repository.PackageRevision, e
 	if closed, err := cd.PackageDraft.Close(ctx); err != nil {
 		return nil, err
 	} else {
-		return cd.cache.update(closed)
+		updated, err := cd.cache.update(closed)
+		if err != nil {
+			return nil, err
+		}
+		_, _, err = cd.cache.refreshAllCachedPackages(ctx)
```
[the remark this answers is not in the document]
natasha41575: The reason was that because the package revision key now changes when it is published (since it gets assigned a revision), the old cache was invalid as it had the wrong key stored in it. But I think you're right that refreshing the whole cache isn't necessary. I instead modified the `update` code in `porch/pkg/cache/repository.go` to delete the old key from the cache before adding the new one, which I think equally fixes the issue.

porch/pkg/git/git.go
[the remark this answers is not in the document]
natasha41575: My most recent commit disables the CLI tests

natasha41575:
@mortent @justinsb I addressed comments, this is ready for another review. PTAL when you can. 

An FYI - this PR disables the CLI tests for reasons described in https://github.com/GoogleContainerTools/kpt/pull/3593#discussion_r986364537 - I don't plan to merge this until I have a working PR that fixes the CLI and can reenable the CLI tests.

porch/pkg/git/git.go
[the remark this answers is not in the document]
mortent: That sounds good.

porch/pkg/git/git.go
```diff
@@ -286,20 +286,39 @@ func (r *gitRepository) CreatePackageRevision(ctx context.Context, obj *v1alpha1
 		return nil, fmt.Errorf("error when resolving target branch for the package: %w", err)
 	}
 
-	draft := createDraftName(obj.Spec.PackageName, obj.Spec.Revision)
+	if obj.Spec.Description == "" {
```
mortent, line 289: We should definitely add this validation in the api layer as well, as well as here. Checks that only affects the current resource (old and new) should probably be in the validator. Checks for uniqueness is a bit more challenging. Do we currently do any checks to verify uniqueness of revisions? I seem to remember that we have a bug for this. It seems like guaranteeing uniqueness will be difficult since users can also create packagerevisions directly in git/oci.

porch/pkg/git/git.go
```diff
@@ -486,7 +506,27 @@ func (r *gitRepository) loadPackageRevision(ctx context.Context, version, path s
...
+	var revision string
+	var description string
+	last := strings.LastIndex(version, "/")
+
+	if strings.HasPrefix(version, "drafts/") || strings.HasPrefix(version, "proposed/") {
+		// the passed in version is a ref to an unpublished package revision
+		description = version[last+1:]
+	} else {
+		// the passed in version is a ref to a published package revision
+		if last < 0 {
```
mortent, line 518: How do we handle packagerevisions on the main branch here? Those will not follow the naming convention of tags.

porch/pkg/git/git.go
```diff
@@ -520,7 +560,11 @@ func (r *gitRepository) discoverFinalizedPackages(ctx context.Context, ref *plum
 
 	var result []repository.PackageRevision
 	for _, krmPackage := range krmPackages.packages {
-		packageRevision, err := krmPackage.buildGitPackageRevision(ctx, revision, ref)
+		description, err := getPkgDescription(commit, krmPackage.path)
```
mortent, line 563: This seems to only check the annotation on the latest commit message. Is that always correct? In particular for packages on the main branch, I think we need to be prepared for there being unrelated commits after the latest commit for the package. Can we use any of the `findLatestPackageCommit` or `packageRevisionHistoryIterator` functions for this?

porch/pkg/git/git.go
mortent: Yeah, I think this sounds good. The key aspect here seems to be that we continue to look up packagerevisions based on the description even after they have been published. I think that sounds ok. We end up with a requirement that the description is unique among all revisions of a package, but I can't really think of any way to avoid that.

porch/pkg/git/package.go
```diff
@@ -55,28 +56,36 @@ var _ repository.PackageRevision = &gitPackageRevision{}
...
-		Package:    p.path,
-		Revision:   p.revision,
+		Repository:  p.repo.name,
+		Package:     p.path,
+		Revision:    p.revision,
+		Description: p.description,
 	}
 }
 
 func (p *gitPackageRevision) uid() types.UID {
```
mortent, line 81: Does this need the same logic as `KubeObjectName` to make sure it is unique for packagerevisions created before this change and therefore can have an empty description?

porch/pkg/oci/mutate.go
```diff
@@ -51,21 +51,40 @@ func (r *ociRepository) CreatePackageRevision(ctx context.Context, obj *api.Pack
 		return nil, err
 	}
 
+	if obj.Spec.Description == "" {
```
mortent, line 54: This seems like the same checks as we do for git?

porch/pkg/git/draft.go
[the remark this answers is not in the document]
mortent: I think that sounds good to me. Does that also align with tags in oci?

porch/api/porch/types_packagerevisionresources.go
```diff
@@ -45,6 +45,9 @@ type PackageRevisionResourcesSpec struct {
 	// PackageName identifies the package in the repository.
 	PackageName string `json:"packageName,omitempty"`
 
+	// Description identifies the description of the package.
+	Description string `json:"description,omitempty"`
```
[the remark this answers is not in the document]
natasha41575: I agree that `description` sounds like some free text, so we should change the name. I think at first glance `revisionName` makes sense, but it might be confusing when you see both the `revision` and `revisionName` fields. It can get even more confusing when you consider that the object is called a `PackageRevision` and has its own `metadata.name` and `packageName`.  I feel like the words `revision` and `name` are becoming overloaded.

WDYT about calling it `tag`?

porch/api/porch/types_packagerevisions.go
```diff
@@ -78,6 +86,10 @@ type ParentReference struct {
 
 // PackageRevisionStatus defines the observed state of PackageRevision
 type PackageRevisionStatus struct {
+	// Revision identifies the version of the package.
+	Revision string `json:"revision,omitempty"`
```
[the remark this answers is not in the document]
natasha41575: SGTM, will move back to spec.

porch/pkg/cache/repository.go
```diff
@@ -214,6 +214,17 @@ func (r *cachedRepository) update(updated repository.PackageRevision) (*cachedPa
 	k := updated.Key()
 	// previous := r.cachedPackageRevisions[k]
 
+	if updated.Lifecycle() == v1alpha1.PackageRevisionLifecyclePublished {
+		oldKey := repository.PackageRevisionKey{
+			Repository:  k.Repository,
+			Package:     k.Package,
+			Description: k.Description,
+		}
+		if _, ok := r.cachedPackageRevisions[oldKey]; ok {
```
justinsb, line 223: Ah - good thinking - I would have missed this one :-)

porch/pkg/git/draft.go
```diff
@@ -33,16 +33,17 @@ import (
...
-	base      *plumbing.Reference // ref to the base of the package update commit chain (used for conditional push)
-	branch    BranchName          // name of the branch where the changes will be pushed
-	commit    plumbing.Hash       // Current HEAD of the package changes (commit sha)
-	tree      plumbing.Hash       // Cached tree of the package itself, some descendent of commit.Tree()
-	tasks     []v1alpha1.Task
+	parent      *gitRepository
+	path        string
+	revision    string
+	description string
+	lifecycle   v1alpha1.PackageRevisionLifecycle // New value of the package revision lifecycle
```
justinsb, line 40: We really should put comments before the fields (in normal go style) so the diffs are more reasonable :-)  But I know it's not strictly related to this PR!

porch/pkg/git/draft.go
justinsb, line 117: tip: If you define a `type WorkspaceName string` or similar, the go compiler should identifying anywhere we need to pass a WorkspaceName.  (I don't see any problems, just that we should get the compiler to help us!)

porch/pkg/git/git.go
justinsb: We could also check uniqueness against the branches stored in git/oci ... that's what matters I believe.

git has the ability to push a branch only if it does not exist, oci does not (yet).  We need to do some work to plumb through the "before" state so we can do a compare-and-swap operations against git, but I don't think we want to tackle that in this PR!

porch/pkg/repository/util.go
```diff
@@ -0,0 +1,101 @@
...
+
+func ValidateDescription(descr string) error {
+	if len(descr) > 63 || len(descr) == 0 {
+		return fmt.Errorf("description %q must be at least 1 and at most 63 characters long", descr)
+	}
+	if strings.HasPrefix(descr, "-") || strings.HasSuffix(descr, "-") {
+		return fmt.Errorf("description %q must start and end with an alphanumeric character", descr)
+	}
+
+	match, err := regexp.MatchString(`^[a-z0-9-]{0,63}$`, descr)
```
justinsb, line 81: Personally I would just skip the regex at this point, given that you're doing all the interesting checking up above, but .. I do detest regex :-)

porch/pkg/repository/util.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return "", err
+	}
+	i++
+	next := "v" + strconv.Itoa(i)
+	return next, nil
+}
+
+func ValidateDescription(descr string) error {
+	if len(descr) > 63 || len(descr) == 0 {
```
justinsb, line 74: We should add a comment to say that `/` should _never_ be allowed, because that is used to different e.g. the `drafts/` prefix.  I don't think anything else is as catastrophic, but we never want to allow that one!

porch/pkg/repository/util.go
```diff
@@ -0,0 +1,101 @@
...
+		// Only consider Published packages
+		if current.Lifecycle() != v1alpha1.PackageRevisionLifecyclePublished {
+			continue
+		}
+
+		currentRev := current.Key().Revision
+
+		if err := ValidateRevisionNumber(currentRev); err != nil {
+			// ignore this revision
+			continue
```
justinsb, line 49: Rather than ignoring, we could collect the max version, and then we could produce the next major version.  So e.g. if we find `v3.1.1` we would do `v4`

justinsb (approved):

justinsb:
I think this looks solid, and I suggest using `workspaceID` or `workspaceName` as the "name" concept for a draft package.

If the tests are hard because of the prebuilt tar files that are the git repos, I think in general we had to evolve the tests over time because we wanted test coverage before we had enough functionality to work with packages end to end.  At this point I think we could likely replace a lot of our tests with e2e tests, that create Repositories and PackageRevisions and drive them through the various changes.  Tests that this works against existing git/oci repos are much trickier, and TBH I don't know if we want to do much more than check that the results are sensible against repos that were manually created (at this early stage).  So for a manually created repo, I would expect that we would recognize the various versions as being "patches" i.e. treat them as direct edits.

One particularly nice thing about doing this is that I think the tests will be simpler and more clearly user-focused, and it should be easy to set them up to loop over various backends (git vs oci) in the various permutations for blueprint repo and deployment repo.

porch/pkg/git/package.go
natasha41575: I think it needs the same logic as `KubeObjectName` to make sure that packages on the main branch have a unique uid (since the description on `main` would be the same as the description of the latest publish revision) - I'll make that change. 

But I don't think the description field would ever be empty; everywhere that a `gitPackageRevision` is initialized, I initialize the description to be the same as the revision if it would otherwise be empty. I don't think we should be letting any package descriptions be empty, it would complicate a lot of the other checks, so this seemed like a reasonable default.

porch/pkg/git/draft.go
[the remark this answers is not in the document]
natasha41575: From https://docs.docker.com/engine/reference/commandline/tag/:

> A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and dashes. A tag name may not start with a period or a dash and may contain a maximum of 128 characters.

The same set of restrictions should be good for OCI.

porch/pkg/git/git.go
mortent: Yeah, I agree that for git we should leverage uniqueness of branches to handle this. Also agree that we should handle it separately.

porch/pkg/repository/util.go
natasha41575: That's a great idea, done!

porch/pkg/repository/util.go
natasha41575: Added!

porch/pkg/repository/util.go
natasha41575: I simplified the regex so that it's not duplicating any of the above checks! For me, what's left is the simplest way to check that all characters are lowercase alphanumeric or '-'.

porch/pkg/git/draft.go
natasha41575: Done

porch/pkg/git/draft.go
natasha41575: Added

porch/pkg/oci/mutate.go
natasha41575: Yeah, it's the same, are we expecting different checks?

porch/pkg/git/package.go
natasha41575: Added the logic to `uid`

porch/pkg/git/git.go
natasha41575: I updated `getPkgWorkspace` so that it uses `findLatestPackageCommit` when we are on the main branch.

porch/pkg/git/git.go
natasha41575: Added the empty description validation in the api layer. I agree that leveraging git branches would be helpful for uniqueness, but I can tackle that in a future PR if we should handle that separately.

porch/api/porch/types_packagerevisions.go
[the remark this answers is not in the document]
natasha41575: Done.

porch/api/porch/types_packagerevisionresources.go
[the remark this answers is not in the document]
natasha41575: Changed to `workspaceName`.

porch/pkg/cache/cache_test.go
```diff
@@ -76,8 +76,8 @@ func TestPublishedLatest(t *testing.T) {
 	_, cached := openRepositoryFromArchive(t, ctx, tarfile, "publish-test")
 
 	revisions, err := cached.ListPackageRevisions(ctx, repository.ListPackageRevisionFilter{
-		Package:  "catalog/gcp/bucket",
-		Revision: "v2",
+		Package:     "catalog/gcp/bucket",
+		Description: "v2",
```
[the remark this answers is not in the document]
natasha41575: It'll break things unless I re-make the tar file it's reading from

mortent (approved):

PR #3595, opened 2022-09-30
Fix make generate target for porch

mortent:


porch/internal/tools/tools.go
```diff
@@ -0,0 +1,22 @@
...
+// limitations under the License.
+
+//go:build tools
+// +build tools
+
+package tools
+
+import (
+	_ "golang.org/x/tools/cmd/stringer"
+)
```
droot, line 22: Why is this `tools` package needed ?

mortent: It allows us to use the import without go complaining about it during a regular build. We already have a similar pattern in use the `porch/api` module. 
This is needed as a result of https://github.com/GoogleContainerTools/kpt/pull/3577.

droot (approved):

PR #3596, opened 2022-10-03
Populate package-path metadata into package context

justinsb:
We construct a path based on the hierarchy of parent packages.  This
is useful for packages/package functions that want to construct values
meaningful beyond just their package - for example domain names, GCP
project IDs, bucket names, descriptions etc.

### DIFF
--- a/internal/builtins/pkg_context.go
+++ b/internal/builtins/pkg_context.go
@@ -30,7 +30,9 @@ import (
 
 const (
 	PkgContextFile = "package-context.yaml"
-	pkgContextName = "kptfile.kpt.dev"
+	PkgContextName = "kptfile.kpt.dev"
+
+	ConfigKeyPackagePath = "package-path"
 )
 
 var (
@@ -42,7 +44,17 @@ var (
 // a KRM object that contains package context information that can be
 // used by functions such as `set-namespace` to customize package with
 // minimal configuration.
-type PackageContextGenerator struct{}
+type PackageContextGenerator struct {
+	// PackageConfig contains the package configuration to set.
+	PackageConfig *PackageConfig
+}
+
+// PackageConfig holds package automatic configuration
+type PackageConfig struct {
+	// PackagePath is the path to the package, as determined by the names of the parent packages.
+	// The path to a package is the parent package path joined with the package name.
+	PackagePath string
+}
 
 // Run function reads the function input `resourceList` from a given reader `r`
 // and writes the function output to the provided writer `w`.
@@ -66,14 +78,14 @@ func (pc *PackageContextGenerator) Process(resourceList *framework.ResourceList)
 	// - Generates a package context resource for each kpt package (i.e Kptfile)
 	for _, resource := range resourceList.Items {
 		gvk := resid.GvkFromNode(resource)
-		if gvk.Equals(configMapGVK) && resource.GetName() == pkgContextName {
+		if gvk.Equals(configMapGVK) && resource.GetName() == PkgContextName {
 			// drop existing package context resources
 			continue
 		}
 		updatedResources = append(updatedResources, resource)
 		if gvk.Equals(kptfileGVK) {
 			// it's a Kptfile, generate a corresponding package context
-			pkgContext, err := pkgContextResource(resource)
+			pkgContext, err := pkgContextResource(resource, pc.PackageConfig)
 			if err != nil {
 				resourceList.Results = framework.Results{
 					&framework.Result{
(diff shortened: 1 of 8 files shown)


porch/pkg/engine/engine.go
```diff
@@ -104,10 +112,56 @@ func (cad *cadEngine) ListPackageRevisions(ctx context.Context, repositorySpec *
 	return repo.ListPackageRevisions(ctx, filter)
 }
 
-func (cad *cadEngine) CreatePackageRevision(ctx context.Context, repositoryObj *configapi.Repository, obj *api.PackageRevision) (repository.PackageRevision, error) {
+func buildPackageConfig(ctx context.Context, obj *api.PackageRevision, parent repository.PackageRevision) (*builtins.PackageConfig, error) {
```
droot, line 115: Can you remind me.. what `parent` means in context of porch ? few things popping up in my mind: `upstream` package, `parent` of a draft package and parent package of a subpackage (which we don't support in porch yet).

internal/builtins/pkg_context.go
```diff
@@ -42,7 +44,17 @@ var (
...
+type PackageContextGenerator struct {
+	// PackageConfig contains the package configuration to set.
+	PackageConfig *PackageConfig
+}
+
+// PackageConfig holds package automatic configuration
+type PackageConfig struct {
+	// PackagePath is the path to the package, as determined by the names of the parent packages.
+	// The path to a package is the parent package path joined with the package name.
+	PackagePath string
```
droot, line 56: Can you share what this is going to look like in `CLI` and `porch` based workflow. Is it relative ?

droot (commented):
drive by questions.

porch/pkg/engine/engine.go
justinsb: Not sure I've got the best name, but `parent` will let us establish a "position" of the package, in the same way that the directory structure does in git.

This proves to be a useful concept, I think, for value propagation and just to organize things.

This isn't done yet, but I would hope we could map this to directories in a git repo, so we would likely need this for CLI/server parity.

internal/builtins/pkg_context.go
justinsb: So as I have it now, it's going to look like a directory path, `myorg/apps/coolapp/dev` or something.  I'm not certain what comes at the start of that (for example, should we try to make these _globally_ unique, or at least allow that, vs being relative to the repository?)  I'm also not certain whether we should include some notion of the package type, in which case the package path would end up looking like a GCP URL `organizations/myorg/folders/apps/applications/coolapp/environments/dev`.

I figure we can iterate a bit here.

I think it would be the same in the CLI, though I'm not yet sure what we should do if you move it (which is the same problem we have if you rename the package - should we update the name in package-config.yaml?)

droot: > for example, should we try to make these _globally_ unique, or at least allow that, vs being relative to the repository?

`relative to the repository` in a deployment repo is the immediate usecase that I can relate with. Others are less clear to me as well. [Since we don't operate in context of repo in pure CLI experience, we can probably keep this no-op in CLI mode.]

About including package-type in the package paths feels like mixing logical view of your packages with the filesystem view (if they map 1:1 well, then we are good, otherwise screwed) :)

I think refactoring a package (like move package to new location or rename), I think the only way to handle them properly, will be to introduce refactoring functionality in our workflows for ex. `pkg rename|move|....` if done through our tooling, we can update the underlying metadata automatically.

internal/builtins/pkg_context.go
```diff
@@ -118,9 +130,16 @@ func pkgContextResource(kf *yaml.RNode) (*yaml.RNode, error) {
...
-		"name": kf.GetName(),
-	})
+	data := map[string]string{
+		"name": kptfile.GetName(),
+	}
+	if packageConfig != nil {
+		if packageConfig.PackagePath != "" {
+			data[ConfigKeyPackagePath] = packageConfig.PackagePath
+		}
+	}
```
droot, line 140: I am guessing this conditional will ensure package-path is not added for `kpt CLI` based workflows ?

porch/pkg/engine/builtin.go
```diff
@@ -33,17 +33,13 @@ type builtinEvalMutation struct {
...
-func newBuiltinFunctionMutation(function string) (mutation, error) {
-	var runner fn.FunctionRunner
-	switch function {
-	case fnruntime.FuncGenPkgContext:
-		runner = &builtins.PackageContextGenerator{}
-	default:
-		return nil, fmt.Errorf("unrecognized built-in function %q", function)
+func newPackageContextGeneratorMutation(packageConfig *builtins.PackageConfig) (mutation, error) {
+	runner := &builtins.PackageContextGenerator{
+		PackageConfig: packageConfig,
```
droot, line 38: I see it now. This confirms this will be done only in porch based workflow.

porch/pkg/engine/engine.go
droot: `positionalParent` is how I will remember this now :)

droot (approved):

internal/builtins/pkg_context.go
justinsb: I believe so, correct.  It'll set it only if we have something to set.

Based on the discussion above, I sort of feel like we shouldn't be persisting either package-path or package-name into the package-context.yaml.  Rather we should be generating it on the fly every time; then we don't have a conflict between directory name / directory path vs package-name/package-path.

The exception to that would be if we wanted the _original_ name, so that if we generated something immutable it would be preserved.  However, we do have other answers there, like trying to capture the name in the parameters in the kpt function pipeline or similar.

Sounds like we need to think more about the challenges here, but that we should merge this so we can gain some experience with how it behaves in practice.

PR #3602, opened 2022-10-06
[WIP] Immediate poll of repo and enable race detector

mortent:
This updates the polling loop for the repositories so it will run during tests. This leads to concurrency problems as demonstrated when enabling the race detector.


yuwenma (approved):
Really cool catch!

natasha41575 (approved):

PR #3603, opened 2022-10-06
porch: restrict `clone` and `init` to only create the first package revision

natasha41575:
Partially implements https://github.com/GoogleContainerTools/kpt/issues/3467#issuecomment-1236430901

Only the first package revision can be made by `init` or `clone`. Subsequent package revisions should be created with `edit`/`copy`.

This adds the check on both the CLI and server side.

### DIFF
--- a/commands/alpha/rpkg/clone/command.go
+++ b/commands/alpha/rpkg/clone/command.go
@@ -114,6 +114,10 @@ func (r *runner) preRunE(cmd *cobra.Command, args []string) error {
 	source := args[0]
 	target := args[1]
 
+	if err := r.packageAlreadyExists(target); err != nil {
+		return err
+	}
+
 	switch {
 	case strings.HasPrefix(source, "oci://"):
 		r.clone.Upstream.Type = porchapi.RepositoryTypeOCI
@@ -197,6 +201,22 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
 	}
 
 	fmt.Fprintf(cmd.OutOrStdout(), "%s created\n", pr.Name)
+	return nil
+}
+
+func (r *runner) packageAlreadyExists(packageName string) error {
+	// only the first package revision can be created from init or clone, so
+	// we need to check that the package doesn't already exist.
+	packageRevisionList := porchapi.PackageRevisionList{}
+	if err := r.client.List(r.ctx, &packageRevisionList, &client.ListOptions{}); err != nil {
+		return err
+	}
+	for _, pr := range packageRevisionList.Items {
+		if pr.Spec.RepositoryName == r.repository && pr.Spec.PackageName == packageName {
+			return fmt.Errorf("`clone` cannot create a new revision for package %q that already exists in repo %q; make subsequent revisions using `copy`",
+				packageName, r.repository)
+		}
+	}
 	return nil
 }
 
--- a/commands/alpha/rpkg/clone/command.go
+++ b/commands/alpha/rpkg/clone/command.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/GoogleContainerTools/kpt/commands/alpha/rpkg/util"
 	"github.com/GoogleContainerTools/kpt/internal/docs/generated/rpkgdocs"
 	"github.com/GoogleContainerTools/kpt/internal/errors"
 	"github.com/GoogleContainerTools/kpt/internal/util/parse"
@@ -114,8 +115,13 @@ func (r *runner) preRunE(cmd *cobra.Command, args []string) error {
 	source := args[0]
 	target := args[1]
 
-	if err := r.packageAlreadyExists(target); err != nil {
+	pkgExists, err := util.PackageAlreadyExists(r.ctx, r.client, r.repository, target, *r.cfg.Namespace)
+	if err != nil {
 		return err
+	}
+	if pkgExists {
+		return fmt.Errorf("`clone` cannot create a new revision for package %q that already exists in repo %q; make subsequent revisions using `copy`",
+			r.target, r.repository)
 	}
 
 	switch {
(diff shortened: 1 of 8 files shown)


natasha41575:
cc @mortent @justinsb

natasha41575:
cc @yuwenma

commands/alpha/rpkg/init/command.go
```diff
@@ -129,3 +128,19 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
 	fmt.Fprintf(cmd.OutOrStdout(), "%s created\n", pr.Name)
 	return nil
 }
+
+func (r *runner) packageAlreadyExists(packageName string) error {
+	// only the first package revision can be created from init or clone, so
+	// we need to check that the package doesn't already exist.
+	packageRevisionList := porchapi.PackageRevisionList{}
+	if err := r.client.List(r.ctx, &packageRevisionList, &client.ListOptions{}); err != nil {
```
yuwenma, line 136: @natasha41575 Does the pacakgeRevision always use `default` namespace? If so, I'm wondering if we shall specify the `namespace` to optimize the request.

yuwenma: Besides, since this is client side command, I think we can consider to set a timeout   (e.g. `.List().Timeout(Expiration)`) to avoid long waiting time. Not a blocker for this PR, just something interesting to think about.

commands/alpha/rpkg/clone/command.go
```diff
@@ -200,6 +204,22 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+	for _, pr := range packageRevisionList.Items {
+		if pr.Spec.RepositoryName == r.repository && pr.Spec.PackageName == packageName {
+			return fmt.Errorf("`clone` cannot create a new revision for package %q that already exists in repo %q; make subsequent revisions using `copy`",
+				packageName, r.repository)
+		}
+	}
+	return nil
+}
```
ChristopherFry, line 221: Should this be a client or server side check? The advantage of making this a server side check is that this will be enforced for all clients, including if someone is applying the package revision resource directly with kubectl.

natasha41575: I wanted to make this a server-side check (I explain this in my PR description). I struggled with how does porch make the determination that the client is trying to clone/init a new package revision? All package revisions (not necessarily just the first ones) will start with a clone/init task, so the existence of a clone/init task doesn't really tell it anything. 

Maybe porch could try to decide that if the task list has 0 or 1 items in it, that means that the client is trying to create the first package revision and we should check for existing package revisions. But with https://github.com/GoogleContainerTools/kpt/pull/3315 on the way, I'm not sure that will always make sense. With that PR, a copy of the initial package revision would look exactly the same as the initial package revision (except for the revision). There's also nothing stopping the initial package revision from having a long list of tasks on it, so the length of the task list doesn't necessarily tell us the intent of the client.

Perhaps there's something I'm missing here, in which case I'm open to suggestions, but the restrictions we want to impose in this PR seems more imperative by nature, which may be difficult to enforce in a more declarative-style server. We can try to make guesses based on the task list about the client's intent, but I don't know if that would be super reliable.

ChristopherFry: Sorry, completely missed the explanation on the pull request's description. I took a quick look myself and didn't see anything obvious either how this can be implemented on the server side, at least without making any assumptions which may not always hold true.

commands/alpha/rpkg/init/command.go
natasha41575: Thanks for the catch about namespace! packageRevision is namespaced and can be in namespaces other than `default`. I forgot to pass in the namespace, I updated it in the most recent commit. 

Re: setting a timeout, that's an interesting idea. There are a few other client-side commands (like `update`) that also list package revisions so we should set the same timeout. I think we can consider adding that in a separate PR.

commands/alpha/rpkg/clone/command.go
mortent: So if it is hard to determine the intention of the user here, could we try to make sure the end state of the request meets our invariants? What we want to make sure is that every packagerevision belonging to a package must:
* have the same origin, although not necessarily the same version. For example, with the replay functionality, it is possible for `v1` of a packagerevision to be based on `v1` from upstream, and for `v2` of the same packagerevision to be based on `v2` from upstream. But all revisions should be based on the same package, which admittedly is a bit easier to reason about with porch packages than with packages directly from git/oci.

But packagerevisions do not have to be a direct derivative of the latest packagerevision, for example it should be possible for `v3` of a package to be based on `v1` rather than `v2`. This would be a typical situation for rollback.

So it seems like what we want to make sure when a packagerevision is created/updated, is that it has the same origin as any existing packagerevisions. If we had the Package resource available, we could just check that one, but otherwise we could check the other packagerevisions. We should think about how we can do this without creating a race when creating a new packagerevision, but I think we have a similar problem when choosing a unique workspaceName or set a version.

Do we think something like this would be possible?

commands/alpha/rpkg/clone/command.go
```diff
@@ -200,6 +204,24 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func (r *runner) packageAlreadyExists(packageName string) error {
```
justinsb, line 207: Nit: This looks to be copied-and-pasted code, other than the error message?  Maybe put in a shared package, and return `(bool, error)`, where error => an error in fetching, and true/false => package found/not-found.

commands/alpha/rpkg/clone/command.go
```diff
@@ -200,6 +204,24 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
...
+	// only the first package revision can be created from init or clone, so
+	// we need to check that the package doesn't already exist.
+	packageRevisionList := porchapi.PackageRevisionList{}
+	if err := r.client.List(r.ctx, &packageRevisionList, &client.ListOptions{
+		Namespace: *r.cfg.Namespace,
+	}); err != nil {
+		return err
+	}
+	for _, pr := range packageRevisionList.Items {
+		if pr.Spec.RepositoryName == r.repository && pr.Spec.PackageName == packageName {
```
justinsb, line 217: I think we support fieldSelectors today, but if/when we get to CRDs we likely won't, so ... fine for now!

justinsb (approved):

justinsb:
This seems fine, though possibly moot once we have workspaces (which I think you're working on in parallel).

If it helps you before workspaces, LGTM!

commands/alpha/rpkg/clone/command.go
natasha41575: I added the check for the invariant to `engine.go` in porch side. PTAL when you can. Thanks for the suggestion @mortent!

commands/alpha/rpkg/clone/command.go
natasha41575: done! Thanks

mortent (approved):

PR #3609, opened 2022-10-11
Include files in subdirs for pull command

mortent:
This fixes the first issue in https://github.com/GoogleContainerTools/kpt/issues/3547, where files in subdirectories are not included in the output `ResourceList` when the content is printed to stdout.

@henderiw


yuwenma (approved):
Looks good!

PR #3614, opened 2022-10-12
Package conditions

mortent:
This adds support for readiness gates and conditions to kpt packages through new fields in the Kptfile, and exposes those fields in Porch through the `PackageRevision` resource.

The two most important aspects about this PR is:
* Is the fields we add to the Kptfile the correct ones? The patterns mirrors the `podReadinessGates` pretty closely, so it aligns well with prior art in Kubernetes.
* Should we expose fields in the Kptfile through the Porch API? We already expose the `UpstreamLock` read-only through the status object, but this allows for updates too by creating patches for the changes to the Kptfile. So in the git history for the Kptfile, it shows up identical to what it looks like if someone updates the Kptfile through `PackageRevisionResources`. Allowing updates through the Porch API directly simplifies updates with controllers.

One aspect of the current implementation that is a bit surprising, is that kpt updates both the conditions and the readiness gates as non-associative lists. This doesn't seem correct, but unfortunately the kyaml library can't identify the merge key and therefore falls back to treating it as a non-associative list. We could handle this either by adding custom merge logic for these fields, or provide a schema that provide this information to kyaml.


porch/api/porch/types_packagerevisions.go
```diff
@@ -302,3 +310,21 @@ type GitLock struct {
 	// This is set by kpt for bookkeeping purposes.
 	Commit string `yaml:"commit,omitempty" json:"commit,omitempty"`
 }
+
+type Condition struct {
+	Type string `yaml:"type" json:"type"`
+
+	Status ConditionStatus `yaml:"type" json:"status"`
+
+	Reason string `yaml:"type,omitempty" json:"reason,omitempty"`
```
justinsb, line 319: Nit: yaml:"type" -> reason

We also should just remove the yaml annotations - they aren't needed by the yaml libraries that apimachinery supports - but we can do that separately!

porch/pkg/engine/engine.go
```diff
@@ -328,14 +330,28 @@ func (cad *cadEngine) UpdatePackageRevision(ctx context.Context, repositoryObj *
...
 
-	// Re-render if we are making changes.
-	mutations = cad.conditionalAddRender(mutations)
-
 	draft, err := repo.UpdatePackageRevision(ctx, oldPackage)
 	if err != nil {
 		return nil, err
 	}
 
+	// If any of the fields in the API that are projections from the Kptfile
```
justinsb, line 338: It's fine to try this, but I suspect we won't be able to maintain this in the Kptfile going forwards.

porch/api/porch/types_packagerevisions.go
```diff
@@ -302,3 +310,21 @@ type GitLock struct {
...
+
+	Message string `yaml:"type,omitempty" json:"message,omitempty"`
```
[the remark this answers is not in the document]
justinsb: Filed https://github.com/GoogleContainerTools/kpt/issues/3618

porch/pkg/git/git.go
```diff
@@ -962,6 +962,35 @@ func (r *gitRepository) loadTasks(ctx context.Context, startCommit *object.Commi
 	return tasks, nil
 }
 
+func (r *gitRepository) getResources(hash plumbing.Hash) (map[string]string, error) {
+	resources := map[string]string{}
+
+	tree, err := r.repo.TreeObject(hash)
+	if err == nil {
```
justinsb, line 969: Nit: We should warn if the hash isn't found, and we should also check the error

porch/pkg/oci/oci.go
```diff
@@ -379,9 +381,11 @@ func (p *ociPackageRevision) Key() repository.PackageRevisionKey {
 	}
 }
 
-func (p *ociPackageRevision) GetPackageRevision() *v1alpha1.PackageRevision {
+func (p *ociPackageRevision) GetPackageRevision(ctx context.Context) *v1alpha1.PackageRevision {
 	key := p.Key()
 
+	kf, _ := p.GetKptfile(ctx)
```
justinsb, line 387: Nit: we shouldn't drop errors without checking them

justinsb (approved):

justinsb:
A few nits, but LGTM

porch/api/porch/types_packagerevisions.go
mortent: Fixed the incorrect names, but we can clean up the tags in a separate PR.

porch/pkg/oci/oci.go
mortent: Yeah, I meant to clean this up but forgot about it. Fixed now.

porch/pkg/git/git.go
mortent: Any reason to treat hash not found different from other errors, other than perhaps trying to get a better message to the user? It seems like almost no matter what goes wrong here, an error would be a better than returning an empty map.

porch/pkg/engine/engine.go
mortent: You mean that projecting fields from the Kptfile into the API here? If so, why? I had some concerns about it, but it actually seems to work a bit better than I feared.

PR #3616, opened 2022-10-13
licensecan: Fix go.mod so we can `go run`

justinsb:
Also mark the --binary flag as required (to avoid a cryptic error
message).


natasha41575 (approved):

yuwenma (approved):
Looks good! Thanks for the fix

PR #3622, opened 2022-10-14
fix non-US-ascii characters

wmuizelaar:
We ran into an issue where applying the resourcegroup.kpt.dev CRD through an old-fashioned system towards our cluster, that the contents of the description weren't fully in US-ASCII format.

This commit fixes that.


justinsb:
Thanks for fixing - lgtm.

I'm also going to open a bug because we probably should at least be using go:embed for this, and probably should think about how we keep this CRD up to date.

PR #3625, opened 2022-10-17
Support for checking sync status for RootSyncSet

mortent:
First step towards supporting progressive rollouts of RootSyncs.


porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
```diff
@@ -41,9 +41,17 @@ import (
 )
 
 var (
-	rootSyncNamespace  = "config-management-system"
-	rootSyncApiVersion = "configsync.gke.io/v1beta1"
-	rootSyncKind       = "RootSync"
+	rootSyncNamespace = "config-management-system"
+	rootSyncGVK       = schema.GroupVersionKind{
```
justinsb, line 45: Thanks for cleaning this up!

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
```diff
@@ -108,59 +116,215 @@ func (r *RootSyncSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 		client, err := r.GetClient(ctx, clusterRef, rootsyncset.Namespace)
 		if err != nil {
-			patchErrs = append(patchErrs, err)
-			continue
-		}
-		rootSyncRes, newRootSync, err := BuildObjectsToApply(&rootsyncset)
-		if err != nil {
-			patchErrs = append(patchErrs, err)
+			result.clientError = err
 			continue
```
justinsb, line 127: I think we skip `results[clusterRefName] = result` in this case?  Maybe pre-assign (though we'd have to make it a pointer)?

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
```diff
@@ -108,59 +116,215 @@ func (r *RootSyncSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	if errs := results.Errors(); len(errs) > 0 {
+		return ctrl.Result{}, errs[0]
 	}
 	return ctrl.Result{}, nil
 }
 
-// BuildObjectsToApply config root sync
-func BuildObjectsToApply(rootsyncset *v1alpha1.RootSyncSet) (schema.GroupVersionResource, *unstructured.Unstructured, error) {
-	gv, err := schema.ParseGroupVersion(rootSyncApiVersion)
+type reconcileResult map[string]clusterRefResult
```
justinsb, line 162: We could also have this implement `error`

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
```diff
@@ -108,59 +116,215 @@ func (r *RootSyncSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	if err != nil {
+		return err
+	}
+	data, err := json.Marshal(newRootSync)
+	if err != nil {
+		return fmt.Errorf("failed to encode root sync to JSON: %w", err)
+	}
+	rs, err := client.Resource(rootSyncGVR).Namespace(rootSyncNamespace).Patch(ctx, name, types.ApplyPatchType, data, metav1.PatchOptions{FieldManager: name})
+	if err != nil {
+		return fmt.Errorf("failed to patch RootSync %w", err)
```
justinsb, line 198: Supernit: missing `:`

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
```diff
@@ -108,59 +116,215 @@ func (r *RootSyncSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	klog.Infof("Create/Update resource %s as %v", name, rs)
+	return nil
+}
+
+// checkSyncStatus fetches the RootSync using the provided client and computes the sync status. The rules
+// for computing status here mirrors the one used in the status command in the nomos cli.
+func (r *RootSyncSetReconciler) checkSyncStatus(ctx context.Context, client dynamic.Interface, rssName string) (string, error) {
+	rs, err := client.Resource(rootSyncGVR).Namespace(rootSyncNamespace).Get(ctx, rssName, metav1.GetOptions{})
+	if err != nil {
+		return "", fmt.Errorf("failed to get RootSync %w", err)
```
justinsb, line 209: `:` (I guess the reason to care is because we'll keep copying and pasting it!)

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
```diff
@@ -108,59 +116,215 @@ func (r *RootSyncSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	s, err := extractStringField(cond, "status")
 	if err != nil {
-		return schema.GroupVersionResource{}, nil, fmt.Errorf("failed to parse group version when building object: %w", err)
+		return "", false, err
 	}
-	rootSyncRes := schema.GroupVersionResource{Group: gv.Group, Version: gv.Version, Resource: "rootsyncs"}
+	return s, true, nil
+}
+
+func getCondition(conditions []interface{}, condType string) (map[string]interface{}, bool, error) {
```
justinsb, line 275: Do we have the type (or could we embed at least the bit we care about?)  If so, it might then be easier to use `runtime.DefaultUnstructuredConverter.FromUnstructured`

justinsb (commented):
Looking good - some drive-by comments :-)

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
mortent: Good catch. Fixed.

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
mortent: Yeah, added that.

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
mortent: Fixed

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
mortent: Yeah, now that Config Sync is OSS we can use the types directly here.

mortent:
> Looking good - some drive-by comments :-)

Thanks, still a work in progress.

mortent:
@justinsb Took another pass at this and added support for watches.

porch/controllers/rootsyncsets/api/v1alpha1/rootsyncset_types.go
```diff
@@ -65,6 +65,16 @@ type SecretReference struct {
...
+
+	ClusterRefStatuses []ClusterRefStatus `json:"clusterRefStatuses,omitempty"`
+}
+
+type ClusterRefStatus struct {
+	ApiVersion string `json:"apiVersion,omitempty"`
+	Kind       string `json:"kind,omitempty"`
+	Name       string `json:"name,omitempty"`
+	Namespace  string `json:"namespace,omitempty"`
+	SyncStatus string `json:"syncStatus,omitempty"`
```
justinsb, line 77: TBD what status we want to expose, but we have to start somewhere :-)

Also not sure if we should give the target clusters an `id` and key off that here (rather than forcing harder matching), but again ... TBD

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
```diff
@@ -108,59 +148,246 @@ func (r *RootSyncSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+		liens:      make(map[types.NamespacedName]struct{}),
+	}
+	go w.watch()
+	r.watchers[clusterRef] = w
+}
+
+// pruneWatches removes the current RootSyncSet from the liens map of all watchers
+// that it no longer needs. If any of the watchers are no longer used by any RootSyncSets,
+// they are shut down.
+func (r *RootSyncSetReconciler) pruneWatches(rssnn types.NamespacedName, clusterRefs []*v1alpha1.ClusterRef) {
```
justinsb, line 344: I do wonder if this could be combined with `setupWatches`, but we can evolve

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
```diff
@@ -176,25 +403,46 @@ func (r *RootSyncSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.RootSyncSet{}).
+		Watches(
+			&source.Channel{Source: r.channel},
+			handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
+				var rssName string
+				var rssNamespace string
+				if o.GetLabels() != nil {
+					rssName = o.GetLabels()[rootSyncSetNameLabel]
```
justinsb, line 412: Nit: you don't have to check if it's nil, you can read a nil map (you just can't write to it))

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/watcher.go
```diff
@@ -0,0 +1,145 @@
...
+	reconnect := newBackoffTimer(minReconnectDelay, maxReconnectDelay)
+	defer reconnect.Stop()
+
+loop:
+	for {
+		select {
+		case <-reconnect.channel():
+			var err error
+			klog.Infof("Starting watch for %s... ", clusterRefName)
+			watcher, err = w.client.Resource(rootSyncGVR).Watch(w.ctx, v1.ListOptions{})
```
justinsb, line 71: Nit: we should add resume-from-last-rv, but .. for the future

justinsb (approved):

justinsb:
Looks great - some ideas for future evolution, but nothing blocking

porch/controllers/rootsyncsets/api/v1alpha1/rootsyncset_types.go
mortent: Yeah, I think at some point we probably want to expose both the synchronization status and reconciliation status from Config Sync, but doing just the sync status at first is easier.

I was considering some options for what to include here to allow us to mach statuses to the clusterRefs from the spec. I ended up duplicating the four identifiers from the clusterRef, because I think this will allow us to remove RootSync from clusters when they are removed from the clusterRef list in the spec. Currently RootSyncSet doesn't handle this, since once the clusterRef is removed, we no longer know which cluster to talk to..

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
mortent: Yeah, it might. I added it separately as pruning watches requires looking at all clusterRefs together, while setting them up fits well within the for loop for getting the client and doing the patching.

PR #3633, opened 2022-10-21
Resolve functions using function objects registered in apiserver

justinsb:
The intent is that many users will run a function mirror or similar.

### DIFF
--- a/porch/pkg/apiserver/apiserver.go
+++ b/porch/pkg/apiserver/apiserver.go
@@ -20,11 +20,11 @@ import (
 
 	"github.com/GoogleContainerTools/kpt/internal/fnruntime"
 	"github.com/GoogleContainerTools/kpt/porch/api/porch/install"
+	porchapi "github.com/GoogleContainerTools/kpt/porch/api/porch/v1alpha1"
 	configapi "github.com/GoogleContainerTools/kpt/porch/api/porchconfig/v1alpha1"
 	internalapi "github.com/GoogleContainerTools/kpt/porch/internal/api/porchinternal/v1alpha1"
 	"github.com/GoogleContainerTools/kpt/porch/pkg/cache"
 	"github.com/GoogleContainerTools/kpt/porch/pkg/engine"
-	"github.com/GoogleContainerTools/kpt/porch/pkg/kpt"
 	"github.com/GoogleContainerTools/kpt/porch/pkg/meta"
 	"github.com/GoogleContainerTools/kpt/porch/pkg/registry/porch"
 	"google.golang.org/api/option"
@@ -73,6 +73,7 @@ type ExtraConfig struct {
 	CoreAPIKubeconfigPath string
 	CacheDirectory        string
 	FunctionRunnerAddress string
+	DefaultImagePrefix    string
 }
 
 // Config defines the config for the apiserver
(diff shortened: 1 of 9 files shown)


porch/pkg/apiserver/apiserver.go
```diff
@@ -147,6 +148,13 @@ func (c completedConfig) getCoreClient() (client.WithWatch, error) {
 	if err := configapi.AddToScheme(scheme); err != nil {
 		return nil, fmt.Errorf("error building scheme: %w", err)
 	}
+
+	// We query the apiserver, even if we could query directly; this will then work with CRDs etc.
+	// TODO: We need to think about priority-and-fairness with loopback queries
```
droot, line 153: This comments seems out of place. Registering porchapi to current scheme doesn't involve querying apiserver. 
Perhaps, misplaced ?

porch/pkg/apiserver/resolvetoimage.go
```diff
@@ -20,25 +20,53 @@ import (
...
 	if !strings.Contains(image, "/") {
-		return fmt.Sprintf("gcr.io/kpt-fn/%s", image), nil
+		var function v1alpha1.Function
+		// TODO: Use fieldSelectors and better lookup
+		name := "functions:" + image + ":latest"
+		key := types.NamespacedName{
+			Namespace: r.namespace,
+			Name:      name,
+		}
+		if err := r.client.Get(ctx, key, &function); err != nil {
```
droot, line 61: Some clarifications:
How's the functions registry API work today ?
I am assuming by default, functions won't be there in any porch installation, so this resolving should be opt-in. Simplify because it is now calling network call for every function with out `/`. WDYT ?

porch/pkg/cmd/server/start.go
```diff
@@ -225,5 +227,6 @@ func (o *PorchServerOptions) AddFlags(fs *pflag.FlagSet) {
 	}
 
 	fs.StringVar(&o.FunctionRunnerAddress, "function-runner", "", "Address of the function runner gRPC service.")
+	fs.StringVar(&o.DefaultImagePrefix, "default-image-prefix", "gcr.io/kpt-fn/", "Default prefix for unqualified function names")
```
droot, line 230: should we make the API based function resolution `opt-in` ?

droot (commented):
Change is looking good. Some minor questions.

porch/pkg/apiserver/apiserver.go
justinsb: This is where we enable it, but fair point - I moved to the callsite!

porch/pkg/apiserver/resolvetoimage.go
justinsb: How it works today: You can register an OCI Repository with `spec.content: Function`, and then any discovered images are registered as Functions.

Performance: Currently I don't _think_ the client.Client is caching (but it's always hard to know with controller-runtime!), but we could make it caching if we're concerned about the performance.

Opt-in: I think this should be the default behaviour and the default way we use functions.  It's somewhat akin to the Function Catalog idea from kustomize.  That said, I expect it to evolve over time (not least because we want to do something on the client side as well).  If we have a flag and we set it for users then everyone will be running with anyway :-)  I think we have a degree of opt-in because the user has to register a Repository of spec.content=Function to activate the behaviour (otherwise we fall through).

I do think I should write a doc about how this works today, we can keep it updated even as it changes...

droot: I think `function registry` is not widely discussed or shown in any of the demos, so I always considered as `aspirational` :) but this discussion suggests it is not the case.

Re: `performance`, I think `functions` are always beaten for slow performance, so just being careful with adding new latencies :)

Yes, I think documentation will certainly help.

For now, let's go with this and see if it causes any trouble in the field.

droot (approved):

PR #3638, opened 2022-10-27
porch: add json struct tags to render results

ChristopherFry:
This pull request adds json struct tags to the render results updating the fields in the render results to be in returned in camel case following the field naming conventions in the [API Conventions Document](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions).


mortent (approved):
We can clean this up separately, but we shouldn't have the yaml tags here. Ref https://github.com/GoogleContainerTools/kpt/issues/3618

ChristopherFry:
> We can clean this up separately, but we shouldn't have the yaml tags here. Ref https://github.com/GoogleContainerTools/kpt/issues/3618

Thanks! Will address this separately.

PR #3642, opened 2022-10-27
Upgrade cli-utils to v0.34.0

rquitales:
This PR updates `cli-utils` to v0.34.0. `cli-utils` now uses Go v1.18 and v1.25 of Kubernetes resources.

Changes in this PR:
* Update cli-utils to v0.34.0
* Add edge case to `create-licenses.sh` to handle `github.com/emicklei/go-restful/v3` dependency
* Added new `Destroy` methods to satisfy upstream Interface changes
* Updated function signature to satisfy upstream Interface changes


scripts/create-licenses.sh
```diff
@@ -167,6 +167,9 @@ find vendor -type f -name LICENSE.md -execdir mv LICENSE.md LICENSE ';'
 V2_LICENSE_DIR="vendor/github.com/cpuguy83/go-md2man"
 mv ${V2_LICENSE_DIR}/v2/LICENSE ${V2_LICENSE_DIR}
 
+GO_RESTFUL_LICENSE_DIR="vendor/github.com/emicklei/go-restful"
```
karlkfi, line 170: Add a comment here to explain why this is necessary

porch/pkg/registry/porch/functions.go
```diff
@@ -48,6 +48,8 @@ func (f *functions) New() runtime.Object {
 	return &v1alpha1.Function{}
 }
 
+func (f *functions) Destroy() {}
```
karlkfi, line 51: Why is this Destroy method now required with a cli-utils update?

Is it ok for it to not be implimented?

/cc @justinsb

scripts/create-licenses.sh
rquitales: L165 above is a general comment for handling the cleanup/edge cases since L167-168 also does a similar handling for the `go-md2man/v2` package.

Will update the comment to be more descriptive!

porch/pkg/registry/porch/functions.go
karlkfi: Can you add comments to the empty methods to explain that? I'm surprised the linter isn't complaining about that, since they're public methods.

natasha41575:
Closing in favor of https://github.com/GoogleContainerTools/kpt/pull/3746.

PR #3644, opened 2022-10-28
Watch for PackageRevisions

mortent:
This moves the watch from PackageRevisionResources to PackageRevisions. It also makes sure to send watch notifications on both updates coming from polling the repository (git/oci) and create/update/delete through the API.


justinsb (approved):

droot (approved):
Looks good.

PR #3645, opened 2022-10-31
cli: rpkg push surface renderstatus

droot:
This PR adds support for displaying render status as part of of `kpt alpha rpkg push` command.

Here is an example output in case of render failure:

```sh
$ ~/work/bin/kpt alpha rpkg push application-blueprints-33f255 /tmp/new-pkg -n default
Package is updated, but failed to render the package.
Error: fn.render: pkg /:
        pkg.render:
        pipeline.run: error: function failure
[RUNNING] "gcr.io/kpt-fn/set-namespace:v0.4.1"
[PASS] "gcr.io/kpt-fn/set-namespace:v0.4.1"
  Results:
    [info]: all namespaces are already "example". no value changed
[RUNNING] "gcr.io/kpt-fn/starlark:v0.4.3"
[FAIL] "gcr.io/kpt-fn/starlark:v0.4.3"
  Results:
    [error]: fail: Validation always fails

```

Here is an example output in case of successful run:

```sh
$ ~/work/bin/kpt alpha rpkg push application-blueprints-33f255a /tmp/new-pkg -n default
[RUNNING] "gcr.io/kpt-fn/set-namespace:v0.4.1"
[PASS] "gcr.io/kpt-fn/set-namespace:v0.4.1"
  Results:
    [info]: all namespaces are already "example". no value changed
```

xref: https://github.com/GoogleContainerTools/kpt/issues/3635


commands/alpha/rpkg/push/command.go
```diff
@@ -129,12 +131,87 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
...
+			}
+			printFnResult(r.ctx, result, printOpt)
+		}
+	}
 	return nil
 }
 
+// printFnResult prints given function result in a user friendly
+// format on kpt CLI.
+func printFnResult(ctx context.Context, fnResult *porchapi.Result, opt *printer.Options) {
```
mortent, line 160: Are we printing this in the same way through the `rpkg push` command as we do with the `kpt fn` commands?

commands/alpha/rpkg/push/command.go
```diff
@@ -129,12 +131,87 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
...
+	pr := printer.FromContextOrDie(ctx)
```
mortent, line 161: Would it be easier to use the printer from the runner here?

mortent (approved):

commands/alpha/rpkg/push/command.go
droot: Done.

commands/alpha/rpkg/push/command.go
droot: There is slight difference in printing the package name, but rest is same.

PR #3653, opened 2022-11-08
Add RootSyncDeployment controller

mortent:
This adds a new CRD and controller that builds upon the RootSyncSet type to support progressive rollouts of PackageRevisions from Porch. It uses label selectors to find remote clusters represented as Config Connector resources in the local cluster and then creates RootSyncSet resources representing each. Whenever the referenced PackageRevision is changed, the update will be rolled out progressively. Sync status is currently the only signal, but we can expand on that to include reconciliation status and possibly other external systems.

Demo: https://drive.google.com/file/d/1zlqlSLNOR25v8Kb9qOAXd-ymRHkxOCRh/view?usp=share_link


porch/controllers/rootsyncdeployments/pkg/controllers/rootsyncdeployment/controller.go
```diff
@@ -0,0 +1,539 @@
...
+	if equality.Semantic.DeepEqual(rds.Status.ClusterRefStatuses, clusterRefStatuses) {
+		klog.Infof("Status has not changed, update not needed.")
+		return nil
+	}
+	rds.Status.ClusterRefStatuses = clusterRefStatuses
+	return r.Status().Update(ctx, rds)
+}
+
+type result struct {
+	apiVersion string
```
natasha41575, line 183: Why make a new type here instead of just using the `ClusterRefStatus` type directly?

natasha41575 (approved):

porch/controllers/rootsyncdeployments/pkg/controllers/rootsyncdeployment/controller.go
mortent: Hmm that is an excellent question. I've updated the PR to just use `ClusterRefStatus` directly.

PR #3655, opened 2022-11-11
Finalizers and OwnerRefs for PackageRevision

mortent:
This adds support for Finalizers and OwnerReferences for the PackageRevision type. It does this by reusing these fields from the internal PackageRev type to the PackageRevision type from the aggregated APIServer.

There are a few caveats to finalizers for PackageRevisions. Since these can be removed out-of-band from the APIServer, either by deletion of the Repository from Porch or removal of the PackageRevision directly from git/oci. In these cases, there isn't any way for us to honor the finalizers, so the resource will be deleted directly.


natasha41575 (approved):
lgtm!

PR #3664, opened 2022-11-17
Add stubs for compiling without wasmtime / without cgo

justinsb:
This allows us to compile kpt without cgo support; wasm support is in
alpha and we have to opt in to wasm support and the wasmtime runner
anyway.

We can likely add an exec-based runner here instead in future (though
we already have the nodejs one).

### DIFF
--- a/internal/fnruntime/wasmtime.go
+++ b/internal/fnruntime/wasmtime.go
@@ -1,3 +1,6 @@
+//go:build cgo
+// +build cgo
+
 // Copyright 2022 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
--- a/internal/fnruntime/wasmtime_unsupported.go
+++ b/internal/fnruntime/wasmtime_unsupported.go
@@ -1,0 +1,37 @@
+//go:build !cgo
+// +build !cgo
+
+// Copyright 2022 Google LLC
+//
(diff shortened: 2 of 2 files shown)


mortent:
Change looks good, but just a few questions:
* This would mean that our regular published binaries will no longer include wasm support. Is the idea that users who want wasm support should build it from source?
* Should we also update the goreleaser spec for this to avoid enabling cgo?

justinsb:
> This would mean that our regular published binaries will no longer include wasm support. Is the idea that users who want wasm support should build it from source?

So today if you compile with GCO_ENABLED=0, kpt fails to build.  This is ensuring it will build, and yes, built-in wasm support will not be available.  However, the nodejs support is still available.

I see this as a first step only; we fix a scenario in which the compile fails.  I would like to continue to support wasm functionality even with CGO_ENABLED=0, one way is to build a separate binary that wraps wasmtime.  But that _would_ have to be built with CGO_ENABLED, and we already have nodejs there, so I don't know whether that is a win.

> Should we also update the goreleaser spec for this to avoid enabling cgo?

We can do, particular if cgo is causing other problems, but we don't have to.  I think given that wasm is in alpha, and given that the nodejs support is available also, it's a reasonable thing to do, but I don't think this PR forces things either way (other than giving us the option to run with CGO_ENABLED=0!)

mortent (approved):

PR #3665, opened 2022-11-17
licensetool: embed license DB, update for more modules

justinsb:
We now embed some licenses using `go:embed`, so we don't need to set up a `modules` DB in every project.  A local modules folder is still checked first, and TODO files are still written locally to a local modules folder.

Also add a bunch more module license information from different places I've been running the tool.

### DIFF
--- a/tools/licensescan/main.go
+++ b/tools/licensescan/main.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"context"
 	"debug/buildinfo"
+	"embed"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -34,6 +35,9 @@ import (
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/yaml"
 )
+
+//go:embed modules/*
+var modulesFS embed.FS
 
 func main() {
 	rootCmd := buildRootCommand()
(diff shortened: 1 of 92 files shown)


tools/licensescan/main.go
```diff
@@ -35,6 +36,9 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+//go:embed modules/*
+var modulesFS embed.FS
```
droot, line 40: interesting: eventually we can have it in its own Go package that will make this DB portable across all the tools :)

tools/licensescan/main.go
```diff
@@ -125,7 +129,14 @@ func RunLicenseScan(ctx context.Context, opts RunLicenseScanOptions) error {
 		modules = append(modules, module)
 
 		p := filepath.Join("modules", module.Name, module.Version+".yaml")
-		b, err := ioutil.ReadFile(p)
+		b, err := os.ReadFile(p)
+		if err != nil && os.IsNotExist(err) {
+			b2, err2 := modulesFS.ReadFile(p)
+			if err2 == nil {
+				b = b2
```
droot, line 136: so this will ensure license info from the moduleFS gets written back to local FS.

tools/licensescan/main.go
```diff
@@ -125,7 +129,14 @@ func RunLicenseScan(ctx context.Context, opts RunLicenseScanOptions) error {
...
+				err = err2
+			}
```
droot, line 138: err2 != nil is being ignored and I believe that's intentional because err is what we care about in that case.

droot (approved):

tools/licensescan/main.go
justinsb: I hope so :-)

tools/licensescan/main.go
justinsb: Yes, ish.  The workflow I use is to run it against the binary.  Then I search for `license: TODO` go through and fill in any missing licenses that are generated.  The link to the go package DB makes this easy, if a little tedious.  Then I commit the files, and then yes the `go:embed` ensures that (once merged) they are automatically embedded in as part of go build.

tools/licensescan/main.go
justinsb: Yes, it is deliberate (I think), but I think I'll take a second look and maybe log a `V(4)` message to make the code at least more self-explanatory.

PR #3666, opened 2022-11-19
Provide version to RESTMapper in WorkloadIdentityBinding controller

mortent:
This fixes an issue with the WorkloadIdentityBinding controller. The new implementation of the RESTMapper interface that was added in https://github.com/GoogleContainerTools/kpt/pull/3567 differs from the regular implementation in that it requires that the version is provided when calling the `RESTMapping` function. This updates the controller to provide the version when use that function.

We should consider fixing the behavior in the RESTMapper implementation, as this differs from the regular implementation.


justinsb:
/approve
/lgtm

Agree we should fix it, though I think that it's very rare that we don't care about the version in a controller (?)  If that's true, making this an explicit error is not a bad behaviour (?)

justinsb (approved):

PR #3675, opened 2022-11-28
Create CRD for Function objects

justinsb:
- Create Function CRD in configapi
- Generated output for Function CRD
- Implement controller for function discovery
- functiondiscovery controller: add test for discovery
- Simple mock OCI server

Currently functions are implemented by the aggregated apiserver.  Copy the definition to a CRD, move the population logic to a controller (sharing most of the code), create a test.

The CRD is in a new group, so we can transition smoothly with the aggregated apiserver and the CRDs running in parallel during the transition.


porch/api/porchconfig/v1alpha1/types_functions.go
```diff
@@ -0,0 +1,107 @@
...
+	// InputTypes specifies to which input KRM types the function applies. Specified as Group Version Kind.
+	// For example:
+	//
+	//    inputTypes:
+	//    - kind: RoleBinding
+	//      # If version is unspecified, applies to all versions
+	//      apiVersion: rbac.authorization.k8s.io
+	//    - kind: ClusterRoleBinding
+	//      apiVersion: rbac.authorization.k8s.io/v1
+	// InputTypes []metav1.TypeMeta
```
johnbelamaric, line 84: Maybe something like this belongs in FunctionConfig type below?

justinsb: Ah - this file is the aggregated definition, just moved here: https://github.com/GoogleContainerTools/kpt/blob/921abb821a8663f5301d7e2f2d87846122dc7e30/porch/api/porch/v1alpha1/types_functions.go

I'm trying to not change it during the transition to CRDs (though I could)

johnbelamaric: Oh, ok. Better in a separate PR if we think we want to change that. I am assuming these as CRDs is intended to allow discovery of functions, so in that case, understanding the CRDs that are inputs to those seems critical.

porch/scripts/create-deployment-blueprint.sh
```diff
@@ -186,6 +186,8 @@ function customize-container-env {
 
 function main() {
```
mortent, line 187: We probably also want to include this controller in the make targets: https://github.com/GoogleContainerTools/kpt/blob/main/porch/Makefile#L52

porch/controllers/functiondiscovery/controller.go
```diff
@@ -0,0 +1,183 @@
...
+
+//go:generate go run sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0 rbac:roleName=porch-controllers-functions webhook paths="." output:rbac:artifacts:config=../../../config/rbac
+
+//+kubebuilder:rbac:groups=config.porch.kpt.dev,resources=repositories,verbs=get;list;watch
+//+kubebuilder:rbac:groups=config.porch.kpt.dev,resources=functions,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=config.porch.kpt.dev,resources=functions/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=config.porch.kpt.dev,resources=functions/finalizers,verbs=update
+
+// Reconcile implements the main kubernetes reconciliation loop.
+func (r *FunctionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
mortent, line 62: I assume we need some kind of regular reconciliation here to pick up new functions added to the oci repo? Alternatively, some other kind of watcher/poller of the oci repo.

mortent (commented):
So with this change, we should be able to remove the Function endpoints on the aggregated APIServer right? Users would just use the regular k8s API to list Function CRs that are mirrored from an OCI repo with the controller. Do we want to make that change in a separate PR?

justinsb:
> So with this change, we should be able to remove the Function endpoints on the aggregated APIServer right? Users would just use the regular k8s API to list Function CRs that are mirrored from an OCI repo with the controller. Do we want to make that change in a separate PR?

Exactly right:  I propose making that change in a separate PR; this is really trying out an incremental approach that _may_ be easier than a big rewrite.  (But it had its problems, for example it was blocked on test flakes on the old aggregated apiserver functions implementation, but I think that was a coincidence as it affected totally unrelated PRs also)

porch/controllers/functiondiscovery/controller.go
```diff
@@ -0,0 +1,183 @@
...
+package functiondiscovery
+
+import (
+	"context"
+	"flag"
+	"fmt"
+
+	kptoci "github.com/GoogleContainerTools/kpt/pkg/oci"
+	api "github.com/GoogleContainerTools/kpt/porch/api/porchconfig/v1alpha1"
+	"github.com/GoogleContainerTools/kpt/porch/controllers/remoterootsyncsets/pkg/applyset"
```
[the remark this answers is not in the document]
justinsb: Good call.  We have one in the declarative pattern lib, perhaps it is time to have kpt use the kdp applyset, and move the work there.  I'll look at this (but a separate PR, I suggest)

porch/scripts/create-deployment-blueprint.sh
justinsb: Done!

porch/controllers/functiondiscovery/controller.go
justinsb: Good point, added a simple re-poll every N minutes mechanism for now.

porch/api/porchconfig/v1alpha1/types_functions.go
```diff
@@ -0,0 +1,107 @@
...
+
+const (
+	FunctionTypeValidator FunctionType = "validator"
+	FunctionTypeMutator   FunctionType = "mutator"
+)
+
+// FunctionSpec defines the desired state of a Function
+type FunctionSpec struct {
+	// Image specifies the function image, such as 'gcr.io/kpt-fn/gatekeeper:v0.2'.
+	Image string `json:"image"`
```
droot, line 56: This is more of a question:
Since this will act as function discovery API, it need to surface all the versions, right ?

For use-cases of resolving the fully qualified domain or migrating to latest version, being able to discover all the versions is important.

porch/api/porchconfig/v1alpha1/types_functions.go
```diff
@@ -0,0 +1,107 @@
...
+// If unspecified, defaults to v1/ConfigMap. For example, function `set-namespace` accepts both `ConfigMap` and `SetNamespace`
+type FunctionConfig struct {
+	metav1.TypeMeta `json:",inline"`
+	// Experimental: requiredFields tells necessary fields and is aimed to help users write the FunctionConfig.
+	// Otherwise, users can get the required fields info from the function evaluation error message.
+	RequiredFields []string `json:"requiredFields,omitempty"`
+}
+
+// FunctionStatus defines the observed state of Function
+type FunctionStatus struct {
```
droot, line 106: This can *eventually* provide info such as `stable` version of `latest` etc.

porch/controllers/functiondiscovery/controller.go
```diff
@@ -0,0 +1,183 @@
...
+	patchOptions := metav1.PatchOptions{
+		FieldManager: "functions-" + subject.GetNamespace() + "-" + subject.GetName(),
+	}
+
+	// We force to overcome errors like: Apply failed with 1 conflict: conflict with "kubectl-client-side-apply" using apps/v1: .spec.template.spec.containers[name="porch-server"].image
+	// TODO: How to handle this better
+	force := true
+	patchOptions.Force = &force
+
+	applyset, err := applyset.New(applyset.Options{
```
droot, line 119: note to self: need to learn about applyset :)

porch/controllers/functiondiscovery/testdata/testfunctiondiscoverycontroller/oci/expected.yaml
```diff
@@ -0,0 +1,163 @@
...
+  repositoryRef:
+    name: kpt-functions
+status: {}
+
+---
+apiVersion: config.porch.kpt.dev/v1alpha1
+kind: Function
+metadata:
+  creationTimestamp: null
+  name: kpt-functions.set-labels.v0
```
droot, line 65: Ignore my previous question, I can see how versions are part of the name.

May be we will have to add label for `function-name` so that one can retrieve all the versions of a function using labelSelector ?

droot (approved):
Awesome. Thanks.

porch/Makefile
```diff
@@ -49,9 +49,9 @@ TEST_GIT_SERVER_IMAGE ?= test-git-server
 # Only enable a subset of reconcilers in porch controllers by default. Use the RECONCILERS
 # env variable to specify a specific list of reconcilers or use
 # RECONCILERS=* to enable all known reconcilers.
-ALL_RECONCILERS="rootsyncsets,remoterootsyncsets,workloadidentitybindings,rootsyncdeployments"
+ALL_RECONCILERS="rootsyncsets,remoterootsyncsets,workloadidentitybindings,rootsyncdeployments,functiondiscovery"
 ifndef RECONCILERS
-  ENABLED_RECONCILERS="rootsyncsets,remoterootsyncsets,workloadidentitybindings"
+  ENABLED_RECONCILERS="rootsyncsets,remoterootsyncsets,workloadidentitybindings,functiondiscovery"
```
mortent, line 54: Do we want to include this controller with porch releases before we remove the endpoints in the aggregated apiserver?

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
```diff
@@ -199,7 +201,9 @@ func (r *RootSyncSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, results
 	}
 
-	return ctrl.Result{}, nil
+	jitter := time.Duration(rand.Intn(60)) * time.Second
+	pollInterval := 15*time.Minute + jitter
+	return ctrl.Result{RequeueAfter: pollInterval}, nil
```
mortent, line 206: Did you mean to add this to the RootSyncSet controller?

porch/api/porchconfig/v1alpha1/types_functions.go
justinsb: Well I think we want to think more about this.  I quite like the "Catalog" approach from kustomize.  I think most end-users just want to run the "foo" function.  I think we might need functionality for the platform-team to (safely) update the "foo" function to a newer version.

In short, function discovery is pretty slow and not necessarily the most useful thing we could be doing.

But I think that is orthogonal to this PR, so I'm going to merge and we should discuss in next f2f meeting!

porch/api/porchconfig/v1alpha1/types_functions.go
justinsb: Agree.  Right now this is a direct copy of the aggregated apiserver type, so we can migrate without worrying about anything more than the group changing, I hope!

porch/controllers/functiondiscovery/controller.go
justinsb: Yeah, I like it ... but I would :-)

porch/controllers/functiondiscovery/testdata/testfunctiondiscoverycontroller/oci/expected.yaml
justinsb: I think we will yes, but I'd like to tackle that as part of thinking about what we want to expose here.  I can see us exposing a list of versions in the status as well, for example, rather than one object per version.

porch/Makefile
justinsb: I think we do yes, it shouldn't matter if we run both, and it should help people move away from the aggregated apiserver endpoints.

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
justinsb: Gah!  Nice catch :-)

justinsb:
@mortent you have a "changes request" on this - are you OK with it now?

mortent (approved):
Yeah, I added request changes due to the changes to the RootSyncSet controller. Looks good now.

PR #3682, opened 2022-12-01
licensescan: Add simple license crosscheck

justinsb:


tools/licensescan/licensedb_test.go
```diff
@@ -0,0 +1,64 @@
...
+				continue
+			}
+
+			if filepath.Dir(f2) != dir {
+				continue
+			}
+
+			if m1.License != m2.License {
+				switch f1 {
+				case "modules/github.com/klauspost/compress/v1.11.2.yaml":
```
droot, line 55: interesting...This block will become the source of truth for license changes for all modules. Another database of value :)

droot: May be we can present it as warning at the scanning step itself.

droot (approved):

PR #3684, opened 2022-12-02
create DownstreamPackage controller

natasha41575:
What this PR does:
- adds a `DownstreamPackage` controller as discussed in https://github.com/GoogleContainerTools/kpt/issues/3488
- implements creation and updating of deployment packages

What this PR does not do (will be coming in a followup PR):
- Implement pruning of deployment packages that are no longer needed
- Implement the proposed AdoptionPolicy field, which will allow the user to choose whether the controller adopts existing package revisions or ignore them


natasha41575:
cc @johnbelamaric

porch/controllers/downstreampackages/api/v1alpha1/downstreampackage_types.go
```diff
@@ -0,0 +1,87 @@
...
+	Upstream   *Upstream   `json:"upstream,omitempty"`
+	Downstream *Downstream `json:"downstream,omitempty"`
+
+	AdoptionPolicy AdoptionPolicy `json:"adoptionPolicy,omitempty"`
+	DeletionPolicy DeletionPolicy `json:"deletionPolicy,omitempty"`
+}
+
+type Upstream struct {
+	Repo     string `json:"repo,omitempty"`
+	Package  string `json:"package,omitempty"`
```
mortent, line 64: It seems like this essentially identifies a single PackageRevision resource. Should we use the name of the resource instead of `Package` and `Revision`?

porch/controllers/downstreampackages/pkg/controllers/downstreampackage/downstreampackage_controller.go
```diff
@@ -0,0 +1,366 @@
...
+
+	if err := r.Client.Get(ctx, req.NamespacedName, &dp); err != nil {
+		return nil, client.IgnoreNotFound(err)
+	}
+
+	var prList porchapi.PackageRevisionList
+	if err := r.Client.List(ctx, &prList, client.InNamespace(dp.Namespace)); err != nil {
+		return nil, err
+	}
+	r.packageRevs = prList
```
mortent, line 98: I suggest passing this list around instead of using a property on the reconciler. I think we only use a single reconcile routine, but we usually try to avoid having the controllers stateful.

porch/controllers/downstreampackages/pkg/controllers/downstreampackage/downstreampackage_controller.go
```diff
@@ -0,0 +1,366 @@
...
+			return &pr
+		} else {
+			latestPublished, latestVersion = compare(&pr, latestPublished, latestVersion)
+		}
+	}
+
+	return latestPublished
+}
+
+func compare(pr, latestPublished *porchapi.PackageRevision, latestVersion string) (*porchapi.PackageRevision, string) {
```
mortent, line 245: Do we need to pass in `latestPublished` and `latestVersion` here? It seems like `latestPublished` is always nil and `latestVersion` is always `v0`.

porch/controllers/downstreampackages/pkg/controllers/downstreampackage/downstreampackage_controller.go
```diff
@@ -0,0 +1,366 @@
...
+		latestPublished = pr.DeepCopy()
+	}
+	return latestPublished, latestVersion
+}
+
+// determine if the downstream PR needs to be updated
+func (r *DownstreamPackageReconciler) isUpToDate(dp *api.DownstreamPackage, downstream *porchapi.PackageRevision) bool {
+	upstreamLock := downstream.Status.UpstreamLock
+	lastIndex := strings.LastIndex(upstreamLock.Git.Ref, "/")
+	if strings.HasPrefix(upstreamLock.Git.Ref, "drafts") {
```
mortent, line 263: Can't we just check the `lifecycle` field on the PackageRevision for this?

porch/controllers/downstreampackages/pkg/controllers/downstreampackage/downstreampackage_controller.go
```diff
@@ -0,0 +1,366 @@
...
+	newPR.Spec.Revision = ""
+	newPR.Spec.WorkspaceName = newWorkspaceName(newPR.Spec.WorkspaceName)
+	newPR.Spec.Lifecycle = porchapi.PackageRevisionLifecycleDraft
+
+	if err := r.Client.Create(ctx, newPR); err != nil {
+		return nil, err
+	}
+
+	// get the newly created package revision back
+	var prList porchapi.PackageRevisionList
```
mortent, line 303: The `newPR` variable should be updated with the latest state from the APIServer as part of the call to `Create`. so I don't think this is needed.

porch/controllers/downstreampackages/api/v1alpha1/downstreampackage_types.go
johnbelamaric: Given that the resource name of package revisions is auto-generated, that would make this resource very hard to create and read for users, and would require the creation of the package revision prior to the creation of this resource. Generally that's probably going to be the case, but frankly the package revision name is more like a machine generated pointer than something useful. Of course, if this resource is being generated by a controller, this is not really an issue. Usually that is probably the case.

The expected general use of this would be to update the Revision number and trigger upgrades (or downgrades). So keeping these separate makes that easier. This does raise the issue of what happens if someone points this resource at a completely different upstream package, with respect to updates.

porch/controllers/downstreampackages/pkg/controllers/downstreampackage/downstreampackage_controller.go
natasha41575: We have the target upstream package revision (the one we want to update to), and the current downstream package revision. What we do not have is the upstream package that the current downstream package is _currently_ based on - which is what we need to check if the downstream package revision is up to date.

All we have is the reference to that current upstream package revision in the downstream package's `upstreamLock.Git.Ref` - so no, we don't have a PackageRevision.Lifecycle field to refer to here.

Fetching that current upstream package revision involves tedious loops and string parsing of `upstreamLock.Git.Ref` (and somewhere in that logic would include checking for the `drafts` prefix so we can parse the ref properly), so IMO just checking for the `drafts` prefix is the best thing to do here.

porch/controllers/downstreampackages/pkg/controllers/downstreampackage/downstreampackage_controller.go
natasha41575: done

porch/controllers/downstreampackages/pkg/controllers/downstreampackage/downstreampackage_controller.go
natasha41575: done

porch/controllers/downstreampackages/api/v1alpha1/downstreampackage_types.go
```diff
@@ -0,0 +1,87 @@
...
+
+type AdoptionPolicy string
+type DeletionPolicy string
+
+const (
+	AdoptionPolicyAdoptExisting AdoptionPolicy = "adoptExisting"
+	AdoptionPolicyAdoptNone     AdoptionPolicy = "adoptNone"
+
+	DeletionPolicyDelete = "delete"
+	DeletionPolicyDisown = "disown"
```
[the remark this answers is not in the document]
natasha41575: changed `disown` to `orphan`. Thanks!

porch/controllers/downstreampackages/api/v1alpha1/downstreampackage_types.go
natasha41575: I see @johnbelamaric's point that if someone is trying to manage a DownstreamPackage resource manually, looking up the package revision resource name for each update can be tedious and perhaps error prone, since it is an auto-generated hash. But the DownstreamPackage is generally going to be created by the UpstreamPolicy controller when it is ready, so this will be less of an issue.

Having the repo/package/revision separate means that the DownstreamPackage controller has put a little more effort into looking up the upstream package revision. If it instead takes the name of the packagerevision resource, the UpstreamPolicy controller will have to do the lookup and pass that to the DownstreamPackage. Either way, we will need to have that lookup somewhere; I think the question is just which controller should we put it in. Personally I don't think that it matters too much, but since the UpstreamPolicy controller is already going to be somewhat complex, maybe it makes sense to delegate the lookup to the DownstreamPackage, and keep the repo/package/revision separate. That said, I don't mind changing it if you feel differently. 

> This does raise the issue of what happens if someone points this resource at a completely different upstream package, with respect to updates.

IIRC porch doesn't let you update a package to a completely different package. We started discussing what that means [here](https://github.com/GoogleContainerTools/kpt/issues/3467#issuecomment-1279221758). I think that in order to completely change the upstream package of the DownstreamPackage, you'd also have to change (or delete) the downstream package target, essentially resulting in a brand new deployment package.

porch/controllers/downstreampackages/api/v1alpha1/downstreampackage_types.go
```diff
@@ -0,0 +1,87 @@
...
+
+type AdoptionPolicy string
+type DeletionPolicy string
+
+const (
+	AdoptionPolicyAdoptExisting AdoptionPolicy = "adoptExisting"
+	AdoptionPolicyAdoptNone     AdoptionPolicy = "adoptNone"
+
+	DeletionPolicyDelete DeletionPolicy = "delete"
+	DeletionPolicyDisown DeletionPolicy = "orphan"
```
mortent, line 50: Nit: We should also update the variable name to use the term `orphan`.

mortent (approved):
Just two small nits, but overall looks good.

We also need to find a pattern for how we want to test the controllers, but that is an issue with all of them at the moment.

johnbelamaric:
> Regardless we need to be prepared for situations where the automatic update fails, since there might be conflicts that can't be automatically resolved

Yeah, I think we'll need to raise an error in the package revision status.

PR #3695, opened 2022-12-15
porch: don't save empty patches

natasha41575:
This makes it so that if we do an `rpkg pull` immediately followed by an `rpkg push` without any changes, we won't end up with any tasks that represent empty patches.


mortent (approved):
Looks good, but can we add a test or two on this?

natasha41575:
> Looks good, but can we add a test or two on this?

Done.

PR #3696, opened 2022-12-16
Rollouts cluster store

droot:
Whats added in this PR:

- Rollouts API now has target cluster selector
- Added basic `clusterstore` package with methods to list kcc gke clusters and establish client connection
- The reconciler code has a basic code that exercises the clusterstore

Overall, I think this is a good checkpoint to get it merge. I plan to iterate more on clusterstore.


ChristopherFry (approved):

.gitignore
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+go.work
+go.work.sum
```
droot, line 16: @ChristopherFry decided to exclude `go.work` and `go.work.sum` to keep the changes outside the `rollouts` directory to minimal.
I can share these from my workspace.

PR #3700, opened 2022-12-21
rollouts: add package cluster matcher

ChristopherFry:
This pull request adds package cluster matching to the rollouts proof of concept.


rollouts/pkg/packageclustermatcher/packageclustermatcher.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	return allClusterPackages, nil
+}
+
+func isPackageClusterMatch(expr string, cluster, rolloutPackage map[string]interface{}) (bool, error) {
+	env, err := cel.NewEnv(
+		cel.Declarations(
+			decls.NewVar("cluster", decls.Dyn),
+			decls.NewVar("rolloutPackage", decls.Dyn),
```
droot, line 91: should we rename `rolloutPackage` to just `package` ?

rollouts/pkg/packageclustermatcher/packageclustermatcher.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+func (m *PackageClusterMatcher) GetClusterPackages(matcher gitopsv1alpha1.PackageToClusterMatcher) ([]ClusterPackages, error) {
+	clusters := m.clusters
+	packages := m.packages
+
+	allClusterPackages := []ClusterPackages{}
+
+	for _, cluster := range clusters {
```
droot, line 51: nit: I think we can constructs list of celCluster and list of celPackages separately and then loop over to run matchers. We can avoid costs of constructing `celPackages` for each clusters.

droot (approved):
Looks good. Minor comment.

rollouts/pkg/packageclustermatcher/packageclustermatcher.go
ChristopherFry: Unfortunately `package` is a resolved word in [CEL](https://github.com/google/cel-spec/blob/master/doc/langdef.md#syntax). I'm up for other suggestions though. `rolloutPackage` is a little wordy.

PR #3705, opened 2022-12-28
rollouts: update package to target matcher to be optional

ChristopherFry:
This pull request updates the packageToTargetMatcher field in the rollouts custom resource definition to be optional since not all rollouts will need to specify advanced package to cluster matching logic.

### DIFF
--- a/rollouts/api/v1alpha1/rollout_types.go
+++ b/rollouts/api/v1alpha1/rollout_types.go
@@ -37,7 +37,10 @@ type RolloutSpec struct {
 	Targets ClusterTargetSelector `json:"targets,omitempty"`
 
 	// PackageToTargetMatcher specifies the clusters that will receive a specific package.
-	PackageToTargetMatcher PackageToClusterMatcher `json:"packageToTargetMatcher"`
+	// +nullable
+	// +optional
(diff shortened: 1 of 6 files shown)


rollouts/api/v1alpha1/rollout_types.go
```diff
@@ -37,7 +37,10 @@ type RolloutSpec struct {
 	Targets ClusterTargetSelector `json:"targets,omitempty"`
 
 	// PackageToTargetMatcher specifies the clusters that will receive a specific package.
-	PackageToTargetMatcher PackageToClusterMatcher `json:"packageToTargetMatcher"`
+	// +nullable
+	// +optional
+	PackageToTargetMatcher *PackageToClusterMatcher `json:"packageToTargetMatcher,omitempty"`
```
droot, line 42: I think this part of the current API is very fragile. It's not very intuitive how packages will be matched with the cluster. I am still looking for some good options here.
But, before we make any field optional, we need to find good defaults. Until we find good defaults, having user explicitly specify the field at least reduces the chances of error (mis-configuration) though at the cost of usability, so it's a trade-off.

One option could be making the `true` matcher as an explicit predefined option so that users don't have to think in terms of `CEL` etc. but they will have explicitly specify it so that they understand how the matching happens.

Still looking for ideas.

PR #3706, opened 2022-12-29
rollouts: add caching for discovered packages

ChristopherFry:
This pull request updates the package discovery logic in the rollouts proof of concept to cache discovered packages for a rollout resource for a short duration.


rollouts/controllers/rollout_controller.go
```diff
@@ -115,15 +128,28 @@ func (r *RolloutReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
-	err := r.reconcileRollout(ctx, &rollout)
+	packageDiscoveryClient := func() *packagediscovery.PackageDiscovery {
```
droot, line 131: nit: I think we can make this function a helper method on the `reconciler`. I think that will be easier for the code reader.

```Go
packageDiscoveryClient := r.GetPackageDiscoveryClient()
```

droot (approved):

rollouts/controllers/rollout_controller.go
ChristopherFry: Done.

PR #3707, opened 2022-12-29
Add PackageVariantSet controller

natasha41575:
This PR adds the UpstreamPolicy (renamed to PackageVariantSet) as described in [this doc](https://docs.google.com/document/d/1OxNon_1ri4YOqNtEQivBgeRzIPuX9sOyu-nYukjwN1Q/edit?usp=sharing&resourcekey=0-2nDYYH5Kw58IwCatA4uDQw) and discussed in https://github.com/GoogleContainerTools/kpt/issues/3488. 

There are some TODOs in this PR that I am working on in a follow-up:
- Implement the option to specify "tag" rather than a "revision" for the upstream package revision
- Using "Conditions" rather than a top-level field in the status and other returned errors
- Implementing the "object set" option for downstream package targets.

This PR also adds a finalizer to the PackageVariant controller to make sure that it cleans up properly before being deleted. (I noticed some issues with its cleanup while trying out the PackageVariantSet controller.)

Relevant issue: https://github.com/GoogleContainerTools/kpt/issues/3488


natasha41575:
cc @johnbelamaric

porch/controllers/packagevariantsets/api/v1alpha1/packagevariantset_types.go
```diff
@@ -0,0 +1,143 @@
...
+
+	Tag string `json:"ref,omitempty"`
+}
+
+type Target struct {
+	// option 1: an explicit repo/package name pair
+	Package *Package `json:"package,omitempty"`
+
+	// option 2: a label selector against a set of repositories
+	Repositories map[string]string `json:"repositories,omitempty"`
```
mortent, line 66: If we want to use label selectors, we probably want to do it the same as in other kubernetes APIs: https://github.com/GoogleContainerTools/kpt/blob/main/porch/controllers/rootsyncrollouts/api/v1alpha1/rootsyncrollout_types.go#L29

porch/controllers/packagevariantsets/api/v1alpha1/packagevariantset_types.go
```diff
@@ -0,0 +1,143 @@
...
+	// APIVersion of the target resources
+	APIVersion string `yaml:"apiVersion,omitempty" json:"apiVersion,omitempty"`
+	// Kind of the target resources
+	Kind string `yaml:"kind,omitempty" json:"kind,omitempty"`
+	// Name of the target resources
+	Name string `yaml:"name,omitempty" json:"name,omitempty"`
+	// Namespace of the target resources
+	Namespace string `yaml:"namespace,omitempty" json:"namespace,omitempty"`
+	// Labels on the target resources
+	Labels map[string]string `yaml:"labels,omitempty" json:"labels,omitempty"`
```
mortent, line 97: Same for this as above. We should see if the normal way to do label selectors can be used here.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -0,0 +1,533 @@
...
+func (r *PackageVariantSetReconciler) objectSet(ctx context.Context,
+	target *api.Target,
+	upstreamPackageName string) ([]*pkgvarapi.Downstream, error) {
+	// TODO: Implement this
+	return nil, fmt.Errorf("specifying a set of objects in the target is not yet supported")
+}
+
+func (r *PackageVariantSetReconciler) getDownstreamPackageName(targetName *api.PackageName,
+	upstreamPackageName string,
+	obj *yaml.RNode) (string, error) {
```
mortent, line 309: I think it might be easier to just use Unstructured here instead of converting to RNode. At least it is easier for future users who might not know kyaml. Fetching fields should be pretty straight-forward using the functions in the unstructured package.

natasha41575: I agree with your sentiment and am glad you mentioned it, so I played around with `Unstructured` for a little bit to see if we could use it here instead, but IMO using RNode will actually be easier. From the [unstructured docs](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured):

```
NestedFieldNoCopy returns a reference to a nested field. Returns false if value is not found and an error if unable to traverse obj.

Note: fields passed to this function are treated as keys within the passed object; no array/slice syntax is supported.
```

From what I can tell, `NestedFieldNoCopy` is used under the hood for most, if not all, of the other `NestedXXX` functions in the library, so I ran into issues with trying to support fetching fields from nested array/slice elements. For example, I want to be able to specify a path like `spec.containers[1].image`, but unless I am missing something (which is very possible), the unstructured package doesn't provide a way to do so easily. In general I do agree that kyaml has a steeper learning curve and probably isn't as popular among kubernetes devs as Unstructured, but I think in this specific case, not using kyaml would mean rewriting nontrivial logic to handle array elements, so I feel that using kyaml directly makes more sense.

porch/controllers/packagevariantsets/api/v1alpha1/packagevariantset_types.go
natasha41575: Done, thanks for the link!

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -0,0 +1,533 @@
...
+
+// PackageVariantSetReconciler reconciles a PackageVariantSet object
+type PackageVariantSetReconciler struct {
+	client.Client
+	Options
+
+	serializer *json.Serializer
+}
+
+const PackageVariantSetOwnerLabel = "internal.porch.kpt.dev/packagevariantset-id"
```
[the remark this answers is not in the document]
natasha41575: Done, thanks!

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -351,25 +382,27 @@ func (r *PackageVariantReconciler) deleteOrOrphan(ctx context.Context,
...
 		klog.Errorf("invalid deletion policy %s", pv.Spec.DeletionPolicy)
 	}
 }
 
-func (r *PackageVariantReconciler) orphanPackageRevision(currentOwners []metav1.OwnerReference,
-	pvUID types.UID) []metav1.OwnerReference {
+func (r *PackageVariantReconciler) orphanPackageRevision(ctx context.Context,
+	pr *porchapi.PackageRevision,
+	pv *api.PackageVariant) {
+	currentOwners := pr.OwnerReferences
```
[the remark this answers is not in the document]
natasha41575: done.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -69,6 +70,36 @@ func (r *PackageVariantReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 
+	if !pv.ObjectMeta.DeletionTimestamp.IsZero() {
+		// This object is being deleted, so we need to make sure the packagerevisions owned by this object
+		// are deleted. Normally, garbage collection can handle this, but we have a special case here because
+		// (a) we cannot delete published packagerevisions and instead have to propose deletion of them
+		// (b) we may want to orphan packagerevisions instead of deleting them.
+		for _, pr := range prList.Items {
+			if r.hasOurOwnerReference(pv, pr.OwnerReferences) {
+				r.deleteOrOrphan(ctx, &pr, pv)
+				if porchapi.LifecycleIsPublished(pr.Spec.Lifecycle) {
```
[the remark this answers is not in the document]
natasha41575: > Does this take advantage of the fact that if a packagerevision still has lifecycle Published after the call to deleteOrOrphan, it must have deletionPolicy Orphan?

It did not, thanks for the suggestion. I have updated it so that it does, and clarified with a comment. This is needed because packagerevisions that are owned by this `PackageVariant` will be automatically deleted when the PackageVariant object is deleted. Since we don't want published package revisions to be deleted, we orphan them first to prevent the automatic deletion.

porch/controllers/packagevariants/api/v1alpha1/packagevariant_types.go
```diff
@@ -48,6 +48,8 @@ const (
 
 	DeletionPolicyDelete DeletionPolicy = "delete"
 	DeletionPolicyOrphan DeletionPolicy = "orphan"
+
+	Finalizer = "config.porch.kpt.dev/finalizer"
```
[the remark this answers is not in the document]
natasha41575: Done, thanks!

porch/controllers/packagevariantsets/api/v1alpha1/packagevariantset_types.go
natasha41575: Updated the type.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
mortent: That makes sense. My only question is if we have to specify a path that traverses into a specific entry in an array, should we actually be looking at all entries.

mortent (approved):

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
natasha41575: Could you elaborate on that? The path specified would come from a user-specified field in `PackageVariantSet` to fetch a particular field, I'm not sure that I understand the scenario where we would need to look at all entries?

mortent: I might have been confused by the `spec.containers[1].image` example. Usually I get suspicious when there is code checking something on only one container, since there might be more than one and we can't usually assume a specific ordering. I see now that we allow the users to specify a field path. I agree that kyaml is much better than unstructured for this.

mortent (approved):

johnbelamaric:
Woohoo!

PR #3712, opened 2023-01-05
Refactor controller logic for getting RESTConfig to a remote cluster

justinsb:
We had two copies, rationalize and take the best of each.

Also remove the HACK_ENABLE_LOOPBACK hack now that we can target remote clusters.

### DIFF
--- a/porch/controllers/remoterootsyncsets/pkg/remoteclient/getremoteclient.go
+++ b/porch/controllers/remoterootsyncsets/pkg/remoteclient/getremoteclient.go
@@ -18,47 +18,91 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
-	"os/exec"
-	"regexp"
+	"os"
 
 	container "cloud.google.com/go/container/apiv1"
-	api "github.com/GoogleContainerTools/kpt/porch/controllers/remoterootsyncsets/api/v1alpha1"
+	"github.com/GoogleContainerTools/kpt/porch/pkg/googleurl"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
+	"google.golang.org/api/option"
 	containerpb "google.golang.org/genproto/googleapis/container/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/discovery"
+	memory "k8s.io/client-go/discovery/cached"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
 	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// GetCCRESTConfig builds a rest.Config for accessing the config controller cluster,
+const (
+	containerClusterKind       = "ContainerCluster"
+	containerClusterApiVersion = "container.cnrm.cloud.google.com/v1beta1"
+
+	configControllerKind       = "ConfigControllerInstance"
+	configControllerApiVersion = "configcontroller.cnrm.cloud.google.com/v1beta1"
+)
+
+type RemoteClientGetter struct {
+	client.Client
+
+	workloadIdentity WorkloadIdentityHelper
+}
+
+// Init performs one-off initialization of the object.
+func (r *RemoteClientGetter) Init(mgr ctrl.Manager) error {
+	r.Client = mgr.GetClient()
+
+	return r.workloadIdentity.Init(mgr.GetConfig())
+}
+
+// getCCRESTConfig builds a rest.Config for accessing the config controller cluster,
 // this is a tmp workaround.
-func GetCCRESTConfig(ctx context.Context, cluster *unstructured.Unstructured) (*rest.Config, error) {
-	gkeResourceLink, exist, err := unstructured.NestedString(cluster.Object, "status", "gkeResourceLink")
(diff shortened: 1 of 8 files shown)


justinsb:
/retest

I don't think this is related, though it is a pod-evaluator bug...

mortent:
Yeah, we have seen some flakes with the pod-evaluator test. I'm triggering another run.

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
```diff
@@ -202,30 +191,6 @@ func (r *RootSyncSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	return ctrl.Result{}, nil
 }
 
-func toCanonicalClusterRef(ref *v1alpha1.ClusterRef, rssNamespace string) (v1alpha1.ClusterRef, error) {
```
mortent, line 205: Are we making all four properties on the clusterRef mandatory in the API here? We use the clusterRef as the key in maps in this controller.

porch/controllers/remoterootsyncsets/pkg/remoteclient/getremoteclient.go
```diff
@@ -18,47 +18,95 @@ import (
...
-// GetCCRESTConfig builds a rest.Config for accessing the config controller cluster,
+const (
+	containerClusterKind       = "ContainerCluster"
+	containerClusterApiVersion = "container.cnrm.cloud.google.com/v1beta1"
+
+	configControllerKind       = "ConfigControllerInstance"
+	configControllerApiVersion = "configcontroller.cnrm.cloud.google.com/v1beta1"
+)
+
+type RemoteClientGetter struct {
```
mortent, line 52: We should figure out a place to put code that is shared between the controllers. It seems a bit strange to have this code under the remoterootsyncset controller when it is pretty generic and also used by the RootSyncSet controller.

porch/controllers/remoterootsyncsets/pkg/remoteclient/getremoteclient.go
```diff
@@ -18,47 +18,95 @@ import (
...
+	projectID := googleURL.Project
+	location := googleURL.Location
+	clusterName := googleURL.Extra["clusters"]
 	klog.Infof("cluster name is %s", clusterName)
+
+	tokenSource, err := r.getConfigConnectorContextTokenSource(ctx, cluster.GetNamespace())
+	if err != nil {
+		return nil, err
+	}
+	// accessToken, err := GetGcloudAccessToken(ctx)
```
mortent, line 88: Should we just delete this?

porch/controllers/remoterootsyncsets/pkg/remoteclient/getremoteclient.go
droot: FWIW, in the rollouts branch, I refactored this into a separate package `clusterstore` to provides a way to list clusters, get clients (dynamic and client.Client) given a cluster. Eventually the plan was clusterstore could be an abstraction that can work with KCC clusters, hub clusters etc.

https://github.com/GoogleContainerTools/kpt/tree/feature-rollouts/rollouts/pkg/clusterstore

justinsb: Yes, agreed.  I was planning to do that as a follow-on.  I was even wondering if we should keep porch as a separate module, as it's a (minor) barrier to sharing with kpt to have to bump the versions.  Although that's less relevant for this logic, which is very server-side!

I like pkg/clusterstore :-)

porch/controllers/remoterootsyncsets/pkg/remoteclient/getremoteclient.go
justinsb: Good call - done!

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
justinsb: I wasn't sure here.  I figured it would still work (?), it might just miss some potential to share caches.  I would like to get this logic into kubebuilder-declarative-pattern - there are a few PRs in flight that support the idea of applying to a remote cluster that we could then just drop in.  I'll try to get those merge-ready today!

I did start introducing a string-clusterKey that e.g. would identify unique clusters based on the URL, but then I spotted that we're not pruning with the canonicalClusterRef (`r.pruneWatches(req.NamespacedName, rootsyncset.Spec.ClusterRefs)`), so I figured the change here was probably more correct anyway?

Happy to upload the clusterKey WIP branch, but I would rather get this into a shared and smaller library, as I think it's useful everywhere and I also think it's quite tricky code and could benefit from being implemented in a minimal standalone library!

justinsb:
/retest

Test failure looks unrelated

porch/controllers/remoterootsyncsets/pkg/remoteclient/getremoteclient.go
mortent: Yeah, I'm fine with this in a follow-up. I think we want to keep the modules separate and instead look into how we can make it easier to work with. At some point we will need to look into separate some of these components, for example splitting out some of the controllers that perhaps isn't tied to Porch.

porch/controllers/rootsyncsets/pkg/controllers/rootsyncset/controller.go
mortent: Moving the logic into a separate library like kubebuilder-declarative-pattern sounds good. The idea behind this was to avoid having to reason about incomplete clusterRefs, we should just make sure they are complete at all times. If we put it back, I can take a pass at simplifying it and make sure it doesn't break the functionality.

mortent (approved):

justinsb:
>  If we put it back, I can take a pass at simplifying it and make sure it doesn't break the functionality.

I wasn't entirely sure whether you were OK with merging this as-is?  I think it's more correct than the previous logic, which was inconsistent in creation vs deletion, but I might be missing something!

PR #3716, opened 2023-01-06
rollouts: added API for ProgressiveRolloutStrategy

droot:
This PR adds API for specifying progressive rollout strategies.


rollouts/PROJECT
```diff
@@ -22,4 +22,13 @@ resources:
   kind: RemoteRootSync
   path: github.com/GoogleContainerTools/kpt/rollouts/api/v1alpha1
   version: v1alpha1
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
```
ChristopherFry, line 28: I'm questing if we actually need a controller for the ProgressiveRolloutStrategy.

ChristopherFry:
Looks good. License header needs to be added to the missing files. And the year updated to 2023 :).

ChristopherFry (approved):

rollouts/PROJECT
droot: As discussed offline, we plan to use it for the following:
- Aggregating what rollouts are using this strategy
- Finalizer to prevent deletion of a strategy in case it is being used by any rollouts.

PR #3720, opened 2023-01-10
rollouts: refine package to cluster matcher

droot:
This PR refines the packageToClusterMatcher field in the rollouts API. Now, there are two matcherTypes

If you want to broadcast a package to all the clusters, you can simply specify

```yaml

packageToTargetMatcher:
    type: AllClusters

```

If you have any custom matcher logic, you will have to specify as:

```yaml

packageToTargetMatcher:
    type: Custom
    matchingExpression: `your CEL expression goes here`

```

Also, we now throw an error if more than one package is targeted for a cluster. We need to think more if we want to support rolling out multiple packages to a cluster.

### DIFF
--- a/rollouts/api/v1alpha1/rollout_types.go
+++ b/rollouts/api/v1alpha1/rollout_types.go
@@ -36,6 +36,7 @@ type RolloutSpec struct {
 
 	// PackageToTargetMatcher specifies the clusters that will receive a specific package.
 	PackageToTargetMatcher PackageToClusterMatcher `json:"packageToTargetMatcher"`
+
 	// Strategy specifies the rollout strategy to use for this rollout.
 	Strategy RolloutStrategy `json:"strategy"`
 }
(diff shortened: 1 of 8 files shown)


ChristopherFry (approved):

rollouts/api/v1alpha1/rollout_types.go
```diff
@@ -84,12 +85,17 @@ type SecretReference struct {
 	Name string `json:"name,omitempty"`
 }
 
-// +kubebuilder:validation:Enum=CEL
+// +kubebuilder:validation:Enum=allClusters;custom
```
ChristopherFry, line 88: Any thoughts using PascalCase here? Similar to Strategy (AllAtOnce, RollingUpdate, Progressive).

droot: Good catch. Fixed it.

PR #3721, opened 2023-01-11
rollouts: update progressive strategy to pause after wave

ChristopherFry:
This pull request updates the rollouts proof of concept, allowing Progressive strategy to be paused after a specific wave completes giving the user more control over the rollout progression. Validation is added to ensure the wave name does exist in the progressive rollout strategy referenced by the rollout.

Rollout status updates will be done in a fast follow-up pull request to keep the changes in this pull request isolated purely to pause logic.


rollouts/controllers/rollout_controller.go
```diff
@@ -192,43 +192,56 @@ func (r *RolloutReconciler) validateProgressiveRolloutStrategy(ctx context.Conte
...
-			return fmt.Errorf("wave %d does not target any clusters", waveIdx)
+			return fmt.Errorf("wave '%s' does not target any clusters", wave.Name)
 		}
 
 		for _, cluster := range waveClusters.Items {
 			currentClusterWave, found := clusterWaveMap[cluster.Name]
 			if !found {
 				// this should never happen
-				return fmt.Errorf("wave %d references cluster %s not selected by the rollout", waveIdx, cluster.Name)
+				return fmt.Errorf("wave '%s' references cluster %s not selected by the rollout", wave.Name, cluster.Name)
```
droot, line 221: tip: %q will give you quoted value of the arg.

rollouts/api/v1alpha1/rollout_types.go
```diff
@@ -116,8 +116,13 @@ type StrategyRollingUpdate struct {
 // StrategyProgressive defines the progressive rollout strategy to use.
 type StrategyProgressive struct {
 	// Reference of ProgressiveRolloutStrategy to use.
-	Name      string `json:"name"`
-	Namespace string `json:"namespace"`
+	Name           string         `json:"name"`
+	Namespace      string         `json:"namespace"`
+	PauseAfterWave PauseAfterWave `json:"pauseAfterWave,omitempty"`
```
droot, line 121: we should add small explanation about the behavior :)

Since this is optional. If not specified, it is assumed to be automatic progression between the waves ?

droot (approved):

rollouts/controllers/rollout_controller.go
ChristopherFry: Thanks! This is really good to know. I updated the print statements to use %q instead of '%s' for rollout name.

rollouts/api/v1alpha1/rollout_types.go
ChristopherFry: Added a description to the field to help explain the behavior in the latest commit.

Currently the field is required (unlike the pre pull request code) to avoid the distraction at the moment in specifying what happens when its not defined, and any defaulting logic that is needed.

PR #3736, opened 2023-01-17
licensescan: Add licenses for more libraries.

justinsb:
Updating our database with the latest libraries, as needed by some other projects sharing this DB!


justinsb:
I re-ran the test and it passed both times, so it seems this does help with the flakes (or I got lucky!)

PR #3737, opened 2023-01-18
rollouts: add ready condition to remote root sync

ChristopherFry:
This pull request updates the rollouts proof of concept, adding a ready condition to the remote root sync status. This condition allows us to understand if the remote root sync has connectivity to the target cluster.

### DIFF
--- a/rollouts/controllers/remoterootsync_controller.go
+++ b/rollouts/controllers/remoterootsync_controller.go
@@ -24,7 +24,6 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -211,7 +210,7 @@ func (r *RemoteRootSyncReconciler) updateStatus(ctx context.Context, rrs *gitops
 				Type:               "Ready",
 				Reason:             readyReason,
 				Status:             rssReady,
-				LastTransitionTime: v1.Now(),
+				LastTransitionTime: metav1.Now(),
 			},
 		}
 	}
--- a/rollouts/controllers/remoterootsync_controller.go
+++ b/rollouts/controllers/remoterootsync_controller.go
@@ -24,6 +24,7 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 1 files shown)


rollouts/controllers/remoterootsync_controller.go
```diff
@@ -24,6 +24,7 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
droot, line 27: nit: In kubernetes code bases, convention is to use `metav1` alias for this package. It's right above this line.

rollouts/controllers/remoterootsync_controller.go
```diff
@@ -136,42 +141,81 @@ func (r *RemoteRootSyncReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		rssReady = metav1.ConditionFalse
+	}
+
 	// Don't update if there are no changes.
-	if rrs.Status.SyncStatus == syncStatus && rrs.Generation == rrs.Status.ObservedGeneration {
+	syncStatusChanged := syncStatus != nil && rrs.Status.SyncStatus != *syncStatus
+	generationChanged := rrs.Generation == rrs.Status.ObservedGeneration
+	readyStatusChanged := len(rrs.Status.Conditions) == 0 || rrs.Status.Conditions[0].Type != "Ready" || rrs.Status.Conditions[0].Status != rssReady
+
+	if !generationChanged && !syncStatusChanged && !readyStatusChanged {
```
droot, line 199: Another way to do this will be: compute the new status and compare the new status with current using `reflect.DeepEqual` (that compares the entire struct recursively). That will future proof this code as we evolve the status.

rollouts/controllers/remoterootsync_controller.go
```diff
@@ -136,42 +141,81 @@ func (r *RemoteRootSyncReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 
+	rssReady := metav1.ConditionTrue
+	if readyReason != "Ready" {
+		rssReady = metav1.ConditionFalse
+	}
+
 	// Don't update if there are no changes.
-	if rrs.Status.SyncStatus == syncStatus && rrs.Generation == rrs.Status.ObservedGeneration {
+	syncStatusChanged := syncStatus != nil && rrs.Status.SyncStatus != *syncStatus
+	generationChanged := rrs.Generation == rrs.Status.ObservedGeneration
```
droot, line 196: Shouldn't this be `!=` or am I missing something ?

rollouts/controllers/remoterootsync_controller.go
```diff
@@ -136,42 +141,81 @@ func (r *RemoteRootSyncReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	gkeCluster, err := r.store.GetCluster(ctx, cr.Name)
 	if err != nil {
+		r.updateStatus(ctx, &remoterootsync, clusterConnectivityFailed, nil)
```
droot, line 144: I need to double check the underlying code but just getting the cluster shouldn't require reaching out to the target cluster so `connectivityFailure` may not be applicable.

droot (commented):
Thanks for improving this. I have some minor comments.

rollouts/controllers/remoterootsync_controller.go
ChristopherFry: It should be `!=`. Updated.

rollouts/controllers/remoterootsync_controller.go
ChristopherFry: Thanks! Fixed with latest.

rollouts/controllers/remoterootsync_controller.go
ChristopherFry: Looking into this. I will need to exclude the `Status.Conditions[].LastTransitionTime` field from the comparison as this can vary from the existing status and any status I compute from scratch.

droot:
@ChristopherFry Do we want to do anything here or close this out ?

ChristopherFry:
Closing this pull request in favor of #3849.

PR #3738, opened 2023-01-18
rollouts: add container cluster watch

ChristopherFry:
This pull request updates the rollouts proof of concept, adding a watch for container cluster resources. This watch allows the reconciliation of rollouts when the target cluster list changes (for instance, a cluster being added or removed from the list).

### DIFF
--- a/rollouts/controllers/rollout_controller.go
+++ b/rollouts/controllers/rollout_controller.go
@@ -63,7 +63,7 @@ const (
 )
 
 var (
-	configConnectorContainerClusterGVK = schema.GroupVersionKind{
+	kccClusterGVK = schema.GroupVersionKind{
 		Group:   "container.cnrm.cloud.google.com",
 		Version: "v1beta1",
 		Kind:    "ContainerCluster",
@@ -709,7 +709,7 @@ func (r *RolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 
 	var containerCluster unstructured.Unstructured
-	containerCluster.SetGroupVersionKind(configConnectorContainerClusterGVK)
+	containerCluster.SetGroupVersionKind(kccClusterGVK)
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&gitopsv1alpha1.Rollout{}).
--- a/rollouts/controllers/rollout_controller.go
+++ b/rollouts/controllers/rollout_controller.go
@@ -28,13 +28,19 @@ import (
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	gkeclusterapis "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/clients/generated/apis/container/v1beta1"
 	gitopsv1alpha1 "github.com/GoogleContainerTools/kpt/rollouts/api/v1alpha1"
@@ -54,6 +60,14 @@ func (o *Options) BindFlags(prefix string, flags *flag.FlagSet) {
 
 const (
 	rolloutLabel = "gitops.kpt.dev/rollout-name"
+)
+
+var (
+	configConnectorContainerClusterGVK = schema.GroupVersionKind{
+		Group:   "container.cnrm.cloud.google.com",
+		Version: "v1beta1",
+		Kind:    "ContainerCluster",
+	}
 )
 
 // RolloutReconciler reconciles a Rollout object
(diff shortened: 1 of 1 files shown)


rollouts/controllers/rollout_controller.go
```diff
@@ -680,13 +707,49 @@ func (r *RolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		selector, err := metav1.LabelSelectorAsSelector(rollout.Spec.Targets.Selector)
+		if err != nil {
+			logger.Error(err, "failed to create label selector", "rolloutName", rollout.Name)
+			continue
+		}
+
+		rolloutDeploysToCluster := rolloutIncludesCluster(rollout, cluster.GetName())
+		clusterInTargetSet := selector.Matches(labels.Set(cluster.GetLabels()))
+
+		if rolloutDeploysToCluster || clusterInTargetSet {
```
droot, line 745: would be good to add a comment explaining why both needs to be done.

rollouts/controllers/rollout_controller.go
```diff
@@ -680,13 +707,49 @@ func (r *RolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	var containerCluster unstructured.Unstructured
+	containerCluster.SetGroupVersionKind(configConnectorContainerClusterGVK)
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&gitopsv1alpha1.Rollout{}).
 		Owns(&gitopsv1alpha1.RemoteRootSync{}).
+		Watches(
+			&source.Kind{Type: &containerCluster},
+			handler.EnqueueRequestsFromMapFunc(r.mapClusterUpdateToRequest),
```
droot, line 719: You are an advance controller author now :)

droot (commented):
Good going so far. I have couple of minor comments.
I am also thinking if we need to do the same thing for RemoteRootSync ? and how to take care of orphaned RemoteRootSync because they will not be captured here probably.

rollouts/controllers/rollout_controller.go
```diff
@@ -56,6 +62,14 @@ const (
 	rolloutLabel = "gitops.kpt.dev/rollout-name"
 )
 
+var (
+	configConnectorContainerClusterGVK = schema.GroupVersionKind{
```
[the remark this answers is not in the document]
ChristopherFry: Updated.

rollouts/controllers/rollout_controller.go
```diff
@@ -588,6 +602,19 @@ func (r *RolloutReconciler) listRemoteRootSyncs(ctx context.Context, rsdName, rs
...
+func (r *RolloutReconciler) listAllRollouts(ctx context.Context) ([]*gitopsv1alpha1.Rollout, error) {
+	var list gitopsv1alpha1.RolloutList
+	if err := r.List(ctx, &list); err != nil {
+		return nil, err
+	}
+	var rollouts []*gitopsv1alpha1.Rollout
+	for i := range list.Items {
+		item := &list.Items[i]
+		rollouts = append(rollouts, item)
+	}
```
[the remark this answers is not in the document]
ChristopherFry: Updated the function to return the list as is. There's no reason it can't.

rollouts/controllers/rollout_controller.go
ChristopherFry: Added a comment explaining the rational when a rollout needs to be reconciled for a cluster update.

rollouts/controllers/rollout_controller.go
```diff
@@ -680,13 +707,49 @@ func (r *RolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	if err := r.store.Init(); err != nil {
 		return err
 	}
-	// TODO: watch cluster resources as well
+
+	var containerCluster unstructured.Unstructured
```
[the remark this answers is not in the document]
ChristopherFry: Two reasons. Copy and paste being the first (not a great reason in itself). And the fact that I need not need to reference any specific fields with the ContainerCluster resource and this is sufficient to pass to the watch source being the second.

Being that we already import the KCC types, I don't think there's any reason I should use the ContainerCluster type, so I've updated the latest code to use the type.

rollouts/controllers/rollout_controller.go
ChristopherFry: Working on it :)

rollouts/controllers/rollout_controller.go
```diff
@@ -680,13 +702,55 @@ func (r *RolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	if err := r.store.Init(); err != nil {
 		return err
 	}
-	// TODO: watch cluster resources as well
+
+	var containerCluster gkeclusterapis.ContainerCluster
+	containerCluster.SetGroupVersionKind(kccClusterGVK)
```
droot, line 707: nit: if kcctypes are added the scheme (which I believe otherwise ListClusters would have failed), this line is redundant :)

rollouts/controllers/rollout_controller.go
```diff
@@ -588,6 +601,15 @@ func (r *RolloutReconciler) listRemoteRootSyncs(ctx context.Context, rsdName, rs
...
 }
 
+func (r *RolloutReconciler) listAllRollouts(ctx context.Context) ([]gitopsv1alpha1.Rollout, error) {
+	var rolloutsList gitopsv1alpha1.RolloutList
+	if err := r.List(ctx, &rolloutsList); err != nil {
+		return nil, err
+	}
+
+	return rolloutsList.Items, nil
+}
```
droot, line 611: nit: I would call `r.List(...)` directly where this method is being invoked :)

ChristopherFry:
> Good going so far. I have couple of minor comments.
I am also thinking if we need to do the same thing for RemoteRootSync ? and how to take care of orphaned RemoteRootSync because they will not be captured here probably.

Thanks! I updated the code so this should be ready for review.

Regarding the `RemoteRootSync`, I believe we already have this covered since the Reconciler considers this as a 'Owned' resource so it automatically watched. For example, anytime I delete a `RemoteRootSync` outside of a rollout, the rollout reconciles and recreated the sync. 

I'll be covering the orphaned case in a separate pull request that will focus on removing a `RemoteRootSync` that is no longer needed by a rollout. The ETA for the pull request is this afternoon. This pull request here is important as it does trigger the rollout reconcile when a cluster is removed from the target set which leads us into this scenario.

droot (approved):
Awesome. Looks good. 🚢 it :) [Don't worry about the minor nits: feel free to address them whenever).

PR #3741, opened 2023-01-19
e2e: add delay after registering Repository

justinsb:
I believe this will help avoid the "failed to list resources" error
immediately after registering a repository.

### DIFF
--- a/porch/test/e2e/e2e_test.go
+++ b/porch/test/e2e/e2e_test.go
@@ -2152,6 +2152,11 @@ func (t *PorchSuite) registerGitRepositoryF(ctx context.Context, repo, name, dir
 		},
 	})
 
+	// TODO: Replace with readiness check or similar, once we get to CRDs
+	// Sometimes we see "failed to list resources" here, I believe because we need to wait for the repository to be crawled.
+	t.Logf("HACK: sleeping for 5 seconds to allow for repository registration")
+	time.Sleep(5 * time.Second)
+
 	t.Cleanup(func() {
 		t.DeleteL(ctx, &configapi.Repository{
 			ObjectMeta: metav1.ObjectMeta{


justinsb:
I retest and encountered a different test (but possibly the same broad class of problems, ResourceGroup CRD not registered.... will try to fix that one separately)

porch/test/e2e/e2e_test.go
```diff
@@ -2152,6 +2152,11 @@ func (t *PorchSuite) registerGitRepositoryF(ctx context.Context, repo, name, dir
 		},
 	})
 
+	// TODO: Replace with readiness check or similar, once we get to CRDs
+	// Sometimes we see "failed to list resources" here, I believe because we need to wait for the repository to be crawled.
+	t.Logf("HACK: sleeping for 5 seconds to allow for repository registration")
```
mortent, line 2157: I'm all for trying this. I tried something similar for the [TestFunctionRepository](https://github.com/GoogleContainerTools/kpt/blob/main/porch/test/e2e/e2e_test.go#L627) test by making it wait for the Repository to become `Ready` (https://github.com/GoogleContainerTools/kpt/blob/af926a09fab38474740870c56199f64c63981db8/porch/test/e2e/e2e_test.go#L2308), but it didn't seem to fix the flakes.

mortent (approved):

porch/test/e2e/e2e_test.go
justinsb: Ah, that's much better than my hack.  I _think_ the issue is that PackageRevisions locks because it is updating synchronously, so adding a "wait for PackageRevisions" poll might help here.

I propose we merge this, see if it helps, and if it seems to be "much better" I can work on adding polling until PackageRevisions are ready as well...

This will likely also get fixed by CRDs :-)

PR #3742, opened 2023-01-19
rollouts: delete remote root sync when no longer needed

ChristopherFry:
This pull request updates the rollouts proof of concept allowing the rollouts controller to remove RemoteRootSync resources which are no longer required for the rollout (for instance, when a cluster is removed from the rollout's target set). The RemoteRootSync resource will be removed in the same wave it was added in by the rollout.

### DIFF
--- a/rollouts/controllers/rollout_controller.go
+++ b/rollouts/controllers/rollout_controller.go
@@ -275,12 +275,25 @@ func (r *RolloutReconciler) getPackageDiscoveryClient(rolloutNamespacedName type
 func (r *RolloutReconciler) reconcileRollout(ctx context.Context, rollout *gitopsv1alpha1.Rollout, strategy *gitopsv1alpha1.ProgressiveRolloutStrategy, packageDiscoveryClient *packagediscovery.PackageDiscovery) error {
 	logger := log.FromContext(ctx)
 
+	targetClusters, err := r.store.ListClusters(ctx, rollout.Spec.Targets.Selector)
 	discoveredPackages, err := packageDiscoveryClient.GetPackages(ctx, rollout.Spec.Packages)
 	if err != nil {
 		logger.Error(err, "failed to discover packages")
 		return client.IgnoreNotFound(err)
 	}
 	logger.Info("discovered packages", "count", len(discoveredPackages), "packages", discoveredPackages)
+
+	packageClusterMatcherClient := packageclustermatcher.NewPackageClusterMatcher(targetClusters.Items, discoveredPackages)
+	allClusterPackages, err := packageClusterMatcherClient.GetClusterPackages(rollout.Spec.PackageToTargetMatcher)
+	if err != nil {
+		logger.Error(err, "get cluster packages failed")
+		return client.IgnoreNotFound(err)
+	}
+
+	targets, err := r.computeTargets(ctx, rollout, allClusterPackages)
+	if err != nil {
+		return err
+	}
 
 	allClusterStatuses := []gitopsv1alpha1.ClusterStatus{}
 
@@ -294,30 +307,16 @@ func (r *RolloutReconciler) reconcileRollout(ctx context.Context, rollout *gitop
 
 	waveStatuses := []gitopsv1alpha1.WaveStatus{}
 
-	for _, wave := range strategy.Spec.Waves {
-		waveClusters, err := r.store.ListClusters(ctx, rollout.Spec.Targets.Selector, wave.Targets.Selector)
-		if err != nil {
-			return err
-		}
+	allWaveTargets, err := r.getWaveTargets(ctx, rollout, targets, strategy.Spec.Waves)
+	if err != nil {
+		return err
+	}
 
-		packageClusterMatcherClient := packageclustermatcher.NewPackageClusterMatcher(waveClusters.Items, discoveredPackages)
-		allClusterPackages, err := packageClusterMatcherClient.GetClusterPackages(rollout.Spec.PackageToTargetMatcher)
-		if err != nil {
-			logger.Error(err, "get cluster packages failed")
-			return client.IgnoreNotFound(err)
-		}
+	for _, waveTargets := range allWaveTargets {
+		wave := waveTargets.Wave
+		waveTargets := waveTargets.Targets
 
-		for _, clusterPackages := range allClusterPackages {
-			clusterName := clusterPackages.Cluster.Name
-			logger.Info("cluster packages", "cluster", clusterName, "packagesCount", len(clusterPackages.Packages), "packages", clusterPackages.Packages)
-		}
-
-		targets, err := r.computeTargets(ctx, rollout, allClusterPackages, waveClusters.Items)
-		if err != nil {
-			return err
-		}
-
-		thisWaveInProgress, clusterStatuses, err := r.rolloutTargets(ctx, rollout, &wave, targets, pauseFutureWaves)
+		thisWaveInProgress, clusterStatuses, err := r.rolloutTargets(ctx, rollout, wave, waveTargets, pauseFutureWaves)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


rollouts/controllers/rollout_controller.go
```diff
@@ -275,13 +275,26 @@ func (r *RolloutReconciler) getPackageDiscoveryClient(rolloutNamespacedName type
...
 		logger.Error(err, "failed to discover packages")
 		return client.IgnoreNotFound(err)
 	}
 	logger.Info("discovered packages", "count", len(discoveredPackages), "packages", discoveredPackages)
 
+	packageClusterMatcherClient := packageclustermatcher.NewPackageClusterMatcher(targetClusters.Items, discoveredPackages)
+	allClusterPackages, err := packageClusterMatcherClient.GetClusterPackages(rollout.Spec.PackageToTargetMatcher)
+	if err != nil {
+		logger.Error(err, "get cluster packages failed")
+		return client.IgnoreNotFound(err)
```
droot, line 290: I am not sure if `client.IgnoreNotFound(err)` is relevant here.

rollouts/controllers/rollout_controller.go
```diff
@@ -275,13 +275,26 @@ func (r *RolloutReconciler) getPackageDiscoveryClient(rolloutNamespacedName type
...
+	}
+
+	targets, err := r.computeTargets(ctx, rollout, allClusterPackages)
+	if err != nil {
+		return err
+	}
```
droot, line 296: I think new flow is more intuitive in my head.

1. compute all the targets
2. Assign targets to waves
3. then, rollout waves

rollouts/controllers/rollout_controller.go
```diff
@@ -294,30 +307,16 @@ func (r *RolloutReconciler) reconcileRollout(ctx context.Context, rollout *gitop
...
+	allWaveTargets, err := r.getWaveTargets(ctx, rollout, targets, strategy.Spec.Waves)
+	if err != nil {
+		return err
+	}
 
-		targets, err := r.computeTargets(ctx, rollout, allClusterPackages, waveClusters.Items)
-		if err != nil {
-			return err
-		}
+	for _, waveTargets := range allWaveTargets {
```
droot, line 315: nit: can we move `pauseFutureWaves` and other waves variable closer to this loop. I think of this phase as `rolloutWaves` so all state related to rolling out waves would be good to have in one section. And potentially we can move this in a function later.

rollouts/controllers/rollout_controller.go
```diff
@@ -294,30 +307,16 @@ func (r *RolloutReconciler) reconcileRollout(ctx context.Context, rollout *gitop
...
+		wave := waveTargets.Wave
+		waveTargets := waveTargets.Targets
 
-		thisWaveInProgress, clusterStatuses, err := r.rolloutTargets(ctx, rollout, &wave, targets, pauseFutureWaves)
+		thisWaveInProgress, clusterStatuses, err := r.rolloutTargets(ctx, rollout, wave, waveTargets, pauseFutureWaves)
```
droot, line 319: Since we are looping over and using the local variable and pointers :), we careful about the pitfall :)

rollouts/controllers/rollout_controller.go
```diff
@@ -378,11 +377,7 @@ func (r *RolloutReconciler) computeTargets(ctx context.Context,
 
 	// initially assume all the keys to be deleted
 	for _, rrs := range existingRRSs {
-		for _, cluster := range allowClusters {
-			if rrs.Spec.ClusterRef.Name == cluster.Name {
-				RRSkeysToBeDeleted[client.ObjectKeyFromObject(rrs)] = rrs
-			}
-		}
+		RRSkeysToBeDeleted[client.ObjectKeyFromObject(rrs)] = rrs
```
droot, line 380: interesting, now that we are computing targets at higher level, this is falling back in place.

rollouts/controllers/rollout_controller.go
```diff
@@ -749,13 +794,57 @@ func (r *RolloutReconciler) mapClusterUpdateToRequest(cluster client.Object) []r
...
+		if _, inWave := clusterNames[rrs.Spec.ClusterRef.Name]; inWave {
+			targets.ToBeUpdated = append(targets.ToBeUpdated, rrs)
+		}
+	}
+
+	for _, rrs := range allTargets.Unchanged {
+		if _, inWave := clusterNames[rrs.Spec.ClusterRef.Name]; inWave {
+			targets.Unchanged = append(targets.Unchanged, rrs)
+		}
+	}
```
droot, line 824: I wonder if we should remove the targets allocated to waves from allTargets ? (and print the counts at the end, might help us in catching any edge cases)

rollouts/controllers/rollout_controller.go
```diff
@@ -749,13 +794,57 @@ func (r *RolloutReconciler) mapClusterUpdateToRequest(cluster client.Object) []r
...
+
+	if len(allTargets.ToBeDeleted) > 0 && len(rollout.Status.WaveStatuses) > 0 {
```
droot, line 826: I couldn't understand the second condition ? (Is it safe to assume that the wave will exist in the status ? or should we attach waveName to rrs either as label or annotation ?)

rollouts/controllers/rollout_controller.go
```diff
@@ -433,6 +428,50 @@ func (r *RolloutReconciler) computeTargets(ctx context.Context,
 	return targets, nil
 }
 
+func (r *RolloutReconciler) getWaveTargets(ctx context.Context, rollout *gitopsv1alpha1.Rollout, allTargets *Targets,
```
droot, line 431: these two function names `getWaveTargets` `getTargetsForWave` confused me a for a bit :)

droot (commented):
I did a quick pass and looking good overall. I have minor comments and notes.

I will take another pass soon :)

rollouts/controllers/rollout_controller.go
ChristopherFry: True. This is a mistake. Updated the code to return the error.

rollouts/controllers/rollout_controller.go
ChristopherFry: Agreed. Updated.

rollouts/controllers/rollout_controller.go
ChristopherFry: True. With the latest code I was able to eliminate the `getTargetsForWave` function.

rollouts/controllers/rollout_controller.go
ChristopherFry: Take a look at the latest code. I was able to eliminate this condition here with new logic that is much easier to follow.

rollouts/controllers/rollout_controller.go
ChristopherFry: I updated the code so the code only needs to iterate through the targets for each list once (created, updated, unchanged, deleted) assigning each target to its respective wave. This simplifies the code and eliminates the possibility where a target is not associated with a wave.

ChristopherFry:
> I did a quick pass and looking good overall. I have minor comments and notes.

Thanks! I updated the code and it is ready for another pass.

rollouts/controllers/rollout_controller.go
```diff
@@ -433,6 +427,65 @@ func (r *RolloutReconciler) computeTargets(ctx context.Context,
...
+
+	for _, rrs := range allTargets.ToBeDeleted {
+		// The remote root sync will be associated back to it's previous wave and then removed as part
+		// of that wave. If the previous wave the remote root sync cannot be determined, then the remote
+		// root sync will be removed with the last wave of the rollout.
+
+		waveName, found := findWaveNameForCluster(rollout, rrs.Spec.ClusterRef.Name)
+
+		if !found {
+			waveName = allWaveTargets[len(allWaveTargets)-1].Wave.Name
```
droot, line 475: You must have thought about it already, I was wondering about the edge case where only one wave is involved.

droot (approved):
Wow, this is so simplified from the previous one I reviewed :).

PR #3746, opened 2023-01-24
chore: Upgrade cli-utils to v0.34.0

droot:
Upgrades cli-utils to v0.34.0 which contains an upgrade to Go v1.18 and Kubernetes v1.25 resources.

Note: I am reviving https://github.com/GoogleContainerTools/kpt/pull/3642 PR.


porch/pkg/registry/porch/packagerevision.go
```diff
@@ -52,6 +52,8 @@ func (r *packageRevisions) New() runtime.Object {
 	return &api.PackageRevision{}
 }
 
+func (r *packageRevisions) Destroy() {}
```
natasha41575, line 55: I was curious about these new `Destroy` methods, it seems to be required in order for `packageRevisions` to implement the `rest.Storage` interface, but I'm not that familiar when these methods are called. From a quick look, it seems like `Destroy` will be called when the apiserver shuts down. I don't think packagerevisions has anything it needs to clean up before being destroyed, so leaving these empty is probably fine.

natasha41575 (approved):
lgtm, but looks like you'll have to change the author of the commit for the cla check to pass

porch/pkg/registry/porch/packagerevision.go
droot: Empty destroy methods shouldn't disrupt existing flow, so I am also assuming them to be safe.

droot:
> lgtm, but looks like you'll have to change the author of the commit for the cla check to pass

Thanks for the tip. Just updated the author, hopefully it will pass this time.

porch/pkg/registry/porch/packagerevision.go
rquitales: From #3642 the `New` methods here initialize objects in memory with pointers to initialized structs so manual cleanup isn't needed since the Go GC should be doing that.
Example code from upstream apiserver with a similar pattern: (https://github.com/kubernetes/apiserver/commit/8010e8e7e364b464575cf26d1ebaafe09b8509b8)

PR #3748, opened 2023-01-24
WIP: Support package version in path

justinsb:
Working with multiple package versions can be difficult when we aren't
already maintaining branches-per-package-version, because it requires
a lot of branches to do effectively.

Add support for the version as the final directory in the path.


liamfallon:
Porch PR

PR #3751, opened 2023-01-24
tests: Add wait for function and packagerevision readiness

justinsb:
The agggregated apiserver can sometimes be slow to respond, for
example if it is polling git synchronously.  After creating a
Repository in our tests, wait for the aggregated apiserver to start
responding before returning.

This should clean up some test flakes.


justinsb:
/retest

Looks like we have more than one class of flake :-)

natasha41575 (approved):

PR #3761, opened 2023-01-27
porch cli: command to reject deletion proposal of published packagerevs

natasha41575:
This adds a way to reject a proposal of deletion of a published package revision (i.e. change the lifecycle from DeletionProposed back to Published) via `kpt alpha rpkg reject`.

### DIFF
--- a/commands/alpha/rpkg/reject/command.go
+++ b/commands/alpha/rpkg/reject/command.go
@@ -62,6 +62,7 @@ type runner struct {
 	ctx     context.Context
 	cfg     *genericclioptions.ConfigFlags
 	client  rest.Interface
+	porchClient client.Client
 	Command *cobra.Command
 
 	// Flags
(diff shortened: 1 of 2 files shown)


mortent:
Using a flag on the `propose-delete` command seems inconsistent with the existing UX. I can think of two other alternatives that I think I like better:
* Let the `reject` command undo both proposed deletion and publishing. So the cli will just check the current state of the packagerevision and then chose to either move it back to `draft` or `published`.
* If we want a flag for reject a proposed deletion, add that flag to the `reject` command rather than the `propose-delete` command.

The third option if of course to just add a separate `reject-delete` command, but it seems a bit unnecessary.

johnbelamaric:
Using the `reject` command for both seems like the better UX to me.

natasha41575:
> Using the reject command for both seems like the better UX to me.

Will do! Ty @johnbelamaric and @mortent for the suggestions

mortent (approved):

PR #3764, opened 2023-01-31
packagevariantset controller: support arbitrary object selectors

natasha41575:
This PR implements support for using arbitrary objects in the cluster as input to the PackageVariantSet controller (option 3 for targets).

Related issue: https://github.com/GoogleContainerTools/kpt/issues/3488


porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -291,11 +306,65 @@ func (r *PackageVariantSetReconciler) repositorySet(
...
+	for _, selector := range selectors {
+		uList := &unstructured.UnstructuredList{}
+		group, version := resid.ParseGroupVersion(selector.APIVersion)
+		uList.SetGroupVersionKind(schema.GroupVersionKind{
+			Group:   group,
+			Version: version,
+			Kind:    selector.Kind,
+		})
+
+		labelSelector, err := metav1.LabelSelectorAsSelector(selector.Labels)
```
mortent, line 346: Nit: How do we want to handle empty selectors here? Presumably it should match everything, but I'm not sure if that is the default behavior here?

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -291,11 +306,65 @@ func (r *PackageVariantSetReconciler) repositorySet(
...
+		if err != nil {
+			return nil, err
+		}
+
+		if err := r.Client.List(ctx, uList,
```
mortent, line 351: I think there are errors that could come here where we should probably expose an error in the status rather than retry. If the provided GVK is not a known type would be one example.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller_test.go
```diff
@@ -286,3 +289,77 @@ packageName:
...
+	downstreams := []*pkgvarapi.Downstream{
+		{Repo: "dn-1", Package: "dn-1"},
+		{Repo: "dn-3", Package: "dn-3"},
+	}
+	pvs := &api.PackageVariantSet{ObjectMeta: metav1.ObjectMeta{Name: "my-pvs"}}
+
+	fc := &fakeClient{}
+	reconciler := &PackageVariantSetReconciler{Client: fc}
+	require.NoError(t, reconciler.ensurePackageVariants(context.Background(), upstream, downstreams, pvs))
+	require.Equal(t, []string{"listing objects",
```
mortent, line 361: This way of testing the behavior seems a bit awkward. It is probably fine for now, but would be be clearer if we kept a an internal cache of objects and test the end state rather than behavior?

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
natasha41575: I updated it so that by default it matches everything!

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
natasha41575: done. Also moved validation errors to status.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller_test.go
natasha41575: done.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -76,13 +81,21 @@ func (r *PackageVariantSetReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, nil
 	}
 
+	var conditions []metav1.Condition
```
mortent, line 84: When we work with conditions, there are a few things we should do to make it easier for clients to use them:
- Always set all the conditions on a resource for every reconcile. So for the `Valid` condition, we should make sure we set it to `True` whenever there are no validation issues.
- We should make sure the `LastTransitionTime` only changes when the value of the `Status` field changes. So we shouldn't recreate the conditions slice from scratch on every reconcile, but update the existing conditions.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -98,7 +111,20 @@ func (r *PackageVariantSetReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	downstreams, err := r.unrollDownstreamTargets(ctx, pvs,
 		upstream.Package)
 	if err != nil {
-		return ctrl.Result{}, err
+		if strings.Contains(err.Error(), "unable to list specified objects in cluster") {
```
mortent, line 114: I think it is primarily if the resource type is unknown that we should handle with conditions rather than retries. Can we do this by relying on `errors.As` rather than string check on our own error?

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -98,7 +111,20 @@ func (r *PackageVariantSetReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+			conditions = append(conditions, metav1.Condition{
+				Type:               "ObjectsListed",
```
mortent, line 116: Do we need a separate conditions for this? If the user has specified an unknown resource type it seems like that could be handled as a validation error?

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
mortent: I think the easiest way to check for this might be to look up the resource type in the RESTMapper as part of the validation step.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
natasha41575: Thanks for the explanation! I'm now using the apimachinery library to update the condition which should take care of `LastTransitionTime`, and made sure that the `Valid` condition is set to true when everything looks good.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
natasha41575: Changed it to be part of the Valid condition.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
natasha41575: I'm not sure how to use the RESTMapper, but I'm now using `meta.IsNoMatchError` to check the returned error type here. Happy to update it if you know of a better way to handle this with RESTMapper and can share some pointers.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -76,36 +82,52 @@ func (r *PackageVariantSetReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, nil
 	}
 
+	defer func() {
+		r.Client.Status().Update(ctx, pvs)
```
mortent, line 86: Nit: We should probably log the error if something fails here.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -76,36 +82,52 @@ func (r *PackageVariantSetReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		// Currently, this code will never be reached, because the upstream.Tag option
+		// is not yet implemented.
 		return ctrl.Result{}, fmt.Errorf("could not find specified upstream")
 	}
 
-	downstreams, err := r.unrollDownstreamTargets(ctx, pvs,
-		upstream.Package)
+	downstreams, err := r.unrollDownstreamTargets(ctx, pvs, upstream.Package)
 	if err != nil {
+		if meta.IsNoMatchError(err) {
```
mortent, line 119: I don't think we want to return the error in this situation. It will trigger a new reconcile, which seems unnecessary for these errors.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
mortent: I think that looks like a good solution.

mortent (approved):
Two small comments, but looks good.

PR #3786, opened 2023-02-04
Add healthcheck crd manifests

ziyue-101:
Open source the HealthCheck CRD.

a description of the change
* As discussed with @justinsb, this PR adds health check CRD manifests to kpt repo.
the motivation for the change
* This change is a starting point for the health check framework library project.
The project will have toolings to help operators to manage and implement health checks.


healthcheck/api/v1alpha1/healthcheck_types.go
```diff
@@ -0,0 +1,89 @@
...
+	// +kubebuilder:validation:Enum=FatalError;NonFatalError
+	Type HealthCheckConditionType `json:"type,omitempty"`
+	// +kubebuilder:validation:Enum=Unknown;Healthy;Unhealthy
+	Status metav1.ConditionStatus `json:"status,omitempty"`
+	// The reason for the condition's last transition.
+	Reason string `json:"reason,omitempty"`
+	// A human readable message indicating details about the transition.
+	Message string `json:"message,omitempty"`
+	// The unique error ID.
+	CanonicalID string `json:"canonicalID,omitempty"`
```
justinsb, line 54: I think it would be good to say something like "this follows metav1.Condition but adds the canonicalID field"

healthcheck/api/v1alpha1/healthcheck_types.go
```diff
@@ -0,0 +1,89 @@
...
+	// The unique error name.
+	CanonicalName string `json:"canonicalName,omitempty"`
```
justinsb, line 56: Is this the same as `Reason` ?

healthcheck/api/v1alpha1/healthcheck_types.go
```diff
@@ -0,0 +1,89 @@
...
+	// The last time the condition transitioned from one status to another.
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
+}
+
+const (
+	// LabelComponent indicates the component to which the health check belongs
+	LabelComponent = "healthcheck.krmapihosting.cloud.google.com/component"
+	// LabelServiceError is set to true if the health check is service level
+	LabelServiceError = "healthcheck.krmapihosting.cloud.google.com/serviceError"
```
justinsb, line 65: Not quite sure how this is used - maybe add an example?

For example, I'm wondering if should be `healthcheck.<apiGroup>/scope = service` or something like that?

healthcheck/hack/boilerplate.go.txt
```diff
@@ -0,0 +1,13 @@
+// Copyright 2023 Google LLC
```
justinsb, line 1: I think we need to sync this with the existing boilerplate in this repo.  However, it's the same license, so not a blocker.

healthcheck/Makefile
```diff
@@ -0,0 +1,106 @@
+# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
```
justinsb, line 1: Would be great to sync this with the other makefiles in the repo.  That said I don't think it's a blocker.

healthcheck/api/v1alpha1/healthcheck_types.go
```diff
@@ -0,0 +1,89 @@
...
+	// LabelUnknown means users haven't set the label value.
+	LabelUnknown = "Unknown"
```
justinsb, line 67: This is the value that is set for e.g. LabelComponent ?

healthcheck/api/v1alpha1/healthcheck_types.go
```diff
@@ -0,0 +1,89 @@
...
+	CanonicalID string `json:"canonicalID,omitempty"`
+	// The unique error name.
+	CanonicalName string `json:"canonicalName,omitempty"`
+	// The last time the condition transitioned from one status to another.
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
+}
+
+const (
+	// LabelComponent indicates the component to which the health check belongs
+	LabelComponent = "healthcheck.krmapihosting.cloud.google.com/component"
```
johnbelamaric, line 63: Why are these Google-specific? Shouldn't this be kpt? These are generally useful. Is the for rollouts?

@s3wong @tliron you may want to look at these in the context of Nephio.

healthcheck/api/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,32 @@
...
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: "healthcheck.cloud.google.com", Version: "v1alpha1"}
```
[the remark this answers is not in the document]
ziyue-101: Yes, there is a google internal bug to track this. Shall I create a seprate OSS issue to track as well?

healthcheck/api/v1alpha1/healthcheck_types.go
ziyue-101: Based on the CRD design, this field only has two possible values Ture and False. And we don't expect to expand the field to include more values in the future, thus we chose boolean here.

healthcheck/api/v1alpha1/healthcheck_types.go
ziyue-101: No, it is a label set by the sever when the labels are not set. Let me remove this Label for now.

healthcheck/api/v1alpha1/healthcheck_types.go
ziyue-101: The CanonicalName is the name for a type of health check error. The val extends the meaning of the CannoicalID. ex https://cloud.google.com/anthos-config-management/docs/reference/errors#knv1000

The Reason field is a set of predefined events that explains the reason for the last transition. ex: https://cloud.google.com/config-connector/docs/how-to/monitoring-your-resources#-specific_events. IIUC, they are more general?

healthcheck/api/v1alpha1/healthcheck_types.go
ziyue-101: Good Catch. changed the api group.

healthcheck/Makefile
ziyue-101: Added the same license header;-)

healthcheck/hack/boilerplate.go.txt
ziyue-101: Switched to use https://github.com/GoogleContainerTools/kpt/blob/main/scripts/update-license.sh#L21 to generate licenses.

healthcheck/api/v1alpha1/groupversion_info.go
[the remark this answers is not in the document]
justinsb: Yeah, creating an issue here would be great thank you (and please send me the internal bug #!)

justinsb:
/ok-to-test

/approve
/lgtm

Let's merge and iterate.  We're not installing this CRD yet, so I figure we can change the api group to something OSS as one of the first things we need to do here :-)

healthcheck/api/v1alpha1/groupversion_info.go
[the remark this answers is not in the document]
ziyue-101: Issue created: https://github.com/GoogleContainerTools/kpt/issues/3800

justinsb (approved):

healthcheck/api/v1alpha1/healthcheck_types.go
```diff
@@ -0,0 +1,93 @@
...
+// json tags for the fields to be serialized.
+// HealthCheckSpec defines the metadata of a single health check.
+type HealthCheckSpec struct {
+}
+
+// HealthCheckStatus defines the status of a single health check.
+type HealthCheckStatus struct {
+	// Conditions represents the status of health check.
+	// +kubebuilder:validation:MaxItems=1
+	Conditions []HealthCheckCondition `json:"conditions,omitempty"`
```
roycaihw, line 31: in future iterations, I think it worth considering if we want to put everything in `Conditions`: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

having `+kubebuilder:validation:MaxItems=1` may suggest we are misusing conditions

healthcheck/api/v1alpha1/healthcheck_types.go
```diff
@@ -0,0 +1,93 @@
...
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
+}
+
+const (
+	// LabelComponent indicates the component to which the health check belongs.
+	LabelComponent = "healthcheck.cloud.google.com/component"
+	// LabelServiceError should be set to true if the health check is service level.
+	// LabelServiceError should be set to false if the health check is user level.
+	// Example: healthcheck.cloud.google.io/serviceError: true
+	LabelServiceError = "healthcheck.cloud.google.com/serviceError"
```
roycaihw, line 71: worth considering if we want to put these attributes to somewhere other than labels: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions, https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

if we want to achieve filtering, we can use field selectors: https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

PR #3787, opened 2023-02-06
add labels and annotations to packagevariant and packagevariantset

natasha41575:
Related to https://github.com/GoogleContainerTools/kpt/issues/3488

This PR adds the fields `spec.labels` and `spec.annotations` to the PackageVariant and PackageVariantSet types. The intention of these fields is so that the target package revisions will end up with these labels and annotations in their metadata. 

Implementation-wise, the PackageVariantSet controller adds the labels and annotations to the specs of the PackageVariant objects that it creates. The PackageVariant controller adds these labels and annotations to `metadata.labels` and `metadata.annotations` of package revisions that it either creates _or adopts_.


johnbelamaric:
Nice, thanks Natasha.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -418,6 +419,27 @@ func removeOwnerRefByUID(ownerRefs []metav1.OwnerReference,
...
+// has our owner reference and also the labels/annotations specified in pv.Spec.
+func (r *PackageVariantReconciler) adoptPackageRevision(ctx context.Context,
+	pr *porchapi.PackageRevision,
+	pv *api.PackageVariant) error {
+	pr.ObjectMeta.OwnerReferences = append(pr.OwnerReferences, constructOwnerReference(pv))
+	if len(pv.Spec.Labels) > 0 && pr.ObjectMeta.Labels == nil {
+		pr.ObjectMeta.Labels = make(map[string]string)
+	}
+	for k, v := range pv.Spec.Labels {
+		pr.ObjectMeta.Labels[k] = v
```
mortent, line 432: If someone removes an annotation or a label from the packagevarient CR, are those also removed from the PackageRevision with this implementation?

mortent (commented):
Just one clarification, otherwise looks good.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
natasha41575: No, they don't get removed. I contemplated this too, and wasn't sure if it's the right thing to do. @johnbelamaric feel free to chime in with your opinion on this too.

1. It would require saving the previous PackageVariant spec and doing a comparison to see if someone has removed the annotation/label from the CR. This makes the controller stateful. 
2. If the same label was added by something else - another controller or even another PackageVariant - then we shouldn't remove it. I don't think there is an easy way to detect whether the desired label was added by this PackageVariant or something else.

johnbelamaric: Oof. Yeah, I think this is the right choice, to not remove these. Let's just be sure we document it.

mortent: I think using SSA on the resource should handle this correctly, since it should only remove labels or annotations that is not included in the apply, but still have the controller as the owner of the fields. So it should require keeping state in the controller. But I think it is probably a corner case, so we can look at this later if necessary.

mortent (approved):

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
natasha41575: Ah thanks! I'm not that familiar with SSA but I can look into that if the need arises.

PR #3792, opened 2023-02-07
rollouts: implement cluster discovery for gcp fleet

ChristopherFry:
This pull request updates the rollouts proof of concept adding cluster discovery for GCP Fleet.


droot (approved):
This is looking good to me.

PR #3798, opened 2023-02-08
rollouts: cache gcp fleet rest config host

ChristopherFry:
This pull request updates the rollouts proof of concept to also cache the REST Config host in addition to project number today when generating a REST Config for a GCP Fleet cluster. This reduces the overall number of requests that need to be sent to Google when fetching a REST Config.


droot (approved):
That's a good optimization.

PR #3809, opened 2023-02-10
rollouts: update root sync watcher to create dynamic client

ChristopherFry:
This pull request updates the root sync watcher to create a new dynamic client every time making a new connection to a target cluster. Creating a new client ensures the client has a valid access token. Previously, the watcher cached the client, leading the watcher to attempt to reconnect to a target cluster with expired access tokens.

### DIFF
--- a/rollouts/controllers/remoterootsync_controller.go
+++ b/rollouts/controllers/remoterootsync_controller.go
@@ -132,12 +132,13 @@ func (r *RemoteRootSyncReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, nil
 	}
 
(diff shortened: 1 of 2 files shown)


droot (approved):
This PR definitely is a step in the right direction. Though, one thing I want to dig deeper is the watch shouldn't be reset and underlying token should be auto-refreshed automatically. I am curious about your result of running this over the weekend.

ChristopherFry:
> This PR definitely is a step in the right direction. Though, one thing I want to dig deeper is the watch shouldn't be reset and underlying token should be auto-refreshed automatically. I am curious about your result of running this over the weekend.

The results are positive. Anytime the watch was reset, the watch was able to successfully reconnect to the target cluster. I am seeing the watches reset every 5 minutes so this is definitely worth looking into.

PR #3813, opened 2023-02-13
rollouts cli: use "default" value for namespace and minor cleanups

droot:
This PR addresses the review comments from the PR https://github.com/GoogleContainerTools/kpt/pull/3807

### DIFF
--- a/commands/alpha/rollouts/advance/advance.go
+++ b/commands/alpha/rollouts/advance/advance.go
@@ -65,7 +65,9 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
 	rolloutName := args[0]
 	waveName := args[1]
 
-	rollout, err := rlc.Get(r.ctx, rolloutName)
+	// TODO(droot): plumb the namespace value from commandline flags
+	ns := "defaut"
+	rollout, err := rlc.Get(r.ctx, ns, rolloutName)
 	if err != nil {
 		fmt.Printf("%s\n", err)
 		return err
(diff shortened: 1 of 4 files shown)


commands/alpha/rollouts/advance/advance.go
```diff
@@ -65,7 +65,9 @@ func (r *runner) runE(cmd *cobra.Command, args []string) error {
 	rolloutName := args[0]
 	waveName := args[1]
 
-	rollout, err := rlc.Get(r.ctx, rolloutName)
+	// TODO(droot): plumb the namespace value from commandline flags
```
mortent, line 68: The `kpt live` code already have functionality for this. Presumably we want to use the default value from the kubeconfig and allowing the user to override this with the `--namespace` flag.

mortent (approved):

mortent:
Can we also update the PR title to be something a bit more descriptive? This will end up in the release notes, but it is not very informative.

droot:
> Can we also update the PR title to be something a bit more descriptive? This will end up in the release notes, but it is not very informative.

Done.

commands/alpha/rollouts/advance/advance.go
droot: Great, we can use whatever existing commands are using.

droot: Added a good first issue as well for this: https://github.com/GoogleContainerTools/kpt/issues/3814

PR #3816, opened 2023-02-14
rollouts: script to generate rollouts kpt package

droot:
Added a script to generate kpt package to deploy `rollouts` component on any k8s cluster. Also added the kpt package under `rollouts/manifests`. I am going to follow up with an installation guide.

This will also sets up for releasing the `rollouts` component.


ChristopherFry (approved):
Looking forward to this! One small nit.

rollouts/manifests/controller/controller.yaml
```diff
@@ -0,0 +1,406 @@
...
+          capabilities:
+            drop:
+            - ALL
+      - args:
+        - --health-probe-bind-address=:8081
+        - --metrics-bind-address=127.0.0.1:8080
+        - --leader-elect
+        command:
+        - /manager
+        image: gcr.io/sunilarora-sandbox/rollouts-controller:v0.0.1
```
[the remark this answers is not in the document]
droot: good catch. fixed it. pointing it to a non-existent image for now `gcr.io/kpt-dev/rollouts-controller:latest`

rollouts/scripts/create-manifests.sh
```diff
@@ -0,0 +1,100 @@
...
+  local NEW="${1}"
+  local TAG="${NEW##*:}"
+  local IMG="${NEW%:*}"
+  kpt fn eval ${DESTINATION}/controller --image set-image:v0.1.1 -- \
+    "name=${OLD}" \
+    "newName=${IMG}" \
+    "newTag=${TAG}"
+}
+
+validate && \
```
sdowell, line 97: nit: the `&&` chaining here shouldn't be necessary since `-e` is set for the script

rollouts/scripts/create-manifests.sh
```diff
@@ -0,0 +1,100 @@
...
+      error "Invalid argument: ${key}"
+    ;;
+  esac
+done
+
+# Defaults
+DESTINATION="manifests"
+
+function validate() {
+  [ -n "${DESTINATION}"       ] || error "--destination is required"
```
sdowell, line 62: nit: bracket spacing

rollouts/manifests/crds/containercluster.yaml
```diff
@@ -0,0 +1,1351 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    cnrm.cloud.google.com/version: 1.100.0
+    kubectl.kubernetes.io/last-applied-configuration: |
```
sdowell, line 20: Is this CRD supposed to have things like annotations (last-applied-configuration), generation, resourceVersion, etc?

rollouts/manifests/crds/containercluster.yaml
```diff
@@ -0,0 +1,1351 @@
+# Copyright 2023 Google LLC
```
sdowell, line 1: I see the script generates `controller/controller.yaml` and `crds/crds.yaml`, but where does this manifest come from?

rollouts/scripts/create-manifests.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# function to generate manifests for deploying rollouts controller
+function generate_controller_manifests {
+    log "generating controller manifests..."
+    kustomize build config/default > ${DESTINATION}/controller/controller.yaml
+}
+
+# update the controller image
+function set_controller_image {
+  local OLD="controller:latest"
```
sdowell, line 87: Should this be `rollouts-controller:latest`?

sdowell (commented):
I don't think I have enough context to do a full review, but just a few passing comments/questions :)

rollouts/scripts/create-manifests.sh
droot: Good pt. fixed it.

rollouts/scripts/create-manifests.sh
droot: Fixed (copy/pasta :))

rollouts/manifests/crds/containercluster.yaml
droot: No, I extracted this from a cluster and forgot to clean it up. Removed runtime info from it.

rollouts/manifests/crds/containercluster.yaml
droot: They are originally coming from kubebuilder generated config under `config` directory. Think of that as the `DRY` or source config and this being a WET package that will be distributed to the end-users.

rollouts/scripts/create-manifests.sh
droot: No, the generated manifests have the image name to be `controller:latest`.

droot (commented):
Thank you @sdowell for the review. Very helpful. Addressed them in the recent commits.

rollouts/manifests/crds/containercluster.yaml
sdowell: I suppose the part I'm confused by is the other manifests in this directory appear to be regenerated from their dry config every time (by `create-manifests.sh`), whereas this one appears to be a static wet config. Is this because there is no "source of truth" dry config for this CRD?

droot: Oh I am sorry, I mis-read your question. 

> Is this because there is no "source of truth" dry config for this CRD?

You are right, there is no SOT dry config for this CRD. This CRD exists on clusters that have KCC installed (for example config-controller clusters). So, I added this for now to complete the package because the package can be installed on any GKE clusters. Eventually, I want to find an elegant solution where this should be an optional installed but has bunch of dependences such as RBAC and also making controller code aware of the presence/absence of the CRDs, but we plan to address further down the line.

rollouts/config/default/kustomization.yaml
```diff
@@ -27,7 +27,7 @@ namePrefix: rollouts-
 #  someName: someValue
 
 bases:
-- ../crd
+# - ../crd
```
mortent, line 30: Maybe just delete the kustomize setup that comes with Kubebuilder. It is kinda complicated, so we can just remove it if we don't need it. The controllers in Porch has a very simple configuration.

rollouts/manifests/README.md
```diff
@@ -0,0 +1,21 @@
+# manifests
```
mortent, line 1: We can probably either update this to have useful information or just delete it.

rollouts/manifests/crds/containercluster.yaml
mortent: Yeah, this is probably ok for now, but we should find a better solution here. This will just install the CRD without the controller. It can also create problems if users try to install this package on a cluster that already have Config Connector installed.

rollouts/manifests/package-context.yaml
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kptfile.kpt.dev
+  annotations:
+    config.kubernetes.io/local-config: "true"
+data:
+  name: example
```
mortent, line 22: Nit: Maybe use a better name here?

rollouts/config/default/kustomization.yaml
droot: I am keeping them around till our APIs becomes stable.  I also found them to be complicated, but just using them for generation purpose for now. Will remove them eventually.

rollouts/manifests/package-context.yaml
droot: Since this is an abstract package, `example` is a placeholder value.

rollouts/manifests/README.md
droot: Ack.

PR #3817, opened 2023-02-15
rollouts: use kubernetes logging library

ChristopherFry:
This pull request updates rollouts to use the Kubernetes logging library, [klog](https://github.com/kubernetes/klog), as the default logger. This is the same logging library that Porch and Config Sync uses.


ChristopherFry:
This will be the first of several pull requests. Essentially I am looking at moving rollouts to use contextual logging fully following https://github.com/kubernetes/community/blob/HEAD/contributors/devel/sig-instrumentation/migration-to-structured-logging.md. We have some of this for free already but part of the changes will be removing any explicit klog logging statements.

droot (approved):

PR #3820, opened 2023-02-16
wip: packagevariant controller: support namespacing target packages

natasha41575:
Not yet ready for review.

Related issue: https://github.com/GoogleContainerTools/kpt/issues/3488

Adds the `set-namespace` function to target downstream packages using the [package-context pattern](https://catalog.kpt.dev/set-namespace/v0.4/set-namespace-kpt-package-context/). This should result in the target packages' resources ending up with the specified namespace set on them.

### DIFF
--- a/porch/controllers/packagevariants/api/v1alpha1/packagevariant_types.go
+++ b/porch/controllers/packagevariants/api/v1alpha1/packagevariant_types.go
@@ -62,6 +62,7 @@ type PackageVariantSpec struct {
 
 	Labels      map[string]string `json:"labels,omitempty"`
 	Annotations map[string]string `json:"annotations,omitempty"`
+	Namespace   *Namespace        `json:"namespace,omitempty"`
 }
 
 type Upstream struct {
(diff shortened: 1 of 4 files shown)


johnbelamaric:
Thanks Natasha. I don't necessarily want to tie this directly to the package context pattern. In fact, I think that pattern is already supported without this change. If the user sets the package name as different from the upstream, and adds set-namespace in the package Kptfile, I think the existing code should already handle this.

The package context pattern for set-namespace is really around packages whose purpose is to provision namespaces - that is, the package name matches the namespace name. The idea of putting namespace in here is for packages that need to be installed in a namespace, we are making namespace and explicit input to the package variant. It's really a special case of general "mutation" that we are building in directly since it is so common.

Along those lines, I have been talking pretty extensively with the Nephio team on what we need from PackageVariant and would like to evolve the API a bit. I will put in a PR for a design proposal that evolves this to the next step, and then once we all agree I'll work on implementation along with whoever is able to contribute.

natasha41575:
Thanks for your feedback @johnbelamaric! If I am interpreting your comment correctly, this PR is not needed and you are planning to submit a design proposal for api changes that you/Nephio need.

In that case, I'll close this PR, focus on getting https://github.com/GoogleContainerTools/kpt/pull/3764 merged, and await your proposal.

johnbelamaric:
Oh, I didn't actually mean that, but I think it's OK to pause on this PR until the proposal is ready and we can debate it.

Issue #3488, opened 2022-08-23
Porch: BaseRevision controller aka Fan Out controller - but more


johnbelamaric:
### Describe your problem

In the [Nephio controller PoC](https://github.com/nephio-project/nephio-controller-poc) I looked at a fan-out controller. We have also discussed this and related issues in #3455, #3387, #3347, #3348 and many other places.

I put together a proposal for a resource and controller that does basic management of the upstream/downstream relationship. Right now it is in gdoc form right now because it's easier to manage comments and collaborate. Please take a look, at some point we can move this to a design proposal.

https://docs.google.com/document/d/1EzUUDxLm5jlEG9d47AQOxA2W6HmSWVjL1zqyIFkqV1I/edit?usp=sharing

natasha41575:
I have a draft of a design, with active discussion here: https://docs.google.com/document/d/1OxNon_1ri4YOqNtEQivBgeRzIPuX9sOyu-nYukjwN1Q/edit?usp=sharing

Feedback is welcome.

justinsb:
The concept behind the proposals sound good.  In terms of functionality that is missing in kpt:

* We have been adding the mechanisms for controllers (watch, etc).  These changes are now merged.
* I have an integration branch that shows how to create Draft packages (it's creating child packages, but we could add functionality / create another controller to update package instances when the upstream changes): https://github.com/GoogleContainerTools/kpt/pull/3575
* I think we probably want to mark these Draft packages as "machine created" in some way.  Labels would be a good start, maybe a structured field.  mortent@ is looking into label support (and more!).

Directionally, I like the idea of storing some sort of metadata around when to propose updates in a CRD.  I suggest we should start with the simple strategy ("propose asap") - there are still plenty of challenges here e.g. what happens when there's a second update before the first is approved.  Then when we have this we can evaluate whether we want to delay proposing updates with a policy CRD.  I could see an alternative world where we generate them all but somehow prioritize them / mark them as blocked (with conditions?).

@natasha41575 if you want to do this, I think you're unblocked from doing this now - I think the klippy PR doesn't have anything left in it that can't be hacked around easily (and I'm whittling it down).  But if there are any blockers, let me know!

natasha41575:
[...]

PR #3827, opened 2023-02-18
PackageVariant design proposal

johnbelamaric:
This documents the final design decisions from all the work on #3488.


johnbelamaric:
cc @ChristopherFry @natasha41575 @mortent @aravind254

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,234 @@
...
+
+When deploying workloads across large fleets of clusters, it is often necessary
+to modify the workload configuration for a specific cluster. Additionally, those
+workloads may evolve over time with security or other patches that require
+updates. [Configuration as Data](06-config-as-data.md) in general and [Package
+Orchestration](07-package-orchestration.md) in particular can assist in this.
+However, they are still centered around manual, one-by-one hydration and
+configuration of a workload.
+
+This proposal introduces concepts and a set of resources that for automating the
```
natasha41575, line 16:
```suggestion
This proposal introduces concepts and a set of resources for automating the
```

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,234 @@
...
+during the creation of the `PackageRevision`. This can assist in additional
+automation workflows.
+
+## Introducing Variance
+Just cloning is not that interesting, so the `PackageVariant` resource also
+allows you to control various ways of mutating the original package to create
+the variant.
+
+### Package Context
+[Note: I believe right now we *always* produce `--for-deployment` package
```
natasha41575, line 105: So the PackageVariant controller doesn't actually do this right now. It just does a clone, without adding or changing any existing `kptfile.kpt.dev` ConfigMap. Whether it _should_ is an interesting idea. It might be useful to have a similar `for-deployment` option in the PackageVariant, since it would keep this controller aligned with the variant constructor pattern that we have with kpt.

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,234 @@
...
+controller as well. Additionally, the author of the `PackageVariant` resource
+can specify additional key-value pairs to insert into the package context [note:
+this additional functionality is proposed here but not yet implemented].
+
+### KRM Function Calls
+[Note: not yet implemented]
+TODO(johnbelamaric): describe adding a KRM function pipeline to the
+PackageVariant to allow arbitrary mutations
+- question: should these allow adding to the package pipeline a la `--save`? Is
+  that a *separate* pipeline?
```
natasha41575, line 125: I like the idea of adding a KRM function pipeline to be added to the PackageVariant, and having that saved to the Kptfile. But there are two questions for me:
1. Does adding it to the package pipeline need to be an option, or should we always add it (or at least add it by default)? Would it be common to run these functions as a one-off thing and not save it to the Kptfile? If they add more resources to the Kptfile later, I would imagine that they'd want to rerun all the same functions again, which is what the Kptfile pipeline helps with.
2. Could you clarify what you mean by a "separate" pipeline? Not sure I fully understand the advantages of having it separate from the package pipeline

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,234 @@
...
+very common operation. However, since namespace provisioning in a cluster is a
+privileged operation, the deployer of a package may not have the authority to
+provision a namespace. For this reason, upstream packages should not directly
+include `Namespace` resources if they want to be truly reusable.
+
+So, the package variant controller provides some convenience features for
+targeting the package at a particular namespace. First, it is possible to use
+the KRM function capability to call the well-known `set-namespace` function.
+For convenience, you can alternatively set the `namespace.value` field in the
+`PackageVariant` resource, and this will be done for you automatically.
```
natasha41575, line 181: I think with the KRM functions capability, adding the `namespace.value` field in the `PackageVariant` resource is not necessary. I feel it is best to have a singular way to achieve this task (to avoid user confusion), and the `set-namespace` function seems convenient enough.

docs/design-docs/08-package-variant.md
johnbelamaric: Ok, good to know. Yeah, I think we probably need a "forDeployment" flag.

docs/design-docs/08-package-variant.md
johnbelamaric: 1. I am not sure we know yet. In some ways `PackageVariant` is trying to emulate what a user might do, which could include imperative calls to KRM functions. That's where this question comes from.
2. We have talked in the past that there are cases (like managing app config), where sets of functions may need to be run at different times or for different operations. So this question is in that context. I am not sure of the answer. 

For both of these, we may want to just pick the simplest option and see if it ends up being sufficient or not.

johnbelamaric: One possible use would be for functions that require network or volume access. For example, we may want to add the ability for `PackageVariant` to render a Helm chart in order to create the downstream package, using the `render-helm-chart` function. But that function cannot be part of the `kpt` pipeline as it is not hermetic and/or requires volume mounts:

https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/functions/go/render-helm-chart

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,234 @@
...
+Similarly, `PackageVariant` can manage the process of updating a package when a
+new version of the upstream package is published. In the human-centered
+workflow, a user would use `kpt pkg update` to pull in changes to their
+derivative package. When using a `PackageVariant` resource, the change would be
+made to the upstream specification in the resource, and the controller would
+propose a new Draft package reflecting the outcome of `kpt pkg update`.
+
+By automating this process, we open up the possibility of performing systematic
+changes that tie back to our different dimensions of scalability. For example,
+we can use data about the specific variant we are creating to lookup additional
```
mortent, line 57: I like thinking about this as a continuous reconciliation where the inputs are a parent package `P` and a context `C` and the output being the derived package `D`. And every time either the version of the upstream package or the context changes, we will get a new revision of the derived package that has the provided revision of the upstream package as upstream and has been customized according to the context.

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,234 @@
...
+creation and lifecycle management of package variants. These are designed to
+address several different dimensions of scalability:
+- Number of different workloads for a given cluster
+- Number of clusters across which those workloads are deployed
+- Complexity of the organizations deploying those workloads (NOTE: actually this
+  is more the `conditions` stuff, will probably move this out of here and create
+  a separate proposal for that)
+- Changes to those workloads over time
+
+## See Also
```
mortent, line 26: I think this is largely an implementation of the bulk package creation (https://github.com/GoogleContainerTools/kpt/issues/3347) and bulk package updates (https://github.com/GoogleContainerTools/kpt/issues/3243).

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,234 @@
...
+revisions out of package variant controller. We will likely need to allows some
+control over that, as we already have a use case in Nephio to use a
+PackageVariant to create a package intended as a base for fan out]
+
+Every kpt package that is designated `--for-deployment` will contain a
+ConfigMap called `kptfile.kpt.dev`. Kpt (or Porch) will automatically add a
+key `name` to the ConfigMap data, with the value of the package name. This
+ConfigMap can then be used as input to functions in the Kpt function pipeline.
+
+This process holds true for package revisions created via the package variant
```
mortent, line 115: I think this could be particularly relevant when combined with the package pipeline and the variant constructor pattern: https://kpt.dev/guides/variant-constructor-pattern.

We have support in PackageVariantSet for creating packages based on other objects referenced by the PackageVariantSet CR. These objects can then provide the context for each variant needed and therefore will be included as parameters in the PackageVariant CRs created by the PackageVariantSet controller. The pipeline included in the packages will then know how to use the context to perform the necessary customizations. It should make it possible to encapsulate the information about customizations and the context in only the targeted objects and the packages themselves.

docs/design-docs/08-package-variant.md
mortent: I think of the pipeline included in kpt packages as maintaining certain invariants of the package as it goes through iterations. So it could make sure that all resources will always belong in a particular namespace since it will be automatically run by porch after every update to the package.

I think of a pipeline that is part of the PackageVariant controller as more of one-off updates to the package. So we are kinda automating the manual changes a user might perform on a newly created packagerevision, so basically imperative function invocations. Therefore I'm not sure if I think it makes sense to add the functions to the package pipeline.

docs/design-docs/08-package-variant.md
johnbelamaric: We have injection for that - to insert those target objects and/or other objects. It's described below. But your point is well taken that we may want to make it super convenient to inject the targeting resource.

As for the package context ConfigMap described here, the problem is that it is only created if the downstream repository is flagged as a deployment repository in Porch. That is, the equivalent of the `--for-deployment` flag  value is determined by an attribute of the downstream repo. So, if the downstream repo is not tagged as a deployment repo, I guess we should error out here, if these values have been set?

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,270 @@
...
+instance of the package to lookup a resource in the Porch cluster, and copy that
+information into the package. Of course, the package has to be ready to receive
+this information. So, there is a protocol for facilitating this dance:
+- Packages may contain resources annotated with `kpt.dev/config-injection`
+- Usually, these will also be `config.kubernetes.io/local-config` resources, as
+  they are likely just used by local functions as input. But this is not
+  mandatory.
+- The package variant controller will look for any resource in the Kubernetes
+  cluster matching the Group, Version, and Kind of the package resource, and
+  satisfying the *injection selector*.
```
johnbelamaric, line 146: To @mortent's point above, perhaps we should inject a `local-config` version of the targeting resource, either automatically or with a simple injection selector - that is, without requiring a pre-existing local object in the package.

johnbelamaric: Actually this is not something I want to do. I'd rather require the package advertise the ways it accepts injections (ie, the GVKs it understands), and only insert those. This keeps the separation of concerns cleaner; the package doesn't build in awareness of the context in which it's being deployed. For example, I don't want a package to accept a Porch Repository resource just because that happens to be the targeting mechanism. If it needs a value from, say, an annotation on the repository, then we should provide a mechanism to inject that value. If, on the other hand, it needs a whole resource of some sort, then we need to define the right injection selectors in the PV/PVS so that the controller can find it and inject it.

johnbelamaric: Current thinking is this (will update the proposal soon):

1) There must be an in-package object with config-injection annotation
2) GVK of in-package and in-cluster object must match
3) PV will contain an ordered list of selectors, first match wins
4) At least for now selectors will be an explicit name (ie, PV will look for an object with that name, in the same namespace as the PV, with the same GVK as the in-package object)
5) These object must have a `spec` field, and only the `spec` will be copied from the in-cluster object to the in-package object
6) the in-package object will be annotated with the name of the object that was selected (recall we can have multiple selectors)
7) the in-package object will be annotated with a hash of the contents of the spec field (so that the PV controller can evaluate whether the spec has been changed in either object during later reconciliations, and fix config drift)
8) PVS will add a "nameFrom" or something similar, that will resolve to the explicity "name" fields in the selectors during fan out. This allows the PVS to, for example, set the injector name based upon the target name or another target field, or even an annotation or label defined on the target.
9) maybe someday we will expand these selectors (both injection on PV and derivation of that in PVS) to label selectors, etc., but right now there is no clear use case for that
10) Regardless of whether any injection selectors are defined, the PV controller will create package conditions corresponding to each resource tagged with `config-injection`. If the PV controller successfully injects the config, that condition gets set to `True`, otherwise `False` with an appropriate message.

docs/design-docs/08-package-variant.md
johnbelamaric: `set-namespace` doesn't have a "Create" option, though.

docs/design-docs/08-package-variant.md
johnbelamaric: The selectors should be able to optionally include GVK. This is way to constrain the injections that will be done. That is, if the package has 10 different objects with `config-injection` annotation, the PV could say it only wants to replace certain GVKs. This also allows a little better control. So, for example, if you had in your cluster

GVK1 foo
GVK1 bar
GVK2 foo
GVK2 bar

you could define a list of selectors like:

 - GVK1 foo
 - GVK2 var

without being able to include GVK in the selector, you would always get "foo" for every injection point. with these selectors we get foo for GVK1 and bar for GVK2.

johnbelamaric: We probably want to differentiate the condition of whether injection succeeded or not based on if the GVK of the in-package resource matches a selector. This allows the PV author to say they only want to "fill in" particular injection points, and leave the others alone (for example, to override a default or not).

This begs the question though - should we have a selector for the in-package resources of interest, and then a separate matcher-selector to match each of those to in-cluster objects. That's sounding complicated.

johnbelamaric: Another option: the `config-injection` annotation can be used to flag whether or not failed injection should create a False condition. In package deployment controller, it is just true/false. But we can do better - for example, if the annotation is "required", then a failed injection results in a False condition. If it is "optional", then even if we fail to find a match, we create the condition as True (and just put in the message that it was left as the default).

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,270 @@
...
+- The package variant controller will copy the `spec` field from the matching
+  in-cluster resource to the in-package resource.
+
+Note that because we are injecting data *from the Kubernetes cluster*, we can
+also monitor that data for changes. For each resource we inject, the package
+variant controller will establish a Kubernetes "watch" on the resource. A change
```
johnbelamaric, line 152: Note that we probably want to initiate a watch on the GVK, not the specific issue, to reduce the number of watches. But it's up for debate (see https://github.com/nephio-project/nephio-controller-poc/issues/24 originally).

docs/design-docs/08-package-variant.md
johnbelamaric: We can probably make it an option, like `--save`, and leave it up to the user.

docs/design-docs/08-package-variant.md
johnbelamaric: Probably the best thing to do would be as you suggest, but add a `create` option to `set-namespace`, or have the user use `upsert-resource`, though the UX on that would be painful.

pgnbguedmvs:
Thank you, boss

docs/design-docs/08-package-variant.md
johnbelamaric: Updated to reflect this.

docs/design-docs/08-package-variant.md
johnbelamaric: Updated to reflect this.

docs/design-docs/08-package-variant.md
johnbelamaric: Updated to reflect this.

docs/design-docs/08-package-variant.md
johnbelamaric: Updated to reflect this.

docs/design-docs/08-package-variant.md
johnbelamaric: Updated to reflect this.

docs/design-docs/08-package-variant.md
johnbelamaric: Updated to reflect this.

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,450 @@
...
+
+ - GVK1 foo
+ - GVK2 bar
+
+That said, often name will be sufficiently unique when combined with the
+in-package resource GVK, and so making the selector GVK optional is more
+convenient.
+
+#### Order of Mutation During Creation
+Note that the KRM function pipeline defined in the Kptfile runs every time Porch
```
johnbelamaric, line 383: @natasha41575 can you take a look at this? Obviously we haven't implemented this yet, but I am wondering if this is the expected behavior. Would we save the package after each step, or just at the end of the reconciliation? If it's the latter, then this becomes:

1. Package Context Injections
1. PackageVariant KRM Function Pipeline
1. Config Injection
1. Kptfile KRM Function Pipeline

WDYT?

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,450 @@
...
+reasons:
+
+1. An injected in-cluster object is updated.
+1. The PackageVariant resource is updated, which could change any of the options
+   for introducing variance, or could also change the upstream package revision
+   referenced.
+1. A new revision of the upstream package has been selected due to a floating
+   tag change, or due to a force retagging of the upstream.
+
+TODO(johnbelamaric): Figure out and document how the mutation process works in
```
johnbelamaric, line 406: @natasha41575 any thoughts on this?

docs/design-docs/08-package-variant.md
johnbelamaric: I think this might be done by Porch, regardless of the `for-deployment` thing. I think the name replacement doesn't happen unless you do `for-deployment`. But otherwise, I think the actual config map gets created automatically already. @mortent do you know?

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,450 @@
...
+- Downstream
+- Annotations
+- Labels
+- Adoption and Deletion Policy
+
+#### Package Context Injection
+
+#### KRM Function Pipeline
+
+#### Configuration Injection
```
johnbelamaric, line 289: @natasha41575 @mortent this section could use a close eye...

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,450 @@
...
+resource.
+
+Injection selectors are defined in the `spec.injectionSelectors` field of the
+`PackageVariant`. This field is an ordered array of structs containing group,
+version, kind, and name. Only the name is required.
+
+The injection process works as follows:
+
+1. The controller will examine all in-package resources, looking for those with
+   an annotation named `kpt.dev/config-injection`, with one of the following
```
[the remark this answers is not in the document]
johnbelamaric: I prefer to keep it as short as possible, but whatever you/Morten/Sunil/Justin think works.

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,450 @@
...
+     with a message indicating that the schema is missing, and processing should
+     proceed to the next injection point. Note that for `optional` injection
+     points, not having the schema may be intentional and not an error.
+   - If the resource schema of the injection point does not contain a `spec`
+     field, then the injection point ConditionType will be set to `False`, with
+     a message explaining the error, and processing should proceed to the next
+     injection point.
+   - The controller will now identify all in-cluster objects in the same
+     namespace as the PackageVariant resource, with GVK matching the injection
+     point (the in-package resource).
```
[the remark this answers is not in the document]
johnbelamaric: Right, the names will not match. For one, the names in the package are the same across for any PackageVariant using that upstream, but we want to inject different resources for each one. For example, in the package it may be named "cluster-info" but when we instantiate the PackageVariant for "cluster A", we want to inject the resource corresponding to "clusterA". We also do not want to *change* the name in the package, because it likely has meaning within the package and will be used by functions in the package.

Also, different roles that are controlling the names. The names in the package are in the control of the package author. The names in the cluster are in the control of whoever populates the cluster (could be, for example, some infrastructure team). The selector is the glue between them, and *that* is in control of the package author, and will be used in function configs, for example.

The GVK on the other hand, has to be the same, because it tells us the API. The namespace has to be the same, just for authorization purposes.

docs/design-docs/08-package-variant.md
johnbelamaric: Yeah, I think that's correct. I'll update the design.

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,450 @@
...
+#### KRM Function Pipeline
+
+#### Configuration Injection
+
+Configuration injection is controlled by a combination of in-package resources
+with annotations, and *injection selectors* defined on the PackageVariant
+resource.
+
+Injection selectors are defined in the `spec.injectionSelectors` field of the
+`PackageVariant`. This field is an ordered array of structs containing group,
```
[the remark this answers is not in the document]
johnbelamaric: That design will come soon...

docs/design-docs/08-package-variant.md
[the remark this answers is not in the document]
johnbelamaric: I added a bunch of this explanation into the file.

docs/design-docs/08-package-variant.md
johnbelamaric: done

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,533 @@
...
+information into the package. Of course, the package has to be ready to receive
+this information. So, there is a protocol for facilitating this dance:
+- Packages may contain resources annotated with `kpt.dev/config-injection`
+- Usually, these will also be `config.kubernetes.io/local-config` resources, as
+  they are likely just used by local functions as input. But this is not
+  mandatory.
+- The package variant controller will look for any resource in the Kubernetes
+  cluster matching the Group, Version, and Kind of the package resource, and
+  satisfying the *injection selector*.
+- The package variant controller will copy the `spec` field from the matching
```
johnbelamaric, line 189: I am not 100% convinced that copying *only* `spec` is the right thing to do. We for sure do not want to copy `metadata` (we don't want to overwrite in-package `metadata`). But everything else is game. For example, if we want to inject a resource representing a cluster that is *discovered*, might that information be in the `status`? Or, if we define a resource that is not used to actuate anything, but really just contains information, does that really belong in a `spec` field?

johnbelamaric: On the other hand I don't want to end up with a bunch of status cruft. Hmm. Maybe we see how this goes for a while, and later if we want to say, add a list of fields to copy, we can do that.

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,551 @@
...
+   an annotation named `kpt.dev/config-injection`, with one of the following
+   values: `required` or `optional`. We will call these "injection points".
+1. For each injection point, a condition will be created in the
+   downstream PackageRevision, with ConditionType set to the dot-delimited
+   concatenation of `config.injection`, with the in-package resource kind and
+   name, and the value set to `False`. Note that since the package author
+   controls the name of the resource, kind and name are sufficient to
+   disambiguate the injection point. We will call this ConditionType the
+   "injection point ConditionType".
+1. For each required injection point, the injection point ConditionType will be
```
mortent, line 408: So if I understand it correctly, it is the responsibility of the package author to add the required injection points to the PackageRevision. I think it might be useful to clarify it here.

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,551 @@
...
+     be set to `False` with a message that no matching in-cluster resource was
+     found, and processing proceeds to the next injection point.
+   - If a matching in-cluster object is selected, then it is injected as
+     follows:
+     - The `spec` field from the in-cluster resource is copied to the `spec`
+       field of the in-package resource (the injection point), overwriting it.
+     - An annotation with name `kpt.dev/injected-resource-name` and value set to
+       the name of the in-cluster resource is added (or overwritten) in the
+       in-package resource.
+1. The PackageVariant resource must have a condition of type `ConfigInjected`.
```
mortent, line 445: The language here doesn't make it clear if the package must always have this condition set, or if it is added by automation. It might be useful to clarify it.

docs/design-docs/08-package-variant.md
```diff
@@ -0,0 +1,551 @@
...
+     field, then the injection point ConditionType will be set to `False`, with
+     a message explaining the error, and processing should proceed to the next
+     injection point.
+   - The controller will now identify all in-cluster objects in the same
+     namespace as the PackageVariant resource, with GVK matching the injection
+     point (the in-package resource).
+   - The controller will look through the list of injection selectors in
+     order and checking if any of the in-cluster objects match the selector. If
+     so, that in-cluster object is selected, and processing of the list of
+     injection selectors stops. Note that in the initial version, the selectors
```
mortent, line 432: So just to clarify here, if we have only one entry in the injector selector list and it only has the name specified, that will apply to all injection points specified in the package right?

And if we add information about GVK in the injector selector list entry, it will only apply if the GVK in the selector matches the GVK of the injection point?

Finally, if there is an injection point, but even after trying all the selectors in the list, we can't find a matching resource in the cluster, then the injection will not happen? And that will be true, even if we can find a unique match based only on GVK?

docs/design-docs/08-package-variant.md
mortent: I agree with the description from @natasha41575 of what should happen here. I think in all three situations, we want to update the package. In practice this means updating a draft if one exists, or create a new draft from the latest published revision if no draft exists.

docs/design-docs/08-package-variant.md
mortent: I agree that spec seems to be the only data that should be copied, assuming that resources follow the convention of using only `metadata`, `spec`, and `status`.

docs/design-docs/08-package-variant.md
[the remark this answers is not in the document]
mortent: For the finalizers, we have just a convention now to reuse the group of the resource type managed by the controller. I think we can use the short form for well-known annotations/labels. Hopefully the number of those will be relatively small.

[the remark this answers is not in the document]
mortent: At least I think this should apply to annotations/labels that are used for general mechanisms in Porch. If it is something that is specific to a controller (like PackageVariantSet keeping track of PackageVariants), we probably want to use something more specific to the controller.

docs/design-docs/08-package-variant.md
mortent: I had a few questions, but overall I feel it is pretty well described. It is a somewhat complicated mechanism, so the examples are very helpful.

docs/design-docs/08-package-variant.md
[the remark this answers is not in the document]
natasha41575: Thanks! The examples are really helpful

docs/design-docs/08-package-variant.md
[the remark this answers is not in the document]
johnbelamaric: So `kpt.dev` or `porch.kpt.dev`?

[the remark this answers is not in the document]
mortent: I think I prefer just `kpt.dev`. Including the name of the product (in this case Porch) in the group doesn't seem necessary and it will just be confusing if we want to use the annotation outside of Porch or if we at some point change the name. It is also shorter.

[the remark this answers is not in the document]
johnbelamaric: ok, will do

docs/design-docs/08-package-variant.md
johnbelamaric: I added some clarifying language, let me know if it's still an issue (after I push).

docs/design-docs/08-package-variant.md
johnbelamaric: I am adding some clarifying language.

docs/design-docs/08-package-variant.md
```diff
@@ -338,20 +337,91 @@ With PackageVariant, we instead create the following resources:
...
+
 #### KRM Function Pipeline
 
-#### Configuration Injection
+PackageVariant resource creators may specify a list of KRM functions to either
+execute on package creation and update, add to the package Kptfile, or both. The
+list of functions to execute as part of package creation or update is listed in
+`spec.execPipeline`, whereas the list of functions to ensure exist in the
+downstream Kptfile pipeline are listed in `spec.prefixPipeline` and
+`spec.suffixPipeline`. These use the Kptfile
```
johnbelamaric, line 398: @mortent @natasha41575 This is probably way overkill and it's kinda hard. I am thinking we just add `execPipeline` for now. This means you couldn't mutate the contents of the upstream pipeline in package variant. I think that's probably OK. We may need a different primitive for that (something like mixins, perhaps).

Similarly, I am also thinking about how much flexibility I want to support along the lines of `fromField` when generating PVs from targets in PVSs. The Function type used in Pipeline is pretty complex; I don't want to have to allow every possible field to be derived fromField in there. I expect we may need the function inputs to vary. I guess I can think about that when I get to PVS design. Maybe we can get away without that.

My guess is we should do the very simplest thing first, and then see what use cases arise. The primary, initial use case for this is just to call `set-namespace` and maybe the occasional `set-labels` or `upsert-resource`.

johnbelamaric:
@droot @mortent this is getting pretty long, would it be useful to merge it once the existing open comments are resolved, then do more as additions? In K8s we set KEPs as `provisional` while we're still working on them...wdyt about something like that?

docs/design-docs/08-package-variant.md
johnbelamaric: I covered this in the detailed design section for the package context feature.

docs/design-docs/08-package-variant.md
johnbelamaric: Ok, simplified this.

docs/design-docs/08-package-variant.md
johnbelamaric: Ok, changing the way the KRM function pipeline works, to just modify the Kptfile pipeline, made this way simpler. Updated.

johnbelamaric:
I have resolved the open comments so far, and have marked this as a "Work-in-Progress" design, so we can merge it. It's getting rather long.

mortent (approved):

PR #3831, opened 2023-02-20
Add delay for TestFunctionRepository to make sure the repo is cached

mortent:


mortent:
@justinsb @droot This is a bit of an ugly hack, but I think it should help with the flaky tests until we can switch off using the aggregated apiserver for functions and use the CRD.

droot (approved):

droot:
> @justinsb @droot This is a bit of an ugly hack, but I think it should help with the flaky tests until we can switch off using the aggregated apiserver for functions and use the CRD.

Sounds good. Thanks for root causing it.

yuwenma:
cc @ziyue-101 
This PR will  help on the healthcheck CI failure.

yuwenma (approved):

PR #3832, opened 2023-02-21
add labels and annotations to rollouts api

natasha41575:
Should resolve https://github.com/GoogleContainerTools/kpt/issues/3797. 

This PR adds labels and annotations to the rollouts api under `syncTemplate`. 

For example:

```
apiVersion: gitops.kpt.dev/v1alpha1
kind: Rollout
metadata:
  name: sample
spec:
  syncTemplate:
    rootSync:
      metadata:
        annotations:
          foo: bar
        labels:
          foo: bar
    type: RootSync
...
```

The specified annotations and labels get added to spec of `RemoteRootSync` objects that the rollout creates. The `RemoteRootSync` objects add them to the metadata of `RootSync` objects that they create.


rollouts/api/v1alpha1/remoterootsync_types.go
```diff
@@ -30,6 +30,7 @@ type RemoteRootSyncSpec struct {
 
 	ClusterRef ClusterRef    `json:"clusterRef,omitempty"`
 	Template   *RootSyncInfo `json:"template,omitempty"`
+	Metadata   *Metadata     `json:"metadata,omitempty"`
```
droot, line 33: why not make `Metadata` part of the RootSyncInfo ?

rollouts/controllers/rollout_controller.go
```diff
@@ -510,11 +510,18 @@ func (r *RolloutReconciler) rolloutTargets(ctx context.Context, rollout *gitopsv
 
 	for _, target := range targets.ToBeCreated {
 		rootSyncSpec := toRootSyncSpec(target.packageRef)
+		var metadata *gitopsv1alpha1.Metadata
+
+		if rollout != nil && rollout.Spec.SyncTemplate != nil &&
+			rollout.Spec.SyncTemplate.RootSync != nil {
+			metadata = rollout.Spec.SyncTemplate.RootSync.Metadata
+		}
```
droot, line 518: If I understand it correctly, if one updates the `metadata` part of the syncTemplate, metadata will only be applied when creating new RootSync.

We should support the `Update` scenario as well. Somewhere there is a check where we decide target to be created or updated and it is on the basis of package `revision` field, we can extend it to compare the metadata part as well.

rollouts/controllers/rollout_controller.go
```diff
@@ -510,11 +510,18 @@ func (r *RolloutReconciler) rolloutTargets(ctx context.Context, rollout *gitopsv
 
 	for _, target := range targets.ToBeCreated {
 		rootSyncSpec := toRootSyncSpec(target.packageRef)
+		var metadata *gitopsv1alpha1.Metadata
+
+		if rollout != nil && rollout.Spec.SyncTemplate != nil &&
```
droot, line 515: nit: `rollout != nil` check should be moved at the beginning of the function.

One more thing I noticed since we do not support `RepoSync` yet, can we throw an error in the reconcileRollout function (I think we have some other validation bits as well).

One more ask: can we also create an issue to support `RepoSync` and Would suggest that to pick that after this.

droot (commented):
Awesome, this is looking good. Some minor comments.

rollouts/controllers/rollout_controller.go
natasha41575: Filed https://github.com/GoogleContainerTools/kpt/issues/3838

rollouts/controllers/rollout_controller.go
natasha41575: Done.

rollouts/controllers/rollout_controller.go
natasha41575: Also moved the check, and added a check to `reconcileRollout` for `RepoSync`.

rollouts/controllers/rollout_controller.go
```diff
@@ -276,6 +277,14 @@ func (r *RolloutReconciler) getPackageDiscoveryClient(rolloutNamespacedName type
 func (r *RolloutReconciler) reconcileRollout(ctx context.Context, rollout *gitopsv1alpha1.Rollout, strategy *gitopsv1alpha1.ProgressiveRolloutStrategy, packageDiscoveryClient *packagediscovery.PackageDiscovery) error {
 	logger := klog.FromContext(ctx)
 
+	if rollout != nil && rollout.Spec.SyncTemplate != nil {
+		if rollout.Spec.SyncTemplate.Type == gitopsv1alpha1.TemplateTypeRepoSync {
+			err := fmt.Errorf("reposync is not yet supported")
```
droot, line 282: just a note for us: we will have to start capturing these as errors as conditions in the status part of the API to surface it to the end users. Today, one has to look at the error logs of the controller :)

droot: Filed https://github.com/GoogleContainerTools/kpt/issues/3839

droot (approved):
Excellent. We can merge this, found a few more things that we need to address.

droot (approved):
Awesome! 🚢

PR #3834, opened 2023-02-22
rollouts: add condition when external sync is created

ChristopherFry:
This pull request updates the remote root sync to add the "ExternalSyncCreated" condition when an external sync is created on the target cluster. Beyond providing a higher-level status to the user, adding this condition allows us to improve the finalization logic to delete the external sync if one was created.

### DIFF
--- a/rollouts/controllers/remoterootsync_controller.go
+++ b/rollouts/controllers/remoterootsync_controller.go
@@ -20,9 +20,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"sync"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


ChristopherFry:
> Beyond providing a higher-level status to the user, adding this condition allows us to improve the finalization logic to delete the external sync if one was created.

Just to expand on this a little, this was the primary reason I am adding this condition. Today the remote root sync controller will throw an error and not allow a remote root sync to be deleted unless the logic is able to successfully the delete the external sync resource. This includes if the external sync resource was never created. This isn't necessarily as simple as adding a `apierrors.IsNotFound` check. The reason for this is the external sync resource may not be able to be created on the external cluster due to connectivity issues, and this issue will be seen in today's finalizer logic as well.

droot (approved):

PR #3845, opened 2023-02-23
rollouts: add condition when external sync is up to date

ChristopherFry:
This pull request adds a new "ExternalSyncUpToDate" condition the remote root sync resource. This condition is set to true when the external sync is up to date with its latest spec as defined by the remote root sync resource. If the external sync is not up to date (reasons can include not being able to connect to the target cluster or not having the neccessary access on the cluster), the condition is updated with the error message.

### DIFF
--- a/rollouts/controllers/remoterootsync_controller.go
+++ b/rollouts/controllers/remoterootsync_controller.go
@@ -63,7 +63,8 @@ var (
 )
 
 const (
-	externalSyncCreatedConditionType = "ExternalSyncCreated"
+	externalSyncCreatedConditionType  = "ExternalSyncCreated"
+	externalSyncUpToDateConditionType = "ExternalSyncUpToDate"
 )
 
 // RemoteRootSyncReconciler reconciles a RemoteRootSync object
@@ -125,6 +126,8 @@ func (r *RemoteRootSyncReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 				// Delete the external sync resource
 				err := r.deleteExternalResources(ctx, &remoterootsync)
 				if err != nil && !apierrors.IsNotFound(err) {
+					r.updateStatus(ctx, &remoterootsync, "", err)
+
 					// if fail to delete the external dependency here, return with error
 					// so that it can be retried
 					return ctrl.Result{}, fmt.Errorf("have problem to delete external resource: %w", err)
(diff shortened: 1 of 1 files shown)


rollouts/controllers/remoterootsync_controller.go
```diff
@@ -125,6 +126,8 @@ func (r *RemoteRootSyncReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 				// Delete the external sync resource
 				err := r.deleteExternalResources(ctx, &remoterootsync)
 				if err != nil && !apierrors.IsNotFound(err) {
+					r.updateStatus(ctx, &remoterootsync, "", err)
```
mortent, line 129: Maybe log the error here if this fails. Not much else we can do at this point, but it would at least give some hint somewhere that something failed.

rollouts/controllers/remoterootsync_controller.go
```diff
@@ -145,42 +148,63 @@ func (r *RemoteRootSyncReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	// Don't update if there are no changes.
-
 	rrsPrior := rrs.DeepCopy()
 
-	rrs.Status.SyncStatus = syncStatus
-	rrs.Status.ObservedGeneration = rrs.Generation
+	if syncError == nil {
+		rrs.Status.SyncStatus = syncStatus
+
+		meta.SetStatusCondition(&rrs.Status.Conditions, metav1.Condition{Type: externalSyncCreatedConditionType, Status: metav1.ConditionTrue, Reason: "SyncCreated"})
```
mortent, line 192: Do these functions automatically handle the LastUpdatedTimestamp?

rollouts/controllers/remoterootsync_controller.go
```diff
@@ -63,7 +63,8 @@ var (
 )
 
 const (
-	externalSyncCreatedConditionType = "ExternalSyncCreated"
+	externalSyncCreatedConditionType  = "ExternalSyncCreated"
```
mortent, line 66: What is the difference between these? It would assume that it being up-to-date would imply that it is also created?

Also, have we thought if we can adopt some of the more common condition types as described in https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus. This might also be relevant: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/308

rollouts/controllers/remoterootsync_controller.go
ChristopherFry: Confirming that these functions do handle the LastTransitionTime field. Reference, https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go.

rollouts/controllers/remoterootsync_controller.go
ChristopherFry: Can do. Will push a commit.

rollouts/controllers/remoterootsync_controller.go
ChristopherFry: > What is the difference between these? It would assume that it being up-to-date would imply that it is also created?

`ExternalSyncCreated` indicates the sync was created on the target cluster and `ExternalSyncUpToDate` indicates the sync is up to date with the latest configuration of the remote root sync. This does similar, but the difference is after you create a rollout resource to rollout v1 of a package and the rollout completes, the rollout resource is updated to rollout the new v2 version of the package, and if, for whatever reason, the reconciler is unable to update the sync on the target cluster (say the cluster is on-premise and there are temporary connectivity issues), the `ExternalSyncCreated` will remain true since the external sync was created and `ExternalSyncUpToDate` would be set to a non-True value since the connection cannot be made to the external sync to update it from v1 of the package to v2.

> Also, have we thought if we can adopt some of the more common condition types as described in https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus. This might also be relevant: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/308

Looking into this. My initial thought is to avoid the Ready condition since we don't have the correct information yet to determine if all sub-resources (including any resources that the sync creates) is in ready status. Similarly with Reconciling. The Stalled condition I am thinking can be a good candidate. I'll likely reach out to discuss more.

ChristopherFry:
Closing this pull request in favor of #3849.

PR #3848, opened 2023-02-23
Rename the api group name to be healthcheck.config.k8s.io

### LINKED ISSUE #3800: Rename the health check API group to something more OSS.
The healthcheck CRD uses `healthcheck.cloud.google.com` as its API group. We should think of a better group name for it as it is in OSS.

ziyue-101:
This PR renames the health check api group to be healthcheck.config.k8s.io, which is more OSS compared to the old one.

Fixes #3800

### DIFF
--- a/healthcheck/api/v1alpha1/groupversion_info.go
+++ b/healthcheck/api/v1alpha1/groupversion_info.go
@@ -14,7 +14,7 @@
 
 // Package v1alpha1 contains API Schema definitions for the healthcheck v1alpha1 API group
 // +kubebuilder:object:generate=true
-// +groupName=healthcheck.cloud.google.com
+// +groupName=config.healthcheck.kpt.dev
 package v1alpha1
 
 import (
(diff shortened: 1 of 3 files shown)


yuwenma:
I assume `config.k8s.io` is reserved for core k/k resources. Since the healthcheck is under kpt repo now, maybe you can consider using `kpt.dev`, like `config.healthcheck.kpt.dev`? How do you plan to suggest HealthCheck to the kpt users?  

CC @droot @mortent @justinsb  for inputs

ziyue-101:
The group is not reserved by kubernetes, at least from this doc. :-) https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#-strong-api-groups-strong-

I happy to change the group name to be more appropriate. waiting for advice;-)

mortent:
We have used `config.k8s.io` in the past, but I think we prefer to use something under the `kpt.dev` domain.

ziyue-101:
Thanks for the context. Changed to config.healthcheck.kpt.dev.

yuwenma (approved):
looks good. Thanks for the change.

PR #3849, opened 2023-02-24
rollouts: add status conditions to remote root sync

ChristopherFry:
This pull request adds the [kstatus](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus#conditions) Reconciling and Stalled conditions status conditions to the Remote Root Sync.

### DIFF
--- a/rollouts/controllers/remoterootsync_controller.go
+++ b/rollouts/controllers/remoterootsync_controller.go
@@ -126,7 +126,7 @@ func (r *RemoteRootSyncReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		// The object is being deleted
 		if controllerutil.ContainsFinalizer(&remoterootsync, myFinalizerName) {
 			// our finalizer is present, so lets handle any external dependency
-			if r.isExternalSyncCreated(&remoterootsync) {
+			if remoterootsync.Status.SyncCreated {
 				// Delete the external sync resource
 				err := r.deleteExternalResources(ctx, &remoterootsync)
 				if err != nil && !apierrors.IsNotFound(err) {
@@ -197,6 +197,7 @@ func (r *RemoteRootSyncReconciler) updateStatus(ctx context.Context, rrs *gitops
 
 	if syncError == nil {
 		rrs.Status.SyncStatus = syncStatus
+		rrs.Status.SyncCreated = true
 
 		meta.RemoveStatusCondition(conditions, conditionReconciling)
 		meta.RemoveStatusCondition(conditions, conditionStalled)
@@ -205,7 +206,7 @@ func (r *RemoteRootSyncReconciler) updateStatus(ctx context.Context, rrs *gitops
 
 		rrs.Status.SyncStatus = "Unknown"
 
-		if !r.isExternalSyncCreated(rrs) {
+		if !rrs.Status.SyncCreated {
 			rrs.Status.SyncStatus = ""
 			reconcileReason = reasonCreateSync
 		}
@@ -365,16 +366,6 @@ func (r *RemoteRootSyncReconciler) getDynamicClientForCluster(ctx context.Contex
 	}
 
 	return dynamicClient, nil
-}
-
-func (r *RemoteRootSyncReconciler) isExternalSyncCreated(rrs *gitopsv1alpha1.RemoteRootSync) bool {
-	reconcilingCondition := meta.FindStatusCondition(rrs.Status.Conditions, conditionReconciling)
-
-	if rrs.Status.ObservedGeneration == 0 || (reconcilingCondition != nil && reconcilingCondition.Reason == reasonCreateSync) {
-		return false
-	}
-
-	return true
 }
 
 // SetupWithManager sets up the controller with the Manager.
--- a/rollouts/controllers/remoterootsync_controller.go
+++ b/rollouts/controllers/remoterootsync_controller.go
@@ -63,7 +63,11 @@ var (
 )
 
 const (
-	externalSyncCreatedConditionType = "ExternalSyncCreated"
+	conditionReady       = "Ready"
+	conditionReconciling = "Reconciling"
+	conditionStalled     = "Stalled"
+
+	reasonSyncNotCreated = "SyncNotCreated"
 )
 
 // RemoteRootSyncReconciler reconciles a RemoteRootSync object
@@ -121,10 +125,16 @@ func (r *RemoteRootSyncReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		// The object is being deleted
 		if controllerutil.ContainsFinalizer(&remoterootsync, myFinalizerName) {
 			// our finalizer is present, so lets handle any external dependency
-			if meta.IsStatusConditionTrue(remoterootsync.Status.Conditions, externalSyncCreatedConditionType) {
+			if r.isExternalSyncCreated(&remoterootsync) {
 				// Delete the external sync resource
 				err := r.deleteExternalResources(ctx, &remoterootsync)
 				if err != nil && !apierrors.IsNotFound(err) {
+					statusError := r.updateStatus(ctx, &remoterootsync, "", err)
+
+					if statusError != nil {
+						logger.Error(statusError, "Failed to update status")
+					}
+
 					// if fail to delete the external dependency here, return with error
 					// so that it can be retried
 					return ctrl.Result{}, fmt.Errorf("have problem to delete external resource: %w", err)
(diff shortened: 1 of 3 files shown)


rollouts/controllers/remoterootsync_controller.go
```diff
@@ -145,42 +155,70 @@ func (r *RemoteRootSyncReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	rrs.Status.ObservedGeneration = rrs.Generation
+	if syncError == nil {
+		rrs.Status.SyncStatus = syncStatus
 
-	meta.SetStatusCondition(&rrs.Status.Conditions, metav1.Condition{Type: externalSyncCreatedConditionType, Status: metav1.ConditionTrue, Reason: "SyncCreated"})
+		meta.SetStatusCondition(conditions, metav1.Condition{Type: conditionReady, Status: metav1.ConditionTrue, Reason: "Ready"})
+		meta.RemoveStatusCondition(conditions, conditionReconciling)
+		meta.RemoveStatusCondition(conditions, conditionStalled)
+	} else {
+		readyReason := "PendingReconcilation"
```
natasha41575, line 204: nit: I suggest putting all the "reasons" into their own consts (so `Ready`, `PendingReconcilation`, and `SyncNotCreated`. It seems a bit odd to me to have a const for `SyncNotCreated` but not the others.

rollouts/controllers/remoterootsync_controller.go
```diff
@@ -333,6 +371,16 @@ func (r *RemoteRootSyncReconciler) getDynamicClientForCluster(ctx context.Contex
 	return dynamicClient, nil
 }
 
+func (r *RemoteRootSyncReconciler) isExternalSyncCreated(rrs *gitopsv1alpha1.RemoteRootSync) bool {
+	readyCondition := meta.FindStatusCondition(rrs.Status.Conditions, conditionReady)
+
+	if readyCondition == nil || (readyCondition.Status != metav1.ConditionTrue && readyCondition.Reason == "SyncNotCreated") {
```
natasha41575, line 377:
```suggestion
	if readyCondition == nil || (readyCondition.Status != metav1.ConditionTrue && readyCondition.Reason == reasonSyncNotCreated) {
```

natasha41575 (commented):
nits

rollouts/controllers/remoterootsync_controller.go
ChristopherFry: Done.

rollouts/controllers/remoterootsync_controller.go
ChristopherFry: Updated.

rollouts/controllers/remoterootsync_controller.go
```diff
@@ -63,7 +63,15 @@ var (
 )
 
 const (
-	externalSyncCreatedConditionType = "ExternalSyncCreated"
+	conditionReady       = "Ready"
+	conditionReconciling = "Reconciling"
```
mortent, line 67: Using both the `Ready` and `Reconciling` conditions seems redundant to me. Are there any situations where the value of the `Ready` condition can not be derived from looking at the `Stalled` and `Reconciling` conditions?

rollouts/controllers/remoterootsync_controller.go
```diff
@@ -121,10 +129,16 @@ func (r *RemoteRootSyncReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		// The object is being deleted
 		if controllerutil.ContainsFinalizer(&remoterootsync, myFinalizerName) {
 			// our finalizer is present, so lets handle any external dependency
-			if meta.IsStatusConditionTrue(remoterootsync.Status.Conditions, externalSyncCreatedConditionType) {
+			if r.isExternalSyncCreated(&remoterootsync) {
```
mortent, line 132: Do we need to make this check here? I would expect us to just make the call to clean up external resources regardless, and it will just check that any external resources are not there. Whenever possible, we should just look at the actual state and determine what do to, unless it is prohibitively expensive.

ChristopherFry: This check was introduced to allow us to recover, as in allowing a remote root sync to be successfully deleted, for a misconfigured external cluster. For instance, if RBAC was not setup correctly on the external cluster, the creation of the root sync will fail to create, but when the remote root sync is to be deleted (rollout resource itself is updated and or removed), the logic to check for the delete the external root sync will fail as well due to the same permissions issue.

rollouts/controllers/remoterootsync_controller.go
ChristopherFry: Correct, the `Ready` and `Reconciling` conditions are redudant. Originally I thought it was recommended to have the `Ready` condition, but re-reading https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus#conditions I see the `Ready` condition is not necessary if the the `Reconciling` and `Stalled` conditions are adopted. I can remove the `Ready` condition.

ChristopherFry: The `Ready` condition is removed in the latest commit.

rollouts/controllers/remoterootsync_controller.go
droot: A note for us: Realized one more edge case where the target cluster never comes back up or is decommissioned without proper cleanup (removing from the targets in the Rollout), the rrs will continue to be hanging in that state. One way to handle that will be to remove all RRS for a cluster when we a cluster is decommissioned (deleting a membership probably).

rollouts/controllers/remoterootsync_controller.go
```diff
@@ -145,42 +156,65 @@ func (r *RemoteRootSyncReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		if !r.isExternalSyncCreated(rrs) {
+			rrs.Status.SyncStatus = ""
+			reconcileReason = reasonCreateSync
+		}
+
+		meta.SetStatusCondition(conditions, metav1.Condition{Type: conditionReconciling, Status: metav1.ConditionTrue, Reason: reconcileReason})
+		meta.SetStatusCondition(conditions, metav1.Condition{Type: conditionStalled, Status: metav1.ConditionTrue, Reason: reasonError, Message: syncError.Error()})
+	}
+
+	rrs.Status.ObservedGeneration = rrs.Generation
```
droot, line 217: I wonder if we can simplify (combining `syncStatus` with other conditions). So, a RRS can be in four states: `{Created, Reconciling, Stalled, Ready}` and state can transition between these four or am I thinking in the wrong way ?

rollouts/controllers/remoterootsync_controller.go
ChristopherFry: +1 Confirming this check doesn't solve everything but does solve any immediate misconfiguration issues. The ones I came across while developing rollouts were mainly connection, authentication, and authorization. Either the cluster not existing or not having the correct credentials/permissions on the target cluster. The setup for GCPFleet can be involved, more so with the Connect Agent, so it's easy to miss a step.

rollouts/controllers/remoterootsync_controller.go
ChristopherFry: My knowledge is still evolving here, but the `Reconciling` and `Stalled` conditions are from [kstatus conditions](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus#conditions) so these do have specific definitions, namely if the Rollout controller has reconciled its changes to the resources it directly owns. In this case, it is if the RootSync the RemoteRootSync owns exists and is up to date with the latest spec (regardless of the status of the RootSync resource itself), so this would deviate from that pattern which the kstatus library depends on.

Beyond that, there's no other reason we could have a custom condition type where the four states could be captured with Reason.

@mortent thoughts?

droot: Thanks for the detailed answer here. Looked at the logic again:

I think `Reconciled`, `Stalled` is enough to convey the status to the end-users. `Ready` is not needed for that. So this is good so far.
However the current logic to determine whether the controller created remote RSync object or not  seems a bit unintuitive implemented using the reason field in `reconciling` condition. If we think more it is is essentially a state that is for internal purpose to be used while finalizing and it is not really useful for the end-user status. Seems like having separate field in the status to track `RSyncObjectCreatedOrNot` would have kept the implementation cleaner. WDYT ?

rollouts/controllers/remoterootsync_controller.go
droot: Related comment in the other thread: https://github.com/GoogleContainerTools/kpt/pull/3849/files#r1117840845

rollouts/controllers/remoterootsync_controller.go
ChristopherFry: Added `Status.SyncCreated` boolean field with the latest commit. This eliminates the need for the isExternalSyncCreated function.

rollouts/controllers/remoterootsync_controller.go
```diff
@@ -145,42 +156,66 @@ func (r *RemoteRootSyncReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 	rrsPrior := rrs.DeepCopy()
+	conditions := &rrs.Status.Conditions
 
-	rrs.Status.SyncStatus = syncStatus
-	rrs.Status.ObservedGeneration = rrs.Generation
+	if syncError == nil {
+		rrs.Status.SyncStatus = syncStatus
+		rrs.Status.SyncCreated = true
+
+		meta.RemoveStatusCondition(conditions, conditionReconciling)
```
droot, line 202: Can you please add a TODO here:
`Reconciling` or `Stalled` condition should be updated (or externalSync conditions should be added) by examining `syncStatus`. `SyncStatus` could have values `Stalled, Reconciling, Error, Pending` even when `syncError` is nil.

droot (approved):

rollouts/controllers/remoterootsync_controller.go
ChristopherFry: I don't think we want actually do this with these being [kstatus conditions](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus). From my understanding, the `Reconciling` and `Stalled` conditions (in the case of the remote root sync) should only consider if the root sync is created and up to date with the latest spec as defined by the remote root sync, regardless of the sync status of the root sync. When the `Reconciling` condition is set to true it implies that the Remote Root Sync still needs to reconcile the root sync with the lates spec from the remote root sync. @mortent can you confirm?

droot: `if the root sync is created and up to date with the latest spec as defined by the remote root sync`

You are right and that's where my comment come from. If you look at `checkSyncStatus`, it can return `Pending` status without any error when `generation != observedGeneration` (meaning rootsync is *not* up to date with the latest spec of rootsync). I think for other values `Stalled, Reconciling, Error`, we will have to create an extra condition to track `externalSyncStatus` or something.

Happy to chat offline :)

PR #3853, opened 2023-03-02
Rollouts gitlab support

droot:
This PR adds support for `GitLab` as package source.

Note: Testing individual components is turning out to be tricky. I have added tests for package discovery for `gitlab` and `github` that avoids running controller but the test setup is not ideal. I would like to use `gitlab` community edition for spinning up git repository with some test data, but I want to tackle that as separate task.


droot:
@ChristopherFry @natasha41575 Would be great if you both of you can take a look at it.

droot:
@ChristopherFry @natasha41575 The tests are not ideal in current form But they are helpful if one is hacking on `packagediscovery` itself. I plan to address them by removing the dependency on gitlab with https://github.com/GoogleContainerTools/kpt/issues/3855.

rollouts/pkg/packagediscovery/github_test.go
```diff
@@ -0,0 +1,95 @@
...
+	gitopsv1alpha1 "github.com/GoogleContainerTools/kpt/rollouts/api/v1alpha1"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGitHubGetPackages_SingleRepo(t *testing.T) {
+	config := gitopsv1alpha1.PackagesConfig{
+		SourceType: gitopsv1alpha1.GitHub,
+		GitHub: gitopsv1alpha1.GitHubSource{
+			Selector: gitopsv1alpha1.GitHubSelector{
+				Org:       "droot",
```
natasha41575, line 32: Should we create these test packages under `https://github.com/GoogleContainerTools/kpt/tree/main/package-examples` (or some other place under GoogleContainerTools)? I think using your username is ok short-term, but long term it would probably be best to have it somewhere more official and not tied to someone's individual account

rollouts/pkg/packagediscovery/packagediscovery.go
```diff
@@ -58,83 +134,118 @@ func NewPackageDiscovery(client client.Client, namespace string) *PackageDiscove
...
+func (d *PackageDiscovery) getPackagesForRepo(gitHubClient *github.Client, ctx context.Context, selector gitopsv1alpha1.GitHubSelector, repo *github.Repository) ([]DiscoveredPackage, error) {
 	discoveredPackages := []DiscoveredPackage{}
-
+	branch := selector.Branch
+	if branch == "" {
+		branch = repo.GetDefaultBranch()
+	}
 	if isSelectorField(selector.Directory) {
-		tree, _, err := gitHubClient.Git.GetTree(ctx, selector.Org, repoName, selector.Revision, true)
+		tree, _, err := gitHubClient.Git.GetTree(ctx, selector.Org, *repo.Name, repo.GetDefaultBranch(), true)
```
natasha41575, line 248: Why are we using `repo.GetDefaultBranch` here instead of `branch`?

rollouts/pkg/packagediscovery/packagediscovery.go
```diff
@@ -149,29 +260,69 @@ func (d *PackageDiscovery) getPackagesForRepository(gitHubClient *github.Client,
...
+func (d *PackageDiscovery) useCache(config gitopsv1alpha1.PackagesConfig) bool {
+	return d.cache != nil && cmp.Equal(config, d.cache.config) && time.Now().Before(d.cache.expiration)
+}
+
+func (d *PackageDiscovery) NewGitLabClient(ctx context.Context, gitlabSource gitopsv1alpha1.GitLabSource) (*gitlab.Client, error) {
+
+	// initialize a gitlab client
+	secretName := gitlabSource.SecretRef.Name
+	if secretName == "" {
+		return nil, fmt.Errorf("GitLab secret reference is missing from the config")
```
natasha41575, line 293: super nit: in this error message, we are camel-casing `GitLab`, but in later error messages, we leave it all lowercase (`gitlab`). I think we should pick one for consistency

rollouts/pkg/packagediscovery/packagediscovery.go
```diff
@@ -149,29 +260,69 @@ func (d *PackageDiscovery) getPackagesForRepository(gitHubClient *github.Client,
...
+	key := client.ObjectKey{
+		Namespace: d.namespace,
+		Name:      secretName,
+	}
+	if err := d.client.Get(ctx, key, &repositorySecret); err != nil {
+		return nil, fmt.Errorf("cannot retrieve gitlab credentials %s: %v", key, err)
+	}
+
+	accessToken := string(repositorySecret.Data["token"])
+	// TODO(droot): BaseURL should also be configurable through the API
```
natasha41575, line 305: question: what other baseURLs are possible with gitlab?

natasha41575 (commented):
Minor comments, but I have a bigger question.

Porch also has some code to discover packages from Github. Is there any overlap here? I'm wondering if we will eventually want to combine them into a git-package-discovery library.

rollouts/api/v1alpha1/rollout_types.go
```diff
@@ -78,32 +78,57 @@ type ClusterSourceGCPFleet struct {
...
+	Revision  string          `json:"revision,omitempty"`
+	Branch    string          `json:"branch,omitempty"`
 	SecretRef SecretReference `json:"secretRef,omitempty"`
 }
 
+// GitLabSource defines the packages source in GitLab.
+type GitLabSource struct {
+	// SecretReference is the reference to a kubernetes secret
+	// that contains GitLab access token
+	SecretRef SecretReference `json:"secretRef,omitempty"`
```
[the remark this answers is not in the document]
droot: Good question. The reasoning for separating `SecretRef` from `Selector` is that `SecretRef` is relevant to `How to connect to the Package Source` and `Selector` is relevant to `How to select packages from a source`. So mixing the two felt odd so decided to do it the right way for `GitLab`.

I intend to do the same for `GitHub` as well but in a separate PR.

rollouts/pkg/packagediscovery/github_test.go
droot: So two things:
1. Absolutely agree on keeping it in an official place. Mine is just a stop gap because of lack of a place.
2. We can't use `...kpt/tree/main/package-examples` for this because it complicates adding/deleting tests. One has to make changes in the test packages and commit it, then update the tests because tests will use whatever is available on `main`. `kpt-samples` could be another place but that is used for reference packages and using it for tests can make it really noisy.

But need to find a solution.

droot (commented):
Thanks for the review @natasha41575

rollouts/pkg/packagediscovery/packagediscovery.go
droot: Two use-cases that I have in my mind:

1. Enterprises can self-host gitlab instances and in that case, capability to point to self-hosted gitlab will be needed
2. Eventually, we will use `gitlab` docker instance for our e2e tests where being able to point to our instance will be needed.

rollouts/pkg/packagediscovery/packagediscovery.go
droot: Good catch, fixed it.

rollouts/pkg/packagediscovery/packagediscovery.go
droot: Good catch, using `lower-case` everywhere for now.

droot:
>  I'm wondering if we will eventually want to combine them into a git-package-discovery library.

Eventually, yes once we understand the usage better.

rollouts/Makefile
```diff
@@ -174,3 +174,14 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
 	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+
```
ChristopherFry, line 177: nit: there's an extra newline here

rollouts/controllers/rollout_controller.go
```diff
@@ -453,8 +458,9 @@ func (r *RolloutReconciler) computeTargets(ctx context.Context,
 
 func pkgNeedsUpdate(rollout *gitopsv1alpha1.Rollout, rrs gitopsv1alpha1.RemoteRootSync, pkg *packagediscovery.DiscoveredPackage) (*gitopsv1alpha1.RemoteRootSync, bool) {
 	// TODO: We need to check other things here besides git.Revision and metadata
+	metadata := getSpecMetadata(rollout)
```
ChristopherFry, line 461: I would consider creating a pull request with just this change since this change is technically unrelated to the rest of the changes in the pull request.

ChristopherFry (approved):
Looks good!

Not needed now, but In a future change I would definitely suggest splitting the GitHub and GitLab logic into separate files. The Package Discovery file is getting long.

rollouts/controllers/rollout_controller.go
droot: Thanks for pointing it out. This will go away once @natasha41575 PR (that has the fix) is merged.

droot:
> Not needed now, but In a future change I would definitely suggest splitting the GitHub and GitLab logic into separate files. The Package Discovery file is getting long.

@ChristopherFry Went ahead and did it in the last commit. At some point, I want to introduce an interface for package selection and this sets up well for that change in the future :)

rollouts/controllers/rollout_controller.go
```diff
@@ -463,6 +469,25 @@ func pkgNeedsUpdate(rollout *gitopsv1alpha1.Rollout, rrs gitopsv1alpha1.RemoteRo
 	return nil, false
 }
 
+func getSpecMetadata(rollout *gitopsv1alpha1.Rollout) *gitopsv1alpha1.Metadata {
```
natasha41575, line 472: I think this PR needs a rebase on main. My PR with this fix, but I believe I moved this function to the end of the file in my PR and we should try to avoid something funny happening when this PR is merged.

natasha41575 (approved):

rollouts/controllers/rollout_controller.go
droot: I removed my changes to facilitate clean merge. I had to make small change to make the code aware of `syncTemplate` being nil.

rollouts/pkg/packagediscovery/github_test.go
natasha41575: Perhaps a `kpt-test` repo?

PR #3857, opened 2023-03-02
rollouts: rename RemoteRootSync to RemoteSync

natasha41575:
Split out from https://github.com/GoogleContainerTools/kpt/pull/3850. This one will likely have merge conflicts with https://github.com/GoogleContainerTools/kpt/pull/3850, so will need a rebase. 

Renames RemoteRootSync to RemoteSync. This controller will eventually support RepoSyncs in addition to RootSyncs, hence the rename.


droot (approved):
Thanks for splitting the PRs out. We should figure out the right time to merge this though because it involves little disruption for devs :)

natasha41575:
> Update the docs (getting-started-guide) specifically and verify it.

If I'm looking at the right doc, it doesn't specify anything about `RemoteRootSync`, so I don't think there is anything to update there.

PR #3860, opened 2023-03-03
add unit tests for remoterootsync controller

natasha41575:
This PR addresses part of of https://github.com/GoogleContainerTools/kpt/issues/3856

The bulk of the business logic in the `RemoteRootSync` reconciler is in the `patchRootSync` function, so this PR adds unit tests for `patchRootSync` using a fake kubernetes dynamic client. 

A note on sequencing, this has some conflicts with https://github.com/GoogleContainerTools/kpt/pull/3857 and https://github.com/GoogleContainerTools/kpt/pull/3850. Whichever PRs are merged last should (a) address the conflicts and (b) should ensure that unit tests for supporting `RepoSync` are also added here.

I plan to add unit tests for `Rollouts` controller separately.


rollouts/controllers/remoterootsync_controller_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	for tn, tc := range testCases {
+		t.Run(tn, func(t *testing.T) {
+			var rrs gitopsv1alpha1.RemoteRootSync
+			require.NoError(t, yaml.Unmarshal([]byte(tc.rrsInput), &rrs))
+
+			client := dynamicfake.NewSimpleDynamicClient(scheme, &rrs)
+			(&RemoteRootSyncReconciler{}).patchRootSync(context.Background(), client, rrs.Name, &rrs)
+
+			require.Equal(t, tc.expectedActions, client.Actions())
```
droot, line 111: I am still thinking if we should just have unit test for `BuildObjectsToApply` and that will give the same effect. Testing the dynamic client part is not the intent here and that can only be tested in an e2e more effectively.

The e2e test for the behavior is, for a given input RRS, corresponding RS in target cluster should exist. Unit test version of that will be for a given RRS, BuildObjectsToApply(RRS) == expected RS.

WDYT ?

natasha41575: SGTM! I will update this to test BuildObjectsToApply instead

natasha41575: Updated this PR to test `BuildObjectsToApply`. I also added tests for two other helper functions that produce important inputs to `BuildObjectsToApply`.

droot (approved):

PR #3871, opened 2023-03-10
rollouts: check entire spec when patching remotesync

natasha41575:
Fixes https://github.com/GoogleContainerTools/kpt/issues/3846

IMO the most robust way to check if the RemoteSync object needs to be updated is to create a new RemoteSync struct based on the current rollout spec and compare it to the existing RemoteSync object in the cluster.

### DIFF
--- a/rollouts/controllers/rollout_controller.go
+++ b/rollouts/controllers/rollout_controller.go
@@ -452,18 +452,13 @@ func (r *RolloutReconciler) computeTargets(ctx context.Context,
 }
 
 // rsNeedsUpdate checks if the underlying remotesync needs to be updated by creating a new RemoteSync object and comparing it to the existing one
-func rsNeedsUpdate(ctx context.Context, rollout *gitopsv1alpha1.Rollout, oldRS *gitopsv1alpha1.RemoteSync, target *clusterPackagePair) (*gitopsv1alpha1.RemoteSync, bool) {
-	newRS := newRemoteSync(rollout,
-		gitopsv1alpha1.ClusterRef{Name: target.cluster.Name},
-		toSyncSpec(target.packageRef),
-		pkgID(target.packageRef),
-		getSpecMetadata(rollout),
-	)
+func rsNeedsUpdate(ctx context.Context, rollout *gitopsv1alpha1.Rollout, currentRS *gitopsv1alpha1.RemoteSync, target *clusterPackagePair) (*gitopsv1alpha1.RemoteSync, bool) {
+	desiredRS := newRemoteSync(rollout, target)
 
 	// if the spec of the new RemoteSync object is not identical to the existing one, then an update is necessary
-	if !equality.Semantic.DeepEqual(oldRS.Spec, newRS.Spec) {
-		oldRS.Spec = newRS.Spec
-		return oldRS, true
+	if !equality.Semantic.DeepEqual(currentRS.Spec, desiredRS.Spec) {
+		currentRS.Spec = desiredRS.Spec
+		return currentRS, true
 	}
 
 	// no update required, return nil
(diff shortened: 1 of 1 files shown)


rollouts/controllers/rollout_controller.go
```diff
@@ -448,18 +451,22 @@ func (r *RolloutReconciler) computeTargets(ctx context.Context,
...
+	)
 
-		rs.Spec.Template.Spec.Git.Revision = pkg.Revision
-		rs.Spec.Template.Metadata = metadata
-		rs.Spec.Type = rollout.GetSyncTemplateType()
-		return &rs, true
+	// if the spec of the new RemoteSync object is not identical to the existing one, then an update is necessary
+	if !equality.Semantic.DeepEqual(oldRS.Spec, newRS.Spec) {
+		oldRS.Spec = newRS.Spec
+		return oldRS, true
```
droot, line 466: Yes, this seems to be more robust. Infact, it is almost equivalent to:

```
newRS := newRemoteSync(...)
// apply newRS using server-side apply where apply is only applicable if there are differences.
client.Apply(newRS)
```

rollouts/controllers/rollout_controller.go
```diff
@@ -448,18 +451,22 @@ func (r *RolloutReconciler) computeTargets(ctx context.Context,
...
+		getSpecMetadata(rollout),
+	)
 
-		rs.Spec.Template.Spec.Git.Revision = pkg.Revision
-		rs.Spec.Template.Metadata = metadata
-		rs.Spec.Type = rollout.GetSyncTemplateType()
-		return &rs, true
+	// if the spec of the new RemoteSync object is not identical to the existing one, then an update is necessary
+	if !equality.Semantic.DeepEqual(oldRS.Spec, newRS.Spec) {
+		oldRS.Spec = newRS.Spec
```
droot, line 465: Some suggestions:
Renaming `oldRS` to `currRS` and `newRS` to `expRS` or `desiredRS` will make the code even more obvious.

rollouts/controllers/rollout_controller.go
```diff
@@ -448,18 +451,22 @@ func (r *RolloutReconciler) computeTargets(ctx context.Context,
...
 
-func pkgNeedsUpdate(ctx context.Context, rollout *gitopsv1alpha1.Rollout, rs gitopsv1alpha1.RemoteSync, pkg *packagediscovery.DiscoveredPackage) (*gitopsv1alpha1.RemoteSync, bool) {
-	// TODO: We need to check other things here besides git.Revision and metadata
-	metadata := getSpecMetadata(rollout)
-	if pkg.Revision != rs.Spec.Template.Spec.Git.Revision ||
-		!reflect.DeepEqual(metadata, rs.Spec.Template.Metadata) ||
-		rs.Spec.Type != rollout.GetSyncTemplateType() {
+// rsNeedsUpdate checks if the underlying remotesync needs to be updated by creating a new RemoteSync object and comparing it to the existing one
+func rsNeedsUpdate(ctx context.Context, rollout *gitopsv1alpha1.Rollout, oldRS *gitopsv1alpha1.RemoteSync, target *clusterPackagePair) (*gitopsv1alpha1.RemoteSync, bool) {
+	newRS := newRemoteSync(rollout,
```
droot, line 456: Note to self: we should consider changing the function signature to accept `newRemoteSync(rollout, target)`

droot (approved):

rollouts/controllers/rollout_controller.go
natasha41575: Makes sense! I can make that change in this PR

rollouts/controllers/rollout_controller.go
natasha41575: Will do

PR #3879, opened 2023-03-13
rollouts: basic integration test

droot:
This PR adds a basic skeleton for rollouts controller tests.

@natasha41575 yet to figure out how to handle the CRD for KCC cluster. If you have some ideas, let me know.

Most the it is created by following docs on [kubebuilder docs](https://kubebuilder.io/cronjob-tutorial/writing-tests.html)


rollouts/config/crd/bases/containercluster.yaml
```diff
@@ -0,0 +1,1336 @@
...
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  labels:
+    cnrm.cloud.google.com/managed-by-kcc: "true"
+    cnrm.cloud.google.com/stability-level: stable
+    cnrm.cloud.google.com/system: "true"
+    cnrm.cloud.google.com/tf2crd: "true"
+    core.cnrm.cloud.google.com/configconnector: configconnector.core.cnrm.cloud.google.com
+  name: containerclusters.container.cnrm.cloud.google.com
```
natasha41575, line 24: Does this file come from another Github source? We could consider using the remote URL in the kustomization in that case, instead of copying it over

droot: I want to find a better solution for this CRD. This is an external CRD (comes from KCC) and is already part of our release manifests, I will try to reuse that instead of adding a new one here.

droot: I am reusing the `containercluster.yaml` from the `manifests` directory for now. So we have one source of truth for now. Eventually we want to make `rollouts` to be able to gracefully handle unavailability of this CRD.

rollouts/controllers/rollout_controller_test.go
```diff
@@ -0,0 +1,373 @@
...
+					Count:  1,
+					Labels: map[string]string{
+						"city": "sjcc",
+					},
+				},
+				{
+					Prefix: "e2e-sfoo-",
+					Count:  1,
+					Labels: map[string]string{
+						"city": "sfoo",
```
natasha41575, line 192: Why `sjcc` and `sfoo` instead of `sjc` and `sfo`? Is there some conflict with the clusters setup in the previous test? 

Since this is otherwise the same function passed to `BeforeEach` as the previous test, I was wondering if it would make sense to define this func elsewhere and reuse it across tests.

rollouts/controllers/rollout_controller_test.go
```diff
@@ -0,0 +1,373 @@
...
+					},
+				},
+			}
+			targetClusterSetup, err = e2eclusters.GetClusterSetup(tt, k8sClient, targets...)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(targetClusterSetup.PrepareAndWait(context.TODO(), 5*time.Minute)).To(Succeed())
```
natasha41575, line 199: Are we expecting to use 5*time.Minute in every test? If so, maybe we can put it in a const

rollouts/controllers/rollout_controller_test.go
droot: We want these two tests to be independent of each other so that we can run them in parallel (though running in parallel requires running with `ginkgo` CLI, but that's a different story). Since we share the APIServer/Etcd between the two tests, keeping a separate `target` clusters is required to keep these tests independent.

Also, each tests may require different target cluster environment (for ex. progressive may have more advance test setup here), so keeping them separate gives that flexibility. So duplication here is the feature not a bug :)

rollouts/controllers/rollout_controller_test.go
droot: So these timeouts really depend on the target cluster setup and that may vary between these tests, so avoiding the reuse.

droot:
@natasha41575 responded to your comments, pl. take another look and let me know.

natasha41575 (approved):

PR #3889, opened 2023-03-20
porch: fix issue with package names when repo dir is specified

natasha41575:
Fixes https://github.com/GoogleContainerTools/kpt/issues/3870.

If a repository is registered with a directory, porch should automatically put the package in the registered directory upon creation, and omit the directory name from the package name when fetching the package revision.

### DIFF
--- a/porch/pkg/git/git.go
+++ b/porch/pkg/git/git.go
@@ -277,10 +277,6 @@ func (r *gitRepository) CreatePackageRevision(ctx context.Context, obj *v1alpha1
 	ctx, span := tracer.Start(ctx, "gitRepository::CreatePackageRevision", trace.WithAttributes())
 	defer span.End()
 
-	if !packageInDirectory(obj.Spec.PackageName, r.directory) {
-		return nil, fmt.Errorf("cannot create %s outside of repository directory %q", obj.Spec.PackageName, r.directory)
-	}
-
 	var base plumbing.Hash
 	refName := r.branch.RefInLocal()
 	switch main, err := r.repo.Reference(refName, true); {
@@ -296,15 +292,16 @@ func (r *gitRepository) CreatePackageRevision(ctx context.Context, obj *v1alpha1
 		return nil, fmt.Errorf("failed to create packagerevision: %w", err)
 	}
 
-	// TODO use git branches to leverage uniqueness
+	packagePath := filepath.Join(r.directory, obj.Spec.PackageName)
 
-	draft := createDraftName(obj.Spec.PackageName, obj.Spec.WorkspaceName)
+	// TODO use git branches to leverage uniqueness
+	draft := createDraftName(packagePath, obj.Spec.WorkspaceName)
 
 	// TODO: This should also create a new 'Package' resource if one does not already exist
 
 	return &gitPackageDraft{
 		parent:        r,
-		path:          obj.Spec.PackageName,
+		path:          packagePath,
 		workspaceName: obj.Spec.WorkspaceName,
 		lifecycle:     v1alpha1.PackageRevisionLifecycleDraft,
 		updated:       time.Now(),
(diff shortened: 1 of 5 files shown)


porch/pkg/git/git.go
```diff
@@ -277,8 +277,10 @@ func (r *gitRepository) CreatePackageRevision(ctx context.Context, obj *v1alpha1
 	ctx, span := tracer.Start(ctx, "gitRepository::CreatePackageRevision", trace.WithAttributes())
 	defer span.End()
 
-	if !packageInDirectory(obj.Spec.PackageName, r.directory) {
-		return nil, fmt.Errorf("cannot create %s outside of repository directory %q", obj.Spec.PackageName, r.directory)
+	// if the repository has been registered with a directory, add the directory name to the name of the package to ensure it is
+	// created in the repository directory
+	if r.directory != "" {
+		obj.Spec.PackageName = r.directory + "/" + obj.Spec.PackageName
```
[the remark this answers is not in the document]
natasha41575: > So it seems like with this solution we are between different patterns for the package name. Is there a way we can avoid this? It seems to create a gotcha that will be surprising for anyone looking at this in the future.

That's fair, I'll update this PR to keep the package name as-is, but have it be created relative to the base directory.

> Do we know how these "folders" in the name are handled for OCI?

AFAICT when I create packages in Artifact Registry, there are no "folders" or directories. If I create an packagerevision with a package name e.g. `foo/bar`, then I see an image named `foo/bar` in the repo without any folder structure, and the [OCI packagerevision struct that we have](https://github.com/GoogleContainerTools/kpt/blob/4cee890e2509a4bc03ae241062aba69415b0c49a/porch/pkg/oci/oci.go#L331) doesn't appear to have any equivalent to directories in git. So I don't think we need to make any changes to the OCI code for this.

porch/pkg/git/git.go
```diff
@@ -296,15 +292,16 @@ func (r *gitRepository) CreatePackageRevision(ctx context.Context, obj *v1alpha1
 		return nil, fmt.Errorf("failed to create packagerevision: %w", err)
 	}
 
-	// TODO use git branches to leverage uniqueness
+	packagePath := filepath.Join(r.directory, obj.Spec.PackageName)
```
natasha41575, line 295: I updated the PR in a way that hopefully makes it more obvious what is going on - the package name is now treated as relative to the directory registered with the repo. 

So the package path stored in our structs contains the base directory. And when we get the package name, we get it relative to the base directory.

natasha41575: This seems to have triggered some test failures, I'm looking into them

natasha41575:
Converting this to a draft while I investigate the test failures

porch/pkg/git/git_test.go
```diff
@@ -1131,7 +1131,7 @@ func (g GitSuite) TestNestedDirectories(t *testing.T) {
 	}{
 		{
 			directory: "sample",
-			packages:  []string{"sample/v1", "sample/v2", "sample/" + g.branch},
+			packages:  []string{"/v1", "/v2", "/" + g.branch},
```
natasha41575, line 1134: This demonstrates an interesting case. If the package path is `foo/bar` and you have a package at the root of `foo/bar`, the package name is empty (`""`). 

This is consistent with porch's current behavior; currently on main, if there is a package at the root of a repository, and you register the repository without specifying a directory, the package is treated as having an empty name:

```
$ kpt alpha rpkg get
NAME                                                   PACKAGE             WORKSPACENAME         REVISION   LATEST   LIFECYCLE   REPOSITORY
deployments-333d0669c638bb12e47928ef0d40303a5503099d                       main                  main       false    Published   deployments
```

mortent: Yeah, I think I've experienced this. I think we should consider this a bug. Even if a repo is registered with an empty directory, all packages should exist in subdirectories. We can fix this separately from this PR.

mortent (approved):
Looks great!

PR #3890, opened 2023-03-21
rollouts: add namespace flag to CLI commands

natasha41575:
Fixes https://github.com/GoogleContainerTools/kpt/issues/3814

Took a look at the `kpt live` code and used the same method of adding the namespace flag via `k8scmdutil.Factory`. This makes it so that the `rollouts` commands use the default value from `kubeconfig` and also supports `-n` or `--namespace` arg.


droot (approved):
Awesome.

PR #3894, opened 2023-03-23
rollouts: add unit test and fake client for simple reconcileRollouts case

natasha41575:
This PR adds a unit test for a simple case for `reconcileRollouts`. To support the unit test, this PR adds a fake client and refactors a few things out of `reconcileRollout`. I'd like to add the remaining unit tests in a followup, so that the review here can focus on structure of the test, fake client code, and refactoring, and the followup PR can focus on code coverage.

I looked at using https://github.com/golang/mock for the fake client, but felt that this fake client is simple enough that another framework wasn't needed. Happy to revisit this idea though.

Related issue: https://github.com/GoogleContainerTools/kpt/issues/3856

### DIFF
--- a/rollouts/controllers/fakeclient.go
+++ b/rollouts/controllers/fakeclient.go
@@ -27,83 +27,82 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-var _ client.Client = &fakeRolloutsClient{}
+var _ client.Client = &fakeRemoteSyncClient{}
 
-type fakeRolloutsClient struct {
+type fakeRemoteSyncClient struct {
 	client.Client
 
-	remotesyncs []gitopsv1alpha1.RemoteSync
+	remotesyncs map[types.NamespacedName]gitopsv1alpha1.RemoteSync
 	actions     []string
 }
 
-func (fc *fakeRolloutsClient) Create(ctx context.Context, obj client.Object, _ ...client.CreateOption) error {
+func newFakeRemoteSyncClient() *fakeRemoteSyncClient {
+	return &fakeRemoteSyncClient{
+		remotesyncs: make(map[types.NamespacedName]gitopsv1alpha1.RemoteSync),
+	}
+}
+
+func (fc *fakeRemoteSyncClient) Create(ctx context.Context, obj client.Object, _ ...client.CreateOption) error {
 	fc.actions = append(fc.actions, fmt.Sprintf("creating object named %q", obj.GetName()))
 
-	if err := fc.Get(ctx, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()}, obj); err == nil {
-		// object already exists
+	namespacedName := types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()}
+	if err := fc.Get(ctx, namespacedName, obj); err == nil {
 		return fmt.Errorf("object %q already exists", obj.GetName())
 	}
-	fc.remotesyncs = append(fc.remotesyncs, *obj.(*gitopsv1alpha1.RemoteSync))
+	fc.remotesyncs[namespacedName] = *obj.(*gitopsv1alpha1.RemoteSync)
 
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


rollouts/controllers/fakeclient.go
```diff
@@ -0,0 +1,109 @@
...
+	gitopsv1alpha1 "github.com/GoogleContainerTools/kpt/rollouts/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ client.Client = &fakeRolloutsClient{}
+
+type fakeRolloutsClient struct {
```
droot, line 32: Since it supports CRUD operations on `RemoteSync` objects, should we name it `fakeRemoteSyncClient` to make it obvious.

rollouts/controllers/fakeclient.go
```diff
@@ -0,0 +1,109 @@
...
+	client.Client
+
+	remotesyncs []gitopsv1alpha1.RemoteSync
```
droot, line 35: It can simplify the code below if we make this a `map[types.NamespacedName]RemoteSync` because that will remove all the loops.

rollouts/controllers/rollout_controller.go
```diff
@@ -154,10 +154,29 @@ func (r *RolloutReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+	logger.Info("Discovered packages", "packagesCount", len(discoveredPackages), "packages", packagediscovery.ToStr(discoveredPackages))
+
+	allClusterStatuses, waveStatuses, err := r.reconcileRollout(ctx, &rollout, strategy, targetClusters, discoveredPackages)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+
+	sortClusterStatuses(allClusterStatuses)
```
droot, line 175: Should we keep `sortClusterStatuses` inside the `reconcileRollout` that will ensure the function being tested return a stable ordered clusterstatuses ?

rollouts/controllers/rollout_controller_test.go
```diff
@@ -371,3 +377,158 @@ var _ = Describe("Rollout", func() {
...
+
+		// first call to reconcileRollout - only one cluster should start progressing
+		_, waveStatus, err := reconciler.reconcileRollout(
+			context.Background(),
+			rollout,
+			strategy,
+			targetClusters,
+			[]packagediscovery.DiscoveredPackage{discoveredPackage},
+		)
+		require.NoError(t, err)
```
droot, line 451: Are we using this style of assertions in other parts of `kpt` project ? otherwise I have found [omega matchers](https://onsi.github.io/gomega/) to be very good for this.

droot (changes requested):
I like the approach. It is much simpler and will help a ton in reproducing edge scenario.
I have some minor comments.

rollouts/controllers/fakeclient.go
natasha41575: good call, done!

rollouts/controllers/fakeclient.go
natasha41575: done.

rollouts/controllers/rollout_controller.go
natasha41575: sgtm! Moved back

rollouts/controllers/rollout_controller_test.go
natasha41575: There are a number of places in kpt repo that we are using the `testify/assert` and `testify/require` packages to do similar assertions in unit tests. `require` is a subtly different flavor of `assert`, except that `require` exits the test immediately, while `assert` continues running the rest of the test even when it encounters an error. 

That said, I am happy to convert to omega matchers if you prefer!

droot: Didn't see usage of `require` before. But the failure behavior of `require` does make more sense.

I am fine using `require` here, so let's ship it.

droot (approved):

PR #3895, opened 2023-03-28
rollouts: unit tests for rollouts controller

natasha41575:
Fixes https://github.com/GoogleContainerTools/kpt/issues/3856

These unit tests for `reconcileRollouts` cover the following: 
- RolloutStrategies
  - AllAtOnce
  - RollingUpdate (with maxConcurrent)
  - Progressive (with waves and maxConcurrent)
- Matching target clusters to packages
  - AllClusters
  - Custom CEL expression matching
- Waves
  - PauseAfterWave
  - Advancing to the next wave
  - Matching clusters to waves with label selection
- RemoteSyncs
   - Creation
   - Updating
   - Deletion
   - Metadata (labels and annotations) get propagated from the Rollout object to RemoteSync

Please feel free to suggest other scenarios to cover here.

### DIFF
--- a/rollouts/controllers/rollout_controller_test.go
+++ b/rollouts/controllers/rollout_controller_test.go
@@ -379,14 +379,396 @@ var _ = Describe("Rollout", func() {
 })
 
 func TestReconcileRollout(t *testing.T) {
-	t.Run("maxconcurrent", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


droot (approved):
This is fantastic. Overall, I think this covers important use-cases that we care about.

Tip: test coverage profile (-cover) will be a good place to look at if you are curious what functional areas are missing.

PR #3898, opened 2023-03-29
porch: packagevariant controller conditions

natasha41575:
PackageVariant controller updates:
 - moves validation errors to conditions
 - surfaces errors from porch in conditions
 - surface the ensured downstream package revision(s) in the status

Partially addresses https://github.com/GoogleContainerTools/kpt/issues/3891 and https://github.com/GoogleContainerTools/kpt/issues/3872

Sample output:

```
# validation error
  status:
    conditions:
    - lastTransitionTime: "2023-04-14T20:15:17Z"
      message: could not find upstream package revision 'foo/v3' in repo 'blueprints'
      reason: ValidationError
      status: "True"
      type: Stalled
    - lastTransitionTime: "2023-04-14T20:15:17Z"
      message: invalid packagevariant object
      reason: Error
      status: "False"
      type: Ready
```

```
# internal error from porch
  status:
    conditions:
    - lastTransitionTime: "2023-04-14T20:17:04Z"
      message: all validation checks passed
      reason: Valid
      status: "False"
      type: Stalled
    - lastTransitionTime: "2023-04-14T20:15:17Z"
      message: 'Internal error occurred: command error on refs/heads/drafts/moo~/packagevariant-1:
        funny refname'
      reason: Error
      status: "False"
      type: Ready
```

```
# no errors, package revision was created successfully
  status:
    conditions:
    - lastTransitionTime: "2023-04-14T20:17:04Z"
      message: all validation checks passed
      reason: Valid
      status: "False"
      type: Stalled
    - lastTransitionTime: "2023-04-14T20:18:55Z"
      message: successfully ensured downstream package variant
      reason: NoErrors
      status: "True"
      type: Ready
    downstreamTargets:
    - name: packages-9453894ecbd684b9f32008d8936e44728646c3c9
```


johnbelamaric:
I wonder if we should change `DownstreamEnsured` to `Ready`?

johnbelamaric:
What happens as updates are published, does the list of downstreams grow indefinitely? So:
 - first one gets created
 - Draft gets approved
 - Upstream changes, or injections change -> new Draft is created : now there are 2 downstreamTargets, yes?
 - Draft gets approved
 - ...repeat 50 times...
 - Are there now 50+ downstreamTargets?

Shouldn't an ownerReference and/or label be sufficient to tie the target to the package revision?

johnbelamaric:
Maybe a `DraftExists` condition type?

natasha41575:
@johnbelamaric In this PR, if there are existing owned draft(s) it will list all the drafts, and not list any of the published downstreams. If there are no drafts, it will list just the latest published revision. 

Re: draftExists condition type, sure I can add that, though would you mind elaborating on the use case? Theoretically with the current PR, you could check the lifecycle of the status.DownstreamTargets to get the same information

johnbelamaric:
Ok, I think that works. If it's not listing all the previous published versions then we're OK. I just wanted to make sure it didn't grow unbounded. `DraftExists` isn't needed if we have this, agreed.

I do wonder though if we should just use `Ready` to be consistent with other resources.

porch/controllers/packagevariants/api/v1alpha1/packagevariant_types.go
```diff
@@ -77,8 +77,17 @@ type Downstream struct {
...
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// DownstreamTargets contains the downstream targets that the PackageVariant
+	// either created or adopted.
+	DownstreamTargets []DownstreamTarget `json:"downstreamTargets,omitempty"`
+}
+
+type DownstreamTarget struct {
+	Name      string `json:"name,omitempty"`
+	Lifecycle string `json:"lifecycle,omitempty"`
```
mortent, line 90: Is this the Lifecycle of the DownstreamTarget? If so, why do we need to cache it here?

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -51,6 +51,9 @@ type PackageVariantReconciler struct {
 
 const (
 	workspaceNamePrefix = "packagevariant-"
+
+	ConditionTypeValid             = "Valid"
+	ConditionTypeDownstreamEnsured = "DownstreamEnsured"
```
mortent, line 56: It seems like this might actually be describing whether the PackageVariant resource could be reconciled. Could we just use the condition types from kstatus: https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus#conditions ?

porch/controllers/packagevariants/api/v1alpha1/packagevariant_types.go
natasha41575: The workflow I had in mind here would be to create the PackageVariant object, then approve the draft (if it is a draft). The exact feedback was:

> So, my first question is about automation -- the PackageVariant reconciliation will create a PackageRevision as a draft. And now I want to approve the draft. What would be the correct way to wait for that PackageRevision to be created? I see nothing in the status of the PackageVariant I can wait on.
The only way I can see this work is by me knowing what the name of this downstream package is going to be and just poll to wait until it appears.

My thought was to give the client an easy way to check if the PackageRevision created needs approval. There might be cases (if the adoption policy is set to adopt already-existing packagerevisions), where you create the PackageVariant and the target package revision is already published. It isn't necessary to cache the lifecycle here though; the client could just as easily grab the name from the status and then look up the lifecycle with an api call, so I could remove it if you think that would be better.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
natasha41575: I will use `Ready`, I think that will be better (esp since @johnbelamaric made the same comment above)

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -51,6 +51,9 @@ type PackageVariantReconciler struct {
 
 const (
 	workspaceNamePrefix = "packagevariant-"
+
+	ConditionTypeValid             = "Valid"
```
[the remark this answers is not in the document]
natasha41575: done

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
johnbelamaric: Not sure. The kstatus and "traditional" Ready status conventions seem to be in direct opposition to one another. :(

mortent: The `Ready` condition and `Reconciling` is essentially the same thing. The kstatus guidance was written back when the Kubernetes API conventions should adhere to abnormal-true polarity, i.e. conditions should have the value `True` for abnormal situations. For reconciling, we decided that a resource being fully reconciled is the "happy" scenario, so the condition should be `False` in this situation. This is why the kstatus guidance use `Reconciling` (rather than `Reconciled`) as the name. `Ready` is widely used and the API conventions were changed a couple of years ago, so I don't see any issue with using `Ready`.

The `Stalled` condition is there to distinguish between "not reconciled, but making progress" and "not reconciled and it is not making progress" (i.e. doomed). The idea is that `Stalled` being true means some kind of external action is probably required in order to make the resource reconcile. It seems like it might be similar to the use of the `Valid` condition here. Typically any resources that fail validation would need external action in order to make progress.

porch/controllers/packagevariants/api/v1alpha1/packagevariant_types.go
mortent: So creating a PackageVariant is currently a synchronous operation, so there shouldn't be any need to wait for it. However, it probably should be asynchronous (ref discussion about CRDs), so the question is still valid. 
I don't see any issues with keeping it here, we just need to make sure we verify the actual lifecycle before we try to make any changes since the cached value might be stale.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
johnbelamaric: Ok, so, it sounds like the summary is:
* Use `Ready` instead of `Reconciling`
* Use `Stalled` instead of `Valid`

Do we want to set `Reconciling` == !`Ready`, just in case there is tooling that cares? Or `Ready` is good enough?

In the #3910 changes, I want to document what we're actually going to do. There are new cases where `Stalled` is appropriate.

Note that @SimonTheLeg from Kubermatic and the Nephio community is going to pick up implementing the Kptfile pipeline editing feature described in #3910 so I want to get that buttoned up before I am off next week.

natasha41575: > Ok, so, it sounds like the summary is:
> 
> Use Ready instead of Reconciling
> Use Stalled instead of Valid

I'm using ready atm so I will update `Valid` to `Stalled`. Seems to me like we don't need `Reconciling` in addition to `Ready` (and we could always add it later), but I'll defer to @mortent on this.

mortent: Yeah, I think using `Stalled` instead of `Valid` seems correct. We can use the `reason` and `message` fields to communicate that the configuration is not valid.

porch/controllers/packagevariants/api/v1alpha1/packagevariant_types.go
natasha41575: I thought about this a bit more and I think it's better not to have the lifecycle value here - better to have the client get a fresh value from the apiserver. I'll update the PR.

natasha41575:
Rebased and updated.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -105,24 +115,23 @@ func (r *PackageVariantReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		return ctrl.Result{}, nil
 	}
-
-	upstream := r.getUpstreamPR(pv.Spec.Upstream, prList)
-	if upstream == nil {
-		return ctrl.Result{}, fmt.Errorf("could not find upstream package revision")
+	upstream, err := r.getUpstreamPR(pv.Spec.Upstream, prList)
+	if err != nil {
+		setValidationConditionsToFalse(pv, err.Error())
+		return ctrl.Result{}, nil
```
johnbelamaric, line 124: As stated above I think we should set this to requeue after a short time in this instance. And if it's easy we should implement backoff on that requeueing. If that part is not easy, it can wait but we probably should open a ticket to reminde us.

johnbelamaric (changes requested):
A few little changes plus some recommendations that may be added now or later.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -105,24 +115,23 @@ func (r *PackageVariantReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-		pv.Status.ValidationErrors = nil
-		for _, validationErr := range errs {
-			if validationErr.Error() != "" {
-				pv.Status.ValidationErrors = append(pv.Status.ValidationErrors, validationErr.Error())
-			}
-		}
-		statusUpdateErr := r.Client.Status().Update(ctx, pv)
-		return ctrl.Result{}, statusUpdateErr
+		setValidationConditionsToFalse(pv, combineErrors(errs))
+		return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
natasha41575: Yeah, I can separate the invalid upstream into a separate condition

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
natasha41575: I can't find official documentation, but from these comments:

- https://github.com/kubernetes-sigs/controller-runtime/issues/808#issuecomment-639845414
- https://github.com/crossplane/crossplane-runtime/issues/40#issuecomment-777145647

it looks like requeueing with exponential backoff is as easy as just returning an error here. So I updated this to return an error.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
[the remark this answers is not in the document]
natasha41575: See https://github.com/GoogleContainerTools/kpt/pull/3898#discussion_r1172832665

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -211,16 +216,42 @@ func validatePackageVariant(pv *api.PackageVariant) field.ErrorList {
...
-			return &pr
+			return &pr, nil
 		}
 	}
-	return nil
+	return nil, fmt.Errorf("could not find upstream package revision '%s/%s' in repo '%s'",
+		upstream.Package, upstream.Revision, upstream.Repo)
+}
+
+func setValidationConditionsToFalse(pv *api.PackageVariant, message string) {
```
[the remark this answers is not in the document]
natasha41575: good catch! fixed.

mortent (approved):
Looks good to me.

johnbelamaric:
LGTM!

johnbelamaric (approved):

PR #3900, opened 2023-04-03
Fix concurrency issue when using the go-git repository

mortent:
We currently use the go-git library as if it can be used concurrently by multiple go-routines. This is not safe, as they go-git library is not thread-safe for using different go-routines on the same repository as described https://github.com/go-git/go-git/issues/48#issuecomment-622090868 and https://github.com/go-git/go-git/pull/186#issuecomment-775842016. We have seen occasional issues caused by this as described in https://github.com/GoogleContainerTools/kpt/pull/3602.

This PR attempts to fix this by removing all places where the `gitRepository` object leaks references to the underlying `git.Repository` and `storage.Storer` objects (from go-git). It then protects all access to the `gitRepository` with a mutex. Some code had to be refactored to avoid calling protected functions with the lock already held (go doesn't support reentrant locks) and to move all functions belonging to the `gitRepository` struct into the `git.go` file.

This also updates the cache background updater to do the initial fetch immediately, instead of waiting a minute for the ticker to trigger. This triggers some concurrency during the unit tests, which has previously led to errors being reported by the golang race detector. With this change, the race detector do not find any issues.

There might be possible to improve performance by making it possible to call some functions without holding the lock, but it needs deeper investigation of the code.

We should move the `gitRepository` into a separate package, so it is clearer which functions are part of the API and needs to be protected with the mutex and which ones are only used internally. It is somewhat difficult to see now since there is a lot of code in the `git` package.

### DIFF
--- a/porch/pkg/git/git.go
+++ b/porch/pkg/git/git.go
@@ -146,6 +146,24 @@ func OpenRepository(ctx context.Context, name, namespace string, spec *configapi
 	return repository, nil
 }
 
+// type GoGitRepository interface {
+// 	References() (storer.ReferenceIter, error)
+// 	Reference(name plumbing.ReferenceName, resolved bool) (*plumbing.Reference, error)
+// 	ResolveRevision(rev plumbing.Revision) (*plumbing.Hash, error)
+// 	Remote(name string) (*git.Remote, error)
+// 	CommitObject(h plumbing.Hash) (*object.Commit, error)
+// 	Branches() (storer.ReferenceIter, error)
+// 	Fetch(o *git.FetchOptions) error
+// 	GetStorer() storage.Storer
+// 	Push(o *git.PushOptions) error
+// 	Log(o *git.LogOptions) (object.CommitIter, error)
+// 	TreeObject(h plumbing.Hash) (*object.Tree, error)
+// 	BlobObject(h plumbing.Hash) (*object.Blob, error)
+// }
+
+// type goGitRepository struct {
+// }
+
 type gitRepository struct {
 	name               string     // Repository resource name
 	namespace          string     // Repository resource namespace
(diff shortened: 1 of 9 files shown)


porch/pkg/git/git.go
```diff
@@ -1167,6 +1205,336 @@ func (r *gitRepository) traverseHistory(startCommit *object.Commit, cb commitCal
...
+	}
+
+	treeHash, err := r.repo.Storer.SetEncodedObject(eo)
+	if err != nil {
+		return plumbing.Hash{}, err
+	}
+	return treeHash, nil
+}
+
+func (r *gitRepository) UpdateDraftResources(ctx context.Context, draft *gitPackageDraft, new *v1alpha1.PackageRevisionResources, change *v1alpha1.Task) error {
```
justinsb, line 1253: I assume you moved this here because we need to do multiple operations ideally under the same mutex.  I believe we probably should be doing optimistic concurrency with a compare and swap against a tag, the problem is that OCI doesn't support that.  I think this is fine, but we probably do want to split out a "pure git" layer as you suggested.

justinsb (approved):

justinsb:
LGTM.  Agree that "We should move the gitRepository into a separate package" but I feel that can be done as a separate PR.

natasha41575 (approved):
lgtm!

porch/pkg/git/git.go
mortent: I moved this functionality to the `gitRepository` so that the `commitHelper` is only used from within `gitRepository` functions. That means we don't need to add synchronization to the `commitHelper` functions, as we know it will only be called when holding the mutex. So it is not about doing multiple git operations under the same mutex, but just make sure that this doesn't happen concurrently with other operations on the same go-git repository.

PR #3907, opened 2023-04-05
Update copyright headers to use 'The kpt Authors' instead of 'Google LLC'

mortent:
This updates the copyright headers in all files as well as the boilerplate files to use 'The kpt Authors' rather than 'Google LLC'. This does not change any copyright claims by contributors.


Makefile
```diff
@@ -115,7 +115,7 @@ test-docker: build
 # KPT_FN_RUNTIME can be set to select the desired function runtime.
 # If unspecified, the default function runtime will be used.
 test-fn-render: build
-	PATH=$(GOBIN):$(PATH) go test -v --tags=docker --run=TestFnRender/testdata/fn-render/$(T) ./e2e/
+	PATH=$(GOBIN):$(PATH) go test -v --tags=docker --run=TestFnSink/testdata/fn-sink/$(T) ./e2e/
```
natasha41575, line 118: Why make this change? It doesn't look correct. It seems like we should have a separate make target `make test-fn-sink` to run the TestFnSink, and leave the `test-fn-render` target as is?

natasha41575 (commented):
My browser had a hard time loading the diff 😅

Just one comment.

Makefile
mortent: Oh, good catch. I did not intend to commit this, it was just something I used for debugging. I will fix it.

mortent: Fixed now.

natasha41575 (approved):

PR #3910, opened 2023-04-07
Continued evolution of the package variant design

johnbelamaric:
This PR will add additional details, including more on PackageVariantSet.


docs/design-docs/08-package-variant.md
```diff
@@ -292,24 +300,51 @@ The package variant set controller is designed to fill this common need. This
...
+| ![Figure 6: PackageVariantSet with Package List](packagevariantset-target-list-with-packages.png) |
+| :---: |
+| *Figure 6: PackageVariantSet with Package List* |
+
+*Figure 7* shows an example that combines a repository label selector with
+configuration injection that various based upon the target. The template for the
+PackageVariant includes a CEL expression for the one of the injectors, so that
+the injection varies systematically based upon attributes of the target.
+
+| ![Figure 7: PackageVariantSet with Repository Selector](packagevariantset-target-repository-selector.png) |
```
mortent, line 345: This is the wrong link. I think it should be `repo` rather than `repository`.

docs/design-docs/08-package-variant.md
```diff
@@ -292,24 +300,51 @@ The package variant set controller is designed to fill this common need. This
...
-
-*Figure 5* shows an example of creating PackageVariant resources based upon the
-explicitly list of repository and package names.
+- An explicit list of repositories and package names
+- A label selector for Repository objects
+- An arbitrary object selector
 
-![Figure 5: List of Targets](packagevariantset-target-package.png)
+Rules for generating a PackageVariant are associated with a list of targets
+using a template. That template can have explicit values for various
```
mortent, line 312: I'm not sure about the use of the word `template` here. I was at first thinking it would be similar to the PodSpec template in a Deployment resource, but this is somewhat different. At first this seems more like an overlay.

docs/design-docs/08-package-variant.md
```diff
@@ -340,17 +375,77 @@ With PackageVariant, we instead create the following resources:
...
+The `Upstream` and `Downstream` fields specify the source package and
+destination repository and package name. The `Repo` fields refer to the names
+Porch Repository resources in the same namespace as the PackageVariant resource.
+The `Downstream` does not contain a revision, because the package variant
+controller will only create Draft packages. The `Revision` of the eventual
+PackageRevision resource will be determined by Porch at the time of approval.
+
+The `Labels` and `Annotations` fields list metadata to include on the created
+PackageRevision. These values are set *only* at the time a Draft package is
+created. They are ignored for subsequent operations, even if the PackageRevision
```
mortent, line 432: So this means that these are only added whenever a new draft PackageRevision is stamped out, but then users or other controllers are free to change these as they want to without the controller interfering right?

docs/design-docs/08-package-variant.md
```diff
@@ -292,24 +300,51 @@ The package variant set controller is designed to fill this common need. This
...
+| *Figure 6: PackageVariantSet with Package List* |
+
+*Figure 7* shows an example that combines a repository label selector with
+configuration injection that various based upon the target. The template for the
+PackageVariant includes a CEL expression for the one of the injectors, so that
+the injection varies systematically based upon attributes of the target.
+
+| ![Figure 7: PackageVariantSet with Repository Selector](packagevariantset-target-repository-selector.png) |
+| :---: |
+| *Figure 7: PackageVariantSet with Repository Selector* |
```
mortent, line 347: Should the template follow the hierarchical structure of the PackageVariant resource spec? So in this case, should we make the template be:
```yaml
...
template:
  downstream:
    package: bar
  injectors:
  - nameExpr: "target.labels['region'] + '-endpoints'"
```

docs/design-docs/08-package-variant.md
johnbelamaric: Let's see what you think once I publish this. It's a lot like that.

docs/design-docs/08-package-variant.md
johnbelamaric: Correct, I will make that clear.

docs/design-docs/08-package-variant.md
johnbelamaric: Ok, take a look at the complete API as published, see what you think. I think `template` is appropriate.

johnbelamaric:
Marking this ready for review. Still to do before we move from "Work In Progress" to "Approved":
 - fill in the PV status section
 - fill in the PVS status section
 - fill in the use cases

johnbelamaric:
I ended up pulling separate use case list out, I discuss a few examples throughout the rest of the design. We can add more later if needed to understand the purpose.

johnbelamaric:
I believe this is complete now, and ready for review.

johnbelamaric:
Rebased

docs/design-docs/08-package-variant.md
```diff
@@ -640,19 +707,530 @@ understanding, we can see that the controller will perform mutations on the
...
+
+#### PackageVariant Status
+
+PackageVariant sets the following status conditions:
+ - `Stalled` is set to True if there has been a failure that most likely
+   requires user intervention.
+ - `Ready` is set to True if the last reconciliation successfully produced an
+   up-to-date Draft.
+
+The PackageVariant resource will also contain a `DownstreamTargets` field,
```
johnbelamaric, line 736: @natasha41575 is this an accurate description?

docs/design-docs/08-package-variant.md
```diff
@@ -664,28 +1242,53 @@ example, a package should not accept a Porch Repository resource just because
...
-    time we do not plan to do so.
-[^pdc]: A prototype version of this was implemented in Nephio PackageDeployment,
-    but this has not been implemented in PackageVariant as of Porch v0.0.16.
-
-## Figure Legend
-
-![Figure Legend](packagevariant-legend.png)
-
+    users.
+[^pvsimpl]: This document describes PackageVariantSet `v1alpha2`, which has not
```
johnbelamaric, line 1279: @mortent I think we should just go ahead an implement the machinery for conversion webhooks. I am assuming they would run in the controller binary. We're likely to need them in the future. The actual roundtrip conversions in this case are pretty easy to do (the proposed `v1alpha2` API is a superset of the `v1alpha1` functionality).

docs/design-docs/08-package-variant.md
natasha41575: Yes, that is accurate.

docs/design-docs/08-package-variant.md
```diff
@@ -640,19 +707,530 @@ understanding, we can see that the controller will perform mutations on the
...
+#### PackageVariant Template
+
+As previously discussed, the list entries generated by the target criteria
+result in PackageVariant entries. If no template is specified, then
+PackageVariant default are used, along with the downstream repository name and
+package name as described in the previous section. The template allows the user
+to have control over all of the values in the resulting PackageVariant. The
+template API is shown below.
+
+```yaml
```
mortent, line 915: Nit: This should be `go` rather than `yaml`. It will give correct color coding and make it easier to read.

mortent (approved):
Just a small nit, but looks good.

PR #3911, opened 2023-04-12
Update generated artifacts and fix formatting issues

mortent:
This fixes some formatting issues and updates some stale generated code.


porch/controllers/packagevariants/config/rbac/role.yaml
```diff
@@ -31,18 +31,6 @@ rules:
   - patch
   - update
   - watch
-- apiGroups:
```
johnbelamaric, line 34: I think this is probably needed. Perhaps there's a missing gen directive?

mortent: Yeah, I saw those. I was going to check with you or @natasha41575 before merging. I can update the directives to include this.

johnbelamaric (approved):

natasha41575 (approved):

PR #3925, opened 2023-04-20
Add KRM func suppport for PackageVariants

SimonTheLeg:
This PR adds the KRM func support discussed in https://github.com/johnbelamaric/kpt/blob/pvs-design/docs/design-docs/08-package-variant.md for PackageVariants.

Fixes: https://github.com/nephio-project/nephio/issues/117

<details>
<summary>Outdated Review Notes</summary>

Couple of notes for the reviewer(s) from my side:

1. With this change, the `Function` struct type of `kptfile.v1` is being used inside PackageVariantSpec. Because PackageVariantSpec is a Kubernetes Object, I think we also need to have DeepCopy on all of its subfields. As a result I believe we need DeepCopy on the `Function` struct type. For now I have just enabled it on the package level. PTAL if that is fine. Another option is to just have it on the Func type and its fields. But then we need to be super careful whenever we use a new struct to not forget the Kubebuilder annotation.
2. I did notice interesting behaviour when writing the tests for this. The spacing of the Kptfile inside the PackageRevisionResources differs whether you have Mutators in the Package revision or not. The reason for this is that I am using kyaml for the Marshal. Kyaml is required to keep the order of the keys in the yaml. But since this can be different from creating a PackageRevisionResource directly from a yaml string, the spacing might be slightly different. For code reference see `add none with existing` testcase.
In practice I don't think it makes a difference. The only good solution I was able to come up with would be to have custom Marshaller/Unmarshallers on the PackageRevisionResources that uses then the kyaml as a preprocessor. But I think this is not necessary. Please let me know what you think.
3. More of a question, than a comment. What is the reason why we run the PackageVariant mutators before the ones specified in the original kptfile? We should probably add the answer to this also to the initial proposal.

Those are my thoughts so far, let me know if there's anything else :)

</details>


SimonTheLeg:
I have to admit, I am running out of creative ideas how to solve the DeepCopy issue (which was the reason why the first CI run failed). Essentially the problem is that kyaml's ResourceMeta/TypeMeta/ObjectMeta do not have DeepCopy funcs implemented on them. But since it is external and we directly embed it, we don't really have control over it. So I thought maybe it would be enough to switch from yaml.ResourceMeta to metav1.TypeMeta and metav1.ObjectMeta as these are DeepCopy compatible. You can see a draft for this in the commit "[Switch ktpfile from yaml metav1 Typemeta and Objectmeta](https://github.com/GoogleContainerTools/kpt/pull/3925/commits/1c3aa4efab3627683a18863d24d98bcfa7534834)". However then the issue is that kyaml handles fields that are missing the `yaml` struct tag differently than sigs.k8s.io/yaml does 🙈  Basically if kyaml cannot find the yaml tag on a struct, it simply takes the name of the struct field and lowercases it ([see here](https://github.com/kubernetes-sigs/kustomize/blob/2ce1c7cce3b714f67a168d6b3929bf39b2dfc294/kyaml/internal/forked/github.com/go-yaml/yaml/yaml.go#L636)). This becomes a problem for the `APIVersion` field as it becomes "apiversion" instead of the required "api**V**ersion". But because we embed metav1.TypeMeta and metav1.ObjectMeta, we cannot control the struct tags for their subfields.

At this point, there are only two ideas I have left:
a) we could take all the fields in metav1.TypeMeta and metav1.ObjectMeta and put them directly into the KptFile and set the yaml struct tag directly on them. Would look a bit ugly and be a bit cumbersome to maintain, but it would work.
b) we could make an upstream contribution to kyaml to have DeepCopy on their ResourceMeta (and/or TypeMeta, ObjectMeta). I think this would be the most elegant solution also going forward (kpt has a lot of types that embed ResourceMeta), but of course then we need to wait for that to merge.

**Maybe I am missing something obvious, please let me know if you can think of a more elegant solution to fix this. I am all for it :)** If not I would proceed with option propose to proceed with option b).

natasha41575:
> we could make an upstream contribution to kyaml to have DeepCopy on their ResourceMeta (and/or TypeMeta, ObjectMeta). I think this would be the most elegant solution also going forward (kpt has a lot of types that embed ResourceMeta), but of course then we need to wait for that to merge.

You can feel free to make this change in kyaml if you'd like. I agree that this is seemingly the most elegant solution. I can review and approve that PR when you make it.

mortent:
So I think the easiest solution here is to just use kyaml to parse and update the Kptfile. It does provide some more control over how each field is added to yaml, it makes sure the formatting is consistent (since other kpt functionality that updates the Kptfile uses kyaml), and it makes sure any comments in the Kptfile is preserved.

It is not quite as convenient as reading the content into a go struct. We can look into whether that is possible, but we could try to do that separately from this change.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -713,3 +720,34 @@ func ensurePackageContext(pv *api.PackageVariant,
 	prr.Spec.Resources["package-context.yaml"] = cm.String()
 	return nil
 }
+
+func ensureKRMFunctions(pv *api.PackageVariant,
+	prr *porchapi.PackageRevisionResources) error {
+
+	if len(pv.Spec.Mutators) == 0 {
```
johnbelamaric, line 727: We should be able to add validators too, not just mutators.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -713,3 +720,34 @@ func ensurePackageContext(pv *api.PackageVariant,
...
+	}
+
+	kptfile, err := internalpkg.DecodeKptfile(strings.NewReader(prr.Spec.Resources[kptfilev1.KptFileName]))
+	if err != nil {
+		return fmt.Errorf("failed to parse %s of PackageRevisionResources '%s/%s'", kptfilev1.KptFileName, prr.Namespace, prr.Name)
+	}
+
+	// TODO q: why are PackageVariant mutators being run before the ones in the kptfile?
+	// Wouldn't that make it harder for the vendors of the the original package, because their mutators now need
+	// to account for the possibility of mutators from end-users?
```
johnbelamaric, line 742: Originally, I considered having the design allowing both prefix and suffix lists of functions, but I didn't see a real use case for it yet. Besides, the function pipeline needs to run and be idempotent. I chose prefix because it allows the downstream author to tweak what gets fed into the original pipeline, though that's somewhat arbitrary. If we have prefix working, suffix is easy to add later.

The original package authors should never make assumptions about the pipeline not having additional items. Users can change it after cloning the package, this is normal usage.

johnbelamaric (changes requested):
Thanks Simon, this is a great start. The main thing that needs to be fixed is to make this idempotent. The first thing that should be done is to remove all the functions added by this PV in past reconciliations. Then you can add them back in. Details are here: 

https://github.com/GoogleContainerTools/kpt/blob/main/docs/design-docs/08-package-variant.md#krm-function-pipeline

natasha41575:
@SimonTheLeg I want to add a warning that sometimes upgrading the kyaml dependency in kpt can be a chore (and probably should go in a separate PR), so while you are waiting for that, I think @mortent's suggestion in https://github.com/GoogleContainerTools/kpt/pull/3925#issuecomment-1518420340 is a good way to work around it for now.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -768,3 +774,57 @@ func ensurePackageContext(pv *api.PackageVariant,
...
+	if len(pv.Spec.Mutators) == 0 {
+		return nil
+	}
+
+	if _, ok := prr.Spec.Resources[kptfilev1.KptFileName]; !ok {
+		return fmt.Errorf("%s not found in PackageRevisionResources '%s/%s'", kptfilev1.KptFileName, prr.Namespace, prr.Name)
+	}
+
+	// parse PackageVariant Mutators
+	mpvmutators, err := kyaml.Marshal(pv.Spec.Mutators)
```
johnbelamaric, line 790: I don't think we need to do this. I think we can just create a list of functions (typed go struct) for each of the mutators and validators as specified in the design, based on the Go structs here. The pv here was loaded from the API server - there are no comments or anything. If there are any comments to preserve they are the ones in the PRR Kptfile.

So, once we have a complete list of "our" mutators, as go structs, we then look at the pipeline.mutators field in the Kptfile KubeObject and get a slice of subobjects back (see https://pkg.go.dev/github.com/GoogleContainerTools/kpt-functions-sdk/go/fn#SubObject.NestedSlice).

If our list of structs is "ourMutators", then something like:

```
 var newMutators fn.SliceSubObjects
 for _, f := range ourMutators {
  ko, err := fn.NewFromTypedObject(&f)
  if err != nil {
    return err
   }
  newMutators = append(newMutators, ko)
}
existingMutators := kptfileKubeObject.NestedSlice([]string{"pipeline", "mutators"})
 for _, ko := range existingMutators {
  name, ok, err := ko.NestedString("name")
// check ok, err
 if name is one of our names continue
 newMutators = append(newMutators, ko)
}
// then SetSlice with the newMutators
```

johnbelamaric (commented):
I think this will be WAY easier using KubeObject. See the comments.

porch/controllers/packagevariants/api/v1alpha1/packagevariant_types.go
```diff
@@ -54,8 +55,9 @@ const (
 
 // PackageVariantSpec defines the desired state of PackageVariant
 type PackageVariantSpec struct {
-	Upstream   *Upstream   `json:"upstream,omitempty"`
-	Downstream *Downstream `json:"downstream,omitempty"`
+	Upstream   *Upstream            `json:"upstream,omitempty"`
+	Downstream *Downstream          `json:"downstream,omitempty"`
+	Mutators   []kptfilev1.Function `json:"mutators,omitempty"`
```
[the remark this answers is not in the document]
johnbelamaric: Sorry, yes, the design is structured as "user experience" and "detailed design". Sorry, I didn't make that clear. And my diagram is incorrect...

johnbelamaric:
Note: I found a bit of a bug where errors in "applyMutation" are not really propagated properly. Or they are, but they are overwritten be subsequent incorrect reconciliations (because findAndUpdate was written before applyMutations, for just doing packageupdates, and I didn't correctly update it when I implemented applyMutations).

I will be fixing it in #3939. I mention it here in case you come across it. You can see it when doing manual/e2e tests. If you focus on your unit tests and get those working, you should be OK.

SimonTheLeg:
alright, everything should be in order now and I have rebased the PR to make it easier to review.
If there is need to view the old revision history in all its 14 commit glory, I have backed it up here https://github.com/SimonTheLeg/kpt/commits/krm-funcs-for-packagevariant-backup2

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -713,3 +720,34 @@ func ensurePackageContext(pv *api.PackageVariant,
 	prr.Spec.Resources["package-context.yaml"] = cm.String()
 	return nil
 }
+
+func ensureKRMFunctions(pv *api.PackageVariant,
+	prr *porchapi.PackageRevisionResources) error {
+
+	if len(pv.Spec.Mutators) == 0 {
+		return nil
```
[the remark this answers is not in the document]
johnbelamaric: We can't just return here - we need to make sure that there are no functions we added to the pipeline in previous reconciliations in the Kptfile. That is, we may be empty now but we may not have been before; if that's the case, we need to make sure we remove the out-of-date functions.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -777,3 +782,158 @@ func getFileKubeObject(prr *porchapi.PackageRevisionResources, file, kind, name
...
+		if err != nil {
+			return err
+		}
+		newPVValidators = append(newPVValidators, &val.SubObject)
+	}
+
+	newValidators = append(newPVValidators, newValidators...)
+
+	// update kptfile
+	if len(newMutators) > 0 {
```
johnbelamaric, line 877: Won't this leave the old ones in place if there are none anymore?

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -777,3 +782,158 @@ func getFileKubeObject(prr *porchapi.PackageRevisionResources, file, kind, name
...
+		mut, err := fn.NewFromTypedObject(newMutator)
+		if err != nil {
+			return err
+		}
+		newPVMutators = append(newPVMutators, &mut.SubObject)
+	}
+
+	newMutators = append(newPVMutators, newMutators...)
+
+	// generate new validators
```
johnbelamaric, line 842: I think you may be able to do mutators and validators in a  loop:
```
for _, field in range []string{"mutators", "validators"} {
...
}
```

That has its pros and cons, but seems do-able if you want to.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -777,3 +782,158 @@ func getFileKubeObject(prr *porchapi.PackageRevisionResources, file, kind, name
...
+		return false, nil
+	}
+
+	// if pv-names don't match, return false
+	if name[1] != pvName {
+		return false, nil
+	}
+
+	// make sure func name is not empty
+	if name[2] == "" {
```
johnbelamaric, line 925: I think it can be empty? The user doesn't need to add a name in the PV function name field, so this would seem to be a problem if they didn't. See the "(if present)" here: https://github.com/GoogleContainerTools/kpt/blob/main/docs/design-docs/08-package-variant.md#kptfile-function-pipeline-editing

Or am I missing something, perhaps there is a need for this or you put a default value in ?

johnbelamaric (changes requested):
Awesome, looking pretty good.

johnbelamaric (changes requested):
Awesome, looking pretty good.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
SimonTheLeg: I see. Yes should be fine. 

Just to clarify the generated name would then look like this right?: `Packagevariant.my-pv..0`

johnbelamaric: Either that or `PackageVariant.my-pv.0` works for me. Just make sure the design is consistent with the implementation.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -768,3 +774,57 @@ func ensurePackageContext(pv *api.PackageVariant,
...
 }
+
+func ensureKRMFunctions(pv *api.PackageVariant,
+	prr *porchapi.PackageRevisionResources) error {
+
+	if len(pv.Spec.Mutators) == 0 {
+		return nil
+	}
+
+	if _, ok := prr.Spec.Resources[kptfilev1.KptFileName]; !ok {
```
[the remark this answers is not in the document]
SimonTheLeg: perfect, switched to it

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
SimonTheLeg: I really like the idea, but I am unsure how to make it viable in Go. The line I am struggling to make work inside the loop is:

```Go
for i, mutator := range pv.Spec.Pipeline.Mutators {
```

Where a different field of pv.Spec.Pipeline should be taken depending on the value of the loop variable. The only solution I could really come up with would include a switch-case statement or some reflection, which I find rather ugly and not nice to maintain.

If you know of any elegant solution for this, I am very happy to change this part :)

johnbelamaric: Well...not sure if it's *elegant*, but here's a solution:

```go
 funclists := map[string][]kptfilev1.Function{
        {"validators": pv.Spec.Pipeline.Validators},
        {"mutators": pv.Spec.Pipeline.Mutators},
}

for field, funclist := range funclists {
...do your thing...
}

```

SimonTheLeg:
okay now everything should be resolved and ready for review

docs/design-docs/08-package-variant.md
```diff
@@ -513,7 +513,7 @@ Then the resulting Kptfile will have these two entries prepended to its
     - image: gcr.io/kpt-fn/set-labels:v0.1
       configMap:
         app: foo
-      name: PackageVariant.my-pv.1
+      name: PackageVariant.my-pv..1
```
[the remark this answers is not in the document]
johnbelamaric: yeah, that's a system generated name, and that particular component is the empty string

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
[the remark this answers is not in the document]
johnbelamaric: If you create a unit test case where the existing pipeline has some functions with our names, and the PV has an empty pipeline, the resulting pipeline should have all the function with our names removed. But I think that test case would fail with this line here.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -851,6 +856,123 @@ func getFileKubeObject(prr *porchapi.PackageRevisionResources, file, kind, name
 	return ko, nil
 }
 
+// TODO known issues:
```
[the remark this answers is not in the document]
SimonTheLeg: That is a good point: I have reworded the section to specify the current behaviour of the func and removed the TODO and known issues section

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -851,6 +856,123 @@ func getFileKubeObject(prr *porchapi.PackageRevisionResources, file, kind, name
...
+// - will move the newly generated name to the top of the mutator sequence
+// - does not preserve indent style. Instead will format if there are mutators/validators in pv, but will preserve if there are none.
+// - will not throw an error if the initial kptfile contains empty pipeline key
```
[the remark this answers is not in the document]
SimonTheLeg: yes you are right. I guess this is a rather theoretical case.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller_test.go
```diff
@@ -1134,3 +1135,409 @@ spec:
...
+spec:
+  upstream:
+    repo: blueprints
+    package: foo
+    revision: v1
+  downstream:
+    repo: deployments
+    package: bar
+  pipeline:
+`[1:]
```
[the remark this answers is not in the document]
SimonTheLeg: Yes normally strings.TrimSpace works for this. Unfortunately it does not work in our case, as the second line is indented with spaces. This will result in trimming out the spaces of that line:


```Go
func Test(t *testing.T) {
	s := `
  second-line indented:
    more indented:
`
	fmt.Println(strings.TrimSpace(s))
	fmt.Println("------")
	fmt.Println(s[1:])
}

// will produce
second-line indented:
    more indented:
------
  second-line indented:
    more indented:
```

But I am also going to look into the idea of your other comment to make use of kyaml, which maybe makes this obsolete all together.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -851,6 +856,122 @@ func getFileKubeObject(prr *porchapi.PackageRevisionResources, file, kind, name
...
+			if err := pipeline.SetSlice(newFieldVal, fieldname); err != nil {
+				return err
+			}
+		} else {
+			if _, err := pipeline.RemoveNestedField(fieldname); err != nil {
+				return err
+			}
+		}
+	}
+
```
johnbelamaric, line 926: If the there are no mutators or validators after this, then we should remove the "pipeline" map altogether.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller_test.go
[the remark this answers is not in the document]
mortent: Not a blocker for this PR. It is just a bit difficult to read with the indentation.

mortent (approved):

PR #3926, opened 2023-04-24
bump kpt to go1.20

yuwenma:
Bump the kpt go from 1.19 to 1.20


.github/workflows/e2eEnvironment.yml
```diff
@@ -29,12 +29,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: ["1.19", "1.20", "1.21", "1.22", "1.23", "1.24"]
+        version: ["1.20", "1.20", "1.21", "1.22", "1.23", "1.24"]
```
mortent, line 32: This should not be updated.

mortent (commented):
This doesn't seem to include changing the version for the goreleaser-cross image?

.github/workflows/e2eEnvironment.yml
yuwenma: missed double checking the .github dir.

mortent (approved):

PR #3930, opened 2023-04-25
Implement PackageVariantSet v1alpha2

johnbelamaric:
This PR implements the PackageVariantSet design described in https://github.com/GoogleContainerTools/kpt/blob/main/docs/design-docs/08-package-variant.md.


johnbelamaric:
I'd like this to start getting reviews. A few notes:
* I think the test failure is a flake?
* This does not yet implement conversion webhooks from v1alpha1 to v1alpha2. I think that's OK, maybe we should punt on that. There is some conversion code but it does not cover the entire API, and getting the webhook creation and setup is a fair bit of work for not much benefit on an alpha that is not used a lot yet. We'll have to release note the change. I can update the existing Nephio references.

porch/controllers/packagevariantsets/api/v1alpha2/packagevariantset_conversion.go
```diff
@@ -0,0 +1,100 @@
...
+	"fmt"
+
+	"sigs.k8s.io/controller-runtime/pkg/conversion"
+
+	pkgvarapi "github.com/GoogleContainerTools/kpt/porch/controllers/packagevariants/api/v1alpha1"
+	v1alpha1 "github.com/GoogleContainerTools/kpt/porch/controllers/packagevariantsets/api/v1alpha1"
+)
+
+func (src *PackageVariantSet) ConvertTo(dstRaw conversion.Hub) error {
+	dst := dstRaw.(*v1alpha1.PackageVariantSet)
```
natasha41575: Is my understanding correct that we won't be using this code until we add the conversion webhooks? If that's the case (and if we are not adding the conversion webhooks in this PR), might it make more sense to leave these files out of this PR, and add this code back in with the conversion webhooks?

johnbelamaric: Yes, that's correct. I could do that.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/render.go
```diff
@@ -0,0 +1,315 @@
...
+	opts = append(opts, cel.Variable(UpstreamVarName, cel.DynType))
+	opts = append(opts, cel.Variable(TargetVarName, cel.DynType))
+	opts = append(opts, cel.Variable(RepositoryVarName, cel.DynType))
+
+	env, err := cel.NewEnv(opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	ast, issues := env.Compile(expr)
```
jpbetz, line 301: Looks great. Once https://github.com/kubernetes/kubernetes/pull/116779 merges, you'll also have the option to more easily use a base environment provided by k8s (which as a few more libs and will evolve over time).  You'd also get some versioning support, if that is useful to you.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -90,349 +90,159 @@ func (r *PackageVariantSetReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-		})
+		setStalledConditionsToTrue(pvs, "ValidationError", combineErrors(errs))
+		return ctrl.Result{}, nil
 	}
 
-	upstream, err := r.getUpstreamPR(pvs.Spec.Upstream)
+	upstreamPR, err := r.getUpstreamPR(pvs.Spec.Upstream, prList)
 	if err != nil {
+		setStalledConditionsToTrue(pvs, "ValidationError", err.Error())
 		return ctrl.Result{}, err
```
mortent, line 100: Should we return the error here? I would expect that if we set the stalled condition, we don't want to trigger an immediate new reconcile.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -90,349 +90,159 @@ func (r *PackageVariantSetReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-	}
-
-	suffix, err := r.fetchValue(targetName.NameSuffix, obj)
-	if err != nil {
-		return "", err
-	}
+		for _, u := range uList.Items {
+			result = append(result, pvContext{
+				template:       target.Template,
+				repoDefault:    u.GetName(),
```
mortent, line 238: If selectors matches arbitrary objects, is it still correct to use the name of the resource here?

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/render.go
johnbelamaric: Cool, thanks for taking a look, Joe.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
johnbelamaric: Yes; at least, that is how it is specified in the design :).

This is just the default. The user can override it using an expression based on the object (for example, `target.annotations['repo-name']`). That comes later in the PackageVariantTemplate.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
johnbelamaric: When we return the error, that means reconcile again with exponential back off, is that correct?

So, in this case, we may just be waiting for the upstream to appear. So, we are stalled in the sense that some user action is needed; but it's not user action against *this* resource.

On other hand, it looks like right now we trigger a complete reconciliation of all PVSs in a namespace whenever there is a change to any PackageRevision. So, we should get triggered without any need for re-queueing, just by the nature of watching PackageRevisions. I wasn't sure though if there could be race conditions that this might resolve.

WDTY?

johnbelamaric: I made this change and add an explanatory note.

porch/controllers/packagevariantsets/api/v1alpha2/packagevariantset_conversion.go
johnbelamaric: Done.

johnbelamaric:
Rebased

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -90,349 +90,161 @@ func (r *PackageVariantSetReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-		})
-	}
-	return result, nil
-}
-
-func (r *PackageVariantSetReconciler) getSelectedObjects(ctx context.Context, selectors []api.Selector) (map[resid.ResId]*yaml.RNode, error) {
-	selectedObjects := make(map[resid.ResId]*yaml.RNode) // this is a map to prevent duplicates
-
-	for _, selector := range selectors {
+		// a selector against a set of arbitrary objects
```
natasha41575, line 213: Is the rest of this supposed to be under a `if objSel != nil`? 

That, or maybe you're missing a `continue` under the `target.RepositorySelector` condition, as I see you have a `continue` under `target.Repositories`?

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -90,349 +90,161 @@ func (r *PackageVariantSetReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-			if err != nil {
-				return nil, err
-			}
-			opts = append(opts, client.MatchingLabelsSelector{Selector: labelSelector})
+		opts := []client.ListOption{client.InNamespace(pvs.Namespace)}
+		labelSelector, err := metav1.LabelSelectorAsSelector(&objSel.LabelSelector)
+		if err != nil {
+			return nil, err
 		}
+		opts = append(opts, client.MatchingLabelsSelector{Selector: labelSelector})
```
natasha41575, line 227: if `objSel.LabelSelector` is empty, it should presumably match everything. Is that what happens?

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
johnbelamaric: No, it's correct. It's a little hard to tell because of all the change, but the logic of the interior of the loop is this:

 - if it's "repositories", handle it and continue
 - if it's repository selector, do the same as "object selector" but set GVK to be a porch Repository

So there are really only two sets of logic - repositories and object selectors, with repository selector just setting the default object info

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
johnbelamaric: Good question. I tried it out, and yes, that's what happens. But while trying it out I found another bug, so pushed a fix for that!

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
natasha41575: Ah got it! Thank you for the explanation.

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
mortent: It seems like triggering reconciliation of all PVSs whenever any PackageRevision has a change seems unnecessary, but probably not a big concern until we have some scale. We do need to have watch on the PackageRevision resource, but we can probably be more selective of which PVSs gets reconciled for each PackageRevision.

johnbelamaric: Yeah, I concur with that.

natasha41575 (commented):
LGTM, but I think we should also get @mortent's approval on this one.

I appreciate the thorough unit tests btw!

mortent (approved):
Looks good. We have completed this months releases of kpt and Porch, so feel free to merge this.

docs/design-docs/08-package-variant.md
```diff
@@ -1183,16 +1226,16 @@ them. That is, there are certain objects that are accessible within the CEL
...
-| repo         | The default repository name based on the targeting criteria. |
-| package      | The default package name based on the targeting criteria.    |
-| upstream     | The upstream PackageRevision.                                |
-| repository   | The downstream Repository.                                   |
+| CEL Variable   | Variable Contents                                            |
+| -------------- | ------------------------------------------------------------ |
+| repoDefault    | The default repository name based on the targeting criteria. |
+| packageDefault | The default package name based on the targeting criteria.    |
+| upstream       | The upstream PackageRevision.                                |
+| repository     | The downstream Repository.                                   |
```
jpbetz, line 1234: Add `target` here too?

johnbelamaric: It's described in the paragraph below, because it's contented varies based on the type of target. But I will add it here in a follow up PR.

PR #3938, opened 2023-04-28
porch CLI: error instead of panic when the upstreamLock is not parseable

natasha41575:
Throws an error instead of panicking in cases like https://github.com/GoogleContainerTools/kpt/issues/3933.

Sample output: 

```
$ kpt alpha rpkg update --discover=downstream
Error: could not parse upstreamLock in Kptfile of package "blueprints-8c50f85abaa5d22110b7c826bcb77f8f33d39dbb": malformed upstreamLock.Git.Ref "v2"
```

When using porch CLI, the upstreamLock.Git.Ref should be in the format of either `drafts/pkgname/workspace` or `pkgname/version`. So we theoretically shouldn't run into any cases where there is no `/` in upstreamLock.Git.Ref unless someone makes a change through something other than porch CLI. In these cases, we should throw an error that we are unable to parse the upstreamLock rather than panicking.


mortent (approved):
Maybe we can add a test here, but otherwise looks good.

PR #3939, opened 2023-05-02
Implement configuration injection in PackageVariant

johnbelamaric:
Implement Configuration Injection as described in https://github.com/GoogleContainerTools/kpt/blob/main/docs/design-docs/08-package-variant.md

Fixes https://github.com/nephio-project/nephio/issues/43


porch/controllers/packagevariants/pkg/controllers/packagevariant/injection.go
```diff
@@ -0,0 +1,425 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	ConfigInjectionAnnotation  = "kpt.dev/config-injection"
+	InjectedResourceAnnotation = "kpt.dev/injected-resource"
+)
+
+func injectionConditionType(object *fn.KubeObject) string {
+	return strings.Join([]string{"config", "injection", object.GetKind(), object.GetName()}, ".")
```
natasha41575, line 42: Can't remember if we discussed this elsewhere, but did we decide against including the apiVersion of the object here? I don't have a strong opinion either way, just noticed that it is omitted

porch/controllers/packagevariants/pkg/controllers/packagevariant/injection.go
```diff
@@ -0,0 +1,425 @@
...
+	return ip
+}
+
+func ensureConfigInjection(ctx context.Context,
+	c client.Client,
+	pv *api.PackageVariant,
+	prr *porchapi.PackageRevisionResources) error {
+
+	if prr.Spec.Resources == nil {
+		return fmt.Errorf("nil resources found for PackageRevisionResources '%s/%s'", prr.Namespace, prr.Name)
```
natasha41575, line 96: For this, in the interest of failing earlier rather than later, could we move this check out to `packagevariant_controller.go` right after we fetch the PackageRevisionResources (L753) , rather than waiting until we get all the way here to throw an error? 

fwiw, I see that `hashFromPackageRevisionResources` also uses `prr.Spec.Resources`

porch/controllers/packagevariants/pkg/controllers/packagevariant/injection.go
```diff
@@ -0,0 +1,425 @@
...
+			Type:    c.Type,
+			Reason:  c.Reason,
+			Status:  metav1.ConditionStatus(c.Status),
+			Message: c.Message,
+		})
+	}
+	return result
+}
+
+var matchResourceContents = append(kio.MatchAll, kptfilev1.KptFileName)
```
natasha41575, line 392: could you add a comment explaining what this does?

natasha41575 (commented):
General code structure and logic makes sense to me. Some drive-by comments

porch/controllers/packagevariants/pkg/controllers/packagevariant/injection.go
johnbelamaric: Yeah, I want these values to be stable across versions. The Kind+Name is sufficient for the package author to avoid ambiguity in the condition types.

porch/controllers/packagevariants/pkg/controllers/packagevariant/injection.go
johnbelamaric: Yeah, this should move to a util function. We've copied it in a few places, I think.

porch/controllers/packagevariants/pkg/controllers/packagevariant/injection.go
johnbelamaric: Sure, good idea.

porch/controllers/packagevariants/pkg/controllers/packagevariant/injection.go
johnbelamaric: for now I added a comment and a TODO. There is some refactoring that can be done with all of this to make it nicer, this would be part of it. But I'd rather do that in a separate PR.

porch/controllers/packagevariants/pkg/controllers/packagevariant/fake_client.go
```diff
@@ -42,3 +44,69 @@ func (f *fakeClient) Update(_ context.Context, obj client.Object, _ ...client.Up
 	f.output = append(f.output, fmt.Sprintf("updating object: %s", obj.GetName()))
 	return nil
 }
+func (f *fakeClient) List(_ context.Context, obj client.ObjectList, _ ...client.ListOption) error {
```
natasha41575, line 47: super nit: add a newline above this

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -328,32 +348,66 @@ func (r *PackageVariantReconciler) findAndUpdateExistingRevisions(ctx context.Co
...
 	for i, downstream := range downstreams {
-		if r.isUpToDate(pv, downstream) {
-			if downstream.Spec.Lifecycle == porchapi.PackageRevisionLifecycleDeletionProposed {
-				// We proposed this package revision for deletion in the past, but now it
-				// matches our target, so we no longer want it to be deleted.
-				downstream.Spec.Lifecycle = porchapi.PackageRevisionLifecyclePublished
-				if err := r.Client.Update(ctx, downstream); err != nil {
-					klog.Errorf("error updating package revision lifecycle: %v", err)
-				}
+		if downstream.Spec.Lifecycle == porchapi.PackageRevisionLifecycleDeletionProposed {
```
natasha41575, line 354: Was there a reason to move this out of the `isUpToDate` condition? If the package revision with lifecycle `DeletionProposed` is not up to date and needs updating, then we are going to make and update a new draft, so I don't think it's needed to update the lifecycle of the old existing package revision?

natasha41575 (approved):
A nit and a minor discussion point that I could go either way on, but looks good to me other than that.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
johnbelamaric: Good question, I meant to ask you to give the changes in this method particular attention.

I was working with the understanding that LifecycleIsPublished would return false for those with deletion proposed, but now that you ask it, maybe that's not true? I was concerned that we would end up cloning a deletion proposed package, and that the published version would then remain in deletion proposed even though it's not supposed to.

My thinking was that we needed to clear this state since it's no longer proposed for deletion, regardless of what else needs to be done to it, and that it was safe to do that even if it's not up to date. That did require saving it before the clone. 

Probably we should have some sort of e2e test or something to see if this works the way we think? Do we have an easy way to do this sort of complex test?

natasha41575: Yeah, LifecycleIsPublished returns true for `Published` or `DeletionProposed`. Those two lifecycles are treated pretty much the same everywhere, with the only exception being the package revision deletion webhook that disallows Published packages to be deleted.

> I was concerned that we would end up cloning a deletion proposed package, and that the published version would then remain in deletion proposed even though it's not supposed to.

The previous logic did copy a deletion proposed package, but the copied version will start out in draft state and will move through the normal package revision lifecycles of Draft -> Proposed -> Published -> DeletionProposed, requiring a client to move it to the next lifecycle at each step, so the copied version shouldn't get to the deletionProposed stage unless someone explicitly changed it to that. 

> My thinking was that we needed to clear this state since it's no longer proposed for deletion, regardless of what else needs to be done to it, and that it was safe to do that even if it's not up to date. That did require saving it before the clone.

Got it. So we are thinking of the proposedDeletion package revision as the target, and therefore we don't want it to be deleted anymore. Copying it later is just part of updating it. I agree that it is safe and I think your reasoning is correct, so your changes make sense. 

> Probably we should have some sort of e2e test or something to see if this works the way we think? Do we have an easy way to do this sort of complex test?

Unfortunately we don't have e2e tests set up for these controllers. We should try to set something up at some point rather than relying on developers to do a bunch of manual testing, but we don't have a concrete plan for that. We do have complex tests for the experimental rollouts controllers (https://github.com/GoogleContainerTools/kpt/blob/565305a20af7ff443d4ec04a98aa37cfcfd08c59/rollouts/controllers/rollout_controller_test.go#L36), so we might be able to draw some inspiration from there.

PR #3945, opened 2023-05-08
Fix the pipeline diagram

johnbelamaric:
There was an error in the PackageVariant definition in the diagram, this fixes it.


natasha41575 (approved):

natasha41575:
Nice catch!

PR #3952, opened 2023-05-10
Implement PackageVariantSet injector template

johnbelamaric:
Since the config injection feature of PackageVariant has been implemented, we need to extend PVS to support setting the injectors field in the PV.

### DIFF
--- a/docs/design-docs/08-package-variant.md
+++ b/docs/design-docs/08-package-variant.md
@@ -957,7 +957,7 @@ type PackageVariantTemplate struct {
 
         // Injectors allows specifying the spec.Injectors field of the generated PackageVariant
         // +optional
-        Injectors     *InfectionSelectorTemplate `json:"injectors,omitempty"`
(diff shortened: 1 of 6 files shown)


johnbelamaric:
@natasha41575 can you review please?

@SimonTheLeg, after your KRM PR merges, we'll need a PR similar to this to allow Pipeline in the PVS. For now, we can skip doing any CEL expressions for Pipeline, though, until we see a real need for it. So instead we would just use a static pipeline in the PVS.

The only real CEL usage I can imagine wanting in the future for Pipeline would be for function.configMap to become a MapExpr instead of a map[string]string.  It's going to be a bit painful, because your template basically has to clone the whole `Pipeline` and just change the type of that one field. That's why I think it could wait until we see a real need. We already can muck with Package Context which can be a config map input to a function...though it's limited to just one and has certain reserved fields already.

mortent (approved):

PR #3956, opened 2023-05-11
Fix missing injectors field in PVS CRD

johnbelamaric:
Looks like I left this out of the recent PR.

### DIFF
--- a/porch/controllers/config/crd/bases/config.porch.kpt.dev_packagevariantsets.yaml
+++ b/porch/controllers/config/crd/bases/config.porch.kpt.dev_packagevariantsets.yaml
@@ -316,7 +316,7 @@ spec:
                 type: array
             type: object
         type: object
-    served: false
+    served: true
     storage: false
     subresources:
       status: {}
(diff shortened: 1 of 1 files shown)


natasha41575 (approved):

porch/controllers/config/crd/bases/config.porch.kpt.dev_packagevariantsets.yaml
```diff
@@ -316,7 +316,7 @@ spec:
                 type: array
             type: object
         type: object
-    served: false
+    served: true
```
yuwenma, line 319: IIRC, this CRD is autogenerated by kubebuilder. So instead of changing the CRD itself, we should make the inline change so in the future when upgrading the CRD this config won't be flipped back.
  
Basically adding ` +kubebuilder:unservedversion` in https://github.com/GoogleContainerTools/kpt/blob/fca5a9170de2808197789e6ef5ff384ca67a7dd4/porch/controllers/packagevariants/api/v1alpha1/packagevariant_types.go#L21
@johnbelamaric

johnbelamaric: Thanks! I will fix that.

PR #3962, opened 2023-05-15
feat: add windows binary to goreleaser.yaml

nadworny:
Support windows binary


mortent:
@nadworny I'm not necessarily against distributing a Windows binary of kpt, but I think we first would have to make sure we have automated tests for Windows. Most of the contributors to kpt doesn't use Windows and don't have easy access to Windows machines for testing during development.

nadworny:
hey @mortent , same here ;) but we have just a few Windows users that would like to use it. I also don't have any experience running Windows tests on Github so not sure how to proceed.

liamfallon:
We have no plans to provide Windows binaries at the moment. See the comment from @mortent
https://github.com/kptdev/kpt/pull/3962#issuecomment-1553480718

PR #3963, opened 2023-05-15
fix: cache directory name too long on windows

nadworny:
New implementation allows to use kpt on windows without getting "Name too long" since the previous implementation would generate very long directory names.

### DIFF
--- a/internal/gitutil/gitutil.go
+++ b/internal/gitutil/gitutil.go
@@ -27,6 +27,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
 	"time"
 
--- a/internal/gitutil/gitutil.go
+++ b/internal/gitutil/gitutil.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/md5"
+	"encoding/base32"
 	"encoding/hex"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


internal/gitutil/gitutil.go
```diff
@@ -323,10 +323,11 @@ func (gur *GitUpstreamRepo) ResolveRef(ref string) (string, bool) {
 }
 
 // getRepoDir returns the cache directory name for a remote repo
-// This takes the md5 hash of the repo uri and then base32 encodes it to make
+// This takes the md5 hash of the repo uri and then hex encodes it to make
 // sure it doesn't contain characters that isn't legal in directory names.
 func (gur *GitUpstreamRepo) getRepoDir(uri string) string {
-	return strings.ToLower(base32.StdEncoding.EncodeToString(md5.New().Sum([]byte(uri))))
+	var hash = md5.Sum([]byte(uri))
```
mortent, line 329: So changing this means we will no longer find existing cached repositories. Would it be an option to check the architecture here and only use the shorter names for windows?

mortent (commented):
Thanks for the contribution. Just one question about the backwards compatibility of this change.

internal/gitutil/gitutil.go
nadworny: done

nadworny:
@mortent not sure if you're aware of it but the current implementation is "broken" and doesn't really generate md5
`strings.ToLower(base32.StdEncoding.EncodeToString(md5.New().Sum([]byte(uri))))`

mortent:
> @mortent not sure if you're aware of it but the current implementation is "broken" and doesn't really generate md5 `strings.ToLower(base32.StdEncoding.EncodeToString(md5.New().Sum([]byte(uri))))`
> 
> it's not part of this PR but maybe it would be worth creating an issue for it.

Yeah, if there is an issue, please file a bug.

nadworny:
hi @mortent , is there anything missing here?

mortent:
@nadworny This looks good. Sorry it took some time, I was out on vacation.

mortent (approved):

PR #3978, opened 2023-05-26
Rollouts: Added support for OCI packages

droot:
This PR adds support for rolling out packages from an OCI registry. Currently it accepts a static OCI package instead of discovering packages (and variants) in the OCI registry.

Few more changes to make it easy to invoke rollouts API as part of bigger system. For example:

RemoteSync objects are now named as `<target-cluster-name>-<rollout-name>` and the external sync objects `Root/RepoSync` are named same as the parent `rollout-name`. So how does it help ?
All of this to make it easy for git-syncer (`config-sync` in this case) to use the workload identity to authenticate with OCI. The git-syncer's reconciler is based on the `root/repoSync's` name so now users can control that by assigning proper rollout name and binding the GCP SA with the corresponding KSA (`root-reconciler-<rollout-name>`).


rollouts/api/v1alpha1/rollout_types.go
```diff
@@ -107,17 +107,21 @@ type ClusterSourceKind struct {
...
+// +kubebuilder:validation:Enum=GitHub;GitLab;OCI
 type PackageSourceType string
 
 // PackagesConfig defines the packages the Rollout should deploy.
 type PackagesConfig struct {
 	SourceType PackageSourceType `json:"sourceType"`
 
-	GitHub GitHubSource `json:"github,omitempty"`
-	GitLab GitLabSource `json:"gitlab,omitempty"`
+	// TODO(droot): Change Github and Gitlab to pointers because
```
mortent, line 120: As long as we are using a discriminator here, it probably doesn't matter that much if they are pointers or not. Nil objects should be fine, since we will be using the discriminator value to determine which one to use rather than check for nil.

rollouts/api/v1alpha1/rollout_types.go
```diff
@@ -157,6 +161,24 @@ type GitLabSelector struct {
...
+	// The image can be pulled by TAG or by DIGEST if it is specified in PACKAGE_NAME.
+	// - Pull by tag: `LOCATION-docker.pkg.dev/PROJECT_ID/REPOSITORY_NAME/PACKAGE_NAME:TAG`.
+	// - Pull by digest: `LOCATION-docker.pkg.dev/PROJECT_ID/REPOSITORY_NAME/PACKAGE_NAME@sha256:DIGEST`.
+	// If neither TAG nor DIGEST is specified, it pulls with the `latest` tag by default.
+	// Required
+	Image string `json:"image"`
+
+	// dir is the absolute path of the directory that contains
+	// the local resources.  Default: the root directory of the image.
+	// Note (droot): We will extend `Dir` to express variants of a package at some point.
```
mortent, line 177: So this doesn't yet take into account variants right? Similar to with git, there are probably several ways for users to organize packages and variants across one or more images, so I would expect we eventually need something similar to the selector pattern from the github/gitlab sources. Do we think we can add this later without needing backwards incompatible changes to the API?

natasha41575 (approved):
lgtm as a POC, though I have a +1 to @mortent's question above about extending this to support variants

rollouts/api/v1alpha1/rollout_types.go
droot: Though the current schema is extensible by supporting glob patterns for `image` and `dir` fields, but the naming was inconsistent with the other package sources `github` and `gitlab`. Changed the schema to add `OCISelector` field which is now consistent with other selectors. We can extend `image` and `dir` to support glob patterns in the future to support variants (that will also require changes to package-matcher component as well).

rollouts/api/v1alpha1/rollout_types.go
droot: You are right. Though, `nil` does help with the serialization of `omitempty` semantics.

droot:
@mortent @natasha41575 updated the PR to address the comments, PTAL. Thanks.

mortent (approved):

PR #3983, opened 2023-06-03
Make sure all access to deletionProposedCache is protected

mortent:
The deletionProposedCache is part of the gitRepository, but there were some places where it was accessed without holding the lock.
In general, we should split up the repository vs package objects into separate go packages so it is easier to enforce access to these objects only through the public functions that make sure any access happens while holding the locks.

Fixes: https://github.com/GoogleContainerTools/kpt/issues/3967


mortent:
This turned out to be a larger issue than expected. The dependencies between `gitRepository`, `gitPackageRevision`, and `gitPackageDraft` is pretty complicated as they are all in the same module. This makes it very difficult to reason around which calls happens holding a lock and which ones doesn't. I will take a stab at cleaning this up in a separate PR.

johnbelamaric:
Does that mean this PR should not be merged? Or is it incremental? There are some CI failures.

PR #3984, opened 2023-06-06
Fix issue with unprotected concurrent access to gitRepository state

mortent:
This is a replacement for https://github.com/GoogleContainerTools/kpt/pull/3983. It moves the logic for fetching and updating the lifecycle of a PackageRevision from the `gitPackageRevision` object to the `gitRepository` object, as it involves manipulating the state of the `gitRepository`.

To make sure all access to the shared state is protected by the `gitRepository` mutex and no calls tries to take the lock twice, all calls from `gitRepository` will call versions of the functions that expects that the caller already has the lock, while calls from elsewhere call the versions of the functions that takes the lock.

### DIFF
--- a/porch/pkg/git/git.go
+++ b/porch/pkg/git/git.go
@@ -205,10 +205,18 @@ func (r *gitRepository) ListPackageRevisions(ctx context.Context, filter reposit
 	r.mutex.Lock()
 	defer r.mutex.Unlock()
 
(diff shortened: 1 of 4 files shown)


natasha41575 (approved):

mortent:
There are still some concurrency issues here, which is the cause of the failing e2e tests. I'm looking into it.

PR #3986, opened 2023-06-07
Change PV naming and fix pointer bug

### LINKED ISSUE #3982: Porch: PackageVariantSet deletes and recreates PackageVariants instead of updating
### Expected behavior
When a change is made to a PackageVariantSet that implies changes to the fields of a PackageVariant, but that does not imply a change in the repo or package name, the expectation is that the PVS will update the existing, previously created PackageVariant.

### Actual behavior
The PVS deletes the PV, and creates a new one, with a new name. This seems to be at least one cause of #3980, though I suspect not the only cause.

This also causes existing published versions to be flagged as DeletionProposed.

### Information
When an update is made to a PV, it will create a new Draft, based on the existing, most recent *published* version of the package, and incorporate the changes to the PV in that new Draft. If we instead delete and recreate, we lose any changes the user may have made to the Draft that were later published. So, delete & recreate can be quite destructive.

See #3980.

### LINKED ISSUE #3985: Porch: PackageVariantSet target.name expressions do not work as expected
### Expected behavior
When using target.name in an expression, the resulting PackageVariants end up with the value corresponding to that target.name.

### Actual behavior
When using target.name in an expression, all resulting PackageVariants end up with the same value.

### Information
@henderiw found this during Nephio testing.

johnbelamaric:
Fixes #3982 
Fixes #3985 

And adds unit tests covering those.

### DIFF
--- a/porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
+++ b/porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
@@ -63,6 +63,9 @@ const (
 
 	ConditionTypeStalled = "Stalled" // whether or not the resource reconciliation is making progress or not
 	ConditionTypeReady   = "Ready"   // whether or not the reconciliation succeeded
+
+	PackageVariantNameMaxLength  = 63
+	PackageVariantNameHashLength = 8
 )
 
 //go:generate go run sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0 rbac:roleName=porch-controllers-packagevariantsets webhook paths="." output:rbac:artifacts:config=../../../config/rbac
@@ -242,7 +245,7 @@ func (r *PackageVariantSetReconciler) unrollDownstreamTargets(ctx context.Contex
 				template:       target.Template,
 				repoDefault:    u.GetName(),
 				packageDefault: upstreamPackageName,
-				object:         &u,
+				object:         u.DeepCopy(),
 			})
 
 		}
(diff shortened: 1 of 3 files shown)


johnbelamaric:
@natasha41575 @mortent PTAL

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
```diff
@@ -315,14 +312,16 @@ func (r *PackageVariantSetReconciler) ensurePackageVariants(ctx context.Context,
...
 		}
-		desiredPackageVariantMap[hash] = &pv
+		desiredPackageVariantMap[pvId] = &pv
 	}
 
-	for existingPvHash, existingPV := range existingPackageVariantMap {
-		if _, found := desiredPackageVariantMap[existingPvHash]; found {
+	for existingPvId, existingPV := range existingPackageVariantMap {
+		if _, found := desiredPackageVariantMap[existingPvId]; found {
+			fmt.Println("Found existing in desired", existingPvId)
```
mortent, line 320: Nit: Use klog for logging.

mortent (approved):

porch/controllers/packagevariantsets/pkg/controllers/packagevariantset/packagevariantset_controller.go
johnbelamaric: Oops, that's debug, I'll remove it

henderiw:
I tested it a few times now and seems to do the job

PR #3993, opened 2023-06-30
Add resource version checks

### LINKED ISSUE #3991: Porch: concurrency issue
### Expected behavior
A client update to PackageRevision or PackageRevisionResources will be rejected if the version it is based on is no longer the current version in the server.

### Actual behavior
There does not seem to be any concurrency control. So, one client can overwrite another client's work. This is causing controller fights, and occasionally broken packages. See #3980 for some more details.

### Information
I think we can fix this if we use the commit hash as the RV in the server and client. So, normally controllers use optimistic concurrency, and let the server kick back any updates that are based on resources not matching the current rv. I don't think we are doing this in PackageRevision or PackageRevisionResources. The RV in most resources looks sequential, but it is explicitly NOT a sequence number. We should be able to use the commit hash for it.

So, when the client tries to write, we know if the resource on the server has changed since the client read the resource if the commit hashes don't match - then we can kick it back and the client can just reconcile again.

I think this would solve a number of issues we are seeing in Nephio.

johnbelamaric:
Adds ResourceVersion handling, including using an annotation to manage it for `rpkg pull` and `rpkg push`.

Fixes #3991

### DIFF
--- a/porch/pkg/engine/engine.go
+++ b/porch/pkg/engine/engine.go
@@ -131,12 +131,7 @@ func (p *PackageRevision) KubeObjectName() string {
 }
 
 func (p *PackageRevision) GetResources(ctx context.Context) (*api.PackageRevisionResources, error) {
-	prr, err := p.repoPackageRevision.GetResources(ctx)
-	if err != nil {
-		return prr, err
-	}
-
-	return prr, err
+	return p.repoPackageRevision.GetResources(ctx)
 }
 
 type Function struct {
@@ -383,11 +378,6 @@ func ensureSameOrigin(ctx context.Context, repo repository.Repository, obj *api.
 		// If there are no tasks, or the first task is not init or clone, then this revision was not
 		// created from another package revision. That means we expect it to be the first revision
 		// for this package.
-		if len(tasks) == 0 {
-			klog.Warningf("not same origin due to no tasks")
-		} else {
-			klog.Warningf("not same origin due to tasks[0].Type: %v", tasks[0].Type)
-		}
 		return false, nil
 	}
 
@@ -405,7 +395,6 @@ func ensureSameOrigin(ctx context.Context, repo repository.Repository, obj *api.
 	for _, rev := range existingRevs {
 		p, err := rev.GetPackageRevision(ctx)
 		if err != nil {
-			klog.Warningf("not same origin due to error getting packagerevision: %s", err.Error())
 			return false, err
 		}
 		revTasks := p.Spec.Tasks
@@ -451,7 +440,6 @@ func ensureSameOrigin(ctx context.Context, repo repository.Repository, obj *api.
 			case firstRevTask.Clone.Upstream.UpstreamRef != nil && firstObjTask.Clone.Upstream.UpstreamRef != nil:
 				revRepoPkgRev, found, err := getPackageRevision(ctx, repo, firstObjTask.Clone.Upstream.UpstreamRef.Name)
 				if err != nil {
-					klog.Warningf("not same origin due to error getting packagerevision to check upstreams: %s", err.Error())
 					return false, err
 				}
 				if !found {
@@ -460,7 +448,6 @@ func ensureSameOrigin(ctx context.Context, repo repository.Repository, obj *api.
 
 				objRepoPkgRev, found, err := getPackageRevision(ctx, repo, firstRevTask.Clone.Upstream.UpstreamRef.Name)
 				if err != nil {
-					klog.Warningf("not same origin due to error getting packagerevision to check upstreams 2: %s", err.Error())
 					return false, err
 				}
 				if !found {
@@ -478,7 +465,6 @@ func ensureSameOrigin(ctx context.Context, repo repository.Repository, obj *api.
 			}
 		}
 	}
-	klog.Warningf("not same origin due to falling through")
 	return false, nil
 }
 
--- a/porch/pkg/engine/engine.go
+++ b/porch/pkg/engine/engine.go
@@ -24,7 +24,6 @@ import (
 	"reflect"
 	"strings"
 
-	fnsdk "github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
 	"github.com/GoogleContainerTools/kpt/internal/builtins"
 	"github.com/GoogleContainerTools/kpt/internal/fnruntime"
 	kptfile "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
--- a/porch/pkg/engine/engine.go
+++ b/porch/pkg/engine/engine.go
@@ -55,8 +55,7 @@ import (
 var tracer = otel.Tracer("engine")
 
 const (
-	OptimisticLockErrorMsg    = "the object has been modified; please apply your changes to the latest version and try again"
-	ResourceVersionAnnotation = "internal.kpt.dev/resource-version"
+	OptimisticLockErrorMsg = "the object has been modified; please apply your changes to the latest version and try again"
 )
 
 type CaDEngine interface {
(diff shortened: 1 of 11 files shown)


johnbelamaric:
Also adds some instrumentation in `ensureSameOrigin`. Right now, in Nephio, our build has that function disabled. However, I am curious if the problems we saw with that function were due to packages getting corrupted by the concurrency issues, so I am leaving it enabled in this PR.

johnbelamaric:
Ok, I still get the error `Error: Internal error occurred: cannot create revision of free5gc-upf with a different origin than other package revisions in the same package` even with this fix. So we will still need to disable that for the Nephio build.

Should I just comment it out in this PR?

mortent:
So this adds the resourceVersion to the Kptfile. Is that to make sure it handles the resourceVersion comparison even if it goes through a `pull` and then later a `pull`?

I think there are three separate issues related to this and we should probably make sure we address them all (although not necessarily in this PR):
* Make sure we correctly check the resourceVersion for normal fields on the PackageRevision resource.
* Make sure we handle this correctly in terms of concurrency so we don't have races between concurrent updates to the same resource. I think this can be challenging due to the way we split data about each resource across both git/oci and a CRD. But we should at least do what we can to minimize the chance of issues here and make sure we understand the cases where it might not be handled correctly.
* Handle resourceVersion for updates to PackageRevisionResources (i.e. what this PR is about).

johnbelamaric:
> So this adds the resourceVersion to the Kptfile. Is that to make sure it handles the resourceVersion comparison even if it goes through a `pull` and then later a `pull`?

So, when using K8s clients, resourceVersion management works as expected simply by setting the metadata.ResourceVersion in the PackageRevision and PackageRevisionResources. However, if you do a `pull`, that metadata does not exist anywhere in the package. We do not store the source git commit ID from a pull, anywhere that I could see. So, in a `pull` I now stick it on an annotation in the Kptfile. But this is ephemeral; it is not stored in git. It is just added/removed by the API so that non-K8s clients (well, really just kpt) can work without modification.

Another option would be to have the kpt pull and push commands do this themselves. That may be preferable.

johnbelamaric:
> * Make sure we correctly check the resourceVersion for normal fields on the PackageRevision resource.

This is [done](https://github.com/GoogleContainerTools/kpt/pull/3993/files#diff-06b20ee91f5cb296f6aa28503b4b2630d3d8f1f17be68f679823a0a73367dcd3R650) in this PR as well.

> * Make sure we handle this correctly in terms of concurrency so we don't have races between concurrent updates to the same resource. I think this can be challenging due to the way we split data about each resource across both git/oci and a CRD. But we should at least do what we can to minimize the chance of issues here and make sure we understand the cases where it might not be handled correctly.

The resourceVersion check done in this PR should handle this in the happy path. The failure paths may be more challenging (if we make a change to the CR, then cannot change Git, or vice-versa, they may get out of sync).

There is nothing we can do about someone out-of-band modifying Git. We could put in a webhook to prevent someone from out-of-band modifying the `packagerev`.

johnbelamaric:
> Another option would be to have the kpt pull and push commands do this themselves. That may be preferable.

I wasn't sure how the UI did things. So, I figured the server-side workflow would work for the UI as well. But maybe it's better to assume that the clients are properly behaving K8s clients. In that case, it should be implemented client side (in push and pull). @ChristopherFry do you know, does the UI do the equivalent of the kpt pull and push commands, OR does it use a normal K8s client to pull out the package contents and push new package contents? Specifically, does it handle metadata on the PackageRevision and PackageRevisionResource objects normally?

mortent:
Yeah, I am also thinking that implementing the logic for including the resourceVersion as an annotation in the Kptfile would be better implemented client-side. That means that for Porch, the resourceVersion will always be available in the resourceVersion field from either the PackageRevision or PackageRevisionResources resources. Do you think there are any concerns about handling it that way?

johnbelamaric:
> Yeah, I am also thinking that implementing the logic for including the resourceVersion as an annotation in the Kptfile would be better implemented client-side. That means that for Porch, the resourceVersion will always be available in the resourceVersion field from either the PackageRevision or PackageRevisionResources resources. Do you think there are any concerns about handling it that way?

No, it's probably better.

johnbelamaric:
Ok, I think this is ready for review. I updated it to do the resource version annotation in pull and use it in push, removing it before pushing.

I haven't tried the UI yet.

johnbelamaric:
Ok, yeah, this break the UI. After a clone I get: "Internal error occurred: resourceVersion must be specified for an update". So, it's not using just the K8s API or if so it's not stashing away all the metadata. We'll need to drill into that.

porch/pkg/engine/engine.go
```diff
@@ -374,6 +383,11 @@ func ensureSameOrigin(ctx context.Context, repo repository.Repository, obj *api.
 		// If there are no tasks, or the first task is not init or clone, then this revision was not
 		// created from another package revision. That means we expect it to be the first revision
 		// for this package.
+		if len(tasks) == 0 {
+			klog.Warningf("not same origin due to no tasks")
```
mortent, line 387: Is it intentional that we add this logging in this PR?

porch/pkg/engine/engine.go
```diff
@@ -127,7 +131,12 @@ func (p *PackageRevision) KubeObjectName() string {
 }
 
 func (p *PackageRevision) GetResources(ctx context.Context) (*api.PackageRevisionResources, error) {
-	return p.repoPackageRevision.GetResources(ctx)
```
mortent, line 130: Is this change just a leftover from previous drafts of the PR?

porch/pkg/engine/engine.go
```diff
@@ -626,6 +644,15 @@ func (cad *cadEngine) UpdatePackageRevision(ctx context.Context, repositoryObj *
 	ctx, span := tracer.Start(ctx, "cadEngine::UpdatePackageRevision", trace.WithAttributes())
 	defer span.End()
 
+	newRV := newObj.GetResourceVersion()
+	if len(newRV) == 0 {
+		return nil, fmt.Errorf("resourceVersion must be specified for an update")
```
mortent, line 649: Is this the normal behavior for resources?

mortent (commented):
Just a few small questions, but looks good.

porch/pkg/engine/engine.go
johnbelamaric: ugh, yes. I guess it's equivalent

johnbelamaric: I can clean it up though, if you want

porch/pkg/engine/engine.go
johnbelamaric: I can remove it. This PR does not fix ensureSameOrigin; I think I will disable in a separate PR.

porch/pkg/engine/engine.go
mortent: Not critical, just wanted to make sure there wasn't something here I was missing.

mortent (approved):

efiacor:
> Ok, yeah, this break the UI. After a clone I get: "Internal error occurred: resourceVersion must be specified for an update". So, it's not using just the K8s API or if so it's not stashing away all the metadata. We'll need to drill into that.

Hi @johnbelamaric , do we have an issue to address this? Seeing it now on deploying packages via UI. Also, blocks new revision creation.

johnbelamaric:
The UI is fixed, we just need to use the new build in the install.

johnbelamaric:
Sorry I forgot to bump that. Will send a PR today

efiacor:
> Sorry I forgot to bump that. Will send a PR today

No problem. Thank you sir. I should learn to use cli more in fairness.

johnbelamaric:
https://github.com/nephio-project/nephio-packages/pull/29

Issue #3980, opened 2023-06-01
Porch: failure cloning a package due to "different origin" even though it is not a different origin


johnbelamaric:
### Expected behavior
I can successfully clone a package.

### Actual behavior
When trying to clone, I get a failure like "cannot create revision of test01-cluster with a different origin than other package revisions in the same package". I am still trying to narrow down the exact circumstances.

It's likely this has something to due with the the way I am testing - I am actually not sure if we should consider it a bug or not.

### Information

I have a repository `example-nephio-packages` which has some packages in it. I have a fork of that repository, which I have registered to Porch. It may have something do due with creating derivates of those packages in the fork. I am not sure yet. I am able to successfully clone packages from my fork to another repo, though.

johnbelamaric:
Actually, this is not during the *clone*. It's when PV controller tries to save the mutated Draft. So the clone works, but the Update doesn't.

natasha41575:
Do the "mutations" in the mutated draft change the initial clone task in the package revision? I think we compare clone tasks among package revisions in the same package to try to make sure that they're coming from the same upstream.

johnbelamaric:
Not intentionally. They just change packagerevisionresources actually. But I think there some concurrency issues going on. Since we do this in controllers, things happen way faster than they would manually. The process is basically:

 - clone
 - pull out package revision resources
 - tweak some values and save them

but the thing is, I often see errors for a while until things stabilize - like "could not find packagerevisionresources" and "removing tasks is not supported". So,  that first one seems to be happening because the reload after clone is not working or something. The second one perhaps is trying to save an outdated version of the package revision or packagerevisionresources (like, a stale cache entry and we are not catching that it is stale and kicking back the update attempt).

johnbelamaric:
At the same time this is all happening, there is another controller watching things. It used to make its own mutations, but I added a delay of 30s before it does anything, and that helped some.

Also, 80% of the time, these things "settle" and work eventually, but not 100% of the time.

johnbelamaric:
I backed the auto-approval controller off to wait 2 minutes (rather than 30s) before doing anything, and this is working a lot better so far.

natasha41575:
> I backed the auto-approval controller off to wait 2 minutes (rather than 30s) before doing anything, and this is working a lot better so far.

@mortent Could this be related to the other package approval issues that we've seen?

johnbelamaric:
Going to implement this which may help with noise, but there is still some issue here:

[...]

PR #4000, opened 2023-07-07
Disable ensureSameOrigin

johnbelamaric:
During our testing in Nephio, we have found that `ensureSameOrigin` seems to erroneously flag updates as invalid, so I think we should disable it for now.


mortent:
Do we know exactly what was causing the problem? I think the `ensureSameOrigin` logic had some potential problems and I'm less convinced about the usefulness of this than when we introduced it, but it would be helpful to understand exactly what happened.

johnbelamaric:
No, not yet. I did add some logging; it showed that in the cases we're talking about, it fell through to the very last return statement.

I can try to capture a specific package that fails. Maybe we can fix it. I disabled it in the Nephio build to work around the problem, but I can spend a little more time on it now.

johnbelamaric:
I added another package this happens with in #3980. But actually it seems to be happening with almost every package created via `clone`. If I clone it, I can't do a `copy`. I am wondering if we have an e2e test that does a clone-and-copy?

I think we probably want to merge this PR while we figure it out.

johnbelamaric:
Ok, based on https://github.com/GoogleContainerTools/kpt/issues/3980#issuecomment-1627905588 I have removed the ensureSameOrigin altogether, and flipped the default for replay strategy.

mortent:
This looks good to me. I also want to see if @natasha41575 has any input on this before we merge it.

johnbelamaric:
May be a legit failure do to replay strategy flip, let me take a closer look.

johnbelamaric:
Ok, yeah, fixed the legit failure and also fixed a flake in the PVS unit tests.

mortent:
I think we've given enough time for feedback on this one, so I think we are good to merge it.

mortent (approved):

PR #4001, opened 2023-07-07
PackageVariant debugging from Nephio

johnbelamaric:
This commit adds a workaround for an issue with Kptfile rendering that is causing a controller battle between PackageVariant controller and the Nephio controllers.

It also adds some instrumentation to help debug some issues with PackageVariant cloning.

### DIFF
--- a/porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
+++ b/porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
@@ -16,8 +16,6 @@ package packagevariant
 
 import (
 	"context"
-	"crypto/sha1"
-	"encoding/hex"
 	"flag"
 	"fmt"
 	"strconv"
@@ -29,6 +27,7 @@ import (
 
 	"github.com/GoogleContainerTools/kpt-functions-sdk/go/fn"
 	kptfilev1 "github.com/GoogleContainerTools/kpt/pkg/api/kptfile/v1"
+	"github.com/GoogleContainerTools/kpt/pkg/kptfile/kptfileutil"
 
 	"golang.org/x/mod/semver"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -42,7 +41,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
-	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
 type Options struct{}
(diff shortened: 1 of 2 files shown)


johnbelamaric:
e2e flaked out. Definitely see occasional issues with the packagerev finalizers, which then causes timeouts. I wonder if it is the test git server failing somehow.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
```diff
@@ -782,12 +785,73 @@ func (r *PackageVariantReconciler) calculateDraftResources(ctx context.Context,
...
+			// a file was deleted
+			klog.Infoln(fmt.Sprintf("PackageVariant %q, PackageRevision %q, resources changed: %q in original files, not in new files", pv.Name, prr.Name, k))
+			return &prr, true, nil
+		}
+
+		if newValue != v {
+			// HACK ALERT - TODO(jbelamaric): Fix this
+			// Currently nephio controllers and package variant controller are rendering Kptfiles slightly differently in YAML
+			// not sure why, need to investigate more. It may be due to different versions of kyaml. So, here, just for Kptfiles,
+			// we will parse and compare semantically.
```
droot, line 806: I asked chatgpt to help :) and it generated the following.


```Go

package main

import (
	"fmt"
	"reflect"

	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
	"k8s.io/apimachinery/pkg/util/yaml"
)

func main() {
	// Example YAML strings
	yamlString1 := `
apiVersion: v1
kind: Pod
metadata:
  name: pod-1
spec:
  containers:
  - name: container-1
    image: nginx:latest
  - name: container-2
    image: redis:latest
`
	yamlString2 := `
apiVersion: v1
kind: Pod
metadata:
  name: pod-2
spec:
  containers:
  - name: container-1
    image: nginx:latest
  - name: container-2
    image: redis:latest
`

	// Convert YAML strings to unstructured objects
	obj1, err := yaml.ToUnstructured([]byte(yamlString1))
	if err != nil {
		panic(err)
	}

	obj2, err := yaml.ToUnstructured([]byte(yamlString2))
	if err != nil {
		panic(err)
	}

	// Compare unstructured objects
	equal := reflect.DeepEqual(obj1, obj2)
	fmt.Println("Objects are equal:", equal)
}
```

droot (approved):
This looks good to me.

porch/controllers/packagevariants/pkg/controllers/packagevariant/packagevariant_controller.go
johnbelamaric: :)

PR #4011, opened 2023-07-24
chore: bump cli-utils

sdowell:
Several fixes have been made in cli-utils that we would like to incorporate downstream. This requires also bumping some of the shared dependencies, such as k8s apis.


natasha41575 (approved):

mortent:
Looks like the last run succeeded. If you remove any logging that we don't want to submit, we can try to get this merged.

PR #4013, opened 2023-07-28
Update to golang 1.20.6

mortent:
Update to golang 1.20.6 due to CVE-2023-29406

### DIFF
--- a/.github/workflows/checkLicenses.yml
+++ b/.github/workflows/checkLicenses.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Set up Go 1.20
         uses: actions/setup-go@v3
         with:
-          go-version: '>=1.20.5'
+          go-version: '>=1.20.6'
       - run: |
           ./scripts/create-licenses.sh
       # Upload the licenses list so it's available if needed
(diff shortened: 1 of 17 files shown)


.github/workflows/checkLicenses.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Set up Go 1.20
         uses: actions/setup-go@v3
         with:
-          go-version: '>=1.20.5'
```
justinsb, line 27: We might be able to use 'stable' here (for things where we are running go but not necessarily building artifacts): https://github.com/actions/setup-go#using-stableoldstable-aliases.  But OTOH that might be _more_ confusing than just having one, explicitly specified version.

justinsb (commented):
LGTM!

/approve
/lgtm

justinsb (approved):

.github/workflows/checkLicenses.yml
mortent: Yeah, I agree. But we have had quite a few CVE's come up about go, so on the other hand it is useful to be explicit about which version we are using.

PR #4014, opened 2023-08-02
Fix release script by removing release targets that require CGO

mortent:
This fixes the release process by removing the release targets that require CGO, in practice these are the WASM targets. There are two reasons for this:
* The complexity of the kpt release process is much higher because of the need to compile with CGO, as we are using ghcr.io/goreleaser/goreleaser-cross and do builds within docker containers on Github Actions. This process is currently broken, and this is a fix that addresses the immediate issue and simplifies the process for the future.
* The ghcr.io/goreleaser/goreleaser-cross image tends to be a bit slow to pick up new Go versions, which makes it hard for us to quickly get out new versions of kpt when there are CVEs in go.

Both of these issues can be fixed, but it requires some effort to figure out how to do it right.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,10 +52,15 @@ jobs:
         id: setup-release-env
         run: |-
           echo 'GITHUB_TOKEN=${{secrets.GITHUB_TOKEN}}' > .release-env
-      - name: Run GoReleaser within Docker
+      - name: Run GoReleaser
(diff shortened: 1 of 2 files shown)


droot:
Can we document instructions for users (in a Github Issue) to build their own kpt binary from a tag with CGO so that users who depend on WASM knows what to do in absence of the WASM release artifacts.

droot (approved):

PR #4018, opened 2023-08-08
Move resource version annotation to a different file

bharat-m1:
Moving resource version annotation to a new file. The newly added file can be used for storing all the metadata going forward. It is added and removed from the client side and will not be present in porch.

Please refer this for more info: https://github.com/GoogleContainerTools/kpt/issues/4017

### DIFF
--- a/commands/alpha/rpkg/pull/command.go
+++ b/commands/alpha/rpkg/pull/command.go
@@ -113,7 +113,7 @@ func (r *runner) runE(_ *cobra.Command, args []string) error {
 		return errors.E(op, err)
 	}
 
-	if err := util.AddResourceVersionAnnotation(&resources); err != nil {
+	if err := util.AddRevisionMetadata(&resources); err != nil {
 		return errors.E(op, err)
 	}
 
@@ -200,7 +200,7 @@ func createScheme() (*runtime.Scheme, error) {
 	return scheme, nil
 }
 
-var matchResourceContents = append(kio.MatchAll, kptfilev1.KptFileName)
+var matchResourceContents = append(kio.MatchAll, kptfilev1.KptFileName, kptfilev1.RevisionMetaDataFileName)
 
 func includeFile(path string) bool {
 	for _, m := range matchResourceContents {
--- a/commands/alpha/rpkg/pull/command_test.go
+++ b/commands/alpha/rpkg/pull/command_test.go
@@ -68,13 +68,24 @@ data:
 apiVersion: config.kubernetes.io/v1
 kind: ResourceList
 items:
+- apiVersion: porch.kpt.dev/v1alpha1
+  kind: KptRevisionMetadata
+  metadata:
+    name: repo-fjdos9u2nfe2f32
+    namespace: ns
+    creationTimestamp: null
+    resourceVersion: "999"
+    annotations:
+      config.kubernetes.io/index: '0'
+      internal.config.kubernetes.io/index: '0'
+      internal.config.kubernetes.io/path: '.KptRevisionMetadata'
+      config.kubernetes.io/path: '.KptRevisionMetadata'
 - apiVersion: kpt.dev/v1
   kind: Kptfile
   metadata:
     name: bar
     annotations:
       config.kubernetes.io/local-config: "true"
-      internal.kpt.dev/resource-version: "999"
       config.kubernetes.io/index: '0'
       internal.config.kubernetes.io/index: '0'
       internal.config.kubernetes.io/path: 'Kptfile'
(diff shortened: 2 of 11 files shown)


commands/alpha/rpkg/pull/command_test.go
```diff
@@ -68,13 +68,22 @@ data:
 apiVersion: config.kubernetes.io/v1
 kind: ResourceList
 items:
+- apiVersion: kpt.dev/v1
+  kind: KptRevisionMetaData
+  metadata:
+    name: bar
+    annotations:
+      internal.kpt.dev/resource-version: "999"
```
mortent, line 76: If we have a separate file, it seems a bit strange to include this information in an annotation. Have we considered putting this in a dedicated field in the spec?

commands/alpha/rpkg/pull/command_test.go
```diff
@@ -121,13 +130,22 @@ data:
 apiVersion: config.kubernetes.io/v1
 kind: ResourceList
 items:
+- apiVersion: kpt.dev/v1
+  kind: KptRevisionMetaData
```
mortent, line 134: I think using the KRM format probably make sense here and is aligned with what we have already done in kpt. But to stay consistent with the naming in Kubernetes, we should probably call this `KptRevisionMetadata`.

pkg/api/kptfile/v1/types.go
```diff
@@ -29,6 +29,10 @@ import (
 const (
 	KptFileName = "Kptfile"
 
+	RevisionMetaDataFileName = "KptRevisionMetaData"
```
mortent, line 32: What is the rationale behind this name? It seems somewhat long, but I'm not what would be a better name. We could also consider if we should use a hidden file for this (some something that starts with `.`.

commands/alpha/rpkg/pull/command_test.go
bharat-m1: Sure will update the name

pkg/api/kptfile/v1/types.go
bharat-m1: As the content in this file is specific to package revision and it will represent the meta data named it as KptRevisionMetaData. If its too long we can name it as KptRevisionFile. Let me know if any other name suits better.

We can make it a hidden file too, let me know if we want do that

johnbelamaric: Ok, so let's think about some pros and cons:
- If it's visible, then users may mess with it. At the very least, they probably would expect it to let them make changes to the metadata. For example, can they add any annotations they want there? Does it only affect PackageRevisionResources or would it affect PackageRevision as well? Maybe we want to allow that? It's more to implement, though.
- If it's not visible, then we can worry less about users messing with it or about setting some expectations. However, it makes it less transparent how things work.

I guess on balance I would recommend:
- We hide it; the name is fine but add the "." to hide it.
- In the "push" command, if we cannot fine the file, we output a very clear message: ".KptRevisionMetaData file not found. This file is created by the pull command and is required by the push command", or something similar.

bharat-m1: Sure it makes more sense to hide the file, I will update it

commands/alpha/rpkg/pull/command_test.go
bharat-m1: As per the discussion, PackageRevisionResources metadata includes the resourceVersion and additional details. So will be dumping the content directly from there to this file.

commands/alpha/rpkg/util/common.go
```diff
@@ -67,43 +68,33 @@ func GetResourceFileKubeObject(prr *api.PackageRevisionResources, file, kind, na
...
+func AddRevisionMetaData(prr *api.PackageRevisionResources) error {
+	kptfileKo, err := GetResourceFileKubeObject(prr, "Kptfile", "Kptfile", "")
 	if err != nil {
 		return err
 	}
 
-	ko.SetAnnotation(ResourceVersionAnnotation, prr.GetResourceVersion())
-	prr.Spec.Resources["Kptfile"] = ko.String()
+	kptMetaDataKo := fnsdk.NewEmptyKubeObject()
+	kptMetaDataKo.SetAPIVersion(kptfileKo.GetAPIVersion())
```
johnbelamaric, line 87: Rather then setting this to the API version of the Kptfile, we probably should set it to the API version of the PackageRevisionResource. Then you don't even need the `kptfileKo`, and it more accurately reflects that this is related to the PRR not the Kptfile.

johnbelamaric (changes requested):
One small change but otherwise LGTM.

johnbelamaric:
The commit hash will change every run, so you probably need to tweak the test code to strip those out before doing deltas. Timestamp too. I had to do that with the prior version.

You may also want to not store all the annotations (in the actual code, not the test code), but I didn't mention it before because it's harmless that they are there; they are ignored by your `push` code, and the file is hidden.

bharat-m1:
> The commit hash will change every run, so you probably need to tweak the test code to strip those out before doing deltas. Timestamp too. I had to do that with the prior version.
> 
> You may also want to not store all the annotations (in the actual code, not the test code), but I didn't mention it before because it's harmless that they are there; they are ignored by your `push` code, and the file is hidden.

Additional things I can find are creationTimestamp & uid Instead of removing specific annotations, I hope we can dump them only. As we have made it a hidden file it should not affect the user too.
Let me know if that is fine or we want to remove some of them.

commands/alpha/rpkg/util/common.go
bharat-m1: Agreed, Just wanted to align the contents with the Kptfile intially. But it makes sense to remove it.

johnbelamaric (approved):

commands/alpha/rpkg/pull/command.go
```diff
@@ -113,7 +113,7 @@ func (r *runner) runE(_ *cobra.Command, args []string) error {
 		return errors.E(op, err)
 	}
 
-	if err := util.AddResourceVersionAnnotation(&resources); err != nil {
+	if err := util.AddRevisionMetaData(&resources); err != nil {
```
mortent, line 116: Nit: We should use consistent naming, so this should probably be `AddRevisionMetadata`.

mortent (approved):
A small comment, but otherwise looks good.

Issue #3672, opened 2022-11-23
kpt.dev/latest-revision label doesn't work well with controllers


mortent:
We use the `kpt.dev/latest-revision` label to tell clients which PackageRevision is the latest for a package. This works well when users use the LIST command so we can always provide a snapshot of all PackageRevisions from the same point in time. But for controllers that rely on watches to build a local cache of PackageRevisions, this doesn't work, since publishing a new PackageRevision that becomes the latest one doesn't notify the clients that the `kpt.dev/latest-revision` label on other revisions are no longer correct.

One possible solution would be that we send watch updates for any PackageRevisions that are no longer the latest and thereby make sure that controller caches get updated correctly. However, this will lead to a race that can lead to controllers for short period of times can either have two PackageRevisions marked as latest, or have no PackageRevision marked as latest. This depends on whether we chose to send the watch events for the new or previous latest PackageRevision first. Also, this approach is difficult to manage with CRDs, since it doesn't give us a good way to do cross-PackageRevision logic.

So I think our best alternative is to drop the `kpt.dev/latest-revision` label and rely on clients to use the `revision` field to determine which PackageRevision is the latest. After https://github.com/GoogleContainerTools/kpt/pull/3593, this should be pretty straightforward to handle in the client.

natasha41575:
I think asking the client to use the `revision` field to determine the latest revision is reasonable. Perhaps we could provide a library function in the kpt repo that does the bulk of the logic, for clients to call?

PR #4021, opened 2023-08-11
Copy the labels map before editing

### LINKED ISSUE #3958: porch: porch-server crash with "concurrent map iteration and map write"
### Expected behavior
No crashing.

### Actual behavior
porch-server crashed with "concurrent map iteration and map write" and "fatal error: I0512 17:06:29.730904       1 watch.go:238] sending watch event {MODIFIED 0xc000524d20}".

### Information

This is during some testing of PackageVariantSet / PackageVariant. In that testing, there were a bunch of errors like incorrect CRDs loaded, missing permissions, etc. that led to a lot of churn of resource creation, deletion, and error handling code getting exercised a lot.

Logs attached, also available at https://cloudlogging.app.goo.gl/q4nwxkeYcfeWtMnn6 - let me know if you need permissions, I think some folks have them to that project already.

This is using a dev build of Porch, though there are no changes in Porch Server compared to v0.0.17, only in the controllers.

[downloaded-logs-20230512-115730.json.gz](https://github.com/GoogleContainerTools/kpt/files/11467029/downloaded-logs-20230512-115730.json.gz)

johnbelamaric:
Fixes #3958 

The logs showed a crash during rendering the metadata.Labels field. This was a crash that was completely outside the path of any of our code - meaning that extra locking won't help, because we can't change the API server code to grab a lock. The code was serving something up from a cache, so we must have been modifying the data in that cache at the same time.

I found a couple places that we tweak the Labels field, and these are areas of the code that were active in other runnable goroutines, so I am pretty sure they are the culprit. I have been running the new build for a few hours now, and no crashes. The released version was crashing ~1 per hour. Let's let it steep over the weekend.

/hold

### DIFF
--- a/porch/pkg/cache/packagerevision.go
+++ b/porch/pkg/cache/packagerevision.go
@@ -40,10 +40,13 @@ func (c *cachedPackageRevision) GetPackageRevision(ctx context.Context) (*v1alph
 		return nil, err
 	}
 	if c.isLatestRevision {
-		if rev.Labels == nil {
-			rev.Labels = map[string]string{}
+		// copy the labels in case the cached object is being read by another go routine
+		labels := make(map[string]string, len(rev.Labels))
+		for k, v := range rev.Labels {
+			labels[k] = v
 		}
-		rev.Labels[v1alpha1.LatestPackageRevisionKey] = v1alpha1.LatestPackageRevisionValue
+		labels[v1alpha1.LatestPackageRevisionKey] = v1alpha1.LatestPackageRevisionValue
+		rev.Labels = labels
 	}
 	return rev, nil
 }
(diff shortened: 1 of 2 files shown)


mortent (approved):
I'm ok with this change if it addresses the crashes. But it seems like if there is concurrent access to the labels map, we will still have potential for concurrent access to the `Labels` field of the PackageRevision objects. Not sure if this will potentially lead to concurrent access errors, or if we risk visibility problems.

Regardless, the usage of the `kpt.dev/latest-revision` label is problematic for other reasons as well, as described in https://github.com/GoogleContainerTools/kpt/issues/3672, so we should try to get rid of it.

johnbelamaric:
Yeah, this fixes the crash but does not fix the issue described in #3672 - which is effectively what you are describing in your comment, @mortent.

After three different instance soaked for > 2.5 days each (so total soak of ~ 8 days), we saw one crash in each, but unrelated to the one this is trying to fix.

PR #4022, opened 2023-08-15
Fix: Throw error while rejecting a package which is not proposed/deletion-proposed

bharat-m1:
Default case of switch should throw an error response while rejecting a package.


johnbelamaric (approved):

e2e/testdata/porch/rpkg-lifecycle/config.yaml
```diff
@@ -158,6 +158,9 @@ commands:
       - --namespace=rpkg-lifecycle
     stderr: |
       cannot reject git-017a8366a5e0d9b35ae6dc489d4d3f68046d6034 with lifecycle 'Published'
+      Error: errors:
```
mortent, line 161: This output is a bit awkward, but it isn't really due to this PR. We should fix that in a separate PR.

mortent (approved):

PR #4028, opened 2023-08-22
Run Porch as non-root

adetalhouet:
Create a `porch` user to run the porch-server as non-root.

### DIFF
--- a/porch/build/Dockerfile.porch
+++ b/porch/build/Dockerfile.porch
@@ -55,5 +55,13 @@ RUN cd porch; go build -v -o /porch ./cmd/porch
 
 FROM debian:bookworm-slim
 RUN apt update && apt install -y ca-certificates && apt install -y git && rm -rf /var/lib/apt && rm -rf /var/cache/apt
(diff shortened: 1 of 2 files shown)


johnbelamaric:
@adetalhouet you probably need to make similar changes to the deployment manifests produced by the Porch build process as you did in the Nephio repo. The manifests in the Nephio repo are just a subset of the ones generated here.

Right now, e2e tests are failing, likely because the manifests used to install Porch [here](https://github.com/GoogleContainerTools/kpt/actions/runs/5943534760/job/16159443837?pr=4028#step:8:15) do not have those changes.

johnbelamaric (approved):

johnbelamaric:
cc @mortent @natasha41575

natasha41575 (approved):
the change itself LGTM and I think it makes sense to do this, but I think we should make sure @mortent agrees before merging

mortent (approved):
Looks good. Thanks for the contribution @adetalhouet

PR #4042, opened 2023-09-18
Disable APF

### LINKED ISSUE #4023: porch: porch-server crash with "Unable to derive new concurrency limits"
### Expected behavior

No crashing

### Actual behavior

```
```

### Information

kpt version --> 1.0.0-beta.38

johnbelamaric:
Fixes #4023 

The upstream bug is currently patched in kubernetes/kubernetes but has not been cherry-picked to release branches. We can try re-enabling this after that is done or we have upgraded to the 1.29 packages. Given that Porch is an aggregated API server, it may be that APF in the primary API server already protects Porch, but that has not been verified.

### DIFF
--- a/porch/deployments/porch/3-porch-server.yaml
+++ b/porch/deployments/porch/3-porch-server.yaml
@@ -73,7 +73,7 @@ spec:
             - --cache-directory=/cache
             - --cert-dir=/tmp/certs
             - --secure-port=4443
-
+            - --feature-gates=APIPriorityAndFairness=false
 ---
 apiVersion: v1
 kind: Service


johnbelamaric:
Note: I am verifying that this fixes the problem, we should know by tomorrow. Without this change, we see a crash about 2-3 times per day.

johnbelamaric:
So my current hypothesis is this:

- we saw APF errors because APF was actually getting triggered and had this bug
- it was getting triggered because we start hammering the api server, probably due to a watch storm of some kind

The watch events being sent just before the crash are 85%+ of the log entries - we're talking over 400k in the 30m before the crash. So, there is something wrong in that watch code somewhere.

johnbelamaric:
I don't think this is needed, #4048 is better.

/close

PR #4043, opened 2023-09-18
GCP fleet sync controller

johnbelamaric:
Updated version of #3944


johnbelamaric:
@mortent I think this is ready for a look, at least, if not merging. It now syncs fleet memberships, scopes, and membership bindings. I don't sync fleet namespaces or RBAC bindings at this time.

We can probably refactor to make less code duplication between types, but I think it's OK for now.

mortent (approved):

PR #4055, opened 2023-10-06
fix: Avoid panic when inventory update errors

karlkfi:
Fix two bugs in the inventory provider:
- Avoid null pointer panic when inventory update errors (appliedObj was nil)
- Avoid losing inventory modifications caused to mutating webhooks (e.g. annotations)

### DIFF
--- a/pkg/live/inventoryrg.go
+++ b/pkg/live/inventoryrg.go
@@ -292,15 +292,36 @@ func (icm *InventoryResourceGroup) ApplyWithPrune(dc dynamic.Interface, mapper m
 	}
 
 	// Update the cluster inventory object.
+	// Since the ResourceGroup CRD specifies the status as a sub-resource, this
+	// will not update the status.
 	appliedObj, err := namespacedClient.Update(context.TODO(), invInfo, metav1.UpdateOptions{})
-
-	// Update status.
-	if statusPolicy == inventory.StatusPolicyAll {
-		invInfo.SetResourceVersion(appliedObj.GetResourceVersion())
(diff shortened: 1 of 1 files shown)


pkg/live/inventoryrg.go
```diff
@@ -292,15 +292,36 @@ func (icm *InventoryResourceGroup) ApplyWithPrune(dc dynamic.Interface, mapper m
...
 	// Update the cluster inventory object.
+	// Since the ResourceGroup CRD specifies the status as a sub-resource, this
+	// will not update the status.
 	appliedObj, err := namespacedClient.Update(context.TODO(), invInfo, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
 
-	// Update status.
+	// Update status, if status policy allows it.
```
sdowell, line 302: Some unit tests here would be nice to have if it's not too cumbersome

sdowell (approved):

mortent (approved):

pkg/live/inventoryrg.go
karlkfi: Adding unit tests would require faking the `dynamic.Interface` and `meta.RESTMapper`. The fakes I wrote are in kpt-config-sync downstream, which make them hard to use here. If we want to use the fake clients in kpt, we'll have to upstream the fakes into cli-utils first.

So adding unit tests might be more effort than it's worth at this point.

karlkfi:
/hold

Blocked on https://github.com/kptdev/kpt/pull/4055 and need to push gke-e2e

PR #4056, opened 2023-10-10
fix: deletionProposed branch throws error for `main` revisions

natasha41575:
TODO: See if I can add tests

When following these steps:
1. Create two PackageRevisions belonging to different packages in the same repo and propose and publish both.
2. Propose deletion of both "main" packagerevisions that get created.
3. Delete the first PackageRevision from "main" and wait over a minute until the sync has happened.
4. Delete the second PackageRevision from "main" and an error occurs complaining that the commit is incorrect.

The error is thrown because when we try to delete the deletionProposed branch, we are passing in the commit from the latest ref on the main branch. In our case, when we try to delete the second packagerevision, we pass in the latest commit from main which was created when the first packagerevision was deleted. This commit doesn't exist in the second deletionProposed branch because the deletion of the first packagerevision happened on the main branch after the second packagerevision was proposed for deletion. 

Since we're trying to delete the deletionProposed branch entirely and are using its existence as a boolean, I don't think we care what commit its on. We can pass in the zero-hash commit for it to delete the branch regardless of what commit its on. This makes the error go away for me locally.

### DIFF
--- a/porch/pkg/git/git.go
+++ b/porch/pkg/git/git.go
@@ -426,7 +426,7 @@ func (r *gitRepository) DeletePackageRevision(ctx context.Context, old repositor
 		refSpecs.AddRefToDelete(ref)
 
 		// If this revision was proposed for deletion, we need to delete the associated branch.
-		if err := r.removeDeletionProposedBranchIfExists(ctx, oldGit.path, oldGit.revision, oldGit.commit); err != nil {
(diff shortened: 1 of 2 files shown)


johnbelamaric:
Makes sense to me. Since we are deleting, the specific commit is not critical.

johnbelamaric (approved):

mortent (approved):

PR #4071, opened 2023-10-24
bumped up the golang/x/net version

droot:


natasha41575 (approved):

droot:
@mortent verified that the porch e2e test is passing locally in this setup, so going ahead with the merge. porch e2e is timing out in the CI, so we will have to keep an eye on it

PR #4076, opened 2023-10-27
make registered repository sync frequency configurable

himanshu19w:

### DIFF
--- a/porch/Makefile
+++ b/porch/Makefile
@@ -159,7 +159,8 @@ run-local: porch
 	--standalone-debug-mode \
 	--kubeconfig="$(KUBECONFIG)" \
 	--cache-directory="$(CACHEDIR)" \
-	--function-runner 192.168.8.202:9445
+	--function-runner 192.168.8.202:9445 \
+	--repo-sync-frequency=60s
 
 .PHONY: run-jaeger
 run-jaeger:
--- a/porch/pkg/apiserver/apiserver.go
+++ b/porch/pkg/apiserver/apiserver.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/GoogleContainerTools/kpt/internal/fnruntime"
 	"github.com/GoogleContainerTools/kpt/porch/api/porch/install"
@@ -76,6 +77,7 @@ type ExtraConfig struct {
 	CacheDirectory        string
 	FunctionRunnerAddress string
 	DefaultImagePrefix    string
+	RepoSyncFrequency     time.Duration
 }
 
 // Config defines the config for the apiserver
@@ -220,7 +222,7 @@ func (c completedConfig) New() (*PorchServer, error) {
 
 	watcherMgr := engine.NewWatcherManager()
 
-	cache := cache.NewCache(c.ExtraConfig.CacheDirectory, cache.CacheOptions{
+	cache := cache.NewCache(c.ExtraConfig.CacheDirectory, c.ExtraConfig.RepoSyncFrequency, cache.CacheOptions{
 		CredentialResolver: credentialResolver,
 		UserInfoProvider:   userInfoProvider,
 		MetadataStore:      metadataStore,
--- a/porch/pkg/cache/cache_test.go
+++ b/porch/pkg/cache/cache_test.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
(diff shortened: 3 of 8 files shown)


porch/pkg/apiserver/apiserver.go
```diff
@@ -76,6 +76,7 @@ type ExtraConfig struct {
 	CacheDirectory        string
 	FunctionRunnerAddress string
 	DefaultImagePrefix    string
+	RepoSyncFrequency     int64
```
johnbelamaric, line 79: We should change this to a time.Duration

johnbelamaric (changes requested):
Nice, thanks. Two things:
- Sign the DCA
- Change it to a time.Duration and allow the config to be passed in as a duration

johnbelamaric:
And add some tests...

porch/pkg/apiserver/apiserver.go
himanshu19w: Done

himanshu19w:
Did the required changes (Signed DCA and passed the config as time.Duration).

Can you please point me to where the required tests should be written? I am not able to find the test setup where I can embed tests related to this change.

porch/pkg/cmd/server/start.go
```diff
@@ -229,4 +232,5 @@ func (o *PorchServerOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.FunctionRunnerAddress, "function-runner", "", "Address of the function runner gRPC service.")
 	fs.StringVar(&o.DefaultImagePrefix, "default-image-prefix", "gcr.io/kpt-fn/", "Default prefix for unqualified function names")
 	fs.StringVar(&o.CacheDirectory, "cache-directory", "", "Directory where Porch server stores repository and package caches.")
+	fs.DurationVar(&o.RepoSyncFrequency, "repo-sync-frequency", 60, "Frequency in seconds at which registered repositories will be synced.")
```
johnbelamaric, line 235: I don't see anywhere to add tests, either. Are you sure this default means 60 seconds? shouldn't it be `60 * time.Second`...is it possible to add a test that at least checks that? Maybe we should add a constraint that the timing can't be less than 30s or something? Up to you, I am OK to merge this once you verify this default value.

porch/pkg/cache/cache_test.go
```diff
@@ -134,7 +134,7 @@ func openRepositoryFromArchive(t *testing.T, ctx context.Context, testPath, name
 	repo, address := git.ServeGitRepository(t, tarfile, tempdir)
 	metadataStore := createMetadataStoreFromArchive(t, "", "")
 
-	cache := NewCache(t.TempDir(), CacheOptions{
+	cache := NewCache(t.TempDir(), 60, CacheOptions{
```
johnbelamaric, line 137: `60 * time.Second` ?

porch/pkg/cmd/server/start.go
himanshu19w: Thanks for pointing. I was only checking the happy scenario where I provide the flag. 60 was considered as 0 second by the code. I have changed it to `60*time.Second`.  Added a unit test to ensure that this time stays above 30s.

porch/pkg/cache/cache_test.go
himanshu19w: Done.

himanshu19w:
Resolved the comments.

johnbelamaric (approved):

himanshu19w:
I have fixed the unit test (license was required at the start of newly added file).

I dont know why e2e tests are failing. They also failed for some of the previous PRs.

porch/Makefile
```diff
@@ -159,7 +159,8 @@ run-local: porch
 	--standalone-debug-mode \
 	--kubeconfig="$(KUBECONFIG)" \
 	--cache-directory="$(CACHEDIR)" \
-	--function-runner 192.168.8.202:9445
+	--function-runner 192.168.8.202:9445 \
+	--repo-sync-frequency=60s
```
mortent, line 163: Did we consider making this a property of the Repository object so it could be set to specific values for different repositories rather than a single value that must be specified at startup?

mortent (approved):

porch/Makefile
himanshu19w: Yes. The current change sets the default value while the property of repo can tweak individual repo sync. So, these two changes need not to be mutually exclusive. I did this change because it satisfies our use case with minimal changes.

PR #4083, opened 2023-11-08
versionbump: golang.org/x/net v0.17.0

justinsb:
Contains a CVE, though we're unlikely to be vulnerable.


justinsb:
Doing this manually as dependabot seems to be struggling with our many modules.

droot:
Change looks good. The CLI test failure is coming from this extra message in `fn render`. 

`Stderr:
            + 	    "Status: Image is up to date for gcr.io/kpt-fn/set-namespace:v0.1.3"` 

Don't think the changes in this PR will cause that. That's probably function of the `docker` runtime I think ? but will have investigate more.

droot (approved):

PR #4085, opened 2023-11-16
Do not refresh if no repo changes

johnbelamaric:
This is a partial fix to the issue of Porch repository polling taking a long time when polling a repo, and holding the lock. It avoids any refresh at all during a steady state.

This helps but is insufficient to resolve the problem completely. Future PR will do partial refreshes, where we hold the lock only long enough to load new or changed revisions.

### DIFF
--- a/porch/pkg/git/git.go
+++ b/porch/pkg/git/git.go
@@ -192,6 +192,8 @@ func (r *gitRepository) Close() error {
 func (r *gitRepository) Version(ctx context.Context) (string, error) {
 	ctx, span := tracer.Start(ctx, "gitRepository::Version", trace.WithAttributes())
 	defer span.End()
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
 
 	if err := r.fetchRemoteRepository(ctx); err != nil {
 		return "", err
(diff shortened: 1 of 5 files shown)


porch/pkg/git/git.go
```diff
@@ -186,6 +189,35 @@ func (r *gitRepository) Close() error {
 	return nil
 }
 
+func (r *gitRepository) Version(ctx context.Context) (string, error) {
```
mortent, line 192: How should we think about the version for the git repository calculated here? If I understand the code correctly it essentially hashes all the refs and uses that as the version. Would this handle commits that happens to existing refs (which I think primarily branches and tags) in git? In particular, would this version change if someone adds an additional commit on the main branch?

johnbelamaric: Yes. This checks the existence of refs, plus the commit hash to which they point. A push to main updates the main head ref to point to that new commit. 

So this will change for any new or deleted tag, for a commit to any branch, for a new or deleted branch, or for a forced push, or for a tag move. AFAIK, there is no change you could make to a remote that this would not pick up. Maybe some sort of magic where you add a commit but somehow have no ref to it. But we would not care about such a change anyway (if it is even possible).

mortent: Yeah, by using both the refs and the commit hash it should cover all scenarios I can think of.

PR #4087, opened 2023-11-20
Revise locking strategy

johnbelamaric:
I am unable to do partial updates of the cache. From what I can see, the way the git repo code works makes this difficult if not impossible. Instead, this PR allows reads to proceed when the cache is in the process of being reconciled, which should help with the concurrency issues we are seeing. To do that, this PR revises the locking strategy for Porch repositories, so that List operations do not block while the cache is being reconciled with the repository.

- List no longer loads the cache if it is empty. Instead, it blocks until the cache is loaded via the reconciliation background go routine.
- The `reconcileMutex` is held during reconciliation of the repo to the cache, and during any operations that mutate the underlying repository (Delete, Update)
- The `mutex` read lock is held whenever the cache *map* is iterated over or checked for nil. ~It is not necessary to hold it when checking if that map is non-nil.~
- The `mutex` write lock is held whenever mutations are happening to the map, or we are change the map the cache points to.

### DIFF
--- a/porch/pkg/cache/repository.go
+++ b/porch/pkg/cache/repository.go
@@ -54,16 +54,25 @@ type cachedRepository struct {
 
 	lastVersion string
 
-	mutex                  sync.Mutex
+	// We use separate mutexes for cache map changes and for the overall
+	// reconcile process. We want update, delete, and reconcile
+	// to all block on the reconcileMutex, which could be held for a long time
+	// during reconcile. For much of that time (during reconcile) we do NOT
+	// want to block reads. There are a few protected areas where we touch map
+	// entries where we need to block reads, so those will also grab the general
+	// mutex.
+	//
+	// Any code that needs to hold both locks MUST get the reconcileMutex first,
+	// or we could end up with deadlocks
+	mutex                  sync.RWMutex
+	reconcileMutex         sync.Mutex
 	cachedPackageRevisions map[repository.PackageRevisionKey]*cachedPackageRevision
-	cachedPackages         map[repository.PackageKey]*cachedPackage
 
 	// TODO: Currently we support repositories with homogenous content (only packages xor functions). Model this more optimally?
 	cachedFunctions []repository.Function
 	// Error encountered on repository refresh by the refresh goroutine.
 	// This is returned back by the cache to the background goroutine when it calls periodicall to resync repositories.
 	refreshRevisionsError error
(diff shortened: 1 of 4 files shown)


johnbelamaric:
Note - I ran this through the Nephio e2e tests and it ran through *functionally* successfully. There were some timeouts because everything took much longer. I think is simply because we have reduced the "chatter" on the watches and so approvals are not happening as quickly. This is actually probably a bug in the Nephio approval controller, rather than an issue with this PR.

porch/pkg/engine/watchermanager.go
```diff
@@ -65,6 +65,18 @@ func (r *watcherManager) WatchPackageRevisions(ctx context.Context, filter repos
 	r.mutex.Lock()
 	defer r.mutex.Unlock()
 
+	// reap any dead watchers
```
johnbelamaric, line 68: This change is needed because of the previous PR that avoided reloads when the repo is unchanged. Basically, since we are not reloading, we are sending fewer events. We were previously reaping only when a send fails. This now reaps any dead watchers whenever we start a new watch, in addition to when a send fails.

porch/test/e2e/e2e_test.go
johnbelamaric, line 1: Not sure these actually ended up being necessary. However, I did notice that the kpt-fn repo takes about 22s to load in my system, so I am not sure 10s wait was enough, and thus am leaving this in.

porch/pkg/cache/repository.go
```diff
@@ -258,63 +264,55 @@ func (r *cachedRepository) update(ctx context.Context, updated repository.Packag
...
 }
 
 func (r *cachedRepository) ListPackages(ctx context.Context, filter repository.ListPackageFilter) ([]repository.Package, error) {
-	packages, err := r.getPackages(ctx, filter, false)
-	if err != nil {
-		return nil, err
-	}
-
-	return packages, nil
+	return nil, fmt.Errorf("not implemented")
```
mortent, line 296: We can probably delete the Package functions. This was added as part of a plan to support Package as a separate resource, but it was never completed.

mortent (approved):

porch/pkg/cache/repository.go
johnbelamaric: Yes, that's my plan eventually but not in this PR.

PR #4093, opened 2023-11-22
Simple kubectl plugin for plan functionality

justinsb:
We would like to modularize kpt into composable functionality.  Start
with creating a kubectl plugin that can hold "plan" functionality.


justinsb:
/assign @mortent for project evolution
/assign @droot for project evolution

cc @seans3 for kubectl plugin angle

cc @diviner524 for possible KCC uses

PR #4113, opened 2024-01-25
docker execution: trim "image up to date" lines

justinsb:
This line was recently added to docker's output, and is not part of
function execution so should be filtered.

It is also breaking our tests!


justinsb:
A little confused as to how #4114 is passing tests without this, given that others failed.

/hold

justinsb:
Looks like it is "random" e.g. https://github.com/kptdev/kpt/pull/4119#issuecomment-1912813388 ... not removing hold until I figure out _why_ it is "random"

PR #4116, opened 2024-01-26
infra: Dump logs from porch-server after tests

justinsb:
This gives us more insight into failures.

Signed-off-by: justinsb <justinsb@google.com>


justinsb:
/hold

Actually, it looks like we already capture this log as an artifact!

justinsb:
/close 

(as porch moved to nephio)

PR #4117, opened 2024-01-26
chore: Cleanup error logging when can't delete webhook

justinsb:
We were using the wrong format string.

Signed-off-by: justinsb <justinsb@google.com>


efiacor:
Should be moved to https://github.com/nephio-project/porch

liamfallon:
Moved to https://github.com/nephio-project/porch/pull/329

PR #4118, opened 2024-01-26
infra: grab logs from all porch-system pods

justinsb:
We saw that the function-runner was not reachable in some test
failures, grab logs for it and all the pods in porch-system.


justinsb:
Closing based on #4153 

/close

PR #4129, opened 2024-03-15
Support flag for status policy

fsommar:
The default behavior is still to use --status-policy=all. The option is there for users that want to avoid the status behavior for various reasons. In our case, we use `kpt live apply` in our CI/CD pipelines, but don't need it saving statuses for all resources into the inventory. Thus, we want to be able to pass `--status-policy=none`.

I created a partner PR in https://github.com/kubernetes-sigs/cli-utils/pull/637. The idea is to have that merged first and then rely on that flagutils func in this PR.

If this flag has been intentional withheld from users, can we consider having this flag hidden instead?


fsommar:
Rebased on main; the failing checks seems unrelated and go generate works locally after rebase.

fsommar:
Failure looks unrelated; let me know if I should rebase again.

justinsb:
Thank you @fsommar - there's some code repetition here in the existing code (it looks like we started live.NewClusterClientFactory but never really finished it).

I'm OK to merge as-is, but I also am checking whether I have permissions to merge https://github.com/kubernetes-sigs/cli-utils/pull/637

justinsb:
Looks like I don't have permissions, so I'm going to apply hold & approve & lgtm.  If you want to merge before the cli-utils PR, just remove the hold :-)

/approve
/lgtm
/hold

fsommar:
I've upgrade to the latest cli-utils release with my changes and cleaned up the history.

/assign justinsb

mortent (approved):

mortent:
The PR that removes Porch from the repo has merged. Can you rebase this one and it should address most of the test failures?

fsommar:
It went through without tests running 😅 At least they were passing locally. I noticed too late that I accidentally included the go version files in the porch directory after rebasing

PR #4165, opened 2024-06-27
Propagate ResourceGroup labels and annotations

fsommar:
When I declare a ResourceGroup on disk, I expect the labels and
annotations defined on it to be propagated to the in-cluster inventory.
That was previously not the case.

kioutil annotations are removed (path, index, etc) since they are
redundant for inventories that are typically defined in one place, and
don't need to be reconstructed by a function pipeline.


fsommar:
Failure is unrelated, due to busybox image being too old. I can change it to the docker hub one, but I assume that you prefer to have it in GCR/GAR?

mortent (approved):

fsommar:
Thanks for the review @mortent! I don't have permissions to merge, can you do it for me? 🙇

PR #4182, opened 2025-02-20
updated license db

droot:
Updated the license db.


efiacor:
Now covered by FOSSA scans

Issue #4235, opened 2025-08-07
Remove the deprecated `kpt alpha x` commands from the documentation


CsatariGergely:
The kpt documentation still has references to the kpt alpha commands which are deprecated. These references should be removed.

Current commands under kpt alpha:
* [live](https://kpt.dev/reference/cli/alpha/live/)
* [repo](https://kpt.dev/reference/cli/alpha/repo/)
* [rpkg](https://kpt.dev/reference/cli/alpha/rpkg/)
* [sync](https://kpt.dev/reference/cli/alpha/sync/)

 "rpkg" and "repo" should be removed as they were moved to prochctl

Should be done in the Hugo documentation set.

PR #4275, opened 2025-10-03
Remove kpt alpha repo|rpkg cmds from docs

efiacor:
https://github.com/kptdev/kpt/issues/4235

Remove the cmd sets from the docs and repoint users to Porch docs.
More work will need to be done to clean up the actual codebase to remove the cmds


CsatariGergely (approved):
Wow, great! Thanks. 
I love the Warning shortcode. We should have Note also.

liamfallon (approved):

PR #4289, opened 2025-10-15
Pull function-catalog to have all docs in the same domain

lapentad:
The makefile will clone the function catalog repo and copy over only function runner folder doc and it's javascript
```
 $ make preview-build
 $ make serve
```

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,8 @@ documentation/content/en/function-catalog
 documentation/layouts/shortcodes/listexamples.html
 documentation/layouts/shortcodes/listfunctions.html
 documentation/layouts/shortcodes/listversions.html
+documentation/layouts/shortcodes/relatedexamples.html
+documentation/layouts/partials/breadcrumb.html
 
 go.work
 go.work.sum


documentation/Makefile
```diff
@@ -1,3 +1,23 @@
...
+CLONE_DIR = tmp-krm-clone
+CONTENT_SUBDIR = documentation/content/en/function-catalog/
+TARGET_HUGO_DIR = content/en/function-catalog/
+
+LAYOUT_SUBDIR = documentation/layouts/shortcodes/
+TARGET_HUGO_LAYOUT_DIR = layouts/shortcodes/
+
+.PHONY: serve production-build preview-build fetch-external-docs
+
+fetch-external-docs:
```
CsatariGergely, line 14: Also copy the examples folder from $(CLONE_DIR) to content/en/examples

Ref https://github.com/kptdev/krm-functions-catalog/blob/28c1a97606bf79c8a5ad574512e304f92c320ae9/documentation/Makefile#L34

CsatariGergely:
Looks good, thanks. Is there a way to trigger the build of kpt.dev in case there is a change in krm-functions-catalog?

lapentad:
> Looks good, thanks. Is there a way to trigger the build of kpt.dev in case there is a change in krm-functions-catalog?

Right, I haven't thought of this

I am reading from docs we should use this build hook https://docs.netlify.com/build/configure-builds/build-hooks/
So we have a github action in the catalog to do a simple curl to this hook so it will trigger the kpt.dev website build, if there are changes to the catalog-docs.

I can generate a PR for the catalog repo, but an OWNER might need to create a secret with the netlify build hook url from kpt.dev site (kpt.dev site on Netlify: Site settings -> Build & deploy -> Build hooks)

documentation/Makefile
lapentad: @CsatariGergely I added them now. 
Thank you because I did not notice they were not getting rendered

lapentad:
> Looks good, thanks. Is there a way to trigger the build of kpt.dev in case there is a change in krm-functions-catalog?

@CsatariGergely https://github.com/kptdev/krm-functions-catalog/pull/1187 added this, but I am assuming an OWNER can create that variable

PR #4302, opened 2025-10-23
Bump versions of go/kubernetes

liamfallon:
Routine PR to bump the versions for go, github actions and kubernetes dependencies.


liamfallon:
The reason for ignoring the int64 warning is that there is a bug in Kubernetes which issues a warning in int64 in CRD definitions. It will be fixed in an upcoming k8s patch.

See https://github.com/kubernetes/kubernetes/pull/133896

PR #4345, opened 2026-01-11
Docs: Update _index.md

peterloane:
Original text is missing a article - so could be edited to "An environment variable..." but looking at the content in the linked references (.../render), these use headings of "Environment variables", so would propose change as "Environment variables"


liamfallon (approved):
Thanks for the contribution :-)

Please sign the commit for the contribution so that we can accept it, Click on the DCO check result below for instructions.

PR #4349, opened 2026-01-13
Fix documentation npm scripts on Windows

Jaisheesh-2006:
## Title

Fix documentation npm scripts on Windows

### What changed
Made the `documentation/` npm scripts cross-platform so Windows contributors can run `npm install` and `npm run serve` without shell-related failures.

### Motivation
On Windows, `npm install` triggers the `prepare` lifecycle, which runs `npm run get:submodule`.  
The previous scripts relied on bash/POSIX-only constructs (for example `set -x`, `${VAR:-default}`), which fail under `cmd.exe` and PowerShell, blocking local documentation site setup.

### Fix
- Replaced shell-specific logic in `documentation/package.json` with portable commands.
- Introduced small Node helpers where environment-variable fallbacks are required (for example preview builds).

### User impact
Unblocks Windows users from setting up and running the docs site locally.  
No behavior change is intended for macOS or Linux.

### Windows (PowerShell)
```bash
cd documentation
npm install
npm run serve


liamfallon (approved):
I'm not familiar with it either, thanks for the change. It'd be good to get a review from @CsatariGergely

PR #4371, opened 2026-02-02
fix(live): validate annotation values to prevent silent data loss

### LINKED ISSUE #3784: kpt live silently fails to apply annotations with errors
### Expected behavior
`kpt live apply` should fail to apply a resource if an annotation has an illegal value

### Actual behavior
`kpt live apply` applied the resource without the annotations with illegal values.

Example resource: 
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: game-config
  namespace: default
  annotations:
    example.com/my-anno: true
data:
  foo: bar
```

This completed without errors. But the resulting resource in the cluster does not include the annotation:

```yaml
apiVersion: v1
data:
  foo: bar
kind: ConfigMap
metadata:
  annotations:
    config.k8s.io/owning-inventory: ed96c94158f88ec28abc7d491b8447fe24853c28-1675358808715604556
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","data":{"foo":"bar"},"kind":"ConfigMap","metadata":{"annotations":{"config.k8s.io/owning-inventory":"ed96c94158f88ec28abc7d491b8447fe24853c28-1675358808715604556"},"name":"game-config","namespace":"default"}}
  creationTimestamp: "2023-02-02T23:46:15Z"
  name: game-config
  namespace: default
  resourceVersion: "665"
  uid: 7502f43b-1a46-4eb4-891c-53fab2801376
```

Jaisheesh-2006:
#### Description
This PR implements strict validation for `metadata.annotations` values during the `live apply` pipeline.

Previously, the Go unmarshaler silently discarded non-string values (e.g., unquoted booleans like `true` or integers like `123`) because they did not match the expected `map[string]string` schema. This resulted in annotations being dropped from the applied resource without warning.

This change adds a validation step that inspects the raw YAML tags of `RNode`s. It now returns a strict error if any annotation value is not a string, while correctly accepting quoted strings (e.g., `"true"`).

#### Motivation
To prevent silent data loss and configuration drift.

Users frequently mistake YAML types (e.g., writing `enabled: true` instead of `enabled: "true"`). Currently, `kpt` ignores these fields silently. This fix ensures the user receives immediate feedback:

**New Error Output:**
`Error: annotation "example.com/enabled" must be a string, got boolean`

#### Which issue does this PR fix?
Fixes #3784

### DIFF
--- a/pkg/live/helpers_test.go
+++ b/pkg/live/helpers_test.go
@@ -79,6 +79,15 @@ metadata:
     example.com/bad-anno: true
 data: {}
 `
+	configMapWithBadLabel = `
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cm
+  labels:
+    app.kubernetes.io/enabled: true
+data: {}
+`
 	crd = `
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
--- a/pkg/live/rgpath.go
+++ b/pkg/live/rgpath.go
@@ -166,7 +166,7 @@ func validateStringMap(metadata *yaml.RNode, fieldName, fieldType string) error 
 	}
 
 	// Handle explicit null (e.g., annotations: null)
-	if mapNode.Tag == "!!null" {
+	if mapNode.Tag == yaml.NodeTagNull {
 		return nil
 	}
 
@@ -177,7 +177,7 @@ func validateStringMap(metadata *yaml.RNode, fieldName, fieldType string) error 
 	for i := 0; i < len(mapNode.Content); i += 2 {
 		keyNode := mapNode.Content[i]
 		valueNode := mapNode.Content[i+1]
-		if valueNode.Kind != yaml.ScalarNode || valueNode.Tag != "!!str" {
+		if valueNode.Kind != yaml.ScalarNode || valueNode.Tag != yaml.NodeTagString {
 			return fmt.Errorf("%s %q must be a string, got %s", fieldType, keyNode.Value, yamlTagToType(valueNode))
 		}
 	}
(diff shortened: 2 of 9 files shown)


pkg/live/rgpath.go
```diff
@@ -109,6 +113,109 @@ func kyamlNodeToUnstructured(n *yaml.RNode) (*unstructured.Unstructured, error)
...
+
+	if mapNode.Kind != yaml.MappingNode {
+		return fmt.Errorf("metadata.%s must be a string map", fieldName)
+	}
+
+	for i := 0; i < len(mapNode.Content); i += 2 {
+		keyNode := mapNode.Content[i]
+		valueNode := mapNode.Content[i+1]
+		if valueNode.Kind != yaml.ScalarNode || valueNode.Tag != "!!str" {
+			return fmt.Errorf("%s %q must be a string, got %s", fieldType, keyNode.Value, yamlTagToType(valueNode))
```
mozesl-nokia, line 181: I think we don't need to convert the type to more human readable, but curious about the others' opinion.

Maybe separate the checks to `kind != scalar` and `tag != !!str`.

Also, is "!!str" not available as a const from somewhere?

pkg/live/helpers_test.go
mozesl-nokia, line 1: Should there be a test for labels here too?

pkg/live/rgpath_test.go
mozesl-nokia, line 1: Should there be a test for labels here too?

pkg/live/rgstream_test.go
mozesl-nokia, line 1: Should there be a test for labels here too?

pkg/live/helpers_test.go
Jaisheesh-2006: Hi @mozesl-nokia , i will add the test for labels and commit it soon

pkg/live/rgpath.go
Jaisheesh-2006: Hi @mozesl-nokia , indeed !!str is available as a constant in yaml.NodeTagString, I will shortly replace the other ones as well and push it

Issue #4388, opened 2026-02-13
Enable skipping pipeline steps based on some condition expression (CEL)


kispaljr:
# Problem statement

As I kpt package author I want to be able to create package with a pipeline step (KRM function) that is only executed if some condition based on the contents of the package is true. 
E.g. I want to take different steps in the pipeline based on which type of secret management the package is configured to do.

# Workarounds with the current version

I can write my own KRM function that evaluates the condition. However, if the step wouldn't need a custom KRM function anyway (since I only want to do a set-annotation) this is very uncomfortable, and complicates my package unnecessary.

# Possible solution

Add an "execution condition" field to the Pipeline schema, that can contain a CEL expression referencing the KRM objects in the package.

OisinJohnston2005:
@SurbhiAgarwal1 is it ok if I pick this issue up?

SurbhiAgarwal1:
Hi @OisinJohnston2005,

Thanks for your interest in this issue. I've already been working on it and have an implementation in progress with multiple commits and PRs linked to this issue. I'm currently updating the work based on the ongoing pipeline rearchitecture discussions and plan to continue pushing updates soon.

I'd appreciate it if this issue could remain with my existing work for now. Thanks for checking first!

PR #4391, opened 2026-02-14
feat: Add CEL-based conditional function execution (#4388)

SurbhiAgarwal1:
Implements #4388

## Changes
- Added `condition` field to Function schema for CEL expressions
- Integrated google/cel-go library for condition evaluation
- Functions are skipped when condition evaluates to false
- Added comprehensive unit and E2E tests

## Example Usage
```yaml
pipeline:
  mutators:
  - image: gcr.io/kpt-fn/set-namespace:v0.4
    condition: "resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'env-config')"
    configMap:
      namespace: production

### DIFF
--- a/internal/fnruntime/condition_test.go
+++ b/internal/fnruntime/condition_test.go
@@ -23,6 +23,7 @@ import (
 	fnresult "github.com/kptdev/kpt/pkg/api/fnresult/v1"
 	kptfile "github.com/kptdev/kpt/pkg/api/kptfile/v1"
 	"github.com/kptdev/kpt/pkg/lib/runneroptions"
+	"github.com/kptdev/kpt/pkg/printer"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
@@ -31,6 +32,7 @@ import (
 
 func TestFunctionRunner_Conditions(t *testing.T) {
 	ctx := context.Background()
+	ctx = printer.WithContext(ctx, printer.New(io.Discard, io.Discard))
 	fsys := filesys.MakeFsInMemory()
 	celEnv, err := runneroptions.NewCELEnvironment()
 	require.NoError(t, err)

### CODE REVIEW

File: .agent/github_comment_4382.md
Comment (liamfallon) [inline .agent/github_comment_4382.md:1]: Did you mean to post these files on the issue rather than on the PR? It looks like AI generated content.

File: internal/fnruntime/celeval.go [hunk 1]
```diff
@@ -0,0 +1,143 @@
...
+)
+
+// CELEvaluator evaluates CEL expressions against KRM resources
+type CELEvaluator struct {
+	env *cel.Env
+}
+
+// NewCELEvaluator creates a new CEL evaluator with the standard environment
+func NewCELEvaluator() (*CELEvaluator, error) {
+	env, err := cel.NewEnv(
```
nagygergo: This is a totally unprotected cel executor. There should be limitations on the number of CPU cycles it can consume, the amount of characters it can output, the max complexity of the ast.

internal/fnruntime/runner.go
```diff
@@ -152,7 +152,27 @@ func NewRunner(
 			}
 		}
 	}
-	return NewFunctionRunner(ctx, fltr, pkgPath, fnResult, fnResults, opts)
+	
+	// Initialize CEL evaluator if a condition is specified
+	var evaluator *CELEvaluator
```
nagygergo: Why do you need to create a new CEL env for each function evaluation? The env can be the same.

internal/fnruntime/runner.go
```diff
@@ -193,10 +213,29 @@ type FunctionRunner struct {
...
 	opts             runneroptions.RunnerOptions
+	condition        string        // CEL condition expression
+	evaluator        *CELEvaluator // CEL evaluator for condition checking
 }
 
 func (fr *FunctionRunner) Filter(input []*yaml.RNode) (output []*yaml.RNode, err error) {
 	pr := printer.FromContextOrDie(fr.ctx)
+	
+	// Check condition before executing function
+	if fr.condition != "" && fr.evaluator != nil {
```
nagygergo: Why check if `fr.evaluator` exists or not. If the function runner was created with a condition appearing for it's Runner, then must have an evaluator. It's ok to run to a panic if it doesn't exist at this point.

internal/fnruntime/celeval_test.go
```diff
@@ -0,0 +1,157 @@
...
+package fnruntime
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
```
nagygergo, line 26: Add a testcase that makes sure that the cel functions can't mutate the resourcelist that is the input. The function signature can allow for it, as it hands over the `*yaml.RNode` list.

internal/fnruntime/conditional_e2e_test.go
```diff
@@ -0,0 +1,265 @@
...
+					if err != nil {
+						return nil, err
+					}
+				}
+				return nodes, nil
+			}
+
+			// Create function runner with condition
+			fnResult := &fnresult.Result{}
+			fnResults := &fnresult.ResultList{}
```
nagygergo: Are these needed for testware when initialising it?

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,143 @@
...
+		cel.Variable("resources", cel.ListType(cel.DynType)),
```
nagygergo: Probably advanced strings libraries would be good to include. https://pkg.go.dev/github.com/google/cel-go/ext#Strings

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	// Convert resources to a format suitable for CEL
+	resourceList, err := e.resourcesToList(resources)
+	if err != nil {
+		return false, fmt.Errorf("failed to convert resources: %w", err)
+	}
+
+	// Evaluate the expression
+	out, _, err := prg.Eval(map[string]interface{}{
```
nagygergo: There should be a context passed to this to protect against long-hanging operations.

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	// Create the program
+	prg, err := e.env.Program(ast)
+	if err != nil {
+		return false, fmt.Errorf("failed to create CEL program: %w", err)
+	}
+
+	// Convert resources to a format suitable for CEL
+	resourceList, err := e.resourcesToList(resources)
```
nagygergo: Is serialising all the yaml.RNode actually needed? As it's a map[string]any type anyways (with no strange subtypes), probably the CEL interpreter can deal with it directly. Serialising the whole package for the cel execution, then not reusing it can cause a significant memory footprint bloat.

container_test_out.txt
nagygergo, line 1: Remove

documentation/content/en/book/03-packages/_index.md
nagygergo, line 1: Are any of these changes relevant to the issue?

documentation/content/en/guides/namespace-provisioning-cli.md
nagygergo, line 1: Are any of these changes relevant to the issue?

documentation/content/en/guides/tenant-onboarding.md
nagygergo, line 1: Not relevant to the issue, please revert.

documentation/content/en/installation/migration.md
nagygergo, line 1: Not relevant to the issue that this pr addresses, please revert.

documentation/content/en/reference/cli/pkg/init/_index.md
nagygergo, line 1: Not relevant for the PR, please revert

internal/fnruntime/CEL_CONDITIONS.md
nagygergo, line 1: Why this file needs to exist? Code should be self-explaining this.

internal/fnruntime/celeval.go
nagygergo: This is still unresolved. Please read up on how cost estimation works in cel environments. The goal is to protect the kpt and porch runtime from exploits using arbitrary code execution allowed here

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	// Pre-compile the condition if provided
+	if condition != "" {
+		ast, issues := env.Compile(condition)
+		if issues != nil && issues.Err() != nil {
+			return nil, fmt.Errorf("failed to compile CEL expression: %w", issues.Err())
+		}
+
+		// Check AST complexity
```
nagygergo: This is not how it should be checked. Read up on how k8s runs cel expressions, and protects it's own runtime from arbitrary code execution.

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,164 @@
...
+	for _, r := range resources {
+		if predicate(r) {
+			return true
+		}
+	}
+	return false
+}
+
+// FilterResources filters resources based on a predicate
+func FilterResources(resources []interface{}, predicate func(interface{}) bool) []interface{} {
```
nagygergo: Not needed

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	return result, nil
+}
+
+// Helper functions for common CEL operations
+
+// ResourceExists checks if any resource matches the given predicate
+// This is exposed as a CEL macro/function
+func ResourceExists(resources []interface{}, predicate func(interface{}) bool) bool {
```
nagygergo: Not needed

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	return result, nil
+}
+
+// resourceToMap converts a single RNode to a map for CEL evaluation
+// RNode doesn't provide a direct method to convert to map[string]interface{},
+// so we serialize to YAML string and unmarshal back. This is the standard approach
+// used throughout the kpt codebase for converting RNode to generic maps.
+func (e *CELEvaluator) resourceToMap(resource *yaml.RNode) (map[string]interface{}, error) {
```
nagygergo: rnode.Map? or rnode.datamap?

liamfallon:
@SurbhiAgarwal1 I could not find you on the kpt channel on slack. I sent an invitation to you there now. If you subscribe on the kpt channel that's where we hang out.

issue_1835_comment.md
efiacor, line 1: Can we remove all of these test and doc files?

internal/fnruntime/CEL_CONDITIONS.md
efiacor, line 1: remove this or move the content to documentation somewhere

internal/fnruntime/celeval.go
SurbhiAgarwal1: Hi @nagygergo! I've researched the CEL cost estimation API and found that `env.EstimateCost()` requires runtime size information about variables (like the `resources` list) which isn't available at compile time. When I tried using it with a nil estimator, it panics during size computation.

I see two approaches:

**Option 1: Runtime Cost Tracking** - Use `cel.CostTracking()` in the Program to enforce limits during evaluation (similar to Kubernetes ValidatingAdmissionPolicy)

**Option 2: Custom Estimator** - Implement `checker.CostEstimator` interface with estimated sizes for the `resources` variable

Which approach would you prefer? I'm happy to implement either one properly. All tests are currently passing with cel-go v0.26.0.

internal/fnruntime/celeval.go
SurbhiAgarwal1: Hi @nagygergo! I've fixed the RNode serialization issue you mentioned.

Changed from `resource.Map()` (which serializes to YAML and back) to using `resource.YNode().Decode()` directly. This avoids the intermediate serialization step and reduces memory overhead.

The new implementation:
- Gets the underlying yaml.Node directly via `YNode()`
- Decodes it to `map[string]interface{}` without serialization
- Maintains the same CEL evaluation behavior

All tests are passing with this change.

internal/fnruntime/celeval_test.go
SurbhiAgarwal1: Hi @nagygergo! Done! I've added `TestEvaluateCondition_Immutability` that verifies CEL functions can't mutate the input resource list.

The test creates a ConfigMap, stores the original YAML, evaluates a CEL condition, then compares before/after to ensure no mutation. Even though the function signature accepts `*yaml.RNode` list (which could be mutated), CEL only receives converted `map[string]interface{}` copies, so the original RNodes remain unchanged.

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,137 @@
...
+
+// NewCELEvaluator creates a new CEL evaluator with the standard environment
+// The environment is created once and reused for all evaluations
+func NewCELEvaluator(condition string) (*CELEvaluator, error) {
+	env, err := cel.NewEnv(
+		cel.Variable("resources", cel.ListType(cel.DynType)),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create CEL environment: %w", err)
+	}
```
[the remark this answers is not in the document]
SurbhiAgarwal1: Fixed the doc comment in `NewCELEvaluator` - it was saying the env is created once and reused but that's not what the code does. Updated it to just describe what the function actually does.

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,134 @@
...
+type CELEvaluator struct {
+	env *cel.Env
+	prg cel.Program // Pre-compiled program for the condition
+}
+
+// NewCELEvaluator creates a new CEL evaluator with the standard environment
+// for the given condition string.
+func NewCELEvaluator(condition string) (*CELEvaluator, error) {
+	env, err := cel.NewEnv(
+		cel.Variable("resources", cel.ListType(cel.DynType)),
```
nagygergo: Did a bit of reading of my own into CEL. I'd propose this set of env setup for it

```suggestion
		cel.Variable("resources", cel.ListType(cel.DynType)),
		// Below is a list of Env settings that is a selection of https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/cel/environment/base.go
		// General rules are for maintaining this list.
		// 1. utility functions should be available. This allows for more compatibility with k8s's own CEL conditions
		// 2. AST validation is not needed as kpt will recompile CEL expressions every time, there is no cost-saving in exiting early
		// 3. Compile time optimisations do not make sense, as each CEL expression will be evaluated once before being discarded.
		// 3. Things that are helping with authorization in k8s are not needed, as they're returning either ResourceCheck or Decision types, which are not needed for kpt
		cel.HomogeneousAggregateLiterals(),
		cel.DefaultUTCTimeZone(true),
		k8scellib.URLs(),
		k8scellib.Regex(),
		k8scellib.Lists(),
		cel.CrossTypeNumericComparisons(true),
		cel.OptionalTypes(),
		k8scellib.Quantity(),
		ext.Strings(ext.StringsVersion(2)),
		ext.Sets(),
		k8scellib.IP(),
		k8scellib.CIDR(),
		k8scellib.Format(),
		ext.TwoVarComprehensions(),
		k8scellib.SemverLib(library.SemverVersion(1)),
		ext.Lists(ext.ListsVersion(3)),
```

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,134 @@
...
+// limitations under the License.
+
+package fnruntime
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types"
```
nagygergo:
```suggestion
	"github.com/google/cel-go/common/types"
	"github.com/google/cel-go/ext"
	"k8s.io/apiserver/pkg/cel/library"
	k8scellib "k8s.io/apiserver/pkg/cel/library"
```

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,134 @@
...
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
```
nagygergo:
```suggestion
const checkFrequency = 100

// This gives about .1 seconds of CPU time for the evaultation to run
const costLimit = 1000000
```

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,134 @@
...
+			return nil, fmt.Errorf("failed to compile CEL expression: %w", issues.Err())
+		}
+
+		// Validate that the expression returns a boolean
+		if ast.OutputType() != cel.BoolType {
+			return nil, fmt.Errorf("CEL expression must return a boolean, got %v", ast.OutputType())
+		}
+
+		// Create the program with cost tracking enabled (no hard limit enforced)
+		prg, err := env.Program(ast, cel.CostTracking(nil))
```
nagygergo: Enforce cost limit by the CEL interpreter, and allow it to be canceled from an upper context.
I'm not sure if Cost tracing for the additional functions is done correctly without the CostEstimator or not. Probably it doesn't have a big runtime cost to add it in, so I'd include it.

```suggestion
	prg, err := env.Program(ast,
			cel.CostLimit(costLimit),
			cel.InterruptCheckFrequency(checkFrequency),
			cel.CostTracking(&k8scellib.CostEstimator{}),
		)
```

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	return evaluator, nil
+}
+
+
+// EvaluateCondition evaluates a CEL condition expression against a list of resources
+// Returns true if the condition is met, false otherwise
+// The program is pre-compiled, so this just evaluates it with the given resources
+func (e *CELEvaluator) EvaluateCondition(_ context.Context, resources []*yaml.RNode) (bool, error) {
```
nagygergo: Propagate context through. This is more relevant when this is going to be integrated into porch, as that is not a one-off cli tool, but a server, and it needs to be able to cancel processes.

```suggestion
func (e *CELEvaluator) EvaluateCondition(ctx context.Context, resources []*yaml.RNode) (bool, error) {
```

internal/fnruntime/celeval.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	// Convert resources to a format suitable for CEL
+	resourceList, err := e.resourcesToList(resources)
+	if err != nil {
+		return false, fmt.Errorf("failed to convert resources: %w", err)
+	}
+
+	// Evaluate the expression
+	out, _, err := e.prg.Eval(map[string]interface{}{
```
nagygergo:
```suggestion
	out, _, err := e.prg.ContextEval(ctx, map[string]interface{}{
```

internal/fnruntime/conditional_e2e_test.go
nagygergo, line 1: Instead of this test, please add e2e testcases to `e2e/testdata/fn-render`

internal/fnruntime/celeval.go
mozesl-nokia, line 1: Having a file just for a type alias seems unnecessary to me

pkg/lib/runneroptions/celenv.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/ext"
+	k8scellib "k8s.io/apiserver/pkg/cel/library"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+const celCheckFrequency = 100
+
+// celCostLimit gives about .1 seconds of CPU time for the evaluation to run
+const celCostLimit = 1000000
```
mozesl-nokia:
```suggestion
const (
	celCheckFrequency = 100
	// celCostLimit gives about .1 seconds of CPU time for the evaluation to run
	celCostLimit = 1000000
)
```

pkg/lib/runneroptions/runneroptions.go
```diff
@@ -58,11 +58,22 @@ type RunnerOptions struct {
...
 func (opts *RunnerOptions) InitDefaults(defaultImagePrefix string) {
 	opts.ImagePullPolicy = IfNotPresentPull
 	opts.ResolveToImage = opts.ResolveToImageForCLIFunc(defaultImagePrefix)
+	celEnv, err := NewCELEnvironment()
+	if err != nil {
+		// CEL environment creation should never fail with the standard config;
+		// panic here surfaces misconfiguration immediately rather than silently skipping conditions.
+		panic(fmt.Sprintf("failed to initialise CEL environment: %v", err))
+	}
+	opts.CELEnvironment = celEnv
```
mozesl-nokia, line 79: Instead of trying to create the CEL env here and panicking on error, I would prefer the caller of `InitDefaults` try to create it and just return with the error.

(You have to change the comment above the `CELEnvironment` field as well if you move it out)

pkg/lib/runneroptions/celenv.go
mozesl-nokia, line 1: I think we should use `any` instead of `interface{}`

internal/fnruntime/runner.go
```diff
@@ -153,7 +153,22 @@ func NewRunner(
...
+	
+	fr, err := NewFunctionRunner(ctx, fltr, pkgPath, fnResult, fnResults, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	// Set condition; the shared CEL environment from opts is used at evaluation time.
+	if f.Condition != "" {
+		if opts.CELEnvironment == nil {
+			return nil, fmt.Errorf("condition specified for function %q but no CEL environment is configured in RunnerOptions", f.Image)
```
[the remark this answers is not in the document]
SurbhiAgarwal1: Fixed in commit ffeadea02. The error message now uses f.Exec as fallback when f.Image is empty:

name := f.Image
if name == "" {
    name = f.Exec
}
return nil, fmt.Errorf("condition specified for function %q but no CEL environment is configured in RunnerOptions", name)

This ensures exec-based functions show a helpful error message.

go.mod
```diff
@@ -109,23 +112,28 @@ require (
...
 	golang.org/x/sys v0.38.0 // indirect
 	golang.org/x/term v0.37.0 // indirect
 	golang.org/x/time v0.14.0 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20250303144028-a0af3efb3deb // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20250303144028-a0af3efb3deb // indirect
 	google.golang.org/protobuf v1.36.10 // indirect
 	gopkg.in/evanphx/json-patch.v4 v4.13.0 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
+	k8s.io/apiserver v0.34.1 // indirect
```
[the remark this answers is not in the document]
SurbhiAgarwal1: Fixed in commit ffeadea02. Removed `// indirect` comment from k8s.io/apiserver v0.34.1 since we import it directly in pkg/lib/runneroptions/celenv.go. Ran `go mod tidy` to ensure dependencies are correct.

internal/fnruntime/celeval.go
SurbhiAgarwal1: Agreed! Removed in ffeadea02. Now using *runneroptions.CELEnvironment directly.

pkg/lib/runneroptions/celenv.go
SurbhiAgarwal1: Fixed in ffeadea02. Constants are now grouped in a const() block.

pkg/lib/runneroptions/runneroptions.go
SurbhiAgarwal1: Removed panic from InitDefaults and created separate InitCELEnvironment() error method. Callers now handle errors gracefully:

opts.InitDefaults(prefix)
if err := opts.InitCELEnvironment(); err != nil {
    return err
}

Updated all callers throughout the codebase. Tests fail on error, CLI commands ignore it (conditions are optional).

pkg/lib/runneroptions/celenv.go
SurbhiAgarwal1: Replaced all interface{} with any throughout celenv.go (resourcesToList, resourceToMap, ContextEval call, and all map types.

krm-functions-catalog
efiacor, line 1: I think this is causing the netlify docs build failure. Should be removed

output.txt
efiacor, line 1: Please remove. Best to check what is being committed before adding/pushing

PR_REVIEW_SUMMARY.md
efiacor, line 1: Please remove. Best to check what is being committed before adding/pushing

e2e/testdata/fn-render/condition/condition-not-met/.expected/diff.patch
efiacor, line 1: I believe the CI failures on this are legitimate. The Kptfile api has been updated to include the renderStatus field

krm-functions-catalog
SurbhiAgarwal1: Thanks for catching this! I accidentally committed the krm-functions-catalog submodule. Removed in commit 19bf57c9a.

PR_REVIEW_SUMMARY.md
SurbhiAgarwal1: You're right, I should have been more careful. I've removed all temporary files in commit 19bf57c9a. Will use git status to check before committing next time.

e2e/testdata/fn-render/condition/condition-not-met/.expected/diff.patch
SurbhiAgarwal1: Fixed in commit 19bf57c9a. I've updated both condition test cases to include the renderStatus field in the expected output. The condition-not-met test now shows empty mutationSteps (function skipped), and condition-met shows the function execution with exitCode 0.

SurbhiAgarwal1:
Hi @nagygergo , 

I have added the documentation in commit b6f3d8ee1! as part of this PR:

1. Updated `kptfile.yaml` schema reference to include the `condition` field with its description.

2. Added a new `### Specifying condition` section in `book/04-using-functions` covering:
   - What the field does and how it works
   - A Kptfile example showing how to use it
   - Terminal output for both condition-met and condition-not-met cases
   - Common CEL expression patterns
   - How it interacts with `selectors` and `exclude`

PR #4396, opened 2026-02-17
Replace random generated "inventory ID"

### LINKED ISSUE #4387: Loose random generated "inventory ID" to make kpt less confusing to new users
# Problem statement

If I simply do `kpt live init mypackage`  currently a random "inventory ID" is assigned to my local package by default. 
If I do `kpt live apply mypackage'  and later loose the local copy of that package (delete the folder from my computer), it is hard to update my deployment in the cluster. 
The naive way to do that would be to `kpt pkg get` the same package again, do my modifications and try to issue `kpt live apply` again. This however won't work, since it requires a new `kpt live init` that will by default assign a different "Inventory ID", that in turn will cause conflicts during `kpt live apply`.

This confused me a lot, when I was a beginner `kpt` user, and I suspect this is confusing to other new users, as well.

Steps to reproduce:
```
kpt pkg get <mypackage URL> mypackage
kpt live init mypackage
kpt live apply mypackage     # this will work as expected

rm -rf mypackage
kpt pkg get <mypackage URL> mypackage
kpt live init mypackage
kpt live apply mypackage     # this won't work due to conflicts (ownership of deployed resources)
```

# How to avoid the problem with the current `kpt` version

Currently the solution that I use to avoid this is to always apply an "inventory ID" manually with all "kpt live init" that I issue. E.g. this works, as expected:

```
kpt pkg get <mypackage URL> mypackage
kpt live init mypackage --inventory-id mydeployment -- name mydeployment
kpt live apply mypackage     # this will work as expected

rm -rf mypackage
kpt pkg get <mypackage URL> mypackage
kpt live init mypackage --inventory-id mydeployment -- name mydeployment
kpt live apply mypackage     # this works as expected
```

# A possible solution

Treat "inventory ID" as  a mandatory parameter and do not generate a random ID by default. Basically I suggest to treat "Inventory ID" as the "release name" in helm (i.e. in `helm install <release-name> ...`).
I also suggest to make --name parameter of `kpt live init` equal to the "inventory ID". (Either by default, or loose the --name parameter and force it to be the same as the "inventory ID")

Jaisheesh-2006:
# Description
This PR replaces the generation of random UUIDs for inventory IDs with a deterministic SHA-1 hashing mechanism. By deriving the inventory ID from the package `namespace` and `--name`, we ensure that the same package configuration always maps to the same inventory object in the cluster.

### Key Changes:
- **Deterministic ID Generation**: Implemented `generateHash(namespace, name)` using length-prefixed SHA-1 to replace `google/uuid`.
- **Mandatory Flag**: The `--name` flag is now mandatory for `kpt live init` to ensure the hash can be consistently generated.
- **Validation**: Added DNS-1123 label validation for the `--name` input to ensure compatibility with Kubernetes resource naming conventions.
- **Legacy Support**: Hidden the `--inventory-id` flag from the help menu to encourage deterministic usage while maintaining backward compatibility for existing workflows.
- **Robustness**: Added guards in `kpt live migrate` to prevent operations on empty or malformed inventory names.

# Motivation
Previously, re-fetching or re-initializing a package would generate a new random UUID. This led to "lost inventory" bugs where `kpt` could no longer track or prune resources previously applied to the cluster because the association (the ID) had changed. Deterministic hashing ensures that as long as the namespace and name remain constant, the inventory remains trackable across different environments and local clones.

# Fixes
Fixes #4387


nagygergo:
@Jaisheesh-2006 This is only a partial review, but what happens in case there was an old `ResourceGroup` on the  cluster that doesn't use the inventoryIDs correctly? The code doesnt' seem to have any handling for it. If possible, I'd appreciate some testcases around it to be clear on what cases would and wouldn't work between the baseline and this PR.

Jaisheesh-2006:
@nagygergo Thank you for the review and for catching this. You are completely right—the current code acts as a strict gatekeeper and completely stops the migration process when it encounters an old ResourceGroup missing its inventory-id.

I will handle both parts of this issue:

I will add the specific test cases you requested to establish the clear baseline behavior for legacy objects.

I will update the migration logic to explicitly handle these missing ID cases so the tool provides a proper reconciliation path instead of just failing.

I will update the PR shortly!

commands/live/init/cmdliveinit.go
```diff
@@ -126,6 +141,11 @@ func (r *Runner) preRunE(_ *cobra.Command, _ []string) error {
 
 func (r *Runner) runE(_ *cobra.Command, args []string) error {
 	const op errors.Op = "cmdliveinit.runE"
+	name, err := validateName(r.Name)
```
liamfallon:
```suggestion
	if err := trimAndValidateName(&r.Name); err != nil {
		return err
	}
```

Maybe change the name of the validate function to better reflect what it does and pass the name as a pointer so that the function can change its value

commands/live/init/cmdliveinit.go
```diff
@@ -302,33 +337,37 @@ func writeRGFile(dir string, rg *rgfilev1alpha1.ResourceGroup, filename string)
...
-	h := sha1.New()
-	if _, err := h.Write([]byte(str)); err != nil {
-		return "", errors.E(op, err)
+	if errs := validation.IsDNS1123Subdomain(trimmed); len(errs) > 0 {
+		return "", fmt.Errorf("--name %q is not a valid Kubernetes resource name: %s",
+			trimmed, strings.Join(errs, "; "))
 	}
-	return fmt.Sprintf("%x", (h.Sum(nil))), nil
+	return trimmed, nil
 }
```
liamfallon:
```suggestion
func validateName(name *string) error {
	trimmed := strings.TrimSpace(*name)
	if trimmed == "" {
		return errNameRequired
}
	if errs := validation.IsDNS1123Subdomain(trimmed); len(errs) > 0 {
		return "", fmt.Errorf("--name %q is not a valid Kubernetes resource name: %s",
			trimmed, strings.Join(errs, "; "))
}
   *name = trimmed
	return nil
}
```

commands/live/init/cmdliveinit.go
```diff
@@ -193,15 +211,25 @@ func (c *ConfigureInventoryInfo) Run(ctx context.Context) error {
...
-		randomSuffix := common.RandomStr()
-		c.Name = fmt.Sprintf("%s-%s", defaultInventoryName, randomSuffix)
+		if c.InventoryID != "" {
+			dirName := filepath.Base(c.Pkg.UniquePath.String())
+			if errs := validation.IsDNS1123Subdomain(dirName); len(errs) > 0 {
+				return errors.E(op, c.Pkg.UniquePath,
+					fmt.Errorf("directory name %q is not a valid Kubernetes resource name and --name was not provided: %s",
+						dirName, strings.Join(errs, "; ")))
+			}
+			c.Name = dirName
```
liamfallon, line 224:
```suggestion
           if trimAndValidatename(filepath.Base(c.Pkg.UniquePath.String()) {
              c.name = filepath.Base(c.Pkg.UniquePath.String()
            } else {
               pakgErrors.WrapF(Err, <blah blah>)
            }
```

PR #4417, opened 2026-02-26
Proofreading of chapter 1.

mpgreaves:


documentation/content/en/book/01-getting-started/_index.md
```diff
@@ -1,67 +1,66 @@
...
+## System requirements
 
 ### kpt
 
-Install the [kpt CLI](installation/kpt-cli):
+Install the [kpt CLI](installation/kpt-cli), using the following command:
 
 ```shell
 kpt version
 ```
```
mozesl-nokia, line 21: That is not the installation command, but to check if you have it installed.

documentation/content/en/book/01-getting-started/_index.md
```diff
@@ -82,63 +81,61 @@ Package "nginx"
 └── [svc.yaml]  Service my-nginx-svc
 ```
 
-As you can see, this package contains 3 resources in 3 files. There is a special file named `Kptfile` which is used by
+As can be seen, this package contains three resources in three files. There is a special file named `Kptfile`. This file is used by
```
mozesl-nokia, line 84: "As can be seen" sounds a little archaic to me.

I don't think we should split "...special file named `Kptfile` which is used by..." into two different sentences.

documentation/content/en/book/01-getting-started/_index.md
```diff
@@ -82,63 +81,61 @@ Package "nginx"
...
 
 ```yaml
 pipeline:
   validators:
     - image: ghcr.io/kptdev/krm-functions-catalog/kubeconform:latest
 ```
 
-You might want to label all resources in the package. To achieve that, you can declare `set-labels` function in the
-`pipeline` section of `Kptfile`. Add this by running the following command:
+It might be a good idea to label all the resources in the package. To achieve this, you can declare the `set-labels` function, in the
```
mozesl-nokia, line 137:
```suggestion
It might be a good idea to label all the resources in the package. To achieve this, you can declare the `set-labels` function in the
```

documentation/content/en/book/01-getting-started/_index.md
```diff
@@ -149,75 +146,69 @@ cat >> Kptfile <<EOF
...
+Delete the package from the cluster, using the following command:
 
 ```shell
 kpt live destroy
 ```
 
-Congrats! You should now have a rough idea of what kpt is and what you can do
-with it. Now, let's delve into the details.
+You should now have a rough idea of what kpt is and what you can do
+with it. Let us now delve into the details.
```
mozesl-nokia, line 214: I agree with copilot here, "let us now" sounds too formal IMO.

PR #4427, opened 2026-03-06
fix(update): preserve Kptfile formatting during upgrades

### LINKED ISSUE #4306: Prevent formatting the kptfile during upgrades
### Describe your problem

As discussed during the Nephio meetup in September 2025, we want to prevent formatting the kptfile during upgrades:
  * Keep the original comments and formatting of kptfile
  * In the new SDK there is a go package called kptfile. Using this go package will preserve most of the formatting.
  * We need an issue to change the kpt to use the kptfile go package -> @lapentad

Jaisheesh-2006:
## Description

This PR fixes Kptfile formatting drift during package upgrades by switching upgrade-time Kptfile reads/updates to the KRM Functions SDK `kptfileko` flow via `kptfileutil`, and by removing the legacy rewrite path from `kptops`. The SDK-backed path preserves user-authored comments, spacing, and ordering when fields are unchanged, while still applying the intended semantic updates.

Key changes include:
- Upgrade path now uses the centralized SDK-backed read/update/write flow in `kptfileutil` instead of custom rewrite logic in `pkg/lib/kptops`.
- `kptfileutil.WriteFile` and `WriteKptfileToFS` preserve formatting by round-tripping through `kptfileko` and applying changes to the existing Kptfile instead of re-serializing from scratch.
- SDK-generated annotations are stripped on decode/update to avoid leaking internal metadata into user files.
- Empty-but-non-nil maps/slices are removed so `labels: {}` and `annotations: {}` are not introduced during updates.

Test and normalization hardening:
- Added/expanded Kptfile preservation tests covering idempotent round-trips, comment retention, and unchanged-field formatting stability.
- Updated Kptfile diff normalization in the test runner to be cross-platform and stable:
  - CRLF -> LF normalization
  - Stable normalization of hunk headers and index lines
  - Order-insensitive comparisons for Kptfile map keys and certain quoted scalars
- Symlink assertions are now environment-aware to avoid false failures on systems where symlinks are not materialized.

## Motivation

During `kpt pkg update`, Kptfile rewrites could unintentionally reformat user-authored content (comments, spacing, map ordering), producing noisy diffs and regressions. Issue #4306 tracks this behavior and expects upgrades to preserve Kptfile structure as much as possible while still applying necessary semantic updates. This change ensures formatting stability in the upgrade path while keeping behavior consistent across platforms.

## Issue

Fixes #4306

## AI use

I have used AI in the creation of this PR. The following AI tool was used

- GitHub Copilot (GPT-5.2-Codex).

I reviewed every change and can explain the design and test assertions independently.


internal/util/render/executor_test.go
```diff
@@ -341,8 +341,12 @@ func TestRenderer_Execute_RenderOrder(t *testing.T) {
...
 
 			fnResults, err := renderer.Execute(ctx)
-			assert.NoError(t, err)
-			assert.NotNil(t, fnResults)
+			if !assert.NoError(t, err) {
+				return
+			}
+			if !assert.NotNil(t, fnResults) {
+				return
+			}
```
mozesl-nokia: I think what you want here is require.NoError and require.NotNil instead.

internal/util/render/executor_test.go
```diff
@@ -273,7 +273,7 @@ kind: Kptfile
 metadata:
   name: root-package
   annotations:
-    kpt.dev/bfs-rendering: %t
+    kpt.dev/bfs-rendering: "%t"
```
mozesl-nokia, line 280: Nice catch!

internal/util/render/executor_test.go
```diff
@@ -440,7 +444,9 @@ metadata:
 
 	// Create a mock hydration context
 	root, err := newPkgNode(mockFileSystem, rootPkgPath, nil)
-	assert.NoError(t, err)
+	if !assert.NoError(t, err) {
+		return
+	}
```
mozesl-nokia:
```suggestion
	require.NoError(t, err)
```

pkg/kptfile/kptfileutil/util.go
```diff
@@ -78,6 +88,20 @@ func (e *UnknownKptfileResourceError) Error() string {
...
+		}
+		return nil
+	}
+
+	if kf, ok := k.(kptfilev1.KptFile); ok {
+		if err := writeKptfilePreservingFormat(dir, &kf); err != nil {
+			return errors.E(op, types.UniquePath(dir), err)
+		}
+		return nil
+	}
```
mozesl-nokia: Would this look better in a switch statement? (actual question, not suggestion)

Something like:
```go
switch k.(type) {
case *kptfilev1.Kptfile:
	...
case kptfilev1.Kptfile:
	...
}
```

pkg/kptfile/kptfileutil/util.go
```diff
@@ -113,6 +137,174 @@ func marshalKptfile(k any) ([]byte, error) {
...
+	if err := existingKptfile.WriteToPackage(existingResources); err != nil {
+		return err
+	}
+	if writeErr := os.WriteFile(kptfilePath, []byte(existingResources[kptfilev1.KptFileName]), 0600); writeErr != nil {
+		return errors.E(errors.IO, types.UniquePath(kptfilePath), writeErr)
+	}
+	return nil
+}
+
+func applyTypedKptfileToSDK(sdkKptfile *kptfileko.KptfileKubeObject, desired *kptfilev1.KptFile) error {
```
mozesl-nokia:
```suggestion
func applyTypedKptfileToKubeObject(sdkKptfile *kptfileko.KptfileKubeObject, desired *kptfilev1.KptFile) error {
```

pkg/kptfile/kptfileutil/util.go
```diff
@@ -113,6 +137,174 @@ func marshalKptfile(k any) ([]byte, error) {
...
+	if sdkKptfile == nil {
+		return fmt.Errorf("cannot update empty sdk Kptfile")
```
mozesl-nokia:
```suggestion
		return fmt.Errorf("cannot update empty Kptfile KubeObject")
```

pkg/kptfile/kptfileutil/util.go
mozesl-nokia, line 1: I feel like this file has gotten very long and should be split up.

pkg/lib/kptops/clone_test.go
```diff
@@ -15,13 +15,22 @@
...
 
 const exampleRepoURL = "https://github.com/example/repo.git"
 
+func normalizeLineEndings(s string) string {
+	return strings.ReplaceAll(s, "\r\n", "\n")
+}
+
+func normalizeAndTrim(s string) string {
+	return strings.TrimSpace(normalizeLineEndings(s))
+}
```
mozesl-nokia: You could probably move these to a more central place and use them elsewhere

internal/builtins/pkg_context_test.go
```diff
@@ -62,7 +63,9 @@ func TestPkgContextGenerator(t *testing.T) {
 			if err != test.expErr {
 				t.Errorf("exp: %v got: %v", test.expErr, err)
 			}
-			if diff := cmp.Diff(string(exp), out.String()); diff != "" {
+			expected := strings.ReplaceAll(string(exp), "\r\n", "\n")
+			actual := strings.ReplaceAll(out.String(), "\r\n", "\n")
```
mozesl-nokia: You could use that `normalizeLineEndings` function here

pkg/lib/kptops/render_test.go
```diff
@@ -78,10 +78,10 @@ func TestRender(t *testing.T) {
 				t.Errorf("Render failed: %v", err)
 			}
 
-			got := output.String()
-			want := readFile(t, filepath.Join(testdata, test.name, test.want))
+			got := strings.ReplaceAll(output.String(), "\r\n", "\n")
+			want := strings.ReplaceAll(string(readFile(t, filepath.Join(testdata, test.name, test.want))), "\r\n", "\n")
```
mozesl-nokia: You could use that `normalizeLineEndings` function here

pkg/kptfile/kptfileutil/util_test.go
```diff
@@ -595,9 +595,438 @@ status:
 				t.FailNow()
 			}
 
-			assert.Equal(t, strings.TrimSpace(tc.expected)+"\n", string(c))
+			expectedObj := map[string]any{}
+			err = yaml.Unmarshal([]byte(strings.TrimSpace(tc.expected)), &expectedObj)
+			if !assert.NoError(t, err) {
+				t.FailNow()
+			}
```
mozesl-nokia: I can see a lot of these, please use require.NoError instead

pkg/kptfile/kptfileutil/util_test.go
```diff
@@ -595,9 +595,438 @@ status:
...
+
+func TestUpdateKptfile_PreservesCommentsAndFormatting(t *testing.T) {
+	writeKptfileToTemp := func(tt *testing.T, content string) string {
+		dir := tt.TempDir()
+		err := os.WriteFile(filepath.Join(dir, kptfilev1.KptFileName), []byte(content), 0600)
+		if !assert.NoError(tt, err) {
+			tt.FailNow()
+		}
+		return dir
+	}
```
mozesl-nokia: I think this is a bit of an overcomplication. Just have function like this in global scope:

```go
writeKptfileToTemp(t *testing.T, content string) string {
	t.Helper()
	dir := t.TempDir()
	err := os.WriteFile(filepath.Join(dir, kptfilev1.KptFileName), []byte(content), 0600)
	require.NoError(t, err)
	return dir
}
```

And reuse it wherever.

pkg/kptfile/kptfileutil/util.go
Jaisheesh-2006: Yes, a type switch is definitely much cleaner and more idiomatic here. I have updated the code to use it. Thank you for pointing this out!

internal/util/render/executor_test.go
Jaisheesh-2006: I have replaced all instances of assert Noerror and NotNil with require for fail-fast

pkg/kptfile/kptfileutil/util.go
Jaisheesh-2006: I completely agree, it has become a bit of a catch-all file. Would you prefer I split it up in a separate, dedicated refactoring PR after this one merges? I want to keep the diff on this current PR focused so it is easy to review.

pkg/kptfile/kptfileutil/util_test.go
Jaisheesh-2006: Implemented

pkg/kptfile/kptfileutil/util_test.go
Jaisheesh-2006: Thank you for pointing out the best practice regarding require versus assert. I have updated all the error and nil checks to use require in the specific test files modified for this PR (executor_test.go and util_test.go).

I noticed there are many other instances of assert.NoError across the broader repository. To prevent scope creep and keep this PR easy to review, I would like to handle those in a separate, dedicated refactoring PR once this feature is merged. Let me know if that works for you!

pkg/lib/kptops/clone_test.go
Jaisheesh-2006: Good idea. I have moved these functions to **_internal/util/strings/strings.go_** file so they can be reused across the test suite.

CsatariGergely:
Please address the comments from Copilot and @mozesl-nokia  and fix the tests.

pkg/kptfile/kptfileutil/util.go
mozesl-nokia: Sure, it can be separate

mozesl-nokia (dismissed):

Jaisheesh-2006:
Hi @CsatariGergely, was working on the issue with tests, a little more left to correct regarding podman test (network issue). I will resolve the podman issue and copilot suggestions within 1-2 days

Jaisheesh-2006:
@CsatariGergely All the failing test issues and copilot issues have been resolved, kindly let me know if anything else to look into regarding this branch fix

pkg/fn/runtime/container_test.go
```diff
@@ -36,9 +36,44 @@ func TestContainerFn(t *testing.T) {
 		err    bool
 	}{
 		{
-			name:   "no-op function",
-			image:  "ghcr.io/kptdev/krm-functions-catalog/no-op:latest",
-			output: "apiVersion: v1\nkind: ResourceList\nmetadata:\n  name: output\nitems: []\n",
+			name:  "no-op function",
```
liamfallon, line 39: Did you try to rebase this PR? This shouldn't be part of this PR so if you rebase off main it should "vanish" off the PR.

Jaisheesh-2006: I rebased it now

CsatariGergely:
@Jaisheesh-2006 do you plan to progress with this?

Jaisheesh-2006:
> @Jaisheesh-2006 do you plan to progress with this?

I dont think I can since this is a bigger PR and will be difficult for the maintainers to evaluate, it will be good if I drop it and work on it freshly later

Issue #4450, opened 2026-03-26
Stabilize kpt API to version 1


ciaranjohnston:
As discussed here:

https://github.com/kptdev/kpt/discussions/4447

* kpt does not have an API on wire, but it has an API on types
* kpt has some API objects in the repo first
* Let's do semantic versioning for kpt, the SDK and for the individual functions in the function catalog

Things what we need to do before announcing v1.0.0
* There are some types copied from Kuberntes and kubctl. We should go back and use the upstream ones. These are in kpt and in Porch also
* Update the documentation
* The SDK and the function catalog have their separate versions
* kpt version should also provide the right version on all architectures
* API and module versions are different
* Add a VERSIONING.md to the kpt repo explaining the versioning strategy

Dependencies
* The functions use the SDK and the SDK uses kpt
* The function use only kpt types, there is no need to bump the version in the functions until the types are backward compatible

aravindtga:
Addressed all the issues and kpt 1.0.0 is released.

PR #4468, opened 2026-04-06
build: Implement semantic versioning in build system

NETIZEN-11:
## Description
This PR implements semantic versioning in the kpt build system using Git tags. It ensures that the version is correctly injected into the binary at build time and provides consistent version output across all architectures.

Part of #4450 - Stabilize kpt API to version 1

### Makefile Updates

1. **Semantic Version Detection**
   - Uses `git describe --tags --match='v*'` to automatically detect version from Git tags
   - Follows semantic versioning format: vMAJOR.MINOR.PATCH
   - Fallback to `v0.0.0-dev+{commit-hash}` for development builds
   - Ensures version consistency across all build environments

2. **Version Injection via ldflags**
   - Injects version into `github.com/kptdev/kpt/run.version` variable
   - Version is embedded at compile time
   - Works seamlessly with the updated version command

## How It Works

```makefile
# Automatic version detection
VERSION := $(shell git describe --tags --match='v*' --abbrev=0 2>/dev/null || echo "v0.0.0-dev+${GIT_COMMIT}")

# Version injection during build
LDFLAGS := -ldflags "-X github.com/kptdev/kpt/run.version=${VERSION}"

### CODE REVIEW

**Review by liamfallon — DISMISSED** [review]

**Comment by liamfallon** [comment]
The solution here  does not insert the correct version string into the binary when building a development version of kpt, see this log:


Before fix:
```
kpt % git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
kpt % make build
go build -ldflags "-X github.com/kptdev/kpt/run.version=2cd4e75fa " -o /Users/liam/go/bin/kpt -v .
github.com/kptdev/kpt
kpt % kpt version
2cd4e75fa
```
After fix:
```
kpt % git status
On branch main
Your branch is up to date with 'origin/main'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   Makefile

no changes added to commit (use "git add" and/or "git commit -a")
kpt % git diff
diff --git a/Makefile b/Makefile
index 7b64949f8..048641216 100644
--- a/Makefile
+++ b/Makefile
@@ -22,9 +22,12 @@ YEAR_GEN          := $(shell date '+%Y')
 GOBIN := $(shell go env GOPATH)/bin
 GIT_COMMIT := $(shell git rev-parse --short HEAD)
 
+# Use git describe to get semantic version, fallback to commit hash for dev builds
+VERSION := $(shell git describe --tags --match='v*' --abbrev=0 2>/dev/null || echo "v0.0.0-dev+${GIT_COMMIT}")
+
 export KPT_FN_WASM_RUNTIME ?= nodejs
 
-LDFLAGS := -ldflags "-X github.com/kptdev/kpt/run.version=${GIT_COMMIT}
+LDFLAGS := -ldflags "-X github.com/kptdev/kpt/run.version=${VERSION}
 ifeq ($(OS),Windows_NT)
        # Do nothing
 else
kpt % make build
go build -ldflags "-X github.com/kptdev/kpt/run.version=v1.0.0-beta.62 " -o /Users/liam/go/bin/kpt -v .
github.com/kptdev/kpt
kpt % kpt version
v1.0.0-beta.62
```
I think the last version string should be `v0.0.0-dev+2cd4e75fa`

**Comment by NETIZEN-11** [comment]
@liamfallon Good catch, thanks for pointing this out!

You're right — the previous implementation was incorrectly picking the latest tag even for development builds.

I've updated the logic to:
- Use the tag only when the commit exactly matches a tag and the working tree is clean
- Fall back to `v0.0.0-dev+<commit>` for development/dirty builds

This should now produce the expected version string. Please take another look.

**Comment by liamfallon** [comment]
This still doesn't seem to be working (on a cheked out tag), see below:

```
kpt % git status
HEAD detached at v1.0.0-beta.62
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   Makefile

no changes added to commit (use "git add" and/or "git commit -a")
kpt % git diff
diff --git a/Makefile b/Makefile
index 7b64949f8..919efb334 100644
--- a/Makefile
+++ b/Makefile
@@ -21,10 +21,23 @@ YEAR_GEN          := $(shell date '+%Y')
 
 GOBIN := $(shell go env GOPATH)/bin
 GIT_COMMIT := $(shell git rev-parse --short HEAD)
+GIT_TAG := $(shell git describe --tags --exact-match 2>/dev/null)
+DIRTY := $(shell git status --porcelain)
+
+# Version logic:
+# - Use tag ONLY if exact match AND clean working tree
+# - Otherwise fallback to dev version
+ifeq ($(GIT_TAG),)
+VERSION := v0.0.0-dev+$(GIT_COMMIT)
+else ifneq ($(DIRTY),)
+VERSION := v0.0.0-dev+$(GIT_COMMIT)
+else
+VERSION := $(GIT_TAG)
+endif
 
 export KPT_FN_WASM_RUNTIME ?= nodejs
 
-LDFLAGS := -ldflags "-X github.com/kptdev/kpt/run.version=${GIT_COMMIT}
+LDFLAGS := -ldflags "-X github.com/kptdev/kpt/run.version=${VESION}
 ifeq ($(OS),Windows_NT)
        # Do nothing
 else
kpt % make build
go build -ldflags "-X github.com/kptdev/kpt/run.version= " -o /Users/liam/go/bin/kpt -v .
kpt % kpt version

kpt % # The kpt version on the tagged branch is blank
```

NETIZEN-11:
Hi @liamfallon 👋

Thanks for pointing this out.

The issue was caused by a typo in the Makefile (`VESION` instead of `VERSION`), which resulted in an empty value being passed during build. I've fixed this now.

Also, regarding your test case: the working tree is currently dirty (`Makefile` modified), so the version logic correctly falls back to the dev version instead of using the tag.

Once the working tree is clean, it should correctly pick up the tag (e.g. `v1.0.0-beta.62`).

Could you please pull the latest changes and verify again? Thanks!

liamfallon:
> Hi @liamfallon 👋
> 
> Thanks for pointing this out.
> 
> The issue was caused by a typo in the Makefile (`VESION` instead of `VERSION`), which resulted in an empty value being passed during build. I've fixed this now.
> 
> Also, regarding your test case: the working tree is currently dirty (`Makefile` modified), so the version logic correctly falls back to the dev version instead of using the tag.
> 
> Once the working tree is clean, it should correctly pick up the tag (e.g. `v1.0.0-beta.62`).
> 
> Could you please pull the latest changes and verify again? Thanks!

Yes, I know it was dirty. I pulled a tagged version and pasted in your change; your change is not merged so it is not in a tagged version yet.

liamfallon:
Also when you have updated a PR, feel free to remove the "awaiting author action" label so we can see that the PR comments are addressed without having to dig into the PR.

liamfallon:
I think we should reconsider the whole approach here. I think the way we did this for `prochctl` in the Porch project is a lot cleaner.

For local Makefile builds, we use the Makefile and code in the `porchctl` command itself, see:
- https://github.com/nephio-project/porch/blob/4509accabdfe3a4f0f92cedbf02f159ef6f7ee2b/make/build.mk#L50
- https://github.com/nephio-project/porch/blob/4509accabdfe3a4f0f92cedbf02f159ef6f7ee2b/cmd/porchctl/run/run.go#L161

For releases and builds on Github we use a Github action:
- https://github.com/nephio-project/porch/blob/main/release/tag/goreleaser.yaml
- https://github.com/nephio-project/porch/blob/main/.github/workflows/release.yaml

NETIZEN-11:
Hi @liamfallon 👋

I've refactored the implementation based on your feedback.

- Simplified the Makefile (removed complex version logic)
- Moved version handling to code
- Aligned the approach with standard release workflows

This should address the concerns around maintainability and separation of local vs release builds.

Could you please take another look and let me know your thoughts? Thanks!

liamfallon:
The core changed in this PR go beyond what is explained in the title and description, there are version changes and path handling changes.

liamfallon:
There has been no update on this PR in quite some time. Please feel free to reopen the PR at any time and to address the comments above.

PR #4469, opened 2026-04-07
feat: Add CEL-based conditional function execution (#4388)

SurbhiAgarwal1:
## Description

This PR adds CEL-based conditional function execution to kpt, implementing #4388.

A new optional `condition` field is added to the `Function` type in the Kptfile pipeline. When specified, the CEL expression is evaluated against the current list of KRM resources. If the expression returns `false`, the function is skipped. If omitted or returns `true`, the function executes normally.

## Motivation

In many real-world scenarios, you want a pipeline function to run only when certain resources exist or meet specific criteria. Without this feature, users had to maintain separate Kptfiles or manually manage which functions run. The `condition` field makes pipelines more dynamic and reduces the need for workarounds.

## Changes

- Added `condition` field to `Function` type in `pkg/api/kptfile/v1/types.go`
- Added `CELEnvironment` in `pkg/lib/runneroptions/celenv.go` using `google/cel-go`
- Integrated condition check in `FunctionRunner.Filter()` in `internal/fnruntime/runner.go`
- Functions skipped due to condition show `[SKIPPED]` in CLI output
- Added `InitCELEnvironment()` method to `RunnerOptions` for proper error handling
- Updated all callers of `InitDefaults()` to also call `InitCELEnvironment()`
- Added E2E testdata for `condition-met` and `condition-not-met` cases
- Added unit tests for CEL evaluation covering all 3 runtimes (builtin, exec, container)
- Updated documentation: kptfile schema reference and book/04-using-functions

## Example Usage

```yaml
pipeline:
  mutators:
    - image: ghcr.io/kptdev/krm-functions-catalog/set-namespace:v0.4
      condition: "resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'env-config')"
      configMap:
        namespace: production

### CODE REVIEW

**Comment by liamfallon** [comment]
Comment from #4391:

Closing this PR in favor of a clean rebase. The branch had accumulated 61 commits including upstream commits from other contributors, making it messy to review.

All the feedback from this review has been addressed in the new PR which has a single clean commit:

    Removed unnecessary type alias file (celeval.go)
    Grouped constants in const() block
    Replaced interface{} with any
    Removed panic from InitDefaults, added InitCELEnvironment() error method
    Added e2e tests for condition-met and condition-not-met
    Added immutability test
    Added documentation (kptfile schema + book/04-using-functions)
    Removed all unwanted files (krm-functions-catalog, output.txt, etc.)
    Fixed diff.patch for renderStatus field

Thank you all for the thorough review!

**Comment by liamfallon** [comment]
@SurbhiAgarwal1

The tests are failing because we have merged the conditions and renederstatus to kpt recently. I thought that by deleting the `diff.patch` files, the comparison would just check the output and the tests would pass. However we also need to add the difference in the Kprfile after render to the diff.patch files. Now that I deleted the `diff.patch` files in the PR, I can't see any option to restore them on the Github GUI.

The following `diff.patch files work for me on the tests locally on my machine:

e2e/testdata/fn-render/condition/condition-met/.expected/diff.patch
```
diff --git a/Kptfile b/Kptfile
index eb90ac3..d305dc0 100644
--- a/Kptfile
+++ b/Kptfile
@@ -5,4 +5,14 @@ metadata:
 pipeline:
   mutators:
     - image: ghcr.io/kptdev/krm-functions-catalog/no-op
-      condition: "resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'app-config')"
+      condition: resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'app-config')
+status:
+  conditions:
+    - type: Rendered
+      status: "True"
+      reason: RenderSuccess
+  renderStatus:
+    mutationSteps:
+      - image: ghcr.io/kptdev/krm-functions-catalog/no-op
+        exitCode: 0
```

e2e/testdata/fn-render/condition/condition-not-met/.expected/diff.patch
```
diff --git a/Kptfile b/Kptfile
index eb90ac3..b055f32 100644
--- a/Kptfile
+++ b/Kptfile
@@ -5,4 +5,14 @@ metadata:
 pipeline:
   mutators:
     - image: ghcr.io/kptdev/krm-functions-catalog/no-op
-      condition: "resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'app-config')"
+      condition: resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'app-config')
+status:
+  conditions:
+    - type: Rendered
+      status: "True"
+      reason: RenderSuccess
+  renderStatus:
+    mutationSteps:
+      - image: ghcr.io/kptdev/krm-functions-catalog/no-op
+        exitCode: 0
+        skipped: true
```

Can you try restoring the two `diff.patch` files with the contents above and see if that fixes the tests?

Sorry for mucking with your PR 😢

**Comment by SurbhiAgarwal1** [comment]
Thank you @liamfallon! I've restored both diff.patch files with exactly the contents you provided in commit df1189f6c. The tests should pass now. Sorry for the confusion!

**Comment by SurbhiAgarwal1** [comment]
Hi @liamfallon, I've updated the diff.patch files again using the exact actual output from the CI logs. The tests are still failing with the same diff - it seems the expected diff.patch I have doesn't match what CI produces.

The actual diff CI produces for condition-met is:
- index eb90ac3..5ccde1c 100644
- @@ -5,4 +5,13 @@
- exitCode: 0 with \ No newline at end of file

And for condition-not-met:
- index eb90ac3..ac7ae33 100644  
- @@ -5,4 +5,14 @@
- Plus resources.yaml diff: index 7806994..c8cdecf

I've set the diff.patch files to exactly match this output. Could you check if there's something else causing the mismatch? The CI run is #8920 if you want to look at the full logs.

**Comment by liamfallon** [comment]
Hi @SurbhiAgarwal1 , Have a look at this test PR, I just cloned this PR across so that I could push changes directly. I think the tests are running on that one so if you copy the diff.patch files from that PR into this PR I hope it'll work:

https://github.com/kptdev/kpt/pull/4472

**Comment by SurbhiAgarwal1** [comment]
Hi @liamfallon, thank you so much for doing that! I've copied the diff.patch files from PR #4472 into this PR. The tests should pass now. Really appreciate you taking the time to debug this!

File: e2e/testdata/fn-render/condition/condition-met/.expected/diff.patch [hunk 1]
```diff
@@ -0,0 +1,19 @@
+diff --git a/Kptfile b/Kptfile
+index eb90ac3..a13955d 100644
```
liamfallon:
```suggestion
index eb90ac3..5ccde1c 100644
```

e2e/testdata/fn-render/condition/condition-met/.expected/diff.patch
```diff
@@ -0,0 +1,19 @@
+diff --git a/Kptfile b/Kptfile
+index eb90ac3..5ccde1c 100644
+--- a/Kptfile
++++ b/Kptfile
+@@ -5,4 +5,14 @@ metadata:
```
liamfallon:
```suggestion
@@ -5,4 +5,13 @@ metadata:
```

liamfallon:
@SurbhiAgarwal1 All tests passing now!

The problems we experienced here are why it is generally not recommended to have text comparison in tests, but to fix all the tests in kpt would be a huge job.

SurbhiAgarwal1:
Got it, makes sense. Thanks!
Yeah I noticed the text comparison was causing these issues. Glad it’s passing now.
Appreciate the help on this — I’ll keep it in mind if I touch similar tests later.

liamfallon:
If you have time to look at the copilot comments that would be great. We can move to formally review and merge the PR then.

internal/util/render/executor.go
```diff
@@ -870,11 +872,14 @@ func (pn *pkgNode) runValidators(ctx context.Context, hctx *hydrationContext, in
 			hctx.validationSteps = append(hctx.validationSteps, preExecFailureStep(function, err))
 			return err
 		}
-		if _, err = validator.Filter(cloneResources(selectedResources)); err != nil {
+		validatorRunner := validator.(*fnruntime.FunctionRunner)
```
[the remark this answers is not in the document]
SurbhiAgarwal1: this can panic if the type isn’t what we expect. I’ll update this to use the comma-ok check and handle the error case properly.

pkg/lib/runneroptions/runneroptions.go
```diff
@@ -57,13 +57,29 @@ type RunnerOptions struct {
 
 	// ResolveToImage will resolve a partial image to a fully-qualified one
 	ResolveToImage ImageResolveFunc
+
+	// CELEnvironment is the shared CEL environment used to evaluate function conditions.
+	// It is initialised by InitDefaults and reused across all function runners.
```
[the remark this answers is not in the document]
SurbhiAgarwal1: Yeah the comment doesn’t match the actual flow. I’ll update it to reflect the InitCELEnvironment usage.

commands/fn/render/cmdrender.go
```diff
@@ -85,6 +85,9 @@ type Runner struct {
 
 func (r *Runner) InitDefaults() {
 	r.RunnerOptions.InitDefaults(runneroptions.GHCRImagePrefix)
+	// Initialize CEL environment for condition evaluation
+	// Ignore error as conditions are optional; if CEL init fails, conditions will error at runtime
+	_ = r.RunnerOptions.InitCELEnvironment()
```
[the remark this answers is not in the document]
SurbhiAgarwal1: Ignoring this error isn’t safe here since it can hide the root failure. I’ll update it to return the error and align other similar cases as well.

documentation/content/en/book/04-using-functions/_index.md
```diff
@@ -375,6 +375,68 @@ will merge each function pipeline list as an associative list, using
...
+
+If the condition is not met:
+
+```shell
+$ kpt fn render wordpress
+Package "wordpress":
+
+[SKIPPED] "ghcr.io/kptdev/krm-functions-catalog/set-labels:latest" (condition not met)
+
+Successfully executed 1 function(s) in 1 package(s).
```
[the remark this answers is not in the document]
SurbhiAgarwal1: Since the step is skipped, it shouldn’t be counted as executed. I’ll update the example output to 0.

e2e/testdata/live-apply/apply-depends-on/config.yaml
```diff
@@ -17,7 +17,7 @@ parallel: true
 kptArgs:
   - "live"
   - "apply"
-  - "--reconcile-timeout=2m"
+  - "--reconcile-timeout=5m"
```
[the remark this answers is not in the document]
SurbhiAgarwal1: This change was just to stabilize the E2E, but you’re right it’s not related to the main PR. I’ll either document it in the PR description or move it to a separate PR.

thirdparty/kyaml/runfn/runfn.go
```diff
@@ -413,5 +414,9 @@ func (r *RunFns) defaultFnFilterProvider(spec runtimeutil.FunctionSpec, fnConfig
 		opts.DisplayResourceCount = true
 	}
 
-	return fnruntime.NewFunctionRunner(r.Ctx, fltr, "", fnResult, r.fnResults, opts)
+	runner, _ := fnruntime.NewFunctionRunner(r.Ctx, fltr, "", fnResult, r.fnResults, opts)
+	if r.Condition != "" {
+		runner.SetCondition(r.Condition, opts.CELEnvironment)
+	}
+	return runner, nil
```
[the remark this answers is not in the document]
SurbhiAgarwal1: I’m currently ignoring the error from NewFunctionRunner, which isn’t safe. I’ll propagate the error and also handle the case where CEL env is nil when a condition is set.

internal/fnruntime/runner.go
```diff
@@ -191,10 +210,44 @@ type FunctionRunner struct {
...
 	pr := printer.FromContextOrDie(fr.ctx)
+	
+	// Check condition before executing function
+	if fr.celEnv != nil && fr.condition != "" {
+		shouldExecute, err := fr.celEnv.EvaluateCondition(fr.ctx, fr.condition, input)
+		if err != nil {
+			return nil, fmt.Errorf("failed to evaluate condition for function %q: %w", fr.name, err)
+		}
+
+		if !shouldExecute {
```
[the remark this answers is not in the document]
SurbhiAgarwal1: You’re right — if the same runner is reused, the skipped state can leak across executions. I’ll reset those fields at the start of Filter().

documentation/content/en/book/04-using-functions/_index.md
```diff
@@ -375,6 +375,68 @@ will merge each function pipeline list as an associative list, using
...
 
+### Specifying `condition`
+
+The `condition` field lets you skip a function based on the current state of the resources in the package.
+It takes a [CEL](https://cel.dev/) expression that is evaluated against the resource list. If the expression
+returns `true`, the function runs. If it returns `false`, the function is skipped.
+
+The expression receives a variable called `resources`, which is a list of all KRM resources passed to
+this function step (after `selectors` and `exclude` have been applied). Each resource is a map with
+the standard fields: `apiVersion`, `kind`, `metadata`, `spec`, `status`.
```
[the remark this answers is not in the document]
SurbhiAgarwal1: the docs should match the actual normalization behavior, and the skipped example output should show 0 executed functions. I’ll update both.

internal/fnruntime/runner.go
```diff
@@ -191,10 +210,44 @@ type FunctionRunner struct {
...
+			}
+			// Append a skipped result so consumers get one result per pipeline step
+			fr.fnResult.ExitCode = 0
+			fr.fnResult.Skipped = true
+			fr.fnResults.Items = append(fr.fnResults.Items, *fr.fnResult)
+			// Return input unchanged - function is skipped
+			fr.skipped = true
+			return input, nil
+		}
+	}
```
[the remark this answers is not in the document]
liamfallon: Should this be fixed?

liamfallon (commented):
Looks good to me.

Could you have a look at that last copilot comment and see if anything needs to be done?

nagygergo (approved):
Looks good to me. Thanks for the contribution @SurbhiAgarwal1

liamfallon (dismissed):

internal/fnruntime/runner.go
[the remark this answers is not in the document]
SurbhiAgarwal1: This is already fixed in the current code. fr.skipped and fr.fnResult .Skipped are reset at the beginning of filter()  before evaluating the condition.

e2e/testdata/fn-render/condition/condition-not-met/.expected/diff.patch
```diff
@@ -0,0 +1,31 @@
...
+ pipeline:
+   mutators:
+     - image: ghcr.io/kptdev/krm-functions-catalog/no-op
+-      condition: "resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'app-config')"
++      condition: resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'app-config')
++status:
++  conditions:
++    - type: Rendered
++      status: "True"
++      reason: RenderSuccess
```
efiacor, line 26: If the render is skipped, should the status.condition not reflect that somehow? I know we capture the "skipped" below but a RenderSuccess reason here seems wrong.

liamfallon: Would this be better?

```
conditions:
    - type: Rendered
      status: "False"
      reason: RenderSkipped
```

But @aravindtga, would this impact the status of the pipeline because the pipeline should still succeed even though one of the renders did not occur>

nagygergo: The render reflects the full pipeline, not the individual function's execution. If you really need to reflect this information somehow, I'd recommend still having `status: True` and `reason: RenderedWithSkippedFunctions`, or a separate set of conditions.
But I'd argue that the rendering pipeline has included the conditions as well, so the render was fully successful.

liamfallon: OK, maybe it's better to leave it as it is so. @efiacor what do you think?

aravindtga: I agree with @nagygergo. Even though the pipeline function is skipped, the overall rendering is successful. Marking `Status: false` gives the misleading impression that rendering has failed. 

From kpt perspective, `status.renderStatus` does include a `skipped:true`(As shown below) field that distinguishes skipped mutations from actual failure. 

However, consumers like Porch don't parse `status.renderStatus`, So they can't differentiate between a skip and a failure. If we want to handle this, it would be better to use something like `reason: RenderedWithSkippedFunctions` rather than setting the status to false.

e2e/testdata/fn-render/condition/condition-not-met/.expected/diff.patch
```diff
@@ -0,0 +1,31 @@
...
++  renderStatus:
++    mutationSteps:
++      - image: ghcr.io/kptdev/krm-functions-catalog/no-op:latest
++        exitCode: 0
```
aravindtga, line 31: It would also be worth capturing the `condition` field in `status.renderStatus`, so the skip condition is recorded alongside the render result.

SurbhiAgarwal1: I agree with @nagygergo and @aravindtga - the render condition should reflect the overall pipeline result, not individual function execution. Since the pipeline completed successfully (just with some functions skipped), keeping status: "True" makes sense.

I can update the reason to RenderedWithSkippedFunctions when at least one function was skipped, so consumers can distinguish between a fully executed render and one with skipped functions. I can also add the condition field to status.renderStatus so the skip reason is recorded alongside the result.

Let me know if you'd like me to go ahead with these changes

e2e/testdata/fn-render/condition/condition-not-met/.expected/diff.patch
liamfallon: We now store the result of each individual krm function execution in a render run in the Kptfile including whether a krm function was skipped or not.

Therefore, if a render run is successful and that render run had skips, then it was still successful and the current overall result is correct.
```
conditions:
    - type: Rendered
      status: "True"
      reason: RenderSuccess
```

If the user wishes to see what krm function was skipped, tehy can look at the kptfile.

Therefore the code is fine as it is.

SurbhiAgarwal1: Make sense! Since the granular execution details (including whether a function was skipped) are already captured for each individual function in the Kptfile, keeping the overall reason as RenderSuccess is perfectly fine. I'll leave the code as it is.

e2e/testdata/fn-render/condition/condition-not-met/.expected/diff.patch
aravindtga: > It would also be worth capturing the `condition` field in `status.renderStatus`, so the skip condition is recorded alongside the render result.

A condition field has been introduced in the mutation pipeline configuration:
```yaml
mutators:
  - image: ghcr.io/kptdev/krm-functions-catalog/no-op
    condition: resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'app-config')
````

Can you also add the condition field to status.renderStatus.mutationSteps so it gets recorded when the pipeline execution is captured? e.g:

```yaml
status:
  conditions:
    - type: Rendered
      status: "True"
      reason: RenderSuccess
  renderStatus:
    mutationSteps:
      - image: ghcr.io/kptdev/krm-functions-catalog/no-op:latest
        condition: resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'app-config')
        exitCode: 0
      ....
```

SurbhiAgarwal1:
Hi @liamfallon, the condition tests are now passing. The remaining Docker failure is TestFnRender/testdata/fn-render/subpkg-fn-failure which expects "statefulset-filter:4:42: got newline, want primary expression" but gets a different starlark error format. This test file is unchanged from upstream - is this a known flaky test?

CsatariGergely:
Please rebase. The test flakyness was fixed last week, so tests should work now. 
Fix the DCO using [this description](https://github.com/kptdev/kpt/pull/4469/checks?check_run_id=73012273092).

SurbhiAgarwal1:
Hi @CsatariGergely,

Yes, I'm still working on this. Unfortunately, I've been facing some issues with my computer recently, which has delayed my progress. I'll get the environment set up again, rebase the branch, and update the PR as soon as possible.

Thank you for your patience.

efiacor:
> Hi @CsatariGergely,
> 
> Yes, I'm still working on this. Unfortunately, I've been facing some issues with my computer recently, which has delayed my progress. I'll get the environment set up again, rebase the branch, and update the PR as soon as possible.
> 
> Thank you for your patience.

Hi @SurbhiAgarwal1 
How can we assist in moving this PR along?

run_e2e.sh
efiacor, line 1: This can be removed

internal/kptops/doc.go
efiacor, line 1: Did these changes in /internal get pulled in by accident?
I think we can remove them

api/kptfile/v1/types.go
```diff
@@ -361,6 +361,20 @@ type Function struct {
...
+	// `Exclude` have been applied) and should return a boolean value.
+	// If omitted or evaluates to true, the function executes normally.
+	// If evaluates to false, the function is skipped.
+	//
+	// Example: Check if a specific ConfigMap exists among the selected resources:
+	//   condition: "resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'my-config')"
+	//
+	// Example: Check resource count among the selected resources:
+	//   condition: "resources.filter(r, r.kind == 'Deployment').size() > 0"
+	Condition string `yaml:"condition,omitempty" json:"condition,omitempty"`
```
efiacor: I think the naming here could be confusing. There is already a Condition type in the file and k8s also has it's own Condition type defined.
Maybe something like the following which clarifies the Go code but keeps the impl agnostic for the user.

CelCondition string `yaml:"when,omitempty" json:"when,omitempty"`

test-local/Kptfile
efiacor, line 1: Should these tests be nested somewhere?
test_condition_not_met also..

go.mod
```diff
@@ -30,6 +31,7 @@ require (
 	k8s.io/api v0.36.1
 	k8s.io/apiextensions-apiserver v0.36.1
 	k8s.io/apimachinery v0.36.1
+	k8s.io/apiserver v0.36.1
```
efiacor: This is expensive just to get the cel/library.
Could we manage without them?
[plain cel-go module?](https://pkg.go.dev/github.com/google/cel-go)

api/kptfile/v1/types.go
```diff
@@ -361,6 +361,20 @@ type Function struct {
 	// `Exclude` are used to specify resources on which the function should NOT be executed.
 	// If not specified, all resources selected by `Selectors` are selected.
 	Exclusions []Selector `yaml:"exclude,omitempty" json:"exclude,omitempty"`
+
+	// `Condition` is an optional CEL expression that determines whether this
```
efiacor: Comment needs to be update to align with new naming and usage

commands/fn/render/cmdrender_test.go
```diff
@@ -32,15 +33,17 @@ func TestCmd_flagAndArgParsing_Symlink(t *testing.T) {
 	err := os.MkdirAll(filepath.Join(dir, "path", "to", "pkg", "dir"), 0700)
 	assert.NoError(t, err)
 	err = os.Symlink(filepath.Join("path", "to", "pkg", "dir"), "foo")
-	assert.NoError(t, err)
+	if err != nil {
```
efiacor: If we have issues with running these tests due to Windows permissions, this needs to be addressed somehow.
These should work in WSL though.

commands/pkg/diff/cmddiff_test.go
```diff
@@ -78,7 +79,9 @@ func TestCmd_flagAndArgParsing_Symlink(t *testing.T) {
 	err := os.MkdirAll(filepath.Join(dir, "path", "to", "pkg", "dir"), 0700)
 	assert.NoError(t, err)
 	err = os.Symlink(filepath.Join("path", "to", "pkg", "dir"), "foo")
-	assert.NoError(t, err)
+	if err != nil {
```
efiacor: If we have issues with running these tests due to Windows permissions, this needs to be addressed somehow.
These should work in WSL though.

documentation/content/en/book/04-using-functions/_index.md
```diff
@@ -339,6 +339,69 @@ pipeline:
 
 It is recommended to use unique function names for all the functions in the Kptfile function pipeline. If the `name` is specified, then the `kpt pkg update` will merge each function pipeline list as an associative list, using `name` as the merge key. An unspecified `name`, or duplicated names, may result in unexpected merges.
 
+### Specifying `condition`
+
+The `condition` field lets you skip a function based on the current state of the resources in the package.
```
efiacor: All of this section will have to re align to the new naming and usage.

CelCondition -> when

Note, the "when" is a convenience for the api user. For me it reads better in the Kptfile.

mutators:
    - image: ghcr.io/kptdev/krm-functions-catalog/set-labels:latest
      configMap:
        app: wordpress
      when: resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'app-config')

pkg/fn/runtime/celeval_test.go
```diff
@@ -0,0 +1,160 @@
+// Copyright 2026 The kpt and Nephio Authors
```
efiacor: All new headers should be:

Copyright 2026 The kpt  Authors

pkg/fn/runtime/condition_test.go
```diff
@@ -0,0 +1,118 @@
...
+	require.NoError(t, err)
+
+	inputNodes := []*yaml.RNode{
+		yaml.MustParse("apiVersion: v1\nkind: ConfigMap\nmetadata:\n  name: app-config"),
+	}
+
+	testCases := []struct {
+		name      string
+		fn        *kptfile.Function
+		condition string
```
efiacor, line 46: celCond

pkg/fn/runtime/runner.go
```diff
@@ -180,7 +180,26 @@ func NewRunner(
...
+	}
+
+	// Set condition; the shared CEL environment from opts is used at evaluation time.
+	if f.CelCondition != "" {
+		if opts.CELEnvironment == nil {
+			name := f.Image
+			if name == "" {
+				name = f.Exec
+			}
+			return nil, fmt.Errorf("condition specified for function %q but no CEL environment is configured in RunnerOptions", name)
```
efiacor: All logs and other "condition" refs should adopt the new name: CelCondition|celCondition

pkg/fn/runtime/runner.go
```diff
@@ -221,10 +240,46 @@ type FunctionRunner struct {
 	fnResult         *fnresultv1.Result
 	fnResults        *fnresultv1.ResultList
 	opts             runneroptions.RunnerOptions
+	condition        string                        // CEL condition expression
```
efiacor: celCond

pkg/fn/runtime/runner.go
```diff
@@ -221,10 +240,46 @@ type FunctionRunner struct {
...
+	// Check condition before executing function
+	if fr.celEnv != nil && fr.condition != "" {
+		shouldExecute, err := fr.celEnv.EvaluateCondition(fr.ctx, fr.condition, input)
+		if err != nil {
+			return nil, fmt.Errorf("failed to evaluate condition for function %q: %w", fr.name, err)
+		}
+
+		if !shouldExecute {
+			if !fr.disableCLIOutput {
+				pr.Printf("[SKIPPED] %q (condition not met)\n", fr.name)
```
efiacor: celCondition

pkg/lib/kptops/render_executor.go
```diff
@@ -26,7 +26,7 @@ import (
 	fnresultv1 "github.com/kptdev/kpt/api/fnresult/v1"
 	kptfilev1 "github.com/kptdev/kpt/api/kptfile/v1"
 	"github.com/kptdev/kpt/pkg/fn"
-	fnruntime "github.com/kptdev/kpt/pkg/fn/runtime"
```
efiacor, line 29: Why drop this? Any reason?

pkg/lib/kptops/render_executor.go
```diff
@@ -713,8 +721,7 @@ func (pn *pkgNode) runPipeline(ctx context.Context, hctx *hydrationContext, inpu
 	// TODO: the DisplayPath is a relative file path. It cannot represent the
 	// package structure. We should have function to get the relative package
 	// path here.
-	prOpts := printer.NewOpt().PkgDisplay(pn.pkg.DisplayPath).PkgName(hctx.rootName)
-	pr.OptPrintf(prOpts, "\n")
+	pr.OptPrintf(printer.NewOpt().PkgDisplay(pn.pkg.DisplayPath), "\n")
```
efiacor: Why the change here?
Also, dropping the context in places below?

pkg/lib/runneroptions/celenv.go
```diff
@@ -0,0 +1,152 @@
+// Copyright 2026 The kpt and Nephio Authors
```
efiacor: Copyright 2026 The kpt Authors

pkg/lib/runneroptions/celenv.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+// CELEnvironment holds a shared CEL environment for evaluating conditions.
+// The environment is created once and reused; programs are compiled per condition call.
+type CELEnvironment struct {
+	env *cel.Env
+}
+
+// NewCELEnvironment creates a new CELEnvironment with the standard KRM variable bindings.
+// Includes cel-go built-in extensions and k8s-specific validators (IP, CIDR, Quantity, SemVer)
```
efiacor, line 34: Comment needs to be updated

pkg/lib/runneroptions/celenv.go
```diff
@@ -0,0 +1,152 @@
...
+	"fmt"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/ext"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+const (
+	celCheckFrequency = 100
```
efiacor: Should these be exposed to be configurable?
@aravindtga @nagygergo

pkg/lib/runneroptions/celenv.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create CEL environment: %w", err)
+	}
+	return &CELEnvironment{env: env}, nil
+}
+
+// EvaluateCondition compiles and evaluates a CEL condition against a list of KRM resources.
+// Returns true if the condition is met, false otherwise.
+// An empty condition always returns true (function executes unconditionally).
+func (e *CELEnvironment) EvaluateCondition(ctx context.Context, condition string, resources []*yaml.RNode) (bool, error) {
```
efiacor: ..., celCond string, ...

pkg/lib/runneroptions/celenv.go
```diff
@@ -0,0 +1,152 @@
...
+
+	result, ok := out.(types.Bool)
+	if !ok {
+		return false, fmt.Errorf("CEL expression must return a boolean, got %T", out)
+	}
+
+	return bool(result), nil
+}
+
+func resourcesToList(resources []*yaml.RNode) ([]any, error) {
```
efiacor, line 99: pretty sure we have some libs that do most of this already.
@mozesl-nokia ? any tips?

pkg/lib/util/get/get_test.go
```diff
@@ -202,6 +202,13 @@ func TestCommand_Run_subdir_symlinks(t *testing.T) {
 	})
 	defer clean()
 
+	// Skip if the test repo doesn't have real symlinks (e.g. Windows/WSL with
```
efiacor: Not sure can just skip these for specific env. Can we get this working on WSL?
@aravindtga

thirdparty/cmdconfig/commands/cmdeval/cmdeval.go
```diff
@@ -117,6 +117,8 @@ func GetEvalFnRunner(ctx context.Context, parent string) *EvalFnRunner {
 		&r.excludeAnnotations, "exclude-annotations", []string{}, "exclude resources matching the given annotations")
 	r.Command.Flags().StringArrayVar(
 		&r.excludeLabels, "exclude-labels", []string{}, "exclude resources matching the given labels")
+	r.Command.Flags().StringVar(
+		&r.CelCondition, "condition", "", "conditional expression to determine if function should be run")
```
efiacor, line 122:
```suggestion
		&r.CelCondition, "when", "", "conditional expression to determine if function should be run")
```

thirdparty/cmdconfig/commands/cmdeval/cmdeval_test.go
```diff
@@ -452,15 +453,17 @@ func TestCmd_flagAndArgParsing_Symlink(t *testing.T) {
 	err = os.MkdirAll(filepath.Join(dir, "path", "to", "pkg", "dir"), 0700)
 	assert.NoError(t, err)
 	err = os.Symlink(filepath.Join("path", "to", "pkg", "dir"), "foo")
-	assert.NoError(t, err)
+	if err != nil {
```
efiacor: Are these symlink tests broken on Windows env?

thirdparty/kyaml/runfn/runfn.go
```diff
@@ -411,5 +412,9 @@ func (r *RunFns) defaultFnFilterProvider(spec runtimeutil.FunctionSpec, fnConfig
 		opts.DisplayResourceCount = true
 	}
 
-	return fnruntime.NewFunctionRunner(r.Ctx, fltr, "", fnResult, r.fnResults, opts)
+	runner, _ := fnruntime.NewFunctionRunner(r.Ctx, fltr, "", fnResult, r.fnResults, opts)
```
efiacor: Should we be checking the err here? _

pkg/fn/runtime/runner.go
```diff
@@ -221,10 +240,46 @@ type FunctionRunner struct {
 	fnResult         *fnresultv1.Result
 	fnResults        *fnresultv1.ResultList
 	opts             runneroptions.RunnerOptions
+	condition        string                        // CEL condition expression
+	celEnv           *runneroptions.CELEnvironment // shared CEL environment for condition evaluation
+	skipped          bool                          // true if function execution was skipped due to condition
+}
+
+func (fr *FunctionRunner) SetCondition(condition string, celEnv *runneroptions.CELEnvironment) {
```
efiacor:
```suggestion
func (fr *FunctionRunner) SetCelCondition(celCond string, celEnv *runneroptions.CELEnvironment) {
```

.gitattributes
efiacor, line 1: Is this file necessary?

go.mod
```diff
@@ -2,14 +2,15 @@ module github.com/kptdev/kpt
...
-//replace github.com/kptdev/kpt/api => ./api
+replace github.com/kptdev/kpt/api => ./api
 
 require (
 	github.com/Masterminds/semver/v3 v3.5.0
 	github.com/bytecodealliance/wasmtime-go v1.0.0
 	github.com/cpuguy83/go-md2man/v2 v2.0.7
 	github.com/distribution/reference v0.6.0
 	github.com/go-errors/errors v1.5.1
+	github.com/google/cel-go v0.26.0
```
efiacor: Could we go with a more recent version?
latest is v0.28.1

api/kptfile/v1/types.go
SurbhiAgarwal1: Done! I've updated the struct comment to refer to `When` instead of `Condition`, and I also updated the YAML examples inside the comment to use when: instead of condition:. This is included in the latest commit.

commands/pkg/diff/cmddiff_test.go
SurbhiAgarwal1: You are completely right! To address this properly without breaking WSL, I have updated the code to explicitly check the OS via runtime.GOOS == "windows". Because WSL registers as "linux", the test will run normally there (where symlinks work), but it will safely skip if a developer runs the suite natively on standard Windows.

documentation/content/en/book/04-using-functions/_index.md
SurbhiAgarwal1: Done! I've updated this entire section in the documentation (04-using-functions/_index.md) to realign with the new naming. All references to condition and CelCondition have been replaced with when, and the YAML examples now correctly demonstrate the usage of when: exactly as you suggested. This is included in the latest commit.

thirdparty/cmdconfig/commands/cmdeval/cmdeval_test.go
SurbhiAgarwal1: Yes, exactly. Native Windows requires either Administrator privileges or Developer Mode to be enabled in order to create symlinks via os.Symlink. Without those, the tests fail with a "privilege not held" error for anyone running them locally on Windows.

To fix this properly, I've updated the latest commit to explicitly check for runtime.GOOS == "windows" and use t.Skip(). This gracefully skips the test for native Windows users while still running normally on WSL, macOS, and our Linux CI pipelines.

thirdparty/kyaml/runfn/runfn.go
SurbhiAgarwal1: Yes, we definitely should. I've updated the code to properly check the error from NewFunctionRunner and return it instead of silently ignoring it with _. This fix is included in my latest commit.

.gitattributes
SurbhiAgarwal1: No, it is not necessary.  I have completely removed the .gitattributes file in the latest commit.

go.mod
SurbhiAgarwal1: Yes, absolutely. I have updated the cel-go dependency to v0.28.1 and ran go mod tidy to update go.mod and go.sum accordingly. This is included in the latest commit!

api/kptfile/v1/types.go
```diff
@@ -361,6 +361,20 @@ type Function struct {
 	// `Exclude` are used to specify resources on which the function should NOT be executed.
 	// If not specified, all resources selected by `Selectors` are selected.
 	Exclusions []Selector `yaml:"exclude,omitempty" json:"exclude,omitempty"`
+
+	// `When` is an optional CEL expression that determines whether this
```
efiacor:
```suggestion
	// `CelCondition` is an optional CEL expression that determines whether this
```

documentation/content/en/reference/schema/kptfile/kptfile.yaml
```diff
@@ -71,6 +71,16 @@ definitions:
                     this is primarily used for merging function declaration with upstream counterparts
                 type: string
                 x-go-name: Name
+            condition:
+                description: |-
+                    `Condition` is an optional CEL expression that determines whether this
```
efiacor:
```suggestion
                    `CelCondition` is an optional CEL expression that determines whether this
```

e2e/testdata/fn-render/condition/condition-met/Kptfile
```diff
@@ -0,0 +1,8 @@
+apiVersion: kpt.dev/v1
+kind: Kptfile
+metadata:
+  name: app
+pipeline:
+  mutators:
+    - image: ghcr.io/kptdev/krm-functions-catalog/no-op
+      when: "resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'app-config')"
```
efiacor:
```suggestion
      when: resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'app-config')
```

e2e/testdata/fn-render/condition/condition-not-met/Kptfile
```diff
@@ -0,0 +1,8 @@
+apiVersion: kpt.dev/v1
+kind: Kptfile
+metadata:
+  name: app
+pipeline:
+  mutators:
+    - image: ghcr.io/kptdev/krm-functions-catalog/no-op
+      when: "resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'app-config')"
```
efiacor:
```suggestion
      when: resources.exists(r, r.kind == 'ConfigMap' && r.metadata.name == 'app-config')
```

pkg/fn/runtime/condition_test.go
```diff
@@ -0,0 +1,118 @@
+// Copyright 2026 The kpt and Nephio Authors
```
efiacor:
```suggestion
// Copyright 2026 The kpt Authors
```

pkg/fn/runtime/runner.go
efiacor:
```suggestion
			return nil, fmt.Errorf("CelCondition specified for function %q but no CEL environment is configured in RunnerOptions", name)
```

pkg/fn/runtime/runner.go
efiacor:
```suggestion
	celCondition        string                        // CEL condition expression
```

pkg/fn/runtime/runner.go
```diff
@@ -221,10 +240,46 @@ type FunctionRunner struct {
...
+	// Check condition before executing function
+	if fr.celEnv != nil && fr.condition != "" {
+		shouldExecute, err := fr.celEnv.EvaluateCondition(fr.ctx, fr.condition, input)
+		if err != nil {
+			return nil, fmt.Errorf("failed to evaluate condition for function %q: %w", fr.name, err)
+		}
+
+		if !shouldExecute {
+			if !fr.disableCLIOutput {
+				pr.Printf("[SKIPPED] %q (condition not met)\n", fr.name)
```
efiacor:
```suggestion
				pr.Printf("[SKIPPED] %q (celCondition not met)\n", fr.name)
```

pkg/lib/runneroptions/celenv.go
efiacor:
```suggestion
// Copyright 2026 The kpt Authors
```

efiacor (commented):
Nice progress made. Please address all comments/suggestions to proceed. Thanks

e2e/testdata/fn-render/condition/condition-met/.expected/config.yaml
```diff
@@ -0,0 +1,12 @@
+actualStripLines:
+  - "    stderr: 'WARNING: The requested image''s platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested'"
+
+stdErrStripLines:
+  - "  Stderr:"
+  - "    \"WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested\""
```
aravindtga: Can these be removed, or does the no-op function throw a compatibility error without them?

e2e/testdata/fn-render/condition/condition-not-met/.expected/config.yaml
```diff
@@ -0,0 +1,11 @@
+actualStripLines:
+  - "    stderr: 'WARNING: The requested image''s platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested'"
+
+stdErrStripLines:
+  - "  Stderr:"
+  - "    \"WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested\""
```
aravindtga: Can these be removed, or does the no-op function throw a compatibility error without them?

e2e/testdata/fn-render/condition/condition-not-met/.expected/diff.patch
aravindtga: `when` should be introduced into the pipeline result as suggested here before.

commands/pkg/get/cmdget_test.go
```diff
@@ -400,7 +400,9 @@ func TestCmd_flagAndArgParsing_Symlink(t *testing.T) {
 	err := os.MkdirAll(filepath.Join(dir, "path", "to", "pkg", "dir"), 0700)
 	assert.NoError(t, err)
 	err = os.Symlink(filepath.Join("path", "to", "pkg", "dir"), "link")
-	assert.NoError(t, err)
+	if err != nil {
+		t.Skipf("skipping test due to symlink creation failure (requires admin/developer mode on Windows): %v", err)
```
aravindtga: This skip should be conditional on Windows only. Skipping unconditionally means we lose the assertion on environments where the error is valid.

commands/pkg/update/cmdupdate_test.go
```diff
@@ -351,7 +351,9 @@ func TestCmd_flagAndArgParsing_Symlink(t *testing.T) {
 	err := os.MkdirAll(filepath.Join(dir, "path", "to", "pkg", "dir"), 0700)
 	assert.NoError(t, err)
 	err = os.Symlink(filepath.Join("path", "to", "pkg", "dir"), "foo")
-	assert.NoError(t, err)
+	if err != nil {
+		t.Skipf("skipping test due to symlink creation failure (requires admin/developer mode on Windows): %v", err)
+	}
```
aravindtga, line 359: This skip should be conditional on Windows only. Skipping unconditionally means we lose the assertion on environments where the error is valid.

pkg/lib/runneroptions/celenv.go
aravindtga: Yes @efiacor It would be preferable to expose these as configurable options rather than hardcoded constants. Since the when field accepts arbitrary user-written CEL expressions, complex expressions could hit the hardcoded limits today. Having configurable limits would provide the flexibility needed without requiring code changes.

pkg/lib/util/get/get_test.go
aravindtga: WSL natively supports symlinks, so these tests should work fine there. The problem only arises when the repo is cloned on Windows (or with `core.symlinks`=`false`) and then accessed from WSL - in that case git stores symlink targets as plain files rather than actual symlinks. The solution for WSL is simply to clone the repo within WSL itself.

For Windows, a `runtime.GOOS == "windows"` check is the appropriate way to skip these tests. I'd prefer this change to be scoped specifically to Windows rather than using a broader file-inspection approach that could inadvertently skip tests on other platforms.

thirdparty/kyaml/runfn/runfn.go
```diff
@@ -411,5 +412,12 @@ func (r *RunFns) defaultFnFilterProvider(spec runtimeutil.FunctionSpec, fnConfig
 		opts.DisplayResourceCount = true
 	}
 
-	return fnruntime.NewFunctionRunner(r.Ctx, fltr, "", fnResult, r.fnResults, opts)
+	runner, err := fnruntime.NewFunctionRunner(r.Ctx, fltr, "", fnResult, r.fnResults, opts)
+	if err != nil {
+		return nil, err
+	}
+	if r.CelCondition != "" {
```
aravindtga, line 443: nil check for opts.CELEnvironment?

SurbhiAgarwal1:
Hi @efiacor  and @aravindtga ,

I finally tracked down the root cause of the CEL condition test failures! It turns out cmdrender.go and cmdeval.go were silently ignoring errors from InitCELEnvironment(). If the CEL environment failed to initialize, the functions would just silently bypass condition evaluations instead of surfacing the error.

I have fixed those files to explicitly handle the initialization error and fail loudly. I've pushed the fixes to the branch. The test suite requires some Unix commands (like cp) so it flakes on my local Windows setup, but these error propagation fixes should resolve the condition evaluation skips we were seeing in CI. Let me know if everything looks green on your end!

efiacor:
> Hi @efiacor and @aravindtga ,
> 
> I finally tracked down the root cause of the CEL condition test failures! It turns out cmdrender.go and cmdeval.go were silently ignoring errors from InitCELEnvironment(). If the CEL environment failed to initialize, the functions would just silently bypass condition evaluations instead of surfacing the error.
> 
> I have fixed those files to explicitly handle the initialization error and fail loudly. I've pushed the fixes to the branch. The test suite requires some Unix commands (like cp) so it flakes on my local Windows setup, but these error propagation fixes should resolve the condition evaluation skips we were seeing in CI. Let me know if everything looks green on your end!

Hi @SurbhiAgarwal1
We do need to try to "resolve" any code review comments left by reviewers. Perhaps you have missed some due to how the github UI "folds" multiple comments. Please expand, reply and|or mark as resolved.
We cannot re review until all previous comments are addressed AND the CI runs are fully passing.

Unfortunately, I have little experience with Windows development and I would suggest that we do not fully support Windows development. 
Some of our contributors are successfully using WSL. Perhaps this could be an option for you.

Again, thanks for your efforts with this but we do need to get this completed asap. Can you address the above so we can proceed swiftly.

thirdparty/kyaml/runfn/runfn.go
SurbhiAgarwal1: Added a nil check here for `opts.CELEnvironment` so that it returns an explicit error if it hasn't been initialized, rather than panicking.

commands/pkg/update/cmdupdate_test.go
SurbhiAgarwal1: Fixed! I've updated the skip logic to explicitly check for `runtime.GOOS == "windows"`. If the symlink creation fails on any other platform, the test will now properly fail with the error.

e2e/testdata/fn-render/condition/condition-met/.expected/config.yaml
SurbhiAgarwal1: Yes, they have been removed! The tests now pass cleanly without needing those strip lines.

e2e/testdata/fn-render/condition/condition-not-met/.expected/config.yaml
SurbhiAgarwal1: Yes, they have been removed! The tests run perfectly fine without needing them.

commands/pkg/get/cmdget_test.go
SurbhiAgarwal1: - Fixed! I've updated the skip condition to explicitly check for `runtime.GOOS == "windows"`. If the symlink creation fails on any other platform, the test will now correctly fail.

SurbhiAgarwal1:
You're completely right about the WSL symlink behavior. I've updated the tests to drop the file-inspection approach and replaced it with a strict `runtime.GOOS == "windows"` check instead. This ensures the tests are only skipped on native Windows, while preserving the test assertions for Linux, macOS, and properly cloned WSL environments. 

(I also ran into the local Docker/Podman failures myself because I had originally cloned the repo on my Windows mount instead of natively inside WSL, but cloning it directly inside WSL fixed it completely!)

aravindtga:
It look like you've addressed most of the review comments, with just a few remaining. There's also a merge conflict and some CI failures that need attention.

We're happy to help you get those resolved and move the reviews along - feel free to reach out if you have any questions.

cc/ @liamfallon @efiacor

liamfallon:
Hi @SurbhiAgarwal1 , have you resolved all the comments above?

aravindtga:
Hi @SurbhiAgarwal1 

This feature has been getting more traction from the community recently and we'd like to prioritise getting it merged. Could you pick this up and address the outstanding items?

1. Rebase onto current main and resolve the merge conflict
2. Address any remaining review comments
3. Get CI passing (including the SonarQube duplication issue)

If you're blocked on anything or need any help, feel free to reach out - we're happy to assist.

cc/ @liamfallon @efiacor

SurbhiAgarwal1:
Hi @aravindtga @efiacor @liamfallon,

I have updated the PR with all fixes and test fixture updates:

1. **Skipped Functions Logic Fix**: Updated `runMutators` and `runValidators` in `render_executor.go` so that functions skipped due to a CEL condition (`when`) are **not** counted toward `executedFunctionCnt` (e.g. outputs `"Successfully executed 0 function(s)"` when all functions are skipped).
2. **Metadata Propagation**: Ensured `When` and `Skipped` fields are populated in `PipelineStepResult` and serialized into `Kptfile` `renderStatus`.
3. **Configurable Limits**: Made `CelCheckFrequency` and `CelCostLimit` configurable via `RunnerOptions` rather than hardcoded constants.
4. **E2E Snapshot Updates**: Updated the expected `stderr` and `diff.patch` test fixtures (`condition-met`, `condition-not-met`, `subpkg-fn-failure`) to match `upstream/main`'s printer output format and `renderStatus` schema.
5. **Test Verification**: Verified unit tests (`pkg/fn/runtime`, `pkg/lib/runneroptions`, `pkg/lib/kptops`) and container e2e test suites pass 100% locally.

All commits are signed off and pushed to the branch. Ready for your review!

efiacor (approved):
We will have to wave this through and pickup the additional pieces afterwards
@aravindtga

aravindtga (approved):

PR #4501, opened 2026-04-24
fix(live): propagate caller ctx into ResourceGroup inventory ops

Jaisheesh-2006:
## Description

`pkg/live/inventoryrg.go` was using `context.TODO()` for every live Kubernetes API call it makes — 7 call sites covering `Apply`, `ApplyWithPrune`, and the `ResourceGroup` CRD match check. This PR threads the caller's real context through each of those calls.

Because the upstream `inventory.Storage` / `inventory.ClientFactory` interfaces do not accept a `ctx` parameter, the fix:

- Carries `ctx` on `InventoryResourceGroup` and `ClusterClientFactory` structs.
- Injects it at wrap time via two new factories: `WrapInventoryObjWithContext(ctx)` and `NewClusterClientFactoryWithContext(ctx)`.
- Falls back to `context.Background()` when the legacy constructors are used, so the change is fully backward-compatible.
- Adds a `ctx` parameter directly to `ResourceGroupCRDMatched` (not interface-constrained).

Six callers updated to pass their existing command ctx: `commands/live/apply`, `commands/live/destroy`, `commands/live/migrate`, `commands/live/livecmd`, `commands/alpha/live/plan`, and `pkg/live/planner`.

Three unit tests added in `pkg/live/inventoryrg_test.go`:
- `TestWrapInventoryObjWithContext_StoresContext` — verifies the factory persists the passed ctx onto the struct.
- `TestWrapInventoryObj_LeavesContextNil` — locks in legacy behavior for callers that haven't migrated.
- `TestContextOrBackground` — asserts the fallback path and, crucially, that cancellation on the caller's ctx propagates through (`cancel()` fires `Done()`).

## Motivation

Before this change, pressing `Ctrl-C` during `kpt live apply` / `destroy` / `plan` against a slow or unreachable cluster could not cancel the in-flight API call — the CLI would block until the network-level timeout fired. Any `context.WithTimeout` the caller wired at the command level was silently discarded at this boundary.

`context.TODO()` is a placeholder from the Go stdlib. This PR fulfills that TODO.

Discovered via a `grep`-level audit of `main` for `context.TODO()` / `context.Background()` usage in production code paths.

## AI use

I have used AI in the creation of this PR.

I have used the following AI tools:
- Claude Code (Claude Opus 4.7) to audit `main` for context-propagation gaps, to plan the interface-safe fix (struct-stored ctx + context-aware factory), to apply the refactor across the 10 affected files, and to author the unit tests.

I reviewed every change and can explain the design (why struct-stored ctx instead of adding it to `Apply`'s signature) and the test assertions independently.

### DIFF
--- a/pkg/live/inventoryrg_test.go
+++ b/pkg/live/inventoryrg_test.go
@@ -1,4 +1,4 @@
-// Copyright 2020 The kpt Authors
+// Copyright 2020,2026 The kpt Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -15,6 +15,7 @@
 package live
 
 import (
+	"context"
 	"testing"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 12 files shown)


pkg/live/inventoryrg_test.go
```diff
@@ -240,3 +242,118 @@ func TestIsResourceGroupInventory(t *testing.T) {
...
+	icm, ok := storage.(*InventoryResourceGroup)
+	if !ok {
+		t.Fatalf("WrapInventoryObjWithContext produced unexpected type %T", storage)
+	}
+	if icm.ctx == nil {
+		t.Fatal("expected ctx on InventoryResourceGroup; got nil")
+	}
+	if got := icm.ctx.Value(ctxKey{}); got != "propagated" {
+		t.Fatalf("expected stored ctx to carry propagated value; got %v", got)
+	}
```
mozesl-nokia: Please replace the rest of the t.Fatalf calls like this too.

```suggestion
	require.Truef(t, ok, "WrapInventoryObjWithContext produced unexpected type %T", storage)
	require.NotNil(t, icm.ctx, "expected ctx on InventoryResourceGroup")
	assert.Equal(t, "propagated", icm.ctx.Value(ctxKey{}), "expected stored ctx to carry propagated value")
```

mozesl-nokia (commented):
An overall comment would be that all the doc comments are a bit too long in my opinion, but great work!

pkg/live/inventoryrg_test.go
Jaisheesh-2006: Thanks for the review and the kind words, @mozesl-nokia! I am on it and will condense the doc comments.

liamfallon (approved):

mozesl-nokia (approved):

PR #4503, opened 2026-04-24
Validate function tag literals before execution

### LINKED ISSUE #4500: Catch blatantly invalid image tags before execution
### Description
Currently, the tag resolving logic in kpt treats the semver tags with a syntax issue and the hashed version tags the same way. It gives a warning [message](https://github.com/kptdev/kpt/blob/main/pkg/fn/runtime/tag_resolution.go#L88) whenever the tag field has syntax issue or the tag is a hash. From tag resolution perspective, the string initialized in the tag field will be added to the image name: https://github.com/kptdev/kpt/blob/main/pkg/fn/runtime/tag_resolution.go#L92-L93. It is beacuse the purpose of handling the hashed tags. However, in case of specifying semantic version directives in the tag field will end up with an image name of '<image_name>:<semver_constraint>'. That means, the funciton execution will continue and tries to find the image with that name. It would be nice to handle this and return with an error before any further execution procedures.

AkashKumar7902:
## Summary

- Validate literal `function.tag` values against image tag syntax before constructing the function image reference.
- Preserve semver constraint resolution and valid non-semver literal tags such as hash-style build tags.
- Add coverage for invalid literal tags that previously reached runtime execution.

## Motivation

Fixes #4500.

Invalid tag values that were neither semantic versions nor semantic version constraints could be appended to the image name and only fail later during function execution. This catches those values during tag resolution with an actionable error.

## Validation

- `go test ./pkg/fn/runtime`
- `go test ./pkg/fn/...`
- `go test -cover ./pkg/fn/runtime`
- `go vet ./pkg/fn/runtime`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run ./pkg/fn/runtime`

## AI Assistance

I used OpenAI GPT-5 for code exploration, implementation assistance, and test drafting. I reviewed the changes and validated them locally.


pkg/fn/runtime/tag_resolution.go
```diff
@@ -25,6 +26,10 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const imageTagError = "must start with an alphanumeric character or underscore, followed by at most 127 alphanumeric characters, underscores, periods, or dashes"
+
+var imageTagRegex = regexp.MustCompile(`^[A-Za-z0-9_][A-Za-z0-9._-]{0,127}$`)
```
mozesl-nokia: I think you could import this instead: https://github.com/distribution/reference/blob/0965666a6ade2e06035fe352e38344be1e68951a/regexp.go#L39

I would actually prefer a validator function from [distribution](https://github.com/distribution/distribution), but I can't find one...

pkg/fn/runtime/tag_resolution.go
```diff
@@ -85,10 +90,22 @@ func (tr *TagResolver) ResolveFunctionImage(ctx context.Context, image, tag stri
 		ref.Tag = allFilteredVersions[0].Original()
 		return ref.CommonName(), nil
 	} else {
+		resolvedImage, tagErr := imageWithLiteralTag(ref, tag)
+		if tagErr != nil {
+			return "", tagErr
+		}
 		klog.Warningf("Tag %q could not be parsed as a semantic version (\"%s\") or constraint (\"%s\"), will use it literally",
 			tag, versionErr, constraintErr)
+		return resolvedImage, nil
```
mozesl-nokia: I don't think this bit is necessary to change, just leave that warning in the else branch

AkashKumar7902:
@mozesl-nokia Thanks for the review! Pushed 9782c09 addressing both points:

- **regex import**: Switched to `reference.TagRegexp` from `github.com/distribution/reference`. That regex is unanchored, so I'm using `FindString(tag) == tag` to keep the previous full-string-match semantics. The exported helpers in `distribution/reference` (`Parse`/`WithTag`) are name+tag oriented and `anchoredTagRegexp` is unexported, so this is the cleanest plug-in.
- **else branch**: Restored the pre-PR shape — warning stays in the else, and the single `imageWithLiteralTag` call at the end handles validation for both branches.

@Copilot the wording suggestion (`must start with...`) was already applied in b47453b8. PTAL.

mozesl-nokia (approved):

nagygergo:
Lgtm. Can't approve for some reason

liamfallon (approved):

liamfallon:
@AkashKumar7902 please resolve the conflicts and then we can proceed to merge this PR. Thanks for the contribution!

go.sum
```diff
@@ -27,8 +27,12 @@ github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSs
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc h1:U9qPSI2PIWSS1VwoXQT9A3Wy9MM3WgvqSxFWenqJduM=
 github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
-github.com/docker/cli v29.4.0+incompatible h1:+IjXULMetlvWJiuSI0Nbor36lcJ5BTcVpUmB21KBoVM=
-github.com/docker/cli v29.4.0+incompatible/go.mod h1:JLrzqnKDaYBop7H2jaqPtU4hHvMKP+vjCwu2uszcLI8=
+github.com/distribution/reference v0.6.0 h1:0IXCQ5g4/QMHHkarYzh5l+u8T3t73zM5QvfrDyIgxBk=
+github.com/distribution/reference v0.6.0/go.mod h1:BbU0aIcezP1/5jX/8MP0YiH4SdvB5Y4f/wlDRiLyi3E=
+github.com/docker/cli v29.2.0+incompatible h1:9oBd9+YM7rxjZLfyMGxjraKBKE4/nVyvVfN4qNl9XRM=
+github.com/docker/cli v29.2.0+incompatible/go.mod h1:JLrzqnKDaYBop7H2jaqPtU4hHvMKP+vjCwu2uszcLI8=
```
liamfallon:
```suggestion
github.com/docker/cli v29.4.0+incompatible h1:+IjXULMetlvWJiuSI0Nbor36lcJ5BTcVpUmB21KBoVM=
github.com/docker/cli v29.4.0+incompatible/go.mod h1:JLrzqnKDaYBop7H2jaqPtU4hHvMKP+vjCwu2uszcLI8=
```

PR #4537, opened 2026-05-21
Make API it's own go module

### LINKED ISSUE #4523: Move kpt APIs into a kpt API module
Move the kpt and Porch APIs into the kpt API repo and create modules for them in there.

mozesl-nokia:
Instead of pulling the API into a separate git repository, this PR exposes it from this one as a separate go module.

This is done in order to lessen the dependency requirements of API users as well as remove circular dependencies between this repo and the SDK repo without the need for duplication in SDK.

Also tried to unify the import alias of the API packages across this repo. Had to move `pkg/lib/types` into `api/kptfile/v1` to avoid some dependency conflicts with SDK.

~~We could look into further removing some dependencies, like `sigs.k8s.io/cli-utils` from which we import a single const.~~ - done

Part of #4523

I have used AI in the creation of this PR:
- Used Cursor to generate the release YAML and docs (which I then had to heavily author myself)


api/go.mod
```diff
@@ -0,0 +1,32 @@
+module github.com/kptdev/kpt/api
+
+go 1.24.0
```
efiacor, line 3: Why are we using an older go version here?

mozesl-nokia: I think it is best for API modules to use as low of a version as possible so dependents don't have to upgrade unnecessarily.

efiacor: But do we then risk pulling in vulnerable indirect dependencies?

mozesl-nokia: Does that matter when these are just API types with a few util functions?

mozesl-nokia:
I realise that we probably need to update the workflows to handle the API as well. If anybody more familiar with GitHub workflows would like to, I would appreciate the help :smiley:.

api/Makefile
```diff
@@ -0,0 +1,54 @@
...
+lint:
+	@if [[ `command -v golangci-lint` && `golangci-lint version --short` == $(GOLANGCI_LINT_VERSION) ]]; then \
+		golangci-lint run -v ./...; \
+	else \
+		go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v$(GOLANGCI_LINT_VERSION) run -v ./...; \
+	fi
+
+.PHONY: test
+test:
+	go test -cover ${LDFLAGS} ./...
```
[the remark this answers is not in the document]
mozesl-nokia: This is not true, but I can remove the var I guess

.github/workflows/release-api.yml
```diff
@@ -0,0 +1,42 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: kpt API release
+on:
+  push:
+    tags:
+    - "api/v[0-9]+.[0-9]+.[0-9]+*"
```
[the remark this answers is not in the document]
mozesl-nokia: The plus symbol does work https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet

release/README.md
```diff
@@ -13,3 +17,26 @@ To cut a new kpt release perform the following:
...
+go get github.com/kptdev/kpt/api@v1.0.0
+```
+
+That `@v1.0.0` form is the **module version**; the corresponding **git tag** in this repo is **`api/v1.0.0`**.
+
+Tags must use the **`api/`** prefix so the version matches the module subdirectory (see [Go modules: VCS version](https://go.dev/ref/mod#vcs-version)).
+
+### What runs in CI
+
+When you push a tag that matches **`api/v[0-9]+.[0-9]+.[0-9]+*`** (for example `api/v1.2.3`), [.github/workflows/release-api.yml](../.github/workflows/release-api.yml) runs: it uses the Go version from [`api/go.mod`](../api/go.mod) and runs **`make api`**.
```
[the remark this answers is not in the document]
mozesl-nokia: See previous comment

liamfallon (approved):

efiacor (approved):

Issue #4516, opened 2026-05-08
Migrate all catalog functions to fn.WithDocs for --help/--doc support


efiacor:
## Context

PR https://github.com/kptdev/krm-functions-sdk/pull/752 adds `fn.WithDocs` to the SDK, enabling `--help` and `--doc` support via `//go:embed`. This issue tracks migrating all catalog functions to use it.

Parent issue: #4485

## Already using fn.AsMain (just need fn.WithDocs added)

These 8 functions already use `fn.AsMain` — migration is adding `//go:embed` + `fn.WithDocs` and deleting `generated/docs.go`:

- [x] set-labels
- [ ] set-namespace
- [ ] set-image
- [ ] delete-annotations
- [ ] apply-replacements
- [ ] render-helm-chart
- [ ] starlark
- [ ] kubeconform

## Still using kyaml/fn/framework + command.Build (full migration needed)

These functions need rewriting from `framework.ResourceList` to `fn.ResourceList`:

### Simple (thin wrappers)
- [ ] sleep
- [ ] remove-local-config-resources
- [ ] set-enforcement-action
- [ ] upsert-resource

### Medium
- [ ] apply-setters
- [ ] create-setters
- [ ] list-setters
- [ ] search-replace
- [ ] annotate-apply-time-mutations

### Complex
- [ ] set-annotations
- [ ] ensure-name-substring
- [ ] gatekeeper
- [ ] generate-kpt-pkg-docs (blocked on list-setters)

## Migration steps per function

1. Add `//go:embed README.md` and `//go:embed metadata.yaml`
2. Pass `fn.WithDocs(readme, metadata)` to `fn.AsMain`
3. Delete `generated/docs.go`
4. For legacy functions: rewrite from `framework.ResourceList` to `fn.ResourceList` / `fn.Runner`
5. Verify `kpt fn doc` works: `docker run --rm <image> --help`

## Done when

All catalog functions support `--help` and `--doc` via the SDK, and `generated/docs.go` files are removed.

PR #4587, opened 2026-06-11
Make mdtogo an independent Go module

aravindtga:
## Description
After the `api` module extraction, the root `go.mod` contains a `replace` directive (`github.com/kptdev/kpt/api => ./api`). 
This breaks external installation of mdtogo package in the repository via `go install ...@main` because Go does not allow `replace` directives in modules installed with a version query.

This affects `krm-functions-catalog`, which uses `go install github.com/kptdev/kpt/mdtogo@main` to generate function documentation.
 
Since the `api` split, that command fails with:

  > The go.mod file for the module providing named packages contains one or more
  > replace directives. It must not contain directives that would cause it to be
  > interpreted differently than if it were the main module.

`mdtogo` has no dependency on any other kpt package.  it only uses `golang.org/x/text` and `testify`. Making it its own module allows external `go install` to resolve it independently of the root `go.mod` and its `replace` directives.

  ## Changes
  - Add `mdtogo/go.mod` and `mdtogo/go.sum` (new independent module)
  - Update `Makefile` install target: `go install ./mdtogo` → `cd mdtogo && go install .`

  ## AI Disclosure

  - [x] I have used AI(Kiro CLI) in the creation of this PR.


efiacor (approved):
I'm hoping this will be semi temp given we will move away from generating docs for the fns
https://github.com/kptdev/kpt/issues/4516

mozesl-nokia (approved):

mozesl-nokia:
This is not just a problem with mdtogo, we have to remove the replace in order to make it "go install"-able

efiacor:
> This is not just a problem with mdtogo, we have to remove the replace in order to make it "go install"-able

Do we merge this anyway so?

mozesl-nokia:
@efiacor
If this is blocking something immediate, sure. Otherwise, we need to establish how to handle the versioning of the API, since it looks like we cannot have the same version for both like this.

efiacor:
> @efiacor If this is blocking something immediate, sure. Otherwise, we need to establish how to handle the versioning of the API, since it looks like we cannot have the same version for both like this.

Not sure, but I think the krm-fn builds/ci are broken given the api change.
@aravindtga could you confirm?
No panic on my side but I am planning to phase the mdtogo in favour of the new sdk api approach.

aravindtga:
> > @efiacor If this is blocking something immediate, sure. Otherwise, we need to establish how to handle the versioning of the API, since it looks like we cannot have the same version for both like this.
> 
> Not sure, but I think the krm-fn builds/ci are broken given the api change. @aravindtga could you confirm? No panic on my side but I am planning to phase the mdtogo in favour of the new sdk api approach.

@efiacor / @mozesl-nokia krm-functions-catalog ci/local build is broken with the api change. Also this is a temporary fix only and can be removed once we phase out mdtogo.

PR #4599, opened 2026-06-23
Add possibility to skip waiting in `kpt live apply`

mozesl-nokia:
Added a `--wait` flag so that it is possible to not wait for resources to be ready. Useful when wait conditions are not always as simple as Ready=True. The default is still `true`.

Further improvements: Add support for `kubectl wait --for=...` style wait conditions, or possibility to define such conditions in the Kptfile or resource metadata.


mozesl-nokia:
This needs to go much further into the depth of cliutils to be possible, we can revisit this later.

Issue #4595, opened 2026-06-18
Ability to customize DefaultImagePrefix at runtime


mozesl-nokia:
Currently, when running the kpt CLI, the default image prefix is hardcoded to be the GHCR prefix.

This should be customizable with and environment variable, a CLI argument or both.

kavirakesh14:
Hi @mozesl-nokia , I'd like to work on this issue, please assign this to me and i'll investigate and raise a PR. Thanks

Catalin-Stratulat-Ericsson:
> Hi [@mozesl-nokia](https://github.com/mozesl-nokia) , I'd like to work on this issue, please assign this to me and i'll investigate and raise a PR. Thanks

Hello @kavirakesh14 are you still working on this? has it been assigned. there are a few individuals starting on the kpt project and im looking for good first issues to assign. please let me know if you are actively working on this.

kavirakesh14:
Sure...please proceed. Thanks!

Catalin-Stratulat-Ericsson:
fixed by #4616

PR #4616, opened 2026-07-02
Make the image prefix configurable

OisinJohnston2005:
This PR is resolving issue #4595, which requests the ability to configure the default prefix used when resolving image names into fully qualified urls, it asks for the ability to do so either by environment variable or command line argument, this PR implements both prioritising cli args and falling back to the environment variable and then the hard coded prefix.

In order to implement this I removed the `ResolveToImageForCLI` function since its approach of currying the default image prefix into the `ResolveToImage` function attribute on the `RunnerOptions` struct didn't allow Cobra to update the ImagePrefix on runtime.

Instead I opted to implement `ResolveToImage` as a method and added the `ImagePrefix` attribute which is updated by cobra whenever a command is run, if no prefix is provided it first checks for the environment variable `KPT_IMAGE_PREFIX` if set it will use that as the image prefix if not it will use `ghcr.io/kptdev/krm-functions-catalog/`


liamfallon:
@OisinJohnston2005 Can you resolve the copilot comments please?

liamfallon (commented):
The implementation looks very good.

I'd be more comfortable with more comprehensive tests, maybe with some valid and invalid prefiexes. 

Does kpt already validate the URL once it's out together in the current code?

It'd be good to check that, see for example
https://stackoverflow.com/questions/31480710/validate-url-with-standard-package-in-go

OisinJohnston2005:
kpt currently has no URL validation other than attempting to pull the image, but since the prefix is hardcoded the possibility of a malformed one is pretty limited.

I'll go ahead and add some validation and some unittests

liamfallon:
> kpt currently has no URL validation other than attempting to pull the image, but since the prefix is hardcoded the possibility of a malformed one is pretty limited.
> 
> I'll go ahead and add some validation and some unittests

You're correct, now that users can specify it we should check it.

liamfallon (approved):

efiacor (approved):

Issue #2123, opened 2021-05-27
kpt live with dry-run doesn't recognize situations where the inventory namespace doesn't exist


mortent:
### Expected behavior
I would expect dry-run to fail if the namespace for the inventory RG doesn't exist. If the namespace is created as part of the package would expect it to succeed though.

### Actual behavior
dry-run succeeds, but actually applying the package fails.

CsatariGergely:
As this is a relatively old issue first check if the problem still persist!

PR #4653, opened 2026-07-20
Add a check for namespace existence in kpt live apply

OisinJohnston2005:
Fixes issue #2123, by adding a check for the existence of the namespace in kpt live apply

### DIFF
--- a/commands/live/apply/cmdapply.go
+++ b/commands/live/apply/cmdapply.go
@@ -28,6 +28,7 @@ import (
 	alphaprinterstable "github.com/kptdev/kpt/pkg/printer/table"
 	"github.com/kptdev/kpt/pkg/status"
 	"github.com/spf13/cobra"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
@@ -53,6 +54,9 @@ func NewRunner(
 		factory:     factory,
 		applyRunner: runApply,
 		alpha:       alpha,
+	}
+	r.namespaceChecker = func(namespace string) error {
+		return checkNamespaceExists(factory, namespace)
 	}
 	c := &cobra.Command{
 		Use:     "apply [PKG_PATH | -]",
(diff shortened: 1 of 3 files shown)


commands/live/apply/cmdapply_test.go
```diff
@@ -148,6 +148,7 @@ func TestCmd(t *testing.T) {
 
 			runner := NewRunner(fake.CtxWithDefaultPrinter(), tf, ioStreams, false)
 			runner.Command.SetArgs(tc.args)
+			runner.namespaceChecker = func(_ string) error { return nil }
```
Catalin-Stratulat-Ericsson, line 195: good stuff. correct me if im wrong though but i belive this only checks the "happy path" side of things aka if namespace is given as intended. but not if it goes wrong aka given a invalid ns or not given a ns at all path and check that the error returned is what we expect?

commands/live/apply/cmdapply.go
```diff
@@ -191,6 +196,12 @@ func (r *Runner) runE(c *cobra.Command, args []string) error {
 		return err
 	}
 
+	if !live.NamespaceInObjects(objs, inv.Namespace) {
```
Catalin-Stratulat-Ericsson, line 204: in the function should we assume the default that if no ns is given we check the default ns the same way as kubectl does? or do we default to invalid return. just a proposition id suggest defaulting same as k8s but just a thought

Catalin-Stratulat-Ericsson (commented):
looks good just 2 minor comments

commands/live/apply/cmdapply_test.go
OisinJohnston2005: I wasn't sure how to go about mocking the kubernetes client for the test framework, without this line all of these tests fail, I can do a bit of digging today and see what options are available.

the applyRunner func seems to be overwritten in a similar way one line below this.

commands/live/apply/cmdapply_test.go
Catalin-Stratulat-Ericsson, line 174: related to my comment about the "happy path" test cases comment. could we not just add 2/3 more tests here one to check valid ns given, one to check invalid and one to maybe check null value given and one to check our expectation. i haven't fully read through this test framework though so i probably am missing something.

commands/live/apply/cmdapply.go
OisinJohnston2005: I'll implement defaulting to "default" if inv.Namespace is "", that was an oversight on my part

mozesl-nokia (dismissed):

mozesl-nokia (approved):

Catalin-Stratulat-Ericsson (approved):
looks good 👍

PR #4660, opened 2026-07-23
Improve error message if inventory does not exist in kpt live status.

### LINKED ISSUE #2124: kpt live status returns a misleading error if the inventory object doesn't exist
### Expected behavior
The error message should describe that an inventory object (RG CR) with the name/namespace from the inventory information couldn't be found

### Actual behavior
The error message is just : `no resources found in the inventory`

OisinJohnston2005:
This PR is to fix #2124.

The error message referred to is propagated from cli-utils where they directly print the message to stdout and don't provide any mechanism to catch it.

I added an api call to verify the inventory's existence in kpt live status allowing me to fix the misleading error message, this adds a redundant API call which may not be worth the cleaner output, but the answer to that question is above my paygrade.

### DIFF
--- a/commands/live/status/cmdstatus.go
+++ b/commands/live/status/cmdstatus.go
@@ -16,6 +16,7 @@ package status
 
 import (
 	"context"
+	"fmt"
 	"os"
 
 	"github.com/kptdev/kpt/internal/docs/generated/livedocs"
@@ -27,6 +28,9 @@ import (
 	"sigs.k8s.io/cli-utils/cmd/status"
 	"sigs.k8s.io/cli-utils/pkg/apply/poller"
 	"sigs.k8s.io/cli-utils/pkg/inventory"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func NewRunner(ctx context.Context, factory util.Factory,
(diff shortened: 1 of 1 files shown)


commands/live/status/cmdstatus.go
```diff
@@ -27,6 +28,9 @@ import (
 	"sigs.k8s.io/cli-utils/cmd/status"
 	"sigs.k8s.io/cli-utils/pkg/apply/poller"
 	"sigs.k8s.io/cli-utils/pkg/inventory"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
liamfallon, line 32: I think we're trying to remove dependencies with apimachinery.

commands/live/status/cmdstatus.go
```diff
@@ -90,5 +94,36 @@ func (rir *RGInventoryLoader) GetInvInfo(cmd *cobra.Command, args []string) (inv
...
 	}
 
+	if err := verifyInventoryExists(rir.ctx, rir.factory, invInfo); err != nil {
+		return nil, err
+	}
+
 	return invInfo, nil
 }
+
+func verifyInventoryExists(ctx context.Context, f util.Factory, invInfo inventory.Info) error {
```
liamfallon, line 104: In this function it would be better to wrap the original error in a more helpful error that you create and that gives context.

the pkgerrors package is useful for this, see:
	pkgerrors "github.com/pkg/errors"


see for example:
https://github.com/Nordix/porch/blob/af771881508cea65f4d7c8247c639ff9f424fe6b/pkg/apiserver/apiserver.go#L410

OisinJohnston2005: the original error in the upstream dependency is dumped directly to stdout [here](https://github.com/kubernetes-sigs/cli-utils/blob/master/cmd/status/cmdstatus.go#L263) so I don't think this solution is viable, this is me implementing [Aravindhan's comment](https://github.com/kptdev/kpt/issues/2124#issuecomment-5057500737) on the original issue.

aravindtga:
@OisinJohnston2005 This can be closed as the decision has been made to not do this. Because the error message is written directly to stdio by the cli-utils library.

OisinJohnston2005:
Decision has been made not to implement

PR #4675, opened 2026-07-31
fixed stale fnconfig bug

### LINKED ISSUE #4674: Validator receives stale functionConfig when configPath resource is mutated by preceding mutators
## Description

 When `kpt fn render` runs a validator that uses `configPath`, the validator's `functionConfig` is always read from the original file on disk, even if a preceding mutator has modified that resource in memory. This means the validator validates against outdated configuration.

This does not affect mutators, which already have a `SetFnConfig` refresh mechanism that picks up in-memory changes before each execution.

  ## Reproduction

  Create the following files in a directory:

  **Kptfile:**
  ```yaml
  apiVersion: kpt.dev/v1
  kind: Kptfile
  metadata:
    name: stale-test
  pipeline:
    mutators:
      - image: ghcr.io/kptdev/krm-functions-catalog/set-labels:latest
        configMap:
          env: production
    validators:
      - image: ghcr.io/kptdev/krm-functions-catalog/starlark:latest
        configPath: validator-config.yaml
```
  **validator-config.yaml:**
  ```yaml
  apiVersion: fn.kpt.dev/v1alpha1
  kind: StarlarkRun
  metadata:
    name: check-config
  source: |-
    fc = ctx.resource_list["functionConfig"]
    labels = fc.get("metadata", {}).get("labels", {})
    if "env" not in labels:
      fail("functionConfig is stale - missing env label that mutator should have added")
```

  **deployment.yaml:**
  ```yaml
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: my-app
  spec:
    replicas: 1
```

### Run: `kpt fn render`

  **Expected:** Validator passes, set-labels added env: production label to validator-config.yaml in memory, and the validator should see it in its functionConfig.

  **Actual:** Validator fails with "functionConfig is stale - missing env label" because it reads the original file from disk (no labels).

### Root Cause

  In `pkg/lib/kptops/render_executor.go`, `runValidators` creates the function runner via `fnruntime.NewRunner`, which calls `newFnConfig()` → `fsys.Open(path)` to read the config from the filesystem. At this point in the render pipeline, the filesystem hasn't been written yet, mutations only exist in the in-memory `[]*yaml.RNode` slice.

Mutators don't have this issue because `runMutators` calls `mutator.SetFnConfig(r)` to refresh the config reference from the in-memory input before each execution. Validators lack this refresh.

### Suggested Fix

After creating the validator runner, refresh its functionConfig from the in-memory input (same pattern as mutators):

```go
  if function.ConfigPath != "" {
      for _, r := range input {
          pkgPath, _ := pkg.GetPkgPathAnnotation(r)
          currPath, _, _ := kioutil.GetFileAnnotations(r)
          if pkgPath == pn.pkg.UniquePath.String() && currPath == function.ConfigPath {
              validator.SetFnConfig(r)
              break
          }
      }
  }
```

OisinJohnston2005:
Fixed #4674 

made the suggested change in the issue and added unit tests to ensure no regression.

### DIFF
--- a/pkg/lib/kptops/render_executor.go
+++ b/pkg/lib/kptops/render_executor.go
@@ -797,6 +797,11 @@ func (pn *pkgNode) runMutators(ctx context.Context, hctx *hydrationContext, inpu
 				return nil, err
 			}
 		}
+
+		if err := pn.refreshFnConfig(mutator, input, pl.Mutators[i].ConfigPath); err != nil {
+			return nil, err
+		}
+
 		// select the resources on which function should be applied
 		selectedInput, err := fnruntime.SelectInput(input, selectors, exclusions, &fnruntime.SelectionContext{RootPackagePath: hctx.root.pkg.UniquePath})
 		if err != nil {
--- a/pkg/lib/kptops/render_executor.go
+++ b/pkg/lib/kptops/render_executor.go
@@ -858,7 +858,7 @@ func (pn *pkgNode) runValidators(ctx context.Context, hctx *hydrationContext, in
 		if err != nil {
 			return err
 		}
-		var validator kio.Filter
+		var validator *fnruntime.FunctionRunner
 		displayResourceCount := false
 		if len(function.Selectors) > 0 || len(function.Exclusions) > 0 {
 			displayResourceCount = true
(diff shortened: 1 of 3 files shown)


pkg/lib/kptops/render_executor.go
```diff
@@ -892,6 +894,29 @@ func cloneResources(input []*yaml.RNode) (output []*yaml.RNode) {
 	return
 }
 
+// refreshFnConfig updates the runner's functionConfig from the in-memory input
+// to pick up any mutations applied earlier in the pipeline.
+func (pn *pkgNode) refreshFnConfig(runner *fnruntime.FunctionRunner, input []*yaml.RNode, configPath string) error {
```
aravindtga, line 882: Can this be used with runMutators function as well ? It is using same logic to refresh the mutator config now ?

OisinJohnston2005: fixed

aravindtga (approved):

mozesl-nokia (approved):

PR #4678, opened 2026-08-03
Improve logging of render pipeline execution (part 1)

mozesl-nokia:
The main visible change in this PR is that now for every log line in render that starts with `[RUNNING]` we also log the package name after it. Also, the ":" in OptPrintf was moved to _after_ the package name. These were supposed to be part of the "port additional changes from Porch" but got lost somewhere in the merges.

The bulk of the changes are actually under the hood, preparing for a more in-depth change of the whole printer interface. This is mainly some new custom printer formatting options which can accomodate Porch's naming scheme for subpackages.

Auxiliary changes:
- added linker flags to all the test make targets

*Update: Looking into the docs revealed that when we merged things from Porch we actually lost a lot of the formatting (newlines, indentation, etc.). I mostly reverted those changes so the formatting now matches what was in the docs. Also updated the docs to contain the "on package..." on the appropriate lines.

AI attribution:
- Cursor's Grok 4.5 was used to update the E2E test configs the second time around.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,24 +80,24 @@ test: test-api
 # KRM_FN_RUNTIME can be set to select the desired function runtime.
 # If unspecified, the default function runtime will be used.
 test-docker: build
-	PATH="$(GOBIN):$(PATH)" go test -cover --tags=docker ./...
+	PATH="$(GOBIN):$(PATH)" go test -cover ${LDFLAGS} --tags=docker ./...
 
 # KPT_E2E_UPDATE_EXPECTED=true (if expected output to be updated)
 # target to run e2e tests for "kpt fn render" command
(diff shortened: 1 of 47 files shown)


efiacor (approved):

efiacor:
I guess this, and subsequent changes will require doc and test updates.

mozesl-nokia:
@efiacor it looks like Dosu did catch at least some of the needed changes (and some that are not), but I don't see a way to accept it's changes...

mozesl-nokia:
Looking through the docs, the *current* formatting of the logs looked way off compared to the docs, so I matched them back to the docs.

efiacor (approved):

liamfallon (approved):

PR #4696, opened 2026-08-12
Error when explicit destination of pkg get already exists

### LINKED ISSUE #2656: kpt pkg get duplicates a package as a subdirectory when attempting to override the local directory name
### Expected behavior

For `kpt pkg get <repo>@main packages/newDirName` to put the pulled package into `/packages/newDirName`, leaving us with:

```shell
|-- packages
|   `-- newDirName
|       |-- patches
|       `-- resources
|-- patches
|   `-- cluster
|-- scratch
`-- scripts
```

When ran a 2nd time expect to see:

> Error: destination directory "packages/newDirName" already exists

### Actual behavior

upon running a 2nd time, we get:

```shell
|-- packages
|   `-- newDirName
|       |-- repoName             <--- why do we get the pkg pulled a 2nd time?
|       |   |-- patches
|       |   `-- resources
|       |-- patches
|       `-- resources
|-- patches
|   `-- cluster
|-- scratch
`-- scripts
```

running a 3rd time provides the correct error, presumeably because `./packages/newDirName/repoName` already exists.

### Information

kpt version `1.0.0beta11`

```shell
kpt pkg get <repo>@main packages/newDirName
kpt pkg get <repo>@main packages/newDirName
```

gangadhar-res:
Fixes #2656

Problem: running kpt pkg get REPO_URI DEST a second time silently fetched the package into DEST/<pkg-name> instead of erroring. getDest() always defaulted to a repo-named subdirectory when the destination existed, even when the user explicitly named it.

Fix: when the destination was explicitly provided, already exists, and is not ., return the same "destination directory already exists" error the defaulted path produces. The . special case and the defaulting behavior for omitted destinations are unchanged.

Validation: ran the reproduction steps from the issue against this branch — first run fetches into packages/newDirName, second run now errors with destination directory "packages/newDirName" already exists, and no nested nginx/ directory is created. kpt pkg get <url> with no destination still defaults to ./nginx. Added a Test_getDest table test covering explicit/defaulted/existing/./file/missing-parent cases.

### DIFF
--- a/pkg/lib/util/parse/parse.go
+++ b/pkg/lib/util/parse/parse.go
@@ -276,6 +276,17 @@ func getDest(v, repo, subdir string, explicitDest bool) (string, error) {
 		return v, nil
 	}
 
+	// The user explicitly named a destination that already holds a fetched
+	// package: error out instead of silently nesting another copy inside it,
+	// so a repeated `kpt pkg get REPO_URI DEST` fails the same way the
+	// defaulted destination does below. An existing directory without a
+	// Kptfile keeps working as a container to fetch the package into.
+	if explicitDest {
+		if _, err := os.Stat(filepath.Join(v, kptfilev1.KptFileName)); err == nil {
+			return "", errors.Errorf("destination directory %q already exists and contains a package", v)
+		}
+	}
+
 	// default the location to a new subdirectory matching the pkg URI base
 	repo = strings.TrimSuffix(repo, "/")
 	repo = strings.TrimSuffix(repo, ".git")
--- a/pkg/lib/util/parse/parse_test.go
+++ b/pkg/lib/util/parse/parse_test.go
@@ -17,6 +17,8 @@ package parse
 import (
 	"context"
 	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
 
 	kptfilev1 "github.com/kptdev/kpt/api/kptfile/v1"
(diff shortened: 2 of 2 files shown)


pkg/lib/util/parse/parse.go
```diff
@@ -276,6 +276,17 @@ func getDest(v, repo, subdir string, explicitDest bool) (string, error) {
 		return v, nil
 	}
 
+	// The user explicitly named a destination that already holds a fetched
+	// package: error out instead of silently nesting another copy inside it,
+	// so a repeated `kpt pkg get REPO_URI DEST` fails the same way the
+	// defaulted destination does below. An existing directory without a
+	// Kptfile keeps working as a container to fetch the package into.
+	if explicitDest {
+		if _, err := os.Stat(filepath.Join(v, kptfilev1.KptFileName)); err == nil {
```
aravindtga, line 285: Is there a specific reason for checking the Kptfile rather than erroring whenever the explicit destination already exists? This could be simpler to just error unconditionally if the directory exists.

aravindtga (changes requested):

pkg/lib/util/parse/parse.go
gangadhar-res: @aravindtga 
Good question. That was actually the first version of this PR: unconditionally erroring when the explicit destination exists. It failed CI because many of the existing command tests rely on fetching into an existing directory as a container, for example TestCmd_execute and most of the cmdget suite pass the existing workspace directory and expect the package to land at workspace/<pkg-name>. Real scripts that mkdir a target and then run kpt pkg get URL target would break the same way. The Kptfile check was the narrowest rule I could find that still catches the repeated-fetch case from #2656: the second run always trips it because the first run wrote a Kptfile at that path, while container directories keep working. If you would rather error unconditionally and treat the container pattern as deprecated, I am happy to make that change and update the affected tests, but that felt like a bigger behavior decision than this bug fix should make on its own imo.

aravindtga (approved):

gangadhar-res:
Hello CODEOWNERs 

@liamfallon @kispaljr @mozesl-nokia @efiacor

Could one of your review and merge this code change ?

gangadhar-res:
@efiacor @mozesl-nokia this has an lgtm from @aravindtga and CI is green. Could one of you take a look when you get a chance? It errors when an explicit pkg get destination already contains a package, instead of silently nesting a second copy inside it.

mozesl-nokia (approved):
Hi, sorry about the very late review

gangadhar-res:
> Hi, sorry about the very late review

no worries, thank you!

PR #4704, opened 2026-08-17
Fix Kptfile pipeline not updating on pkg update with fast-forward/force-delete-replace

### LINKED ISSUE #4012: kpt: pkg update fails to update Kptfile
### Expected behavior

A `kpt pkg get` followed by a `kpt pkg update --strategy force-delete-replace` results in a `Kptfile` that has the git information as specified in the `kpt pkg update` but the pipeline section of the `Kptfile` is from the initial `kpt pkg get` and not the pipeline from the `kpt pkg update`. The consequence is that one cannot update the package pipeline.

The problem also applies to strategy `fast-forward`.

Tested using `kpt` version `1.0.0-beta.38` and `1.0.0-beta.43`.

Detailed description of the problem and steps to reproduce it can be found here: https://github.com/michaelvl/kpt-test

OisinJohnston2005:
Fixes #4012 

With test cases taken from #4030 

Instead of performing a three way merge the KPT file is directly updated from upstream when using strategies that should not respect local changes like fast-forward and force-delete-replace, allowing the upstream to update the pipeline.

### DIFF
--- a/pkg/kptfile/kptfileutil/util_test.go
+++ b/pkg/kptfile/kptfileutil/util_test.go
@@ -1680,3 +1680,141 @@ upstreamLock:
 		})
 	}
 }
+
+func TestReplaceKptfile_errorPaths(t *testing.T) {
+	writeInvalidKptfile := func(tt *testing.T) string {
+		dir := tt.TempDir()
+		err := os.WriteFile(filepath.Join(dir, kptfilev1.KptFileName), []byte(`
+apiVersion: kpt.dev/v1alpha1
+kind: Kptfile
+metadata:
+  name: bad
+`), 0600)
+		if !assert.NoError(tt, err) {
+			tt.FailNow()
+		}
+		return dir
+	}
+
+	writeValidKptfile := func(tt *testing.T) string {
+		dir := tt.TempDir()
+		err := os.WriteFile(filepath.Join(dir, kptfilev1.KptFileName), []byte(`
+apiVersion: kpt.dev/v1
+kind: Kptfile
+metadata:
+  name: valid
+`), 0600)
+		if !assert.NoError(tt, err) {
+			tt.FailNow()
+		}
+		return dir
+	}
+
+	t.Run("error reading local Kptfile", func(t *testing.T) {
+		localDir := writeInvalidKptfile(t)
+		updatedDir := writeValidKptfile(t)
+
+		err := ReplaceKptfile(localDir, updatedDir)
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "kptfileutil.ReplaceKptfile")
+	})
+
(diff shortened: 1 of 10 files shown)


OisinJohnston2005:
SonarQube is only complaining because it's still including `testutil.go` - a test file, in its test coverage tests

efiacor (approved):

pkg/kptfile/kptfileutil/util_test.go
```diff
@@ -1408,3 +1408,413 @@ pipeline:
...
+
+func TestReplaceKptfile(t *testing.T) {
+	writeKptfileToTemp := func(tt *testing.T, content string) string {
+		dir := tt.TempDir()
+		err := os.WriteFile(filepath.Join(dir, kptfilev1.KptFileName), []byte(content), 0600)
+		if !assert.NoError(tt, err) {
+			tt.FailNow()
+		}
+		return dir
+	}
```
mozesl-nokia: I think this can be a scope higher. Also:

```suggestion
	writeKptfileToTemp := func(tt *testing.T, content string) string {
	    tt.Helper()
		dir := tt.TempDir()
		err := os.WriteFile(filepath.Join(dir, kptfilev1.KptFileName), []byte(content), 0600)
		if !assert.NoError(tt, err) {
			tt.FailNow()
		}
		return dir
	}
```

pkg/kptfile/kptfileutil/util_test.go
```diff
@@ -1408,3 +1408,413 @@ pipeline:
...
+apiVersion: kpt.dev/v1
+kind: Kptfile
+metadata:
+  name: valid
+`), 0600)
+		if !assert.NoError(tt, err) {
+			tt.FailNow()
+		}
+		return dir
+	}
```
mozesl-nokia: These could just use `writeKptfileToTemp` with 2 different const contents

pkg/kptfile/kptfileutil/util_test.go
```diff
@@ -1408,3 +1408,413 @@ pipeline:
...
```
mozesl-nokia: Same thing here with `writeKptfileToTemp`

pkg/kptfile/kptfileutil/util_test.go
aravindtga: Done.

pkg/kptfile/kptfileutil/util_test.go
aravindtga: Done.

pkg/kptfile/kptfileutil/util_test.go
aravindtga: Done.

mozesl-nokia (approved):

liamfallon (approved):

PR #4715, opened 2026-08-25
feat: Print skip reasons in live plan output

### LINKED ISSUE #4714: Provide descriptive reasons for skipped resources in the live plan
### Describe your problem

Currently, when running `kpt live plan` and a resource evaluation results in a skipped action (`kptplanner.Skip`), the CLI simply prints out a yellow `=` sign next to the resource without any context.

For example, the output looks something like this:
`=   Group/Kind  namespace/name`

There is an existing `TODO` in `commands/alpha/live/plan/command.go` around line 181 highlighting this missing functionality:
```go
case kptplanner.Skip:
	// TODO: provide more information about why the resource was skipped.
	printEntryWithColor("=", print.YELLOW, action, ioStreams)
```

### Specific use case: 
When managing complex applications, resources may be intentionally skipped during actuation due to specific annotations or missing references. Without a descriptive message explaining why the resource was skipped, developers and operators are left guessing if it was a configuration error, an intentional lifecycle hook, or a bug.

Providing the specific skip reason from the action context would greatly improve transparency and the overall developer experience during kpt live plan.

vishnukothakapu:
**Description of the change**
This PR updates the CLI output logic in `commands/alpha/live/plan/command.go` and the event parsing in `pkg/live/planner/cluster.go` to properly capture and print the skip reason for resources that are skipped during actuation. 

Previously, skipped events with a reason attached to the `Error` field were incorrectly classified as hard errors. They are now appropriately marked as `Skip` with a `SkipReason`, and the `kpt live plan` output prints this reason right below the resource.

**Motivation for the change**
When managing complex applications, resources may be intentionally skipped during actuation due to specific annotations or missing references. Without a descriptive message explaining *why* the resource was skipped, developers and operators are left guessing if it was a configuration error, an intentional lifecycle hook, or a bug. This change improves transparency and developer experience by surfacing the skip reason provided by `cli-utils`.

**Fixes**
Fixes #4714

### DIFF
--- a/commands/alpha/live/plan/command.go
+++ b/commands/alpha/live/plan/command.go
@@ -178,8 +178,10 @@ func printText(plan *kptplanner.Plan, objs []*unstructured.Unstructured, ioStrea
 			printEntry(" ", action, ioStreams)
 			findAndPrintDiff(action.Original, action.Updated, ContentPrefix, ioStreams)
 		case kptplanner.Skip:
-			// TODO: provide more information about why the resource was skipped.
 			printEntryWithColor("=", print.YELLOW, action, ioStreams)
+			if action.SkipReason != "" {
+				printWithPrefix(action.SkipReason, ContentPrefix, ioStreams)
+			}
 		case kptplanner.Error:
 			printEntry("!", action, ioStreams)
 			printWithPrefix(action.Error, ContentPrefix, ioStreams)
--- a/commands/alpha/live/plan/command_test.go
+++ b/commands/alpha/live/plan/command_test.go
@@ -1,0 +1,64 @@
+package plan
+
+import (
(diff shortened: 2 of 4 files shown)


mozesl-nokia (approved):

efiacor:
Hi @vishnukothakapu 
Thanks for your contribution. Please address the linting errors and get the CI to pass.

efiacor:
@vishnukothakapu .
Can you add some unit test coverage for the new code?
See here for example - https://github.com/kptdev/kpt/blob/main/pkg/live/planner/cluster_test.go
Our new sonarqube gating is set at 80%. Try to get close to that please.

efiacor (approved):

PR #4720, opened 2026-08-31
Fix copy-merge to preserve locally-added files in upstream-deleted directories

kushnaidu:
When the upstream revision deletes a directory that still contains files added only in the downstream package, RemoveStaleItems failed with "directory not empty" because it used os.Remove (non-recursive) on directories that legitimately had local content remaining.

This surfaced as an Internal error from the Porch API server during porchctl rpkg upgrade --strategy=copy-merge.

The fix checks whether each stale directory is empty before removing it, consistent with the copy-merge spec: "files that only exist locally are kept." Stale files inside the directory are still removed, but locally-added files and their parent directories are preserved.

The approach mirrors RemovePackageContent in the same file, which already uses Readdirnames(1) to verify emptiness before deletion.

### DIFF
--- a/pkg/lib/pkg/util.go
+++ b/pkg/lib/pkg/util.go
@@ -222,8 +222,8 @@ func RemoveStaleItems(org, src, dst string, _ bool, _ SubpackageMatcher) error {
 	sort.Slice(dirsToDelete, SubPkgFirstSorter(dirsToDelete))
 	for _, dir := range dirsToDelete {
 		// Check whether the directory is empty before removing it. It may still
-		// contain files that were added locally (not present in org) and therefore
-		// intentionally preserved by the walk above. Removing a non-empty
+		// contain files that were added locally (not present in upstream) and therefore
+		// intentionally preserved by the filepath walk above. Removing a non-empty
 		// directory with os.Remove would fail; instead we skip it so those
 		// locally-added files survive the merge.
 		f, err := os.Open(dir)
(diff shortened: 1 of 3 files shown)


pkg/lib/pkg/util.go
```diff
@@ -215,11 +218,28 @@ func RemoveStaleItems(org, src, dst string, _ bool, _ SubpackageMatcher) error {
 	if walkErr != nil {
 		return walkErr
 	}
+	// Sort directories deepest-first so children are processed before parents.
 	sort.Slice(dirsToDelete, SubPkgFirstSorter(dirsToDelete))
 	for _, dir := range dirsToDelete {
-		if err := os.Remove(dir); err != nil {
+		// Check whether the directory is empty before removing it. It may still
+		// contain files that were added locally (not present in org) and therefore
+		// intentionally preserved by the walk above. Removing a non-empty
```
efiacor:
```suggestion
		// contain files that were added locally (not present in orig) and therefore
		// intentionally preserved by the filepath walk above. Removing a non-empty
```

efiacor (approved):

liamfallon (approved):
Looks good. Please fix the typo spotted by @efiacor

pkg/lib/pkg/util.go
kushnaidu: updated and used 'upstream' instead of 'orig.'

Issue #4752, opened 2026-09-15
brew install kpt reports `unknown` git commit hash


aravindtga:
### Problem

  The Homebrew formula (`Formula/kpt.rb`) builds kpt from the **source tarball**, not from a git checkout:

  ```ruby
  url "https://github.com/kptdev/kpt/archive/v1.0.0.tar.gz"
```

The GitHub archive/*.tar.gz contains no .git directory, so Go's VCS stamping (vcs.revision / vcs.modified, read in run/run.go) has nothing to read. As a result kpt version from a Homebrew install cannot report a real commit hash.  It falls back to Git commit: unknown (the version string itself is correct via the ldflag).

```bash
$ kpt version
Version: 1.0.0
Git commit: unknown
```

This is the Homebrew counterpart of #4748

### Expected
`kpt version` should report the correct version and a clean commit reference from all official channels, including Homebrew.

### Proposed fix

1. Build the formula from the tag, not the tarball. In homebrew-core kpt formula switch the source so the build has a real .git and Go can stamp the commit:
```ruby
  url "https://github.com/kptdev/kpt.git",
      tag:      "v1.0.0",
      revision: "<full-sha>"
```

2. Add an authoritative commit ldflag in `run/run.go`. A git checkout can still be dirtied by Homebrew's build. Add an override var that, when set(via ldflag), it skips the VCS build-info lookup; when empty, behaviour is unchanged:
```go
var gitCommit = "" // set via -ldflags; authoritative when non-empty

// in versionCmd:
if gitCommit != "" {
    hash = gitCommit
} else if info, ok := debug.ReadBuildInfo(); ok {
    // ... existing vcs.revision / vcs.modified logic ...
}
```

PR #4757, opened 2026-09-16
run: add authoritative gitCommit ldflag override for version command(#4752)

adi-coderr:
### Description

When building `kpt` from source archives or tarballs (such as in package managers like Homebrew: `https://github.com/kptdev/kpt/archive/refs/tags/v1.0.0.tar.gz`), there is no `.git` directory present in the source tree.

As a result, Go's runtime VCS stamping (`vcs.revision` in `runtime/debug.ReadBuildInfo()`) has no Git metadata to extract, causing `kpt version` to report:
```text
Version: 1.0.0
Git commit: unknown


run/run_test.go
mozesl-nokia, line 1: I think you could rewrite these tests to use subtests, there is a lot of duplicate code like this

mozesl-nokia (approved):

aravindtga (approved):

liamfallon (approved):

liamfallon:
@adi-coderr Please address the comment from @mozesl-nokia and then we can merge the PR.

run/run_test.go
adi-coderr: Thanks for the feedback! I've refactored the tests in run_test.go into a single table-driven test (TestVersionCmd) using subtests via t.Run.

The common setup, teardown/flag reset, execution, and output assertions are now centralized, removing the duplicate code across the test cases. Please Take a Look!!

aravindtga (approved):

adi-coderr:
@liamfallon @mozesl-nokia  I’ve fixed the issues. Please recheck my changes when you get a chance.

Issue #4764, opened 2026-09-21
`go install`-ing kpt doesn't contain version information


mozesl-nokia:
### Expected behavior
```bash
$ kpt version
Version: 1.0.0
Git commit: 4600383a5bf70812f9d1bda2670f034cf8daa470
```
### Actual behavior
```bash
$ kpt version
Version: unknown
Git commit: unknown
```
### Information

When you install kpt with `go install` the version information is not present, because it is only injected with ldflags.

A similar approach to golangci-lint could be used, which manages to fill at least some version information:
```bash
$ golangci-lint --version
golangci-lint has version 2.13.2 built with go1.27.0 from (unknown, modified: ?, mod sum: "h1:bCyq3E4vo9qwzifjpzJqYndEt7Ncva80qkk8G2B4TXU=") on (unknown)
```

(Getting the git hash is probably not possible)

mozesl-nokia:
Related:
- https://github.com/kptdev/kpt/issues/4752
- https://github.com/kptdev/kpt/issues/4748

PR #4765, opened 2026-09-21
run: resolve version and commit metadata from debug.ReadBuildInfo (#4764)

adi-coderr:
## Description

When users install `kpt` via `go install github.com/kptdev/kpt@...`, no `-ldflags` are passed to the compiler and no local `.git` repository exists in the Go module cache. Consequently, `kpt version` previously output:

```text
Version: unknown
Git commit: unknown

### CODE REVIEW

**Comment by adi-coderr** [comment]
@aravindtga I have resolved the issue. Could you please recheck my changes?

File: run/run.go [hunk 1]
```diff
@@ -206,44 +208,98 @@ func newHelp(e []string, c *cobra.Command) func(command *cobra.Command, strings
...
+
+func extractVCSSettings(settings []debug.BuildSetting) (string, string) {
+	var hash, dirty string
+	for _, setting := range settings {
+		switch setting.Key {
+		case "vcs.revision":
+			hash = setting.Value
+		case "vcs.modified":
+			if strings.EqualFold(setting.Value, "true") {
+				dirty = " (dirty)"
```
aravindtga: `dirty` can be a boolean, so that the string formatting to include can be done at the caller.

aravindtga (commented):
Consider extracting the resolveVersion logic into a shared pkg/version package. porch needs the same logic as well, so both CLIs could reuse it (follow-up, non-blocking).

run/run.go
adi-coderr: @aravindtga could you please recheck again.

mozesl-nokia (approved):

aravindtga (approved):

